PR #3, opened 2025-02-01
Init

n8mgr:


cmd/indexd/main.go
```diff
@@ -0,0 +1,330 @@
...
+	var logLevelOverride string
+
+	rootCmd := flagg.Root
+	rootCmd.Usage = flagg.SimpleUsage(rootCmd, ``)
+	rootCmd.StringVar(&cfg.Directory, "dir", cfg.Directory, "directory to store indexd metadata in")
+	// syncer
+	rootCmd.StringVar(&cfg.Syncer.Address, "syncer.address", cfg.Syncer.Address, "address to listen for p2p connections")
+	rootCmd.BoolVar(&cfg.Syncer.Bootstrap, "syncer.bootstrap", cfg.Syncer.Bootstrap, "when true, will connect to default peers for syncing consensus")
+	// consensus
+	rootCmd.StringVar(&cfg.Consensus.Network, "network", cfg.Consensus.Network, "network name (mainnet, zen, etc)")
```
peterjan: Do we have a rule on what config fields to expose as a CLI flag? I like the ones you've selected, just common sense really, just figured to draw attention to it because you could also argue we can get away with not having any CLI flags and force users to configure the indexer through a YAML config file.

cmd/indexd/main.go
```diff
@@ -0,0 +1,330 @@
...
+// checkFatalError prints an error message to stderr and exits with a 1 exit code. If err is nil, this is a no-op.
+func checkFatalError(context string, err error) {
+	if err == nil {
+		return
+	}
+	os.Stderr.WriteString(fmt.Sprintf("%s: %s\n", context, err))
+	os.Exit(1)
+}
+
+func parseLogLevel(level string) zap.AtomicLevel {
```
peterjan: we can use `zapcore.ParseLevel`

peterjan (dismissed):

cmd/indexd/main.go
n8mgr: There's not yet. I could also see dropping all but `dir` and `-log` -- those are the only ones I feel confident users should be able to override them.

n8mgr: Dropped all but log and dir for now. We can add more and try to come up with a standard for making that decision in a f/u.

PR #4, opened 2025-02-03
Init postgres

n8mgr:
Requires #3


persist/postgres/init.sql
```diff
@@ -0,0 +1,22 @@
+CREATE TABLE hello_world ( -- TODO: remove
```
peterjan: is this a leftover `TODO` or do you want to keep it in a while longer?

persist/postgres/init.sql
```diff
@@ -0,0 +1,22 @@
...
+    peer_address INET PRIMARY KEY,
+    first_seen TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
+);
+
+CREATE TABLE syncer_bans (
+    net_cidr CIDR PRIMARY KEY,
+    expiration TIMESTAMP WITH TIME ZONE NOT NULL,
+    reason TEXT NOT NULL
+);
+CREATE INDEX syncer_bans_expiration_index_idx ON syncer_bans (expiration);
```
peterjan: should we adopt the following pattern `[TABLE_NAME]_[FIELD_NAME]_idx` to ensure we're all consistent when it comes to index names?

persist/postgres/store_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	err := store.transaction(context.Background(), func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(`INSERT INTO hello_world VALUES ($1)`, sqlCurrency(amount))
+		return err
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var value types.Currency
```
peterjan: `nit` missing the assert to check whether value equals the persisted amount

persist/postgres/store_test.go
```diff
@@ -0,0 +1,64 @@
...
+			panic(err)
+		} else if err := db.Close(); err != nil {
+			panic(err)
+		}
+	})
+	return db
+}
+
+func TestPostgresCurrency(t *testing.T) {
+	amount := types.Siacoins(uint32(frand.Uint64n(math.MaxUint32)))
```
peterjan: doesn't hurt to also test `NewCurrency(math.MaxUint64, math.MaxUint64)` to assert we can persist values that exceed a `BIGINT`

peterjan (approved):
LGTM 🔥

persist/postgres/init.sql
n8mgr: We can remove it in a future PR that adds to the store. The goal was to have one working test

persist/postgres/init.sql
n8mgr: I'm not strongly attached to any format as long as we're consistent, but that's what I use in `hostd`.

persist/postgres/init.sql
n8mgr: Removed the placeholder and changed the test to create a temp table for testing currency storage and retrieval

persist/postgres/init.sql
n8mgr: It looks like `renterd` does `[PREFIX]_[TABLE NAME]_[COLUMN NAME(s)]`.

For reference, the automatic names in Postgres are: `[TABLE NAME]_[COLUMN NAME(s)]_[SUFFIX]`

Where suffix is one of the following:
- `pkey` for a Primary Key constraint
- `key` for a Unique constraint
- `excl` for an Exclusion constraint
- `idx` for any other kind of index
- `fkey` for a Foreign key
- `check` for a Check constraint
- `seq` for all sequences

peterjan: Is it possible you missed `_index_idx`?

n8mgr: I did miss that entirely

PR #5, opened 2025-02-03
Use AtomicLevel in config

n8mgr:


cmd/indexd/main.go
```diff
@@ -182,19 +170,15 @@ func main() {
 
 	// attempt to load the config file, command line flags will override any
 	// values set in the config file
-	configPath := tryLoadConfig()
-	if configPath != "" {
-
-	}
-	// set the data directory to the default if it is not set
+	tryLoadConfig()
```
peterjan, line 173: `nit` since this is the only caller we might want to adjust the function signature and not return the path at all

peterjan (approved):
good stuff 👍

PR #7, opened 2025-02-04
Add HTTP server

peterjan:


cmd/indexd/main.go
```diff
@@ -22,12 +22,17 @@ import (
 )
 
 const (
-	configFileEnvVar = "INDEXD_CONFIG_FILE"
-	dataDirEnvVar    = "INDEXD_DATA_DIR"
+	apiPasswordEnvVar = "INDEXD_API_PASSWORD"
```
n8mgr: I think we should force this into the config file

cmd/indexd/main.go
```diff
@@ -22,12 +22,17 @@ import (
...
+	configFileEnvVar  = "INDEXD_CONFIG_FILE"
+	dataDirEnvVar     = "INDEXD_DATA_DIR"
 )
 
 var cfg = config.Config{
 	Directory: os.Getenv(dataDirEnvVar), // default to env variable
+	HTTP: config.HTTP{
+		Address:  "127.0.0.1:9980",
```
n8mgr: Should we default to `:9980` here? The API is designed to be exposed publicly anyway.

PR #9, opened 2025-02-05
Add Syncer

peterjan:


persist/postgres/init.sql
```diff
@@ -1,6 +1,10 @@
 CREATE TABLE syncer_peers (
-    peer_address INET PRIMARY KEY,
-    first_seen TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
+    ip_address INET PRIMARY KEY,
+    port INTEGER NOT NULL CHECK (port BETWEEN 1 AND 65535),
+    first_seen TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    last_connect TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    synced_blocks INTEGER NOT NULL DEFAULT 0 CHECK (synced_blocks >= 0),
+    sync_duration INTEGER NOT NULL DEFAULT 0 CHECK (sync_duration >= 0)
```
peterjan, line 7: I went back and forth on whether or not to suggest these updates but ended up going for it. `UpdatePeerInfo` updates this information and I feel it's kind of nice to have it. The `INET` column type in postgresql doesn't have the port so I ended up splitting it up in an ip_address and port... I think the only real alternative we have is going back to `TEXT`.

n8mgr: I keep these in memory in `hostd` because it causes a lot of database noise for very little user benefit. `last_connect` is probably the only one that is missing I would consider valuable. `synced_blocks` and `synced_duration` don't benefit from being persisted, imo.

persist/postgres/store.go
```diff
@@ -35,15 +35,15 @@ func (ci ConnectionInfo) String() string {
 	return fmt.Sprintf("host=%s port=%d user=%s password=%s dbname=%s sslmode=%s", ci.Host, ci.Port, ci.User, ci.Password, ci.Database, ci.SSLMode)
 }
 
-func (s *Store) transaction(ctx context.Context, fn func(context.Context, *txn) error) error {
+func (s *Store) transaction(ctx context.Context, fn func(*txn) error) error {
```
n8mgr: This is outside the scope of this PR and undesirable. It's more idiomatic to derive and use a scoped context that we pass into the callback rather than relying directly on the outer-scoped context. Makes it clearer that we should be defaulting to `QueryContext` and `ExecContext`. Pretty sure that cancelling the transaction's `context` doesn't interrupt running queries.

n8mgr (changes requested):

persist/postgres/peers_test.go
```diff
@@ -0,0 +1,89 @@
...
+		info.LastConnect = lastConnect
+		info.SyncedBlocks = syncedBlocks
+		info.SyncDuration = syncDuration
+	}); err != nil {
+		t.Fatal("unexpected error", err)
+	}
+
+	if info, err := db.PeerInfo(addr); err != nil {
+		t.Fatal("unexpected error", err)
+	} else if !info.LastConnect.Equal(lastConnect) {
```
n8mgr, line 51: Postgres only goes up to microseconds, should probably `.Truncate(time.Second)` for the comparison.

persist/postgres/peers.go
```diff
@@ -0,0 +1,193 @@
...
+// AddPeer adds a peer to the store. If the peer already exists, nil is
+// returned.
+func (s *Store) AddPeer(addr string) error {
+	host, port, err := net.SplitHostPort(addr)
+	if err != nil {
+		return fmt.Errorf("invalid peer address: %w", err)
+	}
+	return s.transaction(context.Background(), func(ctx context.Context, tx *txn) error {
+		const query = `INSERT INTO syncer_peers (ip_address, port, first_seen, last_connect, synced_blocks, sync_duration) VALUES ($1, $2, $3, '0001-01-01'::TIMESTAMP WITH TIME ZONE, 0, 0) ON CONFLICT (ip_address) DO NOTHING`
+		_, err := tx.Exec(query, host, port, time.Now())
```
n8mgr: Should probably enforce defaulting to `tx.*Context()` everywhere.

persist/postgres/types.go
```diff
@@ -191,3 +191,19 @@ func (sc *sqlCurrency) Scan(src any) error {
 		return fmt.Errorf("cannot scan %T to Currency", src)
 	}
 }
+
+type sqlDurationMS time.Duration
```
n8mgr, line 195: ~~This wrapper seems unnecessary since the only place the precision loss matters is in tests.~~

Ignore this one. I was under the impression we could pass `time.Duration` directly instead of casting it to an `int64` first. This is more clear.

persist/postgres/peers.go
```diff
@@ -0,0 +1,193 @@
...
+
+// AddPeer adds a peer to the store. If the peer already exists, nil is
+// returned.
+func (s *Store) AddPeer(addr string) error {
+	host, port, err := net.SplitHostPort(addr)
+	if err != nil {
+		return fmt.Errorf("invalid peer address: %w", err)
+	}
+	return s.transaction(context.Background(), func(ctx context.Context, tx *txn) error {
+		const query = `INSERT INTO syncer_peers (ip_address, port, first_seen, last_connect, synced_blocks, sync_duration) VALUES ($1, $2, $3, '0001-01-01'::TIMESTAMP WITH TIME ZONE, 0, 0) ON CONFLICT (ip_address) DO NOTHING`
```
n8mgr: You can use the built-in `NOW()` instead of passing in `time.Now()`. It might save a CPU cycle or two for the statement preparation.

persist/postgres/init.sql
peterjan: `LastConnect` complicates keeping peers in memory. We could of course persist the field if we find it's being set in `UpdatePeerInfo` or we could save peers in the database on shutdown or something. Did you make this change because someone reported it? Or because you noticed it became a problem? If not I would persist all peer info regardless to keep it simple.

In general I feel the `PeerStore` interface is a little awkward. For one, we don't require a context for `Peers` and `PeerInfo` which makes us pass a background context, which is sad. But also in `UpdatePeerInfo` we don't update all fields, we ignore the address, and now it turns out we don't want to update all fields and it makes us keep peers in memory.

I also removed the logging in the peer store. It should not be the responsibility of the store to log that. The caller, in this case the syncer, should log those events and it does. Not as much detail though but not sure if we need that to begin with? Don't mind reverting these changes, just wanted to propose this version because it's the most straight forward impl. of the peer store for me.

persist/postgres/peers.go
peterjan: Yes! Thanks for catching that. I will F/U with that because there's several and it's not in the scope of the PR.

persist/postgres/init.sql
n8mgr: The main reasoning was that the Syncer goroutines were locking the database excessively in `SQLite` during the initial sync. I felt like I had three options:
1. batch the updates which required me to store them in memory anyway
2. only persist fields that are used to reconnect to peers
3. Store nothing and always use the bootstrap list

Option three is *probably* untenable, so I chose option two. We *shouldn't* have as bad of lock contention problems, so I'm fine trying it out. Having said that, I still feel the information is not worth the IO or storage space 😅

PR #12, opened 2025-02-05
Switch to pgx postgres driver

n8mgr:
https://github.com/lib/pq is noted to be in maintenance mode and recommends a different Postgres driver https://github.com/jackc/pgx

Happens to require a context for all the query methods, so that's nice.

> This package is currently in maintenance mode, which means:
> 
> It generally does not accept new features.
> It does accept bug fixes and version compatability changes provided by the community.
> Maintainers usually do not resolve reported issues.
> Community members are encouraged to help each other with reported issues.
> For users that require new features or reliable resolution of reported bugs, we recommend using [pgx](https://github.com/jackc/pgx) which is under active development.


peterjan (approved):
nice

PR #13, opened 2025-02-10
Add test host

peterjan:


internal/testutils/host.go
```diff
@@ -0,0 +1,151 @@
...
+	)
+}
+
+// PublicKey returns the host's public key.
+func (h *Host) PublicKey() types.PublicKey {
+	return h.pk.PublicKey()
+}
+
+// NewHost creates a new host.
+func NewHost(pk types.PrivateKey, n *consensus.Network, genesis types.Block, log *zap.Logger) (*Host, error) {
```
ChrisSchinnerl: I think this would be nicer to use if we added a `testing.TB` as the first parameter instead of returning an error. As an external dependency we don't ever expect the creation of a host to fail in our tests.

Maybe we can even add our own `testing.TT` in this PR to make this even more concise. With the helpers we added such as `tt.OK` and `tt.OKAll` which were very convenient.

internal/testutils/host.go
```diff
@@ -0,0 +1,151 @@
...
+	return &Host{
+		c:       c,
+		s:       s,
+		ss:      ss,
+		sr:      sr,
+		w:       w,
+		pk:      pk,
+		address: rhp4Listener.Addr().String(),
+
+		closeFn: func() error {
```
ChrisSchinnerl: Instead of tracking a `closeFn` I think we can get away with using `t.Cleanup` in here. So that the host always gets cleaned up at the end of the test.

For the tests that require a host to be offline I'd explicitly track the `rhp4Listener`. If we want a host to be offline we can close the listener in a dedicated `Shutdown` function ahead of time but leave the remaining components running since they'll be closed by the `Cleanup`.

peterjan: I went back and forth on this a couple of times for the exact reason you mentioned. Ended up deciding that I preferred an explicit `Close` method over calls to `Cleanup` but reverted it so it matches your proposal.

internal/testutils/host.go
peterjan: It's convenient but I'm not the biggest fan of `testing.TT`

ChrisSchinnerl (dismissed):

internal/testutils/tt.go
```diff
@@ -0,0 +1,56 @@
+package testutils
+
+type (
+	// TT is a wrapper around testing.T that provides additional helper methods.
+	TT interface {
```
n8mgr: This wrapper feels unnecessary.

OK() is three lines and OKAll() is five. Most test logic won't benefit and the line savings are negligible.

internal/testutils/tt.go
```diff
@@ -0,0 +1,56 @@
...
+		TestingCommon
+
+		OK(err error)
+		OKAll(vs ...interface{})
+	}
+
+	// TestingCommon is an interface that describes the common methods of
+	// testing.T and testing.B ensuring this testutil can be used in both
+	// contexts.
+	TestingCommon interface {
```
n8mgr: Use `testing.TB`

internal/testutils/host.go
```diff
@@ -0,0 +1,130 @@
...
+	return h.l.Addr().String()
+}
+
+// PublicKey returns the host's public key.
+func (h *Host) PublicKey() types.PublicKey {
+	return h.pk.PublicKey()
+}
+
+// NewHost creates a new host.
+func NewHost(tt TT, pk types.PrivateKey, n *consensus.Network, genesis types.Block, log *zap.Logger) *Host {
```
n8mgr: testing.TB

n8mgr (changes requested):

ChrisSchinnerl (approved):

PR #14, opened 2025-02-10
Shutdown http server gracefully

ChrisSchinnerl:
Was going over the code while catching up and noticed that we are just killing the http servers underlying connection. This PR updates that to a more graceful shutdown that handles all ongoing connections while preventing new ones before exiting.  
Giving users the chance to handle shutdowns gracefully on production setups.

Without it, commands like `docker stop` will just drop any incoming connections even though they intend to nicely handle graceful shutdowns via a `SIGTERM` followed by a `SIGKILL` after a user-configurable timeout.

For convenience sending `SIGTERM` twice will still interrupt the shutdown immediately.

### DIFF
--- a/cmd/indexd/run.go
+++ b/cmd/indexd/run.go
@@ -123,7 +123,8 @@ func runRootCmd(ctx context.Context, cfg config.Config, log *zap.Logger) error {
 		Handler: webRouter{
 			api: jape.BasicAuth(cfg.HTTP.Password)(api.NewServer(cm, s, store, apiOpts...)),
 		},
-		ReadTimeout: 30 * time.Second,
+		ReadTimeout:  30 * time.Second,
+		WriteTimeout: 30 * time.Second,
 	}
 	defer web.Close()
 
--- a/cmd/indexd/run.go
+++ b/cmd/indexd/run.go
@@ -6,8 +6,11 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"os"
+	"os/signal"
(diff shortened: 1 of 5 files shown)


cmd/indexd/run.go
```diff
@@ -122,7 +125,6 @@ func runRootCmd(ctx context.Context, cfg config.Config, log *zap.Logger) error {
 		},
 		ReadTimeout: 30 * time.Second,
 	}
-	defer web.Close()
```
n8mgr, line 125: Leave the defer. It won't make much difference here, but good practice in case of a panic and consistent about expectations for Close methods.

internal/testutils/host_test.go
```diff
@@ -11,6 +11,10 @@ import (
 	"go.uber.org/zap"
 )
 
+func TestMain(m *testing.M) {
+	VerifyTestMain(m)
```
peterjan, line 15: 👀 neat

peterjan (approved):
I would've probably added a sane timeout because `web.Shutdown` will now hang indefinitely if something is keeping a rquest/connection open.

ChrisSchinnerl:
> I would've probably added a sane timeout because `web.Shutdown` will now hang indefinitely if something is keeping a rquest/connection open.

Yeah I'm intentionally leaving the interruption of a clean shutdown up to the user to avoid an opinionated approach but individual requests should eventually time out due to the timeouts on the `http.Server`.

PR #17, opened 2025-02-11
Add SingleAddressStore

peterjan:


persist/postgres/init.sql
```diff
@@ -15,6 +15,28 @@ CREATE TABLE syncer_bans (
...
 
+CREATE TABLE wallet_events (
+    id BYTEA PRIMARY KEY,
+    chain_index BYTEA NOT NULL,
+    maturity_height INTEGER NOT NULL,
+    confirmations INTEGER NOT NULL,
+    event_type TEXT NOT NULL,
+    event_data BYTEA NOT NULL,
+    event_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    relevant BYTEA
```
peterjan: Do we want to store relevant addresses? Same question for `confirmations`.

persist/postgres/encoding.go
```diff
@@ -0,0 +1,95 @@
...
+		types.EncodeSlice(e, obj)
+		e.Flush()
+		return buf.Bytes()
+	case types.EncoderTo:
+		var buf bytes.Buffer
+		e := types.NewEncoder(&buf)
+		obj.EncodeTo(e)
+		e.Flush()
+		return buf.Bytes()
+	case wallet.Event:
```
peterjan: can be removed if we end up updating `coreutils`

persist/postgres/encoding.go
```diff
@@ -0,0 +1,95 @@
...
+	default:
+		return fmt.Errorf("cannot scan %T to %T", src, d.v)
+	}
+}
+
+func decode(obj any) sql.Scanner {
+	return &decodable{obj}
+}
+
+func encode(obj any) any {
```
n8mgr: I’m in favor of keeping individual types. Encode is nice, but it’s bitten me a few times with columns.

persist/postgres/init.sql
n8mgr: All you need is:
- id
- Chain index
- Maturity height
- Event type
- Event data

We don’t want to persist confirmations since it needs to be calculated every block, there will ever only be one constant relevant address for the SAW, and timestamp is irrelevant.

n8mgr (changes requested):

persist/postgres/encoding.go
peterjan: Hm, so I definitely think that if something has bitten you before we should probably avoid it. It's so nice though, was it really nasty stuff or things you could've foreseen/prevented. It saves so much lines and handling `NULL` becomes trivial, although I can see how that could end up biting us at one point.

ChrisSchinnerl: Definitely prefer the types as well. Maybe the only thing that `renterd` has done better than `hostd` when it comes to DB code^^

peterjan: \* sad noises *

persist/postgres/types.go
```diff
@@ -1,140 +1,236 @@
...
+	sql.Scanner
+}
 
-	// A txn wraps a *sql.Tx, logging slow queries.
-	txn struct {
-		pgx.Tx
-		log *zap.Logger
+// ChainIndex is a wrapper around types.ChainIndex that implements the
+// scannerValuer interface.
+type ChainIndex types.ChainIndex
```
n8mgr: We shouldn't export these. We should convert them to and from their base types, like we were originally doing with `sqlDuration`. Having a bunch of duplicate wrapping types floating around is no fun.

persist/postgres/init.sql
```diff
@@ -15,6 +15,25 @@ CREATE TABLE syncer_bans (
 CREATE INDEX syncer_bans_expiration_idx ON syncer_bans (expiration);
 CREATE INDEX syncer_bans_net_cidr_idx ON syncer_bans USING gist (net_cidr inet_ops); -- fast subnet matches
```
ChrisSchinnerl, line 17: Left some comments regarding the schema on the RFC

persist/postgres/rows.go
```diff
@@ -0,0 +1,104 @@
...
+	"context"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgconn"
+	"go.uber.org/zap"
+)
+
+const (
+	longQueryDuration = 10 * time.Millisecond
```
ChrisSchinnerl: Damn that is ambitious. I don't mind but I do predict that we will update this in the future after we got annoyed enough on our test nodes ^^

persist/postgres/rows.go
ChrisSchinnerl, line 1: I'd probably go for `postgres.go` or `sql.go` here as a place that includes all the wrappers for the database. `rows.go` doesn't seem to do it justice.

Could you also make this move in a separate PR that we can insta merge? That way we clear this diff up quite a bit.

persist/postgres/rows.go
```diff
@@ -0,0 +1,104 @@
...
+// any placeholder parameters in the query.
+func (tx *txn) Exec(ctx context.Context, query string, args ...any) (pgconn.CommandTag, error) {
+	start := time.Now()
+	result, err := tx.Tx.Exec(ctx, query, args...)
+	if dur := time.Since(start); dur > longQueryDuration {
+		tx.log.Debug("slow exec", zap.String("query", query), zap.Duration("elapsed", dur), zap.Stack("stack"))
+	}
+	return result, err
+}
+
```
ChrisSchinnerl: Surprised to see no `Prepare` here

peterjan: Adding it in next PR. We don't have any prepared statements yet. Prepared statements work differently in postgres, I'm not sure it makes sense to wrap it and log something like a "slow prepare".. we'll never see it.

persist/postgres/init.sql
```diff
@@ -15,6 +15,27 @@ CREATE TABLE syncer_bans (
...
 
+CREATE TABLE wallet_events (
+    id SERIAL PRIMARY KEY,
+    chain_index BYTEA NOT NULL,
+    maturity_height INTEGER NOT NULL,
```
n8mgr, line 21: Go ahead and add the index on `maturity_height DESC, id DESC`. We know the query will use it.

persist/postgres/init.sql
```diff
@@ -15,6 +15,27 @@ CREATE TABLE syncer_bans (
...
+CREATE TABLE wallet_siacoin_elements (
+    id SERIAL PRIMARY KEY,
+    output_id BYTEA UNIQUE NOT NULL,
+    value NUMERIC(50,0) NOT NULL,
+    address BYTEA NOT NULL,
+    merkle_proof BYTEA NOT NULL,
+    leaf_index INTEGER NOT NULL,
+    maturity_height INTEGER NOT NULL
+);
+CREATE INDEX wallet_siacoin_elements_output_id_idx ON wallet_siacoin_elements(output_id);
```
n8mgr: UNIQUE columns get an auto index

persist/postgres/wallet.go
```diff
@@ -0,0 +1,115 @@
...
+
+// WalletEvents returns a paginated list of transactions ordered by maturity
+// height, descending. If no more transactions are available, (nil, nil) should
+// be returned.
+func (s *Store) WalletEvents(offset, limit int) (events []wallet.Event, err error) {
+	if limit == 0 || limit == -1 {
+		limit = math.MaxInt64
+	}
+	err = s.transaction(context.Background(), func(ctx context.Context, tx *txn) error {
+		rows, err := tx.Query(ctx, `SELECT chain_index, maturity_height, event_id, event_type, event_data FROM wallet_events ORDER BY maturity_height DESC LIMIT $1 OFFSET $2`, limit, offset)
```
n8mgr: This should be `ORDER BY maturity_height DESC, id DESC` since insertion order is the reverse of display order if events have the same maturity height.

persist/postgres/init.sql
peterjan: I dreamt about this mistake.

persist/postgres/wallet.go
```diff
@@ -0,0 +1,115 @@
...
+		return err
+	})
+	return
+}
+
+func scanEvent(rows scanner) (event wallet.Event, _ error) {
+	var buf []byte
+	err := rows.Scan((*sqlHash256)(&event.ID), (*sqlChainIndex)(&event.Index), &event.MaturityHeight, &event.Type, &buf)
+	if err != nil {
+		return
```
ChrisSchinnerl: This is a very good example of why we shouldn't use named return vars unless it actually increases the safety of our code by allowing us to defer handle an error.

Hint: https://play.golang.com/p/GYKo_4FURtO

persist/postgres/wallet.go
```diff
@@ -0,0 +1,115 @@
...
+// WalletEventCount returns the total number of events relevant to the wallet.
+func (s *Store) WalletEventCount() (count uint64, err error) {
+	err = s.transaction(context.Background(), func(ctx context.Context, tx *txn) error {
+		err := tx.QueryRow(ctx, `SELECT COUNT(*) FROM wallet_events`).Scan(&count)
+		return err
+	})
+	return
+}
+
+func scanEvent(rows scanner) (event wallet.Event, _ error) {
```
ChrisSchinnerl: unused function?

persist/postgres/wallet.go
```diff
@@ -0,0 +1,115 @@
...
+		e.DecodeFrom(dec)
+		event.Data = e
+	case wallet.EventTypeV2Transaction:
+		var e wallet.EventV2Transaction
+		e.DecodeFrom(dec)
+		event.Data = e
+	default:
+		return wallet.Event{}, fmt.Errorf("unknown event type %v", event.Type)
+	}
+	return
```
ChrisSchinnerl: We aren't checking `dec.Err()`.

ChrisSchinnerl (changes requested):

persist/postgres/types.go
```diff
@@ -1,42 +1,208 @@
...
+	case wallet.EventV1Transaction:
+		data.EncodeTo(e)
+	case wallet.EventV1ContractResolution:
+		data.EncodeTo(e)
+	case wallet.EventV2ContractResolution:
+		data.EncodeTo(e)
+	case wallet.EventV2Transaction:
+		types.V2Transaction(data).EncodeTo(e)
+	default:
+		return nil, fmt.Errorf("unknown event type: %s", event.eventType) // should never happen
```
n8mgr: Better to panic when something absolutely should not be possible

persist/postgres/types.go
```diff
@@ -1,42 +1,208 @@
...
+		case wallet.EventTypeV1Transaction:
+			var e wallet.EventV1Transaction
+			e.DecodeFrom(dec)
+			*ed.data = e
+		case wallet.EventTypeV2Transaction:
+			var e wallet.EventV2Transaction
+			e.DecodeFrom(dec)
+			*ed.data = e
+		default:
+			return fmt.Errorf("unknown event type %v", ed.eventType)
```
n8mgr: Ditto

persist/postgres/wallet_test.go
```diff
@@ -0,0 +1,157 @@
...
+)
+
+func TestSingleAddressWalletStore(t *testing.T) {
+	store := initPostgres(t, zaptest.NewLogger(t).Named("postgres"))
+
+	for _, m := range []struct {
+		method string
+		test   func(*Store) func(*testing.T)
+	}{
+		{"Tip", testSingleAddressWalletStoreTip},
```
n8mgr: I don't really like this structure over individual test functions since we added individual functions anyway. The single line savings don't seem significant enough to warrant it.

n8mgr (changes requested):

persist/postgres/wallet.go
```diff
@@ -0,0 +1,78 @@
...
+	})
+	return
+}
+
+// WalletEvents returns a paginated list of transactions ordered by maturity
+// height, descending. If no more transactions are available, (nil, nil) should
+// be returned.
+func (s *Store) WalletEvents(offset, limit int) (events []wallet.Event, err error) {
+	if limit == 0 || limit == -1 {
+		limit = math.MaxInt64
```
n8mgr: I don't think we should ever allow returning all transactions. Set a sane limit, like 100.

persist/postgres/wallet.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+		defer rows.Close()
+
+		for rows.Next() {
+			var event wallet.Event
+			err := rows.Scan((*sqlChainIndex)(&event.Index), &event.MaturityHeight, (*sqlHash256)(&event.ID), &event.Type, sqlDecodeEvent(&event.Data))
+			if err != nil {
+				return fmt.Errorf("failed to scan wallet event: %w", err)
+			}
+			events = append(events, event)
```
n8mgr, line 86: Need to get current scan height and set event.Confirmations

n8mgr (changes requested):

persist/postgres/types.go
peterjan: This was done on purpose actually. Don't get me wrong, I would usually make the exact same review comment but I was thinking about this and I wonder if `panic(x) // developer error` really makes a lot of sense. I personally have never ran into one for example, like it never helped me to add an extra case that I forgot to add. It also causes panic fatigue or at least we're used to seeing an occasional panic in our code here and there. The reason we do add it is to catch these developer errors before they reach a release but what if we don't, then we ship a panic and for what.. because some exotic case isn't covered in some switch statement. I'm not sure if that tradeoff still makes sense. I'm probably wrong though and the answer to all of the above questions is still a resounding "yes, we do"... but more and more I think I'm favouring just not panics ever with maybe the one exception being database corruption.

persist/postgres/wallet.go
peterjan: Fair enough, removed it entirely, going to add a parse + validation in the API handlers when we expose the wallet endpoints.

ChrisSchinnerl (dismissed):

persist/postgres/types.go
ChrisSchinnerl: I don't mind either way but I think we are going to panic in the chain subscriber code anyway when we can't apply a change so returning an error here might help us keep our  panics localized to these critical spots.

peterjan: In a perfect world you grep the codebase for `panic` and exclude tests and it should show you exactly the exception cases that we would list in an engineering guideline.

persist/postgres/wallet.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+		defer rows.Close()
+
+		for rows.Next() {
+			var event wallet.Event
+			err := rows.Scan((*sqlChainIndex)(&event.Index), &event.MaturityHeight, (*sqlHash256)(&event.ID), &event.Type, sqlDecodeEvent(&event.Data))
+			if err != nil {
+				return fmt.Errorf("failed to scan wallet event: %w", err)
+			}
+			if ci.Height > event.Index.Height {
```
n8mgr:
```
if ci.Height >= event.Index.Height {
 	event.Confirmations = 1 + ci.Height - event.Index.Height
}
```

If the event has been included in a block, I would consider that one confirmation. Would it be safe to always assume `ci.Height >= event.Index.Height` since the event would be deleted in a revert.

persist/postgres/wallet.go
```diff
@@ -0,0 +1,212 @@
...
+	// sanity check input
+	if offset < 0 {
+		return nil, fmt.Errorf("offset must be positive")
+	} else if limit < 0 {
+		return nil, fmt.Errorf("limit must be positive")
+	} else if limit == 0 {
+		return nil, nil
+	}
+
+	ci, err := s.Tip()
```
n8mgr: This should be moved into the transaction

persist/postgres/wallet.go
peterjan: Crazy, I was actually under the impression I was just following how confirmations generally work. Turns out the first confirmation is the block in which the transaction got mined, I never knew. I think it's indeed safe to assume the height is always bigger but I kept the check, it doesn't hurt to keep it as a sanity check and avoid the underflow at all costs?

persist/postgres/types.go
```diff
@@ -1,42 +1,207 @@
...
 	case int64:
-		*sd = sqlDurationMS(time.Duration(src) * time.Millisecond)
+		*d = sqlDurationMS(time.Duration(src) * time.Millisecond)
 		return nil
 	default:
 		return fmt.Errorf("cannot scan %T to Duration", src)
 	}
 }
+
+type sqlEventData struct {
```
n8mgr, line 89: nit: There's a bit of benefit when scanning by duplicating the event type here, but I'd lean towards using the existing column over storing the event type twice even though it's slightly less clean.

n8mgr (approved):

persist/postgres/types.go
peterjan: I would normally just go and implement your suggestion but since you approved it and it's a `nit` I kind of want to keep it as is 😛 I coded up the alternative and it felt like taking a step back. If you regret making it a `nit` feel free to unresolve and I'll update it.

ChrisSchinnerl (approved):

PR #18, opened 2025-02-11
Add Indexer type for testing

ChrisSchinnerl:
Adds an `Indexer` helper type for integration testing which connects to Postgres. Each test getting its own database to avoid conflicts and conveniently debug failed tests.


internal/testutils/indexer.go
```diff
@@ -0,0 +1,138 @@
...
+
+// Client returns a ready-to-go API client for the indexer.
+func (i *Indexer) Client() *api.Client {
+	return i.client
+}
+
+// NewIndexer creates a new indexer for testing. It returns a cleanup function
+// that closes all of its resources and causes the test to fail if any of them
+// fail to close.
+func NewIndexer(t testing.TB, n *consensus.Network, genesis types.Block, log *zap.Logger) (*Indexer, func()) {
```
peterjan: Do we really want to go for this hybrid approach where we return a cleanup function the caller has to defer? When I added the test host I figured there's only two options, either you wrap things in a struct and define a `Close` method to house the close/shutdown functions or you pass `testing.TB` and call `t.Cleanup` to ensure all dependencies get closed properly. This hybrid approach feels like the worst of both worlds? We're passing `testing.TB` anyway so I don't see a good reason to return a cleanup function.

internal/testutils/indexer.go
```diff
@@ -0,0 +1,138 @@
...
+	"go.sia.tech/jape"
+	"go.uber.org/zap"
+	"lukechampine.com/frand"
+)
+
+// Indexer is a test utility combining an indexer, an http client for the
+// indexer and useful helpers for testing.
+type Indexer struct {
+	client   *api.Client
+	closeFns []func() error
```
peterjan: Not used, if this field is gone we're essentially wrapping only the client, so should there be an `Indexer` type at all? This is exactly what I went through in my PR, defining this struct is the `renterd` approach while using `t.Cleanup` and have the caller close the dependencies and treat the server/client as the indexer is the `hostd` approach.

internal/testutils/indexer.go
```diff
@@ -0,0 +1,138 @@
...
+	// peers will reject us if our hostname is empty or unspecified, so use loopback
+	s := syncer.New(syncerListener, cm, testutil.NewEphemeralPeerStore(), gateway.Header{
+		GenesisID:  genesis.ID(),
+		UniqueID:   gateway.GenerateUniqueID(),
+		NetAddress: syncerListener.Addr().String(),
+	},
+		syncer.WithSendBlocksTimeout(2*time.Second),
+		syncer.WithRPCTimeout(2*time.Second),
+	)
+	t.Cleanup(func() { s.Close() })
```
peterjan: this is a leftover I guess because the `cleanup` function does the same?

internal/testutils/indexer.go
```diff
@@ -0,0 +1,138 @@
...
+
+	apiOpts := []api.ServerOption{
+		api.WithLogger(log.Named("api")),
+	}
+
+	// prepare store
+	store := initTestDB(t, log)
+
+	password := hex.EncodeToString(frand.Bytes(16))
+	web := http.Server{
```
peterjan, line 67: doesn't hurt setting read/write timeout here as well, I wonder if it makes sense to deduplicate this code and extract it into `node.go` that wraps creating the chain manager, syncer, server and handles the graceful shutdown

internal/testutils/indexer.go
```diff
@@ -0,0 +1,138 @@
...
+			}
+			if err := store.Close(); err != nil {
+				t.Errorf("failed to close store: %v", err)
+			}
+		}
+}
+
+func initTestDB(t testing.TB, log *zap.Logger) *postgres.Store {
+	// parse connection info from env vars
+	ci := postgres.ConnectionInfo{
```
peterjan, line 100: little sad we're duplicating this between here and `stores_test.go`

api/client.go
```diff
@@ -0,0 +1,24 @@
+package api
+
+import (
+	"go.sia.tech/jape"
+)
+
+// A Client provides methods for interacting with an autopilot.
```
peterjan: copy pasta

api/client.go
```diff
@@ -0,0 +1,24 @@
...
+type Client struct {
+	c jape.Client
+}
+
+// NewClient returns a new autopilot client.
```
peterjan: copy pasta

api/client.go
```diff
@@ -0,0 +1,24 @@
...
+func NewClient(addr, password string) *Client {
+	return &Client{jape.Client{
+		BaseURL:  addr,
+		Password: password,
+	}}
+}
+
+// State returns the current state of the autopilot.
```
peterjan: copy pasta

internal/testutils/indexer.go
n8mgr: I think you should stick with `t.Cleanup()` here instead of returning a close func that will be deferred. That simplifies the struct, keeps the usage consistent with other test types and you can still panic if one of the closes takes more than 30s or returns an error.

Something like:

```go
t.Cleanup(func() {
  closed := make(chan struct{}, 1)
  time.AfterFunc(30 * time.Second, func() { 
    select {
    case <- closed:
    default:
      panic("chain manager did not close")
    }
  })
  if err := cm.Close(); err != nil {
    panic(err)
  }
})
```

You could probably distill it into a helper that takes an `io.Closer` to reduce duplication.

peterjan (commented):
(sorry forgot to submit the review yesterday)

internal/testutils/indexer.go
ChrisSchinnerl: Yeah not sure how to deduplicate that. I thought about using that init function first even though I don't like the cleanup of the database but it gets circular pretty fast. So maybe it's fine to just have both for now.

internal/testutils/indexer.go
ChrisSchinnerl: I think prod and test setups are different enough to warrant separate functions. e.g. all the signal handling is stuff we don't need in testing and I'd argue that in testing we also want to be less forgiving overall when it comes to unblocking the shutdown. e.g. if something blocks the shutdown we want the test to time out and print a stack trace or panic ourselves but not necessarily hide it.

internal/testutils/indexer.go
ChrisSchinnerl: Right now that's true but I think pretty soon we are going to add test helpers to this type. Sure they could probably be functions as well but having something like an `indexer.Restart()` might be inevitable.

internal/testutils/indexer.go
ChrisSchinnerl: The main reason for this is tests that require closing the indexer. We had a few of those in the past. Although it's only a few of them. What do you think of having two functions. 
1. `NewIndexer` -> cleans up after itself
2. `NewIndexerNoCleanup` -> used for those specific tests

internal/testutils/indexer.go
```diff
@@ -0,0 +1,168 @@
...
+
+func closeWithTimeout(c io.Closer) error {
+	closed := make(chan struct{})
+	defer close(closed)
+
+	time.AfterFunc(30*time.Second, func() {
+		select {
+		case <-closed:
+		default:
+			panic("indexer ")
```
peterjan: incomplete message

internal/testutils/indexer.go
```diff
@@ -0,0 +1,168 @@
...
+	go func() {
+		if err := web.Serve(httpListener); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			log.Error("http server failed", zap.Error(err))
+		}
+	}()
+
+	return &Indexer{
+			Client: api.NewClient(fmt.Sprintf("http://%s", httpListener.Addr().String()), password),
+		}, func() {
+			if err := closeWithTimeout(shutdownCloser{&web}); err != nil {
```
peterjan: This is sufficient and slightly more correct imo. The cleanup function should take a context and I would just inline this and avoid adding something like "the shutdowner".

```go
func(ctx context.Context) {
    doneChan := make(chan struct{})
    defer close(doneChan)

    go func() {
        select {
            case <-doneChan:
            case <-ctx.Done():
                t.Error("timed out")
        }
    }()

    if err := errors.Join(
        web.Shutdown(ctx)
        s.Close()
        store.Close()
    ); err != nil {
        t.Error(err)
    }
}
```

peterjan (dismissed):
Aside from the remark I left this LGTM. I'd probably get rid of the `Indexer` type and return the client until we get to the point we actually have to wrap it.

internal/testutils/indexer.go
ChrisSchinnerl: Problem with this is that
1. you pollute the call sites with unnecessary contexts
2. we purposefully don't want to interrupt the webserver shutdown on `ctx` 
3. you don't get the stack trace from the panic which tells you which shutdown handler is responsible

As a compromise I refactored it to pass the close function rather than using the `io.Closer` interface. That makes it a bit less awkward.

internal/testutils/indexer.go
n8mgr: If we are specifically testing closing, don't use the wrapper is my opinion. Especially if it's only a handful of tests.

internal/testutils/indexer.go
```diff
@@ -0,0 +1,168 @@
...
+	indexer, cleanup := NewIndexerNoCleanup(t, n, genesis, log)
+	t.Cleanup(cleanup)
+	return indexer
+}
+
+// NewIndexerNoCleanup creates a new indexer for testing. It returns a cleanup
+// function that closes all of its resources and causes the test to fail if any
+// of them fail to close. Useful for tests that require closing the indexer
+// ahead of time.
+func NewIndexerNoCleanup(t testing.TB, n *consensus.Network, genesis types.Block, log *zap.Logger) (*Indexer, func()) {
```
n8mgr: Keeping this around because we might need it later is not worth it. I feel that way about a lot of the code we're effectively copying from the other repos. Add it when we need it, try not to anticipate -- there may be a better solution.

peterjan (approved):

n8mgr (approved):

PR #20, opened 2025-02-11
Add schema RFC

peterjan:
Kicking things off with an RFC for the indexer schema. I purposefully kept it short, wanted to get a round of reviews in first to check whether this was more or less what we had in mind.

### DIFF
--- a/doc/rfc/schema.md
+++ b/doc/rfc/schema.md
@@ -40,26 +40,25 @@ Notable columns/types:
 
 ```postgresql
 CREATE TABLE wallet_events (
-    id BYTEA PRIMARY KEY,
+    id SERIAL PRIMARY KEY,
     chain_index BYTEA NOT NULL,
     maturity_height INTEGER NOT NULL,
-    confirmations INTEGER NOT NULL,
+    event_id BYTEA UNIQUE NOT NULL,
     event_type TEXT NOT NULL,
-    event_data BYTEA NOT NULL,
-    event_time TIMESTAMP WITH TIME ZONE NOT NULL,
-    relevant BYTEA
+    event_data BYTEA NOT NULL
 );
-CREATE INDEX wallet_events_chain_index ON wallet_events(chain_index);
-CREATE INDEX wallet_events_maturity_height ON wallet_events(maturity_height DESC);
+CREATE INDEX wallet_events_chain_index_idx ON wallet_events(chain_index);
 
 CREATE TABLE wallet_siacoin_elements (
-    id BYTEA PRIMARY KEY,
+    id SERIAL PRIMARY KEY,
+    output_id BYTEA UNIQUE NOT NULL,
     value NUMERIC(50,0) NOT NULL,
     address BYTEA NOT NULL,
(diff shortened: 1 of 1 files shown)


doc/rfc/schema.md
```diff
@@ -0,0 +1,90 @@
...
+The wallet stores events as well as the state of the wallet in the form of its
+outputs. There's several types of events, the raw event data is stored in the
+`event_data` column, which is a `BYTEA` since it doesn't have to be searchable.
+
+Notable columns/types:
+- **`NUMERIC(50,0)` for `value`**: Allows precise representation of siacoins.
+
+```postgresql
+CREATE TABLE wallet_events (
+    id BYTEA PRIMARY KEY,
```
ChrisSchinnerl: It's usually best practice to keep the primary key as small as possible and to not use "natural keys" of your use-case. So I'd honestly make it a rule to default to an auto incremented `SERIAL`/`BIGSERIAL` (depending on number of rows we expect) + a unique constraint on the natural key.
It's especially useful for large join operations since it significantly reduces the amount of data to be read from disk when using e.g. a 4 byte `SERIAL` or 8 byte `BIGSERIAL` compared to a 1+32 byte hash. So join operations are potentially quite a bit faster.

doc/rfc/schema.md
```diff
@@ -0,0 +1,90 @@
...
+    chain_index BYTEA NOT NULL,
+    maturity_height INTEGER NOT NULL,
+    event_type TEXT NOT NULL,
+    event_data BYTEA NOT NULL
+);
+CREATE INDEX wallet_events_chain_index ON wallet_events(chain_index);
+CREATE INDEX wallet_events_maturity_height ON wallet_events(maturity_height DESC);
```
ChrisSchinnerl: One thing I feel we haven't been great about is managing indices in `renterd`. We just added them willy-nilly without any benchmarking and might have had more indices than we even needed.

To avoid being too wasteful this time around I'd prefer it if we could introduce indices alongside queries using them. e.g. we might never want an index on the maturity height since we really need a composite index on let's say the event id + the maturity height.  But to determine that we really need to see the query that index is for.

doc/rfc/schema.md
```diff
@@ -0,0 +1,90 @@
...
+
+### 2.2 Syncer
+
+The syncer connects to other peers and syncs the Sia blockchain. It keeps a list
+of peers alongside some information per peer, as well as a list of peers it
+banned.
+
+Notable columns/types:
+- **`INET` for `ip_address`**: Efficient storage and validation of IP addresses.
+- **`CIDR` for `net_cidr`**: Allows subnet-level banning and optimized lookups.
```
ChrisSchinnerl, line 72: Not gonna lie this is pretty awesome. We will probably make good use of that in our slab health code

doc/rfc/schema.md
peterjan: Removed `wallet_events_maturity_height` for now.

n8mgr (approved):

PR #23, opened 2025-02-12
Implement `wallet.UpdateTx` in Store

peterjan:


persist/postgres/sql.go
```diff
@@ -69,22 +69,32 @@ func (r *row) Scan(dest ...any) error {
...
+	result, err := tx.Tx.Exec(ctx, sql, args...)
 	if dur := time.Since(start); dur > longQueryDuration {
-		tx.log.Debug("slow exec", zap.String("query", query), zap.Duration("elapsed", dur), zap.Stack("stack"))
+		tx.log.Debug("slow exec", zap.String("sql", sql), zap.Duration("elapsed", dur), zap.Stack("stack"))
 	}
 	return result, err
 }
 
+// Prepare creates a prepared statement for later queries or executions.
+func (tx *txn) Prepare(ctx context.Context, name, sql string) (*pgconn.StatementDescription, error) {
```
n8mgr: Pgx recommends not using prepared statements because of its built-in caching.

persist/postgres/chain.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"context"
+
+	"go.sia.tech/coreutils/wallet"
+	"go.uber.org/zap"
+)
+
+// UpdateTx defines the interface for updating the state when processing chain
+// updates.
+type UpdateTx interface {
```
n8mgr: Shouldn’t need to wrap the interface here. I get where this is coming from, but this should probably be in another package.

persist/postgres/sql.go
```diff
@@ -69,22 +69,32 @@ func (r *row) Scan(dest ...any) error {
...
+	if dur := time.Since(start); dur > longQueryDuration {
+		tx.log.Debug("slow prepare", zap.String("sql", sql), zap.Duration("elapsed", dur), zap.Stack("stack"))
+	}
+	return desc, err
+}
+
 // Query executes a query that returns rows, typically a SELECT. The
 // args are for any placeholder parameters in the query.
-func (tx *txn) Query(ctx context.Context, query string, args ...any) (*rows, error) {
+func (tx *txn) Query(ctx context.Context, sql string, args ...any) (*rows, error) {
```
n8mgr: All these are outside the scope of this pr and I don’t really see a huge benefit to changing it. Maybe in a follow up so it doesn’t mess with the diff

n8mgr (changes requested):

persist/postgres/sql.go
ChrisSchinnerl: Is that like GORM's caching of prepared statements? We ended up turning it off because it didn't work for statements created via string manipulation. e.g. when you conditionally wanted to apply a filter it would use the cached query from before and throw random "invalid number of arguments" errors. 

I'd still give Pgx's caching a shot but definitely something to keep in mind once we get to that point.

persist/postgres/chain.go
ChrisSchinnerl: +1 on that. Usually better to keep the interface in the place where we consume it or a shared location if needed rather than where it is implemented.

persist/postgres/wallet.go
```diff
@@ -77,6 +86,143 @@ func (s *Store) WalletEventCount() (count uint64, err error) {
...
+		if _, err := u.tx.Exec(u.ctx, stmt, se.LeafIndex, sqlMerkleProof(se.MerkleProof), id); err != nil {
+			return fmt.Errorf("failed to update siacoin element: %w", err)
+		}
+	}
+	return nil
+}
+
+func (u *updateTx) WalletApplyIndex(index types.ChainIndex, created, spent []types.SiacoinElement, events []wallet.Event, timestamp time.Time) (err error) {
+	log := u.log.With(zap.Uint64("height", index.Height), zap.Stringer("id", index.ID))
+	defer func() {
```
ChrisSchinnerl: I wouldn't defer logging like this. The function returns an error so we can log that error where we call it.
That achieves the same thing without paying the price of a defer and the potentially increased surface for errors caused by a named return var.

ChrisSchinnerl (changes requested):

persist/postgres/sql.go
peterjan: I reverted it. I changed it because `pgx` asks for `sql` and we just renamed it to `query` for no reason.

persist/postgres/sql.go
n8mgr: > We ended up turning it off because it didn't work for statements created via string manipulation

Unless I'm misunderstanding what was actually going on, that sounds like something we shouldn't be doing regardless.

n8mgr (approved):

persist/postgres/wallet.go
```diff
@@ -44,6 +52,15 @@ func (s *Store) UnspentSiacoinElements() (sces []types.SiacoinElement, err error
 // height, descending. If no more transactions are available, (nil, nil) should
 // be returned.
 func (s *Store) WalletEvents(offset, limit int) (events []wallet.Event, err error) {
+	// sanity check input
```
peterjan, line 55: This will move or be updated when I expose the wallet in the API and figure out how to parse these parameters in the API handlers. Figured the `Store` should probably sanity check the input regardless of whether we already validate this higher up, not sure yet.

ChrisSchinnerl: I'd honestly argue that they should be evaluated at the lowest point of the call stack. 
Because if we do it in the API, we just end up adding it here too to be safe.

But if we add it here and maybe decorate it with a `var ErrBadInput = errors.New("bad inputs")`, we can check for that error in the API handler and turn it into the corresponding status code. Hopefully avoiding a lot of duplicate code in the process.

ChrisSchinnerl (approved):

persist/postgres/wallet.go
peterjan: Yeah that's the plan for now. I'll turn these into `ErrWhatever` and catch and 400 like we usually do.

PR #24, opened 2025-02-12
Add cluster type for testing

### LINKED ISSUE #1: Setup Repository
Set up the basics

- [x] go module
- [x] vetting and linting (match hostd's, I think renterd is missing comment linting)
- [x] common ci workflows
- [x] knope
- [x] cli parsing
- [x] yaml parsing
- [x] Dockerfile
- [x] http server startup and shutdown
- [x] connection to postgresql and shutdown
- [x] run test suite on CI with postgresql
- [x] Indexer type that holds the basics such as a chain manager, syncer and a store that connects to a database
- [x] `Host` type for integration tests similar to `internal/test/e2e/host.go` in renterd but rhp4 and ephemeral types only
- [ ] `Cluster` type for spinning up an indexer and multiple hosts, syncing them and funding them

ChrisSchinnerl:
Add a cluster skeleton type for testing. The idea is to extend this as we go. For now it only creates a few hosts, funds them, syncs up and closes itself.

Closes #1


n8mgr:
My recommendation is to think about this when we actually need it not anticipate the need ahead of time.

ChrisSchinnerl:
> My recommendation is to think about this when we actually need it not anticipate the need ahead of time.

We can hold off on merging it until we need it for testing contract formation, uploads or whatever but overall it's really just the basics. Creating the indexer, hosts and connecting them while mining a few blocks for money.

internal/testutils/cluster.go
```diff
@@ -0,0 +1,122 @@
...
+// WithLogger allows for attaching a custom logger to the cluster for debugging
+// if necessary
+func WithLogger(logger *zap.Logger) ClusterOpt {
+	return func(cfg *clusterOpts) {
+		cfg.logger = logger
+	}
+}
+
+// WithNHosts allows for overriding the default number of hosts in the cluster
+func WithNHosts(n int) ClusterOpt {
```
peterjan: `nit` `WithHosts`

internal/testutils/cluster.go
```diff
@@ -0,0 +1,122 @@
...
+// NewCluster creates a cluster for testing. A cluster contains an indexer and
+// multiple hosts.
+func NewCluster(t testing.TB, opts ...ClusterOpt) *Cluster {
+	cfg := defaultClusterOpts
+	for _, opt := range opts {
+		opt(&cfg)
+	}
+
+	n, genesis := testutil.V2Network()
+	indexer := NewIndexer(t, n, genesis, zap.NewNop())
```
peterjan: Should probably pass a child logger? `cfg.logger.Named("indexer")`. Maybe `ClusterOpts` needs indexer options.

internal/testutils/cluster_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+func TestNewCluster(t *testing.T) {
+	cluster := NewCluster(t)
+
+	// assert blocks were mined
+	indexer := cluster.Indexer
+	state, err := indexer.TipState()
+	if err != nil {
+		t.Fatal(err)
+	} else if state.Index.Height == 0 {
```
peterjan: we can assert we're passed the allowheight here

internal/testutils/cluster.go
```diff
@@ -0,0 +1,122 @@
...
+	for _, h := range hosts {
+		indexer.MineBlocks(t, h.w.Address(), 1)
+	}
+
+	// mine more blocks to get outputs to mature
+	indexer.MineBlocks(t, types.Address{}, int(n.MaturityDelay))
+
+	// TODO: implement as needed
+	// - add volumes to hosts
+	// - announce hosts
```
peterjan, line 89: I feel this PR should include getting a host discoverable and ready to form a contract with.

internal/testutils/cluster.go
```diff
@@ -0,0 +1,122 @@
...
+	// - wait for contracts
+	// - sync cluster
+
+	Retry(t, 100, 100*time.Millisecond, func() error {
```
peterjan, line 93: Do we really want to reintroduce `Retry` so early on? I have mixed feelings about `Retry`, on the one hand it's  to be expected you need something like a retry simply because of all the goroutines and async processing of things, but on the other hand I feel we could rewrite most testing infrastructure in a way we can avoid it for quite a long time.

There's also things like https://github.com/jonboulle/clockwork which we could consider. I think we had another one but I couldn't immediately find the library. We brought it up couple of months ago but blanking on the name (or on the thread where we discussed it).

internal/testutils/cluster.go
```diff
@@ -0,0 +1,122 @@
...
+		hosts = append(hosts, h)
+	}
+
+	// fund hosts with one block each
+	for _, h := range hosts {
+		indexer.MineBlocks(t, h.w.Address(), 1)
+	}
+
+	// mine more blocks to get outputs to mature
+	indexer.MineBlocks(t, types.Address{}, int(n.MaturityDelay))
```
peterjan, line 85: `MineBlocks` looks very misplaced on the indexer

internal/testutils/cluster.go
ChrisSchinnerl: Maybe you are thinking of https://danp.net/posts/synctest-experiment/ ?
We can definitely experiment with it once released this month but for now I think the least NDF prone solution that still allows our tests to pass quickly is a simple retry.

internal/testutils/cluster.go
ChrisSchinnerl: Didn't want to go too far. Host announcing can be added once we actually scan for hosts on chain.

api/api.go
```diff
@@ -39,6 +42,7 @@ func NewServer(chain ChainManager, syncer Syncer, store Store, opts ...ServerOpt
 	}
 
 	return jape.Mux(map[string]jape.Handler{
-		"GET /state": a.handleGETState,
+		"GET /state":         a.handleGETState,
+		"GET /consensus/tip": a.handleGETConsensusTip,
```
peterjan: It should be `/consensus/tipstate` because we're returning `.TipState()`. Having both `Tip` and `TipState` always somewhat bothered me, even though I understand why that is the case.

Do we really have to add this route though? I see why you added it, you want to compare the height with the allow height in a test but  eventually we'll want to return the height up until which the indexer is synced (I usually call this the subscribe-height). It's a pretty big commitment to make a pretty lame assertion in a test right? If we have to add it maybe we can do `/consensus/state` instead so it exactly matches the type that's being returned.

We store the `last_synced_index` on the `global_settings` table, so maybe we can use that to our advantage here and expose it through the `/state` endpoint (an alternative would be `/settings` but I like that less)... That way we can still assert the cluster mined up until the allow height (providing we expose the network on the network) and we avoid needlessly adding a somewhat confusing endpoint.

internal/testutils/cluster.go
n8mgr: To me, `Retry` usually indicates a poorly written or misunderstood test. I think it should absolutely be avoided.

api/client.go
```diff
@@ -22,3 +23,9 @@ func (c *Client) State() (state State, err error) {
 	err = c.c.GET("/state", &state)
 	return
 }
+
+// TipState returns the consensus tip of the indexer.
+func (c *Client) TipState() (state consensus.State, err error) {
```
n8mgr: These API changes feel out of scope for this PR

internal/testutils/cluster.go
ChrisSchinnerl: I definitely agree for unit tests but for integration tests there are just a lot more parts that we don't want to interact with directly and instead only through the API to keep the integration tests valuable. 

e.g. in this case we want to return from `NewCluster` when all components agree on the block height. Not just the chain managers. Meaning we need to wait for syncers doing their thing, the ephemeral contractor and other components picking up on the new blocks from the chain managers.

Once the cluster is complete it gets even worse. We need announcements to be read from chain, the scan loop to do its thing, the contract maintenance to run, accounts to be funded and so on. Depending on how many hosts you want this takes different amounts of time. Also the CI is often very inconsistent when it comes to I/O.

Sure we can say this takes 1s in the worst case and sleep for that amount of time but is that better than retrying for 1s and checking every 100ms?

peterjan (approved):

internal/testutils/cluster.go
n8mgr: Why was this merged without resolution or discussion? A generic retry function makes it easier to write lazy tests or hide poorly designed concurrency. All of the problems listed above can be resolved with better test architecture. Generically calling `Retry` to fix a test should absolutely be a last resort rather than an expectation.

Having a `Cluster` type at all at this stage is unnecessary. Current tests should be small, reproducible, and focused on the implementation of each system, not a huge all encompassing integration test.

PR #25, opened 2025-02-12
Add RHP4 client

ChrisSchinnerl:
This adds renterd's RHP4 client implementation without any of the extras we added such as decorating errors with custom errors from the utils package, the custom dialer or any of that.

### DIFF
--- a/internal/rhp/v4/client.go
+++ b/internal/rhp/v4/client.go
@@ -10,16 +10,20 @@ import (
 	rhp "go.sia.tech/coreutils/rhp/v4"
 )
 
+// Client is a client for RHP4. It is a wrapper around the coreutils client that
+// adds pooling of transports to reuse TCP connections.
 type Client struct {
 	tpool *transportPool
 }
 
+// New creates a new RHP4 client.
 func New() *Client {
 	return &Client{
 		tpool: newTransportPool(),
 	}
 }
 
+// Settings executes the RPCSettings RPC on the host.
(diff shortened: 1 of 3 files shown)


internal/rhp/v4/transport.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+type transportPool struct {
+	mu   sync.Mutex
+	pool map[string]*transport
+}
+
+func newTransportPool() *transportPool {
+	return &transportPool{
+		pool: make(map[string]*transport),
```
n8mgr: Wouldn't this be better as `map[types.PublicKey]*transport`?

n8mgr:
My recommendation is to think about this when we actually need it not anticipate the need ahead of time.

n8mgr (approved):

internal/rhp/v4/client.go
```diff
@@ -0,0 +1,157 @@
...
+	rhp "go.sia.tech/coreutils/rhp/v4"
+)
+
+// Client is a client for RHP4. It is a wrapper around the coreutils client that
+// adds pooling of transports to reuse TCP connections.
+type Client struct {
+	tpool *transportPool
+}
+
+// New creates a new RHP4 client.
```
peterjan, line 19: `nit` would've preferred `NewClient` here

peterjan (approved):

peterjan:
> My recommendation is to think about this when we actually need it not anticipate the need ahead of time.

+1

PR #26, opened 2025-02-13
Fix syncer peers

peterjan:


persist/postgres/peers.go
```diff
@@ -50,7 +50,7 @@ func (s *Store) PeerInfo(addr string) (info syncer.PeerInfo, err error) {
 // Peers returns the set of known peers.
 func (s *Store) Peers() (infos []syncer.PeerInfo, err error) {
 	err = s.transaction(context.Background(), func(ctx context.Context, tx *txn) error {
-		const query = `SELECT ip_address, port, first_seen, last_connect, synced_blocks, sync_duration FROM syncer_peers`
+		const query = `SELECT host(ip_address), port, first_seen, last_connect, synced_blocks, sync_duration FROM syncer_peers`
```
peterjan, line 53: This was a little surprising. I casted it to `::text` first because we try and scan into a string, but then it has the mask in the address which the syncer can't handle.

n8mgr (approved):

ChrisSchinnerl (approved):

PR #27, opened 2025-02-13
RFC: User accounts

ChrisSchinnerl:
This PR adds an RFC related to user accounts, what they are for and how their accounts are refilled.
I also added a section for discussions since I ran some numbers (also in the rfc) which seem to require further refinement of this spec.

### DIFF
--- a/doc/rfc/0005_account_management.md
+++ b/doc/rfc/0005_account_management.md
@@ -19,8 +19,8 @@ that right before we allow users to access third party indexers
 
 Unlike `renterd`, the `indexer` doesn't locally track the expected balance of
 accounts to know when to fund them. Instead, it uses a ratelimit-based approach.
-Accounts are funded with at most 1SC per hour as needed. To do so, the following
-state will be stored for every account and host:
+Accounts are funded with at most 1SC per hour. To do so, the following state
+will be stored for every account and host:
 
 | **Field** | **Description** |
 |-------|-------------|
--- a/doc/rfc/0005_account_management.md
+++ b/doc/rfc/0005_account_management.md
@@ -30,75 +30,23 @@ will be stored for every account and host:
 
 The `indexer` will then perform the following steps for each host it has a contract with:
 1. Fetch all accounts for that host where `TimeToFund` is in the past
-2. In batches of up to 100, fund the accounts with 1SC each. If the remaining funds in the contract are too low, fund fewer accounts.
+2. In batches of up to 1000, fund the accounts with the replenish RPC to restore
+their balances to 1SC again. If the remaining funds in the contract are too low,
+fund fewer accounts.
 3. For each account in the batch:
-  3a. Upon success, update `TimeToFund` to the `MIN(6 hours , current time + MAX(new account balance, 1SC) / 1SC * 1 hour (rounded up))` to avoid funding accounts at a higher rate than 1SC/hour
+  3a. Upon success, update `TimeToFund` to 1 hour in the future
   3b. Upon failure, log the error and update `TimeToFund` using an exponential backoff (1, 2, 4, 8, 16, 32, 64 and 128 minutes)
 4. Sleep until the lowest `TimeToFund` for an account on this host is reached or until the next account is registered
 
-NOTE: `MAX(new account balance, 1SC)` is used to prevent hosts that return `0SC`
-as the new balance from sending us into a hot refill-loop.
+**Malicious Hosts**
 
-**Costs**
+Using the replenish RPC, account funding is a pretty efficient operation that
+won't waste any funds assuming a host is honest.
 
-While the rate-based approach has many advantages over the approach used in
-`renterd`, it also has one drawback. The `indexer` never fetches the balance of
-an account before trying to fund it. So we will continuously pay hosts. So let's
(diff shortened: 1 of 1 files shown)


doc/rfc/0005_account_management.md
```diff
@@ -0,0 +1,104 @@
...
+  3a. Upon success, update `TimeToFund` to the `MIN(6 hours , current time + MAX(new account balance, 1SC) / 1SC * 1 hour (rounded up))` to avoid funding accounts at a higher rate than 1SC/hour
+  3b. Upon failure, log the error and update `TimeToFund` using an exponential backoff (1, 2, 4, 8, 16, 32, 64 and 128 minutes)
+4. Sleep until the lowest `TimeToFund` for an account on this host is reached or until the next account is registered
+
+NOTE: `MAX(new account balance, 1SC)` is used to prevent hosts that return `0SC`
+as the new balance from sending us into a hot refill-loop.
+
+**Costs**
+
+While the rate-based approach has many advantages over the approach used in
```
n8mgr: If this is difficult using current RHP4 RPC, this may be a good candidate for rethinking how we think about account balances and funding.

The primary issue is we currently have no way to batch-check account balances before deciding how much to fund. A secondary issue is even if we check the balance, there would still be a small race between the calculation of how much to fund and the account being funded.

How impactful we think the secondary issue is should guide how much we want to change about accounts. Two solutions come to mind:

#### Add an RPC to batch-check account balances

1. Call RPC batch account balance with a list of accounts
  1.1 Host replies with the balance of each account
2. Calculate the funding amount
3. Call RPC fund accounts with a list of accounts, the funding amounts, and a signed revision
 3.1 Host replies with the signature and new balances

This has a slight race between checking the balance and calling a separate RPC to actually fund the accounts. Each of these RPC is a single round trip.

#### Add an additional RPC

1. Call RPC set balance with a list of accounts and a target of 1 SC
  1.1 Host replies with the amount necessary to get all accounts to that target
  1.2 Renter sends a revision
  1.3 Host signs the revision and returns the signature to the renter

This one gives a lot of control to the host, but we can add other guard rails to prevent too much abuse. Either way, we add an additional roundtrip.

doc/rfc/0005_account_management.md
```diff
@@ -0,0 +1,104 @@
...
+### Authentication
+
+TODO: Probably don't need special authentication for now. Probably fine to add
+that right before we allow users to access third party indexers
+
+### Account Funding
+
+Unlike `renterd`, the `indexer` doesn't locally track the expected balance of
+accounts to know when to fund them. Instead, it uses a ratelimit-based approach.
+Accounts are funded with at most 1SC per hour as needed. To do so, the following
```
peterjan: This says "as needed" but the funding happens on a time-based trigger.

doc/rfc/0005_account_management.md
peterjan: I have always been a fan of that second approach. I feel checking a balance and then depositing the difference between the desired amount and whatever you get back needless introduces a race that you can avoid by communicating your desired balance, then signing whatever the host sends back.  What's nice about it is that you can keep track of drift on a per-host basis and not a per-account basis. Funding up until the desired amount also feels like a better fit for our use case where we essentially want to "top off" our accounts, not blindly deposit more funds into them.

ChrisSchinnerl: I think having RPCBalance fetch 1..n balances would be nice regardless of the other RPC. Because we potentially don't always want to fund an account right after.

I'm not a fan of the Balance + Fund RPC for the following reasons:

1. It requires protection against malicious actors. Because knowing an apps account key I can just lock that account on every host simply by starting the Balance+Fund RPC and then keeping the connection open after the initial roundtrip.
2. It doesn't fix any race regarding to drift management. The client still needs to do client-side locking even if the RPC itself is atomic for exact accounting. If the client has to do that in the first place, the atomic RPC is redundant.

peterjan: Wait where is there account locking? In my head there is no account level locking and the eventual balance of the account is close to the requested target balance, but it might be off slightly. The same goes for drifting, there's only fuzzy accounting and no exact accounting at all.

ChrisSchinnerl: Yeah the `indexer` doesn't do any locking or exact accounting. It can't do that at all considering it's only funding the accounts and not spending from them. Which means even if we had an RPC that supported atomic balance+fund we wouldn't benefit.

n8mgr: > It requires protection against malicious actors. Because knowing an apps account key I can just lock that account on every host simply by starting the Balance+Fund RPC and then keeping the connection open after the initial roundtrip.


Even if we didn't make it atomic, it would still be more performant than two separate streams and saves some bytes from only sending a single currency instead of `n` and only sending the list of accounts once.

ChrisSchinnerl: Yeah that is fine as long as we don't lock the account for withdrawals.

n8mgr: Implementation for RPCReplenish here: https://github.com/SiaFoundation/coreutils/pull/180

peterjan (approved):

n8mgr (approved):

PR #28, opened 2025-02-13
Add chain subscriber

peterjan:


n8mgr (dismissed):

internal/threadgroup/threadgroup.go
ChrisSchinnerl, line 1: Use the one expoed by `coreutils` instead.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

PR #29, opened 2025-02-14
RFC: Slab Pinning

ChrisSchinnerl:
This PR adds the RFC for slab pinning by users.

The RFC makes some higher level assumptions of the database which I think is fine since we probably should sign off on the higher level "what we want from a feature" and then start updating the database schema RFC.

### DIFF
--- a/doc/rfc/006_slab_pinning.md
+++ b/doc/rfc/006_slab_pinning.md
@@ -4,8 +4,8 @@
 
 ## API
 
-For user apps the API to pin uploaded slab is pretty straightforward. All we
-need is a `POST /slabs/pin` endpoint its `DELETE` counterpart.
+For user apps, the API to pin uploaded slabs is pretty straightforward. All we
+need is a `POST /slabs/pin` endpoint and its `DELETE` counterpart.
 
 ### Request
 
@@ -29,7 +29,7 @@ Where every slab is an object of the following form:
 
 The `Key` is necessary because the `indexer` can't repair slabs without the
 ability to decrypt shards. However, the first layer of encryption is applied by
-the client the corresponding key is never sent to the `indexer`.
+the client, so the corresponding key is never sent to the `indexer`.
 
 Each shard within a slab is an object of the following form:
 
@@ -59,8 +59,8 @@ The `ID` is special because it is derived from the roots of the shards and key.
 This means that two slabs with the same shards are effectively the same. The
 reason for including the key is to prevent two slabs from being treated as the
 same if one of them was purposefully created with the wrong key. One nice
-side-effect of this way of creating the `ID` is that slabs shared between
-accounts will have the same `ID`.
+side-effect creating the `ID` this way is that slabs shared between accounts
+will have the same `ID`.
 
 If everything goes well, the slabs are now in the `indexer`'s database and
 uniquely identified via their `ID`s. The shards are also persisted as "unpinned
--- a/doc/rfc/006_slab_pinning.md
+++ b/doc/rfc/006_slab_pinning.md
@@ -72,26 +72,18 @@ However, the slabs are not yet pinned to contracts. That happens asynchronously.
 Slabs are scoped to the account key that pinned them. So one account can't unpin
 another account's slabs or even fetch that slab's metadata.
 
-## Pinning Loops
+## Pinning
 
-The pinning loops are the goroutines within the `indexer` that actually pin
-slabs to hosts. We run one loop for each host we have a contract with. Since the
-`indexer` stores the unpinned sectors alongside their expiry, the pinning loops
-are pretty straightforward.
+Pinning slabs happens as part of periodic maintenance that the indexer performs
+on contracts (see [Pinning and Pruning](006_slab_pinning.md). The process looks
+like this:
 
 For each host we have a contract with we do the following:
 1. Fetch a batch of unpinned sectors (at most 1<<40) sorted by expiration time
 2. Attempt to append the sectors to any of the contracts we have with the host
 3. Update the database
-  a. For successful appends: Atomically remove the unpinned sectors from the database and add the sector<->contract link
+  a. For successful appends: Update the sector<->host link to contain the contract ID the sector was pinned to
   b. For "missing sectors" errors: Remove the sectors from the database
-
-Every host gets its own loop since contracts can only be interacted with
-sequentially and we usually only have one contract with each host that we want
-to use. The biggest advantage of this approach is its simplicity. Where
-`renterd` required distributed locking since contracts were used by multiple
-entities, every contract in the `indexer` is only modified within the pinning
-loop. So we don't need any contract locking at all.
 
 ## Slab Health and unpinned Sectors
 
--- a/doc/rfc/006_slab_pinning.md
+++ b/doc/rfc/006_slab_pinning.md
@@ -1,0 +1,102 @@
+# Slab Pinning
+
+## Abstract
+
+## API
+
+For user apps the API to pin uploaded slab is pretty straightforward. All we
+need is a `POST /slabs/pin` endpoint its `DELETE` counterpart.
+
+### Request
+
+The payload for the `POST` request should look like this:
+
+```json
+{
+  "slabs": [...]
+}
+```
+
+Where every slab is an object of the following form:
+
+```json
+{
+  "Key": "<32-random-bytes>", // key used to encrypt shards after erasure-coding
+  "MinShards": 10, // number of data pieces used for uploading
+  "Shards": [...] // matches number of data pieces + parity pieces
+}
+```
+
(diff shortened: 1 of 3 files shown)


doc/rfc/006_slab_pinning.md
```diff
@@ -0,0 +1,102 @@
...
+However, the slabs are not yet pinned to contracts. That happens asynchronously.
+
+### Slab Scope
+
+Slabs are scoped to the account key that pinned them. So one account can't unpin
+another account's slabs or even fetch that slab's metadata.
+
+## Pinning Loops
+
+The pinning loops are the goroutines within the `indexer` that actually pin
```
n8mgr: I bet we'd get pretty far with a single goroutine. If we assume worst case for each iteration is 60 seconds, that's up to 60TiB of pins per hour or ~17 hours to append 1PiB of sectors. It removes some overhead and complexity from managing an additional 100-200 goroutines with little material benefit.

doc/rfc/006_slab_pinning.md
```diff
@@ -0,0 +1,102 @@
...
+slabs to hosts. We run one loop for each host we have a contract with. Since the
+`indexer` stores the unpinned sectors alongside their expiry, the pinning loops
+are pretty straightforward.
+
+For each host we have a contract with we do the following:
+1. Fetch a batch of unpinned sectors (at most 1<<40) sorted by expiration time
+2. Attempt to append the sectors to any of the contracts we have with the host
+3. Update the database
+  a. For successful appends: Atomically remove the unpinned sectors from the database and add the sector<->contract link
```
n8mgr: When we get to schema design, I want to ensure we are using references as much as possible over deleting from one table and inserting into another.

doc/rfc/006_slab_pinning.md
```diff
@@ -0,0 +1,102 @@
...
+entities, every contract in the `indexer` is only modified within the pinning
+loop. So we don't need any contract locking at all.
+
+## Slab Health and unpinned Sectors
+
+Unpinned sectors count towards the health of a slab. That's because the repair
+code will reuse the same pinning loop used for user slabs. e.g. if a slab needs
+repairs on 5 sectors, the repair loop uploads 5 sectors to temp storage and
+notifies the pinning loop. Until the pinning is done, the repair loop shouldn't
+revisit the same slab.
```
n8mgr, line 94: > Until the pinning is done, the repair loop shouldn't revisit the same slab.

Ideally, the schema will be capable of preventing the repair loop from considering shards that are unpinned as needing repair. To the repair loop, they should look like a newly uploaded shard after they've been migrated to a new host.

n8mgr (changes requested):

doc/rfc/006_slab_pinning.md
```diff
@@ -0,0 +1,102 @@
+# Slab Pinning
+
+## Abstract
+
+## API
+
+For user apps the API to pin uploaded slab is pretty straightforward. All we
```
peterjan:
```suggestion
For user apps, the API to pin uploaded slabs is pretty straightforward. All we
```

doc/rfc/006_slab_pinning.md
```diff
@@ -0,0 +1,102 @@
...
+need is a `POST /slabs/pin` endpoint its `DELETE` counterpart.
```
peterjan:
```suggestion
need is a `POST /slabs/pin` endpoint and its `DELETE` counterpart.
```

doc/rfc/006_slab_pinning.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Request
+
+The payload for the `POST` request should look like this:
+
+```json
+{
+  "slabs": [...]
```
peterjan, line 16: Could consider getting rid of the nesting here.

doc/rfc/006_slab_pinning.md
```diff
@@ -0,0 +1,102 @@
...
+{
+  "Key": "<32-random-bytes>", // key used to encrypt shards after erasure-coding
+  "MinShards": 10, // number of data pieces used for uploading
+  "Shards": [...] // matches number of data pieces + parity pieces
+}
+```
+
+The `Key` is necessary because the `indexer` can't repair slabs without the
+ability to decrypt shards. However, the first layer of encryption is applied by
+the client the corresponding key is never sent to the `indexer`.
```
peterjan:
```suggestion
the client and that key is never sent to the `indexer`.
```
or something else, but it was a bit of a weird sentence

doc/rfc/006_slab_pinning.md
```diff
@@ -0,0 +1,102 @@
...
+}
+```
+
+### Response
+
+The response to the request contains one item for each slab:
+
+```json
+{
+  "slabs": [
```
peterjan, line 49: same

doc/rfc/006_slab_pinning.md
```diff
@@ -0,0 +1,102 @@
...
+    ...
+  ]
+}
+```
+
+The `ID` is special because it is derived from the roots of the shards and key.
+This means that two slabs with the same shards are effectively the same. The
+reason for including the key is to prevent two slabs from being treated as the
+same if one of them was purposefully created with the wrong key. One nice
+side-effect of this way of creating the `ID` is that slabs shared between
```
peterjan:
```suggestion
side-effect of creating the `ID` this way is that slabs shared between
```

doc/rfc/006_slab_pinning.md
ChrisSchinnerl: I guess this RFC doesn't give a full picture since I called it `Pinning Loops`, but the idea was that each contract has one loop like a state machine that is allowed to update it to avoid locking. Which means, once we have all RFCs in place, for every contract the loop would look something like this:

```
loop {
  pinSectors()

  pruneSectors()

  verifyDataIntegrity()
}
```

So one contract might prune while another one pins while another one performs some integrity checks. Instead of having a single loop where coordinating this becomes more complex.

Considering that the MVP probably doesn't need pruning or data integrity checks, we could start with a single loop regardless though. Just wanted to share the bigger picture on my mind.

doc/rfc/006_slab_pinning.md
ChrisSchinnerl: Yeah in the database I want the slabs and sectors to already be in the spot where we need them to be after pinning. Only difference is a potential reference in an "unpinned sectors" table and a missing reference in a "sector_contracts" table. But the slabs and sectors would already be in the tables that they'll be in after pinning.

The actual metadata that the repair loop gets will probably implicitly indicate that a sector is not pinned since it doesn't contain any contracts for that sector. e.g. if you imagine the extreme edge case of a slab with 15 missing shards and 15 unpinned shards, we want the repair loop to only repair the missing ones since the unpinned ones will get pinned soon anyway.

doc/rfc/006_slab_pinning.md
ChrisSchinnerl: Doing that on purpose to allow for adding fields without breaking the API

doc/rfc/006_slab_pinning.md
n8mgr: My primary concern with how it's written is that I want to avoid spawning goroutines unless necessary. The pin loop itself being a goroutine makes sense. If we can get enough performance in the worst case scenario out of a single goroutine, we should avoid spawning 50+. Go makes it very easy to add concurrency without thinking about the technical/mental overhead of managing or testing it.

Ideally we'd set a target: 
- if we can meet that target without additional goroutines, great. 
- if not, add a semaphore and spawn n goroutines. 

```
for {
  sleep(5m)
  loop hosts {
    pinSectors()
    pruneSectors()
  }
}
```

vs

```
for {
  loop hosts (
    go {
        sleep(5m)
        pinSectors()
        pruneSectors()
    }
  }
}
```

`verifyDataIntegrity` doesn't require a contract, it uses an account and a sector root. We don't have to worry about locking for it, so it can live anywhere.

ChrisSchinnerl: That's fair but just to clarify I intended for the goroutines to be long-lived worker tasks

So every goroutine is essentially your first example
```go
for {
  sleep(5m)
  loop hosts {
    pinSectors()
    pruneSectors()
  }
}
```

But we launch this loop once per host on startup or when we form a new contract with a host instead of in the inner loop as per your second example.

n8mgr: Doing long lived workers adds state and requires managing those goroutines whenever you add or drop hosts. Short-lived is simpler. Since the indexer isn’t performance bound, we don’t really benefit from having a constant goroutine per host

n8mgr (dismissed):

n8mgr (dismissed):

PR #30, opened 2025-02-14
Add wallet endpoints

### LINKED ISSUE #2: Implement Single Address Wallet
Implement the interface required for the Single Address wallet and add it to the indexer. 
Probably a good idea to split this one up as needed since it requires

- Implementing tracking of state in the database (subscribing to chain updates and reverts) and testing
- Exposing the wallet through the API and integration testing

peterjan:
Closes https://github.com/SiaFoundation/indexd/issues/2


api/endpoints.go
```diff
@@ -19,5 +51,112 @@ func (a *api) handleGETState(jc jape.Context) {
...
+	} else if limit < -1 || limit > maxLimit {
+		jc.Error(ErrInvalidLimit, http.StatusBadRequest)
+		return
+	} else if limit == -1 {
+		limit = maxLimit
+	}
+
+	ok = true
+	return
+}
```
ChrisSchinnerl:
```suggestion
func parseOffsetLimit(jc jape.Context) (offset int, limit int, ok bool) {
	if jc.DecodeForm("offset", &offset) != nil {
		return 0, 0, false
	} else if offset < 0 {
		jc.Error(ErrInvalidOffset, http.StatusBadRequest)
		return 0, 0, false
	}

	limit = defaultLimit
	if jc.DecodeForm("limit", &limit) != nil {
		return 0, 0, false
	} else if limit < -1 || limit > maxLimit {
		jc.Error(ErrInvalidLimit, http.StatusBadRequest)
		return 0, 0, false
	} else if limit == -1 {
		limit = maxLimit
	}

	return offset, limit, true
}
```

api/state.go
```diff
@@ -15,5 +15,6 @@ type (
 	State struct {
 		StartTime time.Time `json:"startTime"`
 		BuildState
+		ScanHeight uint64 `json:"scanHeight"`
```
ChrisSchinnerl: I'd order it like so:

```
BuildState // embedded stuff first

ScanHeight // everything else alphabetized 
StartTime
```

api/endpoints.go
ChrisSchinnerl: I'd also avoid `-1` as a sentinel value in `indexd` since we have defaults and a max now. It just doesn't work as well anymore considering that `-1` usually means "no limit" but we still enforce one.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):
Lgtm but I'd either here or in a f/u add the first actual integration test for the wallet to make sure the new endpoints (especially sending funds) are actually tested.

e.g. to `internal/tests/wallet_test.go`

peterjan:
> Lgtm but I'd either here or in a f/u add the first actual integration test for the wallet to make sure the new endpoints (especially sending funds) are actually tested.
> 
> e.g. to `internal/tests/wallet_test.go`

I'll F/U with this, but yeah good call. I actually had it at one point but I didn't like how it looked and removed it 😂

n8mgr:
> Lgtm but I'd either here or in a f/u add the first actual integration test for the wallet to make sure the new endpoints (especially sending funds) are actually tested.
> 
> e.g. to `internal/tests/wallet_test.go`

Why add an additional `internal/tests` package over putting the API endpoint tests in the `api` package? Test systems where they're implemented so the codebase is predictable. Focus on small, thorough, reproducible tests before adding the big integration tests.

ChrisSchinnerl:
> > Lgtm but I'd either here or in a f/u add the first actual integration test for the wallet to make sure the new endpoints (especially sending funds) are actually tested.
> > e.g. to `internal/tests/wallet_test.go`
> 
> Why add an additional `internal/tests` package over putting the API endpoint tests in the `api` package? Test systems where they're implemented so the codebase is predictable. Focus on small, thorough, reproducible tests before adding the big integration tests.

I honestly don't care about the exact location. But we probably need to make it an `api_test` package to avoid the circular dependency with `NewIndexer`.

ChrisSchinnerl (approved):

peterjan:
Mind merging this as is? I'll F/U with a PR to find a better spot for the tests and add the one that's missing.

n8mgr (approved):

PR #32, opened 2025-02-19
Add host announcements

### LINKED ISSUE #19: Host Announcements
The indexer should...

- Scan the blockchain for new hosts (ignore hosts that haven't announced in the last 12 months)
- Add new hosts to the store
- Have 2 store methods to fetch a single host or query multiple

No need for API endpoints yet but we should have a test that creates an indexer and hosts, announces the hosts and makes sure the store eventually returns these hosts.

peterjan:
Closes https://github.com/SiaFoundation/indexd/issues/19


hosts/manager.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+type (
+	// UpdateTx defines what the host manager needs to atomically process a
+	// chain update in the database.
+	UpdateTx interface {
+		AddHostAnnouncement(hk types.PublicKey, ha chain.V2HostAnnouncement, ts time.Time) error
+	}
+
+	// Manager manages the host announcements.
```
ChrisSchinnerl: If possible I'd like to avoid the `Manager` naming hell we have in `hostd`. Depending on what we want the type to do there are better names like `HostSubscriber`, `AnnouncementSubscriber`,`Hosts`, `Announcements` or even `HostManager`. Just some additional context to allow for better lookups.  
When doing a global symbol search or worse, a grep, nobody wants to arrow up/down through 15 managers and rely on the file path to figure out which one to go to.

hosts/manager.go
```diff
@@ -0,0 +1,85 @@
...
+		log:                zap.NewNop(),
+	}
+	for _, opt := range opts {
+		opt(m)
+	}
+	return m
+}
+
+// Close closes the manager.
+func (m *Manager) Close() error {
```
ChrisSchinnerl: Do we expect this package to also manage scanning? Because if it's only used for announcements it feels like the type is a bit overkill. Especially since we will probably never add anything to `Close` here.

hosts/manager.go
```diff
@@ -0,0 +1,85 @@
...
+// NewManager creates a new host manager.
+func NewManager(opts ...Option) *Manager {
+	m := &Manager{
+		announcementMaxAge: time.Hour * 24 * 365,
+		log:                zap.NewNop(),
+	}
+	for _, opt := range opts {
+		opt(m)
+	}
+	return m
```
ChrisSchinnerl: I'd make sure that `announcementMaxAge` isn't 0 here. That would make for some awkward issues.

hosts/manager.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+
+		has := make(map[types.PublicKey]chain.V2HostAnnouncement)
+		chain.ForEachV2HostAnnouncement(update.Block, func(hk types.PublicKey, addrs []chain.NetAddress) {
+			filtered := make(map[chain.Protocol][]chain.NetAddress)
+			for _, addr := range addrs {
+				if addr.Address == "" {
+					m.log.Debug("ignoring host announcement with empty address", zap.Stringer("hk", hk))
+					continue
+				} else if addr.Protocol != rhp4.ProtocolTCPSiaMux {
```
ChrisSchinnerl: We also got `ProtocolQUIC` now to check for.

internal/testutils/host.go
```diff
@@ -17,6 +18,10 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	announcementTxnSize = 1000
```
ChrisSchinnerl: I'd inline this because it's only used in one spot

persist/postgres/host.go
```diff
@@ -0,0 +1,102 @@
...
+	"errors"
+	"fmt"
+	"time"
+
+	"go.sia.tech/core/types"
+	"go.sia.tech/coreutils/chain"
+	"go.sia.tech/indexd/api"
+)
+
+type dbHost struct {
```
ChrisSchinnerl: How about a `scanHost` function that returns an `api.Host` and takes a scanner as its input?
Then we can inline the type as `host` in that function and call it every time we need to scan query results.

Imo that's better than potentially calling `rows.Scan` in multiple places.

persist/postgres/host.go
```diff
@@ -0,0 +1,102 @@
...
+	id int64
+	api.Host
+}
+
+func (u *updateTx) AddHostAnnouncement(hk types.PublicKey, ha chain.V2HostAnnouncement, ts time.Time) error {
+	var hostID int64
+	err := u.tx.QueryRow(u.ctx, `INSERT INTO hosts (public_key, last_announcement) VALUES ($1, $2) ON CONFLICT (public_key) DO UPDATE SET last_announcement = $3 RETURNING id;`, sqlPublicKey(hk), ts, ts).Scan(&hostID)
```
ChrisSchinnerl: Maybe we can reuse `$2` rather than declaring a `$3` and passing `ts` twice?

persist/postgres/host.go
```diff
@@ -0,0 +1,102 @@
...
+
+	err = rows.Err()
+	return
+}
+
+func queryHostAddresses(ctx context.Context, tx *txn, hostID int64) (addresses []chain.NetAddress, err error) {
+	rows, err := tx.Query(ctx, `SELECT net_address, protocol FROM host_addresses WHERE host_id = $1`, hostID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to query host addresses: %w", err)
+	}
```
ChrisSchinnerl: missing close

persist/postgres/host.go
```diff
@@ -0,0 +1,102 @@
...
+	for rows.Next() {
+		var host dbHost
+		if err := rows.Scan(&host.id, (*sqlPublicKey)(&host.PublicKey), &host.LastAnnouncement); err != nil {
+			return nil, fmt.Errorf("failed to scan host: %w", err)
+		}
+		hosts = append(hosts, host)
+	}
+
+	err = rows.Err()
+	return
```
ChrisSchinnerl: Please stop defaulting to named return vars. The one line you safe ends up being an awkward looking `err = rows.Err()` here that's easy to forget.

persist/postgres/host.go
ChrisSchinnerl, line 1: Naming is a bit inconsistent. Should be `hosts.go`.

persist/postgres/host.go
```diff
@@ -0,0 +1,102 @@
...
+			}
+			hosts = append(hosts, h.Host)
+		}
+		return nil
+	})
+	return
+}
+
+func queryHosts(ctx context.Context, tx *txn, offset, limit int) (hosts []dbHost, err error) {
+	rows, err := tx.Query(ctx, `SELECT h.id, h.public_key, h.last_announcement FROM hosts h ORDER BY h.last_announcement DESC LIMIT $1 OFFSET $2`, limit, offset)
```
ChrisSchinnerl: Is there a specific reason for sorting by announcement date rather than leaving it unsorted for now?

ChrisSchinnerl (changes requested):

hosts/manager.go
peterjan: Yeah my F/U adds scanning here

persist/postgres/host.go
peterjan: Could you write it out in pseudo code, just very roughly. I know how to transform this into a `scanHost` of course, I considered it even, but converting it immediately makes it less clean for me. I think I don't see how you could mess up `rows.Scan` in my approach and I like that the queries are in their simplest forms.

hosts/manager.go
peterjan: I renamed it to `HostManager`. I personally want to try the `Manager` thing, very similar to how `hostd` does it. It's personal preference of course, I don't mind doing whatever you feel most comfortable with. I think we'll end up with 6 managers, each in their own package. We have roughly 6 classes of entities I think.

persist/postgres/host.go
ChrisSchinnerl: Yeah while writing it out I noticed that it gets messy with the addresses and passing around an incomplete `api.Host` is arguably worse.  

Alternatively we could call `queryHostAddresses` in `queryHosts` and inline the type in `queryHosts` but it's probably not worth the trouble right now.

persist/postgres/init.sql
```diff
@@ -1,8 +1,29 @@
...
+    last_announcement TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT '0001-01-01 00:00:00+00',
+    last_scan_success BOOLEAN NOT NULL DEFAULT FALSE,
+    next_scan TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT '0001-01-01 00:00:00+00',
+    uptime INTERVAL NOT NULL DEFAULT '0 seconds',
+    downtime INTERVAL NOT NULL DEFAULT '0 seconds'
+);
+
+CREATE TABLE host_addresses (
+    id SERIAL PRIMARY KEY,
+    host_id INTEGER REFERENCES hosts(id) ON DELETE CASCADE,
```
ChrisSchinnerl, line 17: I think we are missing an index for this. Apparently foreign keys aren't automatically indexed by Postgres.

internal/testutils/cluster_test.go
```diff
@@ -66,5 +66,15 @@ func TestNewCluster(t *testing.T) {
 		t.Fatalf("expected miner payout, %+v", events[0])
 	}
 
+	// assert host announcements were persisted
+	hosts, err := indexer.db.Hosts(context.Background(), 0, 10)
+	if err != nil {
+		t.Fatal(err)
+	} else if len(hosts) != defaultClusterOpts.hosts {
+		t.Fatalf("expected %d hosts, got %d", defaultClusterOpts.hosts, len(hosts))
+	} else if len(hosts[0].Addresses) == 0 {
```
ChrisSchinnerl, line 75: would be great to have a dedicated integration test with hosts announcing themselves with
1. too many addresses to make sure we cap it at 2
2. invalid protocols to make sure we filter them out
3. not a single valid address to make sure we don't record it in the first place

internal/testutils/host.go
```diff
@@ -36,6 +39,48 @@ func (h *Host) Addr() string {
...
+	h.w.SignV2Inputs(&txn, toSign)
+	basis, txnset, err := h.cm.V2TransactionSet(basis, txn)
+	if err != nil {
+		h.w.ReleaseInputs(nil, []types.V2Transaction{txn})
+		return fmt.Errorf("failed to create transaction set: %w", err)
+	} else if _, err := h.cm.AddV2PoolTransactions(basis, txnset); err != nil {
+		h.w.ReleaseInputs(nil, []types.V2Transaction{txn})
+		return fmt.Errorf("failed to add transaction to pool: %w", err)
+	}
+
```
ChrisSchinnerl, line 71: Isn't this missing adding the txn to the `h.cm`'s pool first to validate it? Probably works without validation but might be useful to have regardless to rule out errors. Otherwise the txn might just get dropped without a trace.

ChrisSchinnerl (changes requested):

persist/postgres/init.sql
peterjan: wow

internal/testutils/host.go
peterjan: Wait, did you miss the spot where I add it to the pool on line `:74`?

internal/testutils/host_test.go
```diff
@@ -15,6 +17,106 @@ func TestMain(m *testing.M) {
 	goleak.VerifyTestMain(m)
 }
 
+func TestAnnounceHost(t *testing.T) {
```
ChrisSchinnerl: `testutils` isn't the right place for tests apart from testing that the helpers we have here work. 

You could move this to `hosts/hosts_test.go` but keep it the same otherwise.

internal/testutils/host_test.go
```diff
@@ -15,6 +17,106 @@ func TestMain(m *testing.M) {
...
+		if a.Protocol == rhp4.ProtocolQUIC {
+			nQUIC++
+		} else if a.Protocol == rhp4.ProtocolTCPSiaMux {
+			nSiaMux++
+		}
+	}
+	if nQUIC != 2 || nSiaMux != 2 {
+		t.Fatalf("expected 2 QUIC and 2 siamux addresses, got %d QUIC and %d siamux", nQUIC, nSiaMux)
+	}
+}
```
ChrisSchinnerl: Can you add one more case for reannouncing a host to make sure the last announcement gets removed?

internal/testutils/cluster_test.go
ChrisSchinnerl, line 1: `TestNewCluster` seems to fail

ChrisSchinnerl (changes requested):

internal/testutils/cluster_test.go
peterjan: First NDF  👀  (comparing times)

hosts/hosts_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"go.sia.tech/indexd/internal/testutils"
+)
+
+func TestAnnounceHost(t *testing.T) {
+	c := testutils.NewCluster(t, testutils.WithHosts(0))
+	db := c.Indexer.DB()
+
+	hosts := c.NewHosts(t, 3)
+	c.AddHosts(t, hosts...)
+	c.FundHosts(t, hosts...)
```
n8mgr: Spawning a full cluster for something so simple is overkill. I'd like to avoid using it at all right now so the scope of what we're testing stays focused. That likely means a bit of duplication, but that's not inherently a bad thing.

You don't really need a host for this test at all. You're not scanning them at this point. Just build and broadcast announcements.

n8mgr (changes requested):

hosts/hosts_test.go
```diff
@@ -0,0 +1,155 @@
...
+		{
+			Protocol: rhp4.ProtocolQUIC,
+			Address:  "[::]:4848",
+		},
+	}); err != nil {
+		t.Fatal(err)
+	}
+
+	// assert addresses get replaced
+	testutils.Retry(t, 100, 100*time.Millisecond, func() error {
```
n8mgr: All of this in a retry loop is lazy. 

1. Mine n blocks
2. Check that the subscriber is synced
3. Do your assertions once

If that fails, something is probably wrong.

hosts/manager_test.go
```diff
@@ -0,0 +1,186 @@
...
+	"go.sia.tech/coreutils/chain"
+	rhp4 "go.sia.tech/coreutils/rhp/v4"
+	"go.sia.tech/indexd/api"
+	"go.sia.tech/indexd/hosts"
+	"go.sia.tech/indexd/internal/testutils"
+	"go.sia.tech/indexd/subscriber"
+	"go.uber.org/zap"
+)
+
+func TestHostManager(t *testing.T) {
```
peterjan, line 19: image

n8mgr (approved):

ChrisSchinnerl (approved):

PR #33, opened 2025-02-19
RFC: Database schema for slabs and sectors

ChrisSchinnerl:
Adds tables relevant to storing slabs and sectors as well as pinning sectors

### DIFF
--- a/doc/rfc/001_db_schema.md
+++ b/doc/rfc/001_db_schema.md
@@ -172,14 +172,14 @@ CREATE TABLE slabs (
 
 CREATE TABLE sectors (
     id BIGSERIAL PRIMARY KEY,
-    slab_id BIGINT REFERENCES slabs(id) NOT NULL ON DELETE CASCADE,
+    slab_id BIGINT REFERENCES slabs(id) NOT NULL,
     slab_index SMALLINT NOT NULL, -- index within corresponding slab to retrieve sectors in right order
     root BYTEA UNIQUE NOT NULL
 )
 
 CREATE TABLE host_sectors (
     host_id INTEGER REFERENCES hosts(id) NOT NULL,
-    sector_id BIGINT REFERENCES sectors(id) NOT NULL ON DELETE CASCADE,
+    sector_id BIGINT REFERENCES sectors(id) NOT NULL,
     PRIMARY KEY (host_id, sector_id), -- a sector should only exist once per host
 
     contract_id INTEGER REFERENCES contracts(id) ON DELETE SET NULL, -- determines whether the sector is pinned
--- a/doc/rfc/001_db_schema.md
+++ b/doc/rfc/001_db_schema.md
@@ -170,32 +170,17 @@ CREATE TABLE slabs (
 
 CREATE TABLE sectors (
     id BIGSERIAL PRIMARY KEY,
+    sector_root BYTEA UNIQUE NOT NULL
+
+    -- uploading
+    host_id INTEGER REFERENCES hosts(id) NOT NULL, -- host that stores sector
+    contract_id INTEGER REFERENCES contracts(id), -- null if not pinned
+    uploaded_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW() -- allow sorting by upload time
+
+    -- slab
     slab_id BIGINT REFERENCES slabs(id) NOT NULL,
     slab_index SMALLINT NOT NULL, -- index within corresponding slab to retrieve sectors in right order
-    root BYTEA UNIQUE NOT NULL
-)
--- enforce one sector per index per slab
-CREATE UNIQUE INDEX sectors_slab_id_slab_index_idx ON sectors(slab_id, slab_index ASC)
-```
-
-CREATE TABLE host_sectors (
-    id BIGSERIAL PRIMARY KEY,
-    host_id INTEGER REFERENCES hosts(id) NOT NULL,
-
-    -- NULL if sector isn't pinned yet
-    contract_id INTEGER REFERENCES contracts(id),
-
-    -- a sector should only ever be stored on one host and pinned to one contract
-    -- which is why this is UNIQUE. If a sector is migrated, the previous row is
-    -- overwritten.
-    sector_id BIGINT REFERENCES sectors(id) UNIQUE NOT NULL,
-
-    -- NOTE: instead of expiration, we track the upload time and update
-    -- contract_id after successfully pinning them or when the host reports that
-    -- they don't have the sector. That way, we can still prioritize sectors that
-    -- expire soon but might get lucky when pinning after what we thought was the
-    -- expiration date.
-    uploaded_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
+    UNIQUE(slab_id, slab_index) -- enforce one sector per index per slab
 )
 -- quick lookup of sectors to pin prioritized by upload time
 CREATE INDEX host_sectors_contract_id_uploaded_at_idx ON host_sectors(contract_id, uploaded_at ASC)
--- a/doc/rfc/001_db_schema.md
+++ b/doc/rfc/001_db_schema.md
@@ -182,7 +182,7 @@ CREATE TABLE host_sectors (
     sector_id BIGINT REFERENCES sectors(id) NOT NULL,
     PRIMARY KEY (host_id, sector_id), -- a sector should only exist once per host
 
-    contract_id INTEGER REFERENCES contracts(id) ON DELETE SET NULL, -- determines whether the sector is pinned
+    contract_id INTEGER REFERENCES contracts(id) ON DELETE, -- determines whether the sector is pinned
 
     -- NOTE: instead of expiration, we track the upload time and remove sectors
     -- after successfully pinning them or when the host reports that they don't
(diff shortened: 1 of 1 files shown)


doc/rfc/001_db_schema.md
```diff
@@ -146,3 +146,50 @@ CREATE TABLE host_settings (
...
+
+CREATE TABLE sectors (
+    id BIGSERIAL PRIMARY KEY,
+    slab_id BIGINT REFERENCES slabs(id) NOT NULL ON DELETE CASCADE,
+    slab_index SMALLINT NOT NULL, -- index within corresponding slab to retrieve sectors in right order
+    root BYTEA UNIQUE NOT NULL
+)
+
+CREATE TABLE host_sectors (
+    host_id INTEGER REFERENCES hosts(id) NOT NULL ON DELETE CASCADE,
```
n8mgr: Absolutely do not want a CASCADE on any sector data table. It should be explicitly cleaned up or throw a FK error. Foreign keys are about enforcing relationships, ensuring we are deleting those relationships explicitly is part of it. Throwing that away is a waste.

doc/rfc/001_db_schema.md
```diff
@@ -146,3 +146,50 @@ CREATE TABLE host_settings (
...
+    slab_index SMALLINT NOT NULL, -- index within corresponding slab to retrieve sectors in right order
+    root BYTEA UNIQUE NOT NULL
+)
+
+CREATE TABLE host_sectors (
+    host_id INTEGER REFERENCES hosts(id) NOT NULL,
+    sector_id BIGINT REFERENCES sectors(id) NOT NULL,
+    PRIMARY KEY (host_id, sector_id), -- a sector should only exist once per host
+
+    contract_id INTEGER REFERENCES contracts(id) ON DELETE SET NULL, -- determines whether the sector is pinned
```
n8mgr: I feel similarly about `SET NULL` here. Forcing devs to be explicit about destroying the relationship is a good thing.

n8mgr (changes requested):

doc/rfc/001_db_schema.md
```diff
@@ -146,3 +146,50 @@ CREATE TABLE host_settings (
...
+
+### 2.5 Slabs and Sectors
+
+```postgresql
+CREATE TABLE slabs (
+    id BIGSERIAL PRIMARY KEY, -- internal db id
+
+    digest BYTEA UNIQUE NOT NULL, -- unique identifier for the slab derived from sector roots
+    encryption_key BYTEA NOT NULL,
+    health SMALLINT NOT NULL DEFAULT 100 CHECK(health >= 0 AND health <= 100), -- 0 unhealthy, 100 healthy
```
n8mgr: Should this be stored on the slab or should this be something like `healthy BOOL` in the `sectors` table since part of the health code will be using `RPCVerify` on the host.

Assuming there are at most 255 sectors per shard and we don't need to do object health with `n` slabs, it should be a pretty efficient query with a single level of joins.

doc/rfc/001_db_schema.md
```diff
@@ -146,3 +146,50 @@ CREATE TABLE host_settings (
...
+    slab_index SMALLINT NOT NULL, -- index within corresponding slab to retrieve sectors in right order
+    root BYTEA UNIQUE NOT NULL
+)
+
+CREATE TABLE host_sectors (
+    host_id INTEGER REFERENCES hosts(id) NOT NULL,
+    sector_id BIGINT REFERENCES sectors(id) NOT NULL,
+    PRIMARY KEY (host_id, sector_id), -- a sector should only exist once per host
+
+    contract_id INTEGER REFERENCES contracts(id) ON DELETE, -- determines whether the sector is pinned
```
n8mgr: `ON DELETE` should be removed

doc/rfc/001_db_schema.md
```diff
@@ -146,3 +146,50 @@ CREATE TABLE host_settings (
...
+
+    -- NOTE: instead of expiration, we track the upload time and remove sectors
+    -- after successfully pinning them or when the host reports that they don't
+    -- have the sector. That way, we can still prioritize sectors that expire
+    -- soon but might get lucky when pinning after what we thought was the
+    -- expiration date.
```
n8mgr: > and remove sectors after successfully pinning them

Is this accurate even with the FK on `contract_id`? I could be misunderstanding, but this feels contradictory.

doc/rfc/001_db_schema.md
```diff
@@ -111,21 +111,21 @@ CREATE TABLE hosts (
...
     id SERIAL PRIMARY KEY,
-    host_id INTEGER REFERENCES hosts(id) ON DELETE CASCADE,
+    host_id INTEGER REFERENCES hosts(id) NOT NULL ON DELETE CASCADE,
 
     cidr CIDR NOT NULL
 )
 
 CREATE TABLE host_settings (
-    host_id INTEGER PRIMARY KEY REFERENCES hosts(id) ON DELETE CASCADE,
+    host_id INTEGER PRIMARY KEY REFERENCES hosts(id) NOT NULL ON DELETE CASCADE,
```
peterjan: NOT NULL is implicit on PKs

doc/rfc/001_db_schema.md
```diff
@@ -146,3 +146,50 @@ CREATE TABLE host_settings (
...
+### 2.5 Slabs and Sectors
+
+```postgresql
+CREATE TABLE slabs (
+    id BIGSERIAL PRIMARY KEY, -- internal db id
+
+    digest BYTEA UNIQUE NOT NULL, -- unique identifier for the slab derived from sector roots
+    encryption_key BYTEA NOT NULL,
+    health SMALLINT NOT NULL DEFAULT 100 CHECK(health >= 0 AND health <= 100), -- 0 unhealthy, 100 healthy
+    health_refresh_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT '0001-01-01 00:00:00+00',
```
peterjan: `next_health_refresh` to be consistent with `next_scan`? or does this represent the time at which the slab health was refreshed

doc/rfc/001_db_schema.md
```diff
@@ -111,21 +111,21 @@ CREATE TABLE hosts (
...
+    host_id INTEGER PRIMARY KEY REFERENCES hosts(id) NOT NULL ON DELETE CASCADE,
 
     net_address TEXT NOT NULL,
     protocol SMALLINT NOT NULL
 )
 
 CREATE TABLE host_resolved_cidrs (
     id SERIAL PRIMARY KEY,
-    host_id INTEGER REFERENCES hosts(id) ON DELETE CASCADE,
+    host_id INTEGER REFERENCES hosts(id) NOT NULL ON DELETE CASCADE,
```
peterjan:
```suggestion
    host_id INTEGER NOT NULL REFERENCES hosts(id) ON DELETE CASCADE,
```

doc/rfc/001_db_schema.md
```diff
@@ -111,21 +111,21 @@ CREATE TABLE hosts (
 
 CREATE TABLE host_addresses (
     id SERIAL PRIMARY KEY,
-    host_id INTEGER PRIMARY KEY REFERENCES hosts(id) ON DELETE CASCADE,
+    host_id INTEGER PRIMARY KEY REFERENCES hosts(id) NOT NULL ON DELETE CASCADE,
```
peterjan, line 114:
```suggestion
    host_id INTEGER NOT NULL PRIMARY KEY REFERENCES hosts(id) ON DELETE CASCADE,
```

doc/rfc/001_db_schema.md
peterjan: probably worth throwing all sql in chat gpt to fix syntax errors, there's missing semi colons and couple of other inconsistencies

doc/rfc/001_db_schema.md
ChrisSchinnerl: Nope, that's outdated now

n8mgr (dismissed):

doc/rfc/001_db_schema.md
```diff
@@ -146,3 +146,57 @@ CREATE TABLE host_settings (
...
+-- enforce one sector per index per slab
+CREATE UNIQUE INDEX sectors_slab_id_slab_index_idx ON sectors(slab_id, slab_index ASC)
+```
+
+CREATE TABLE host_sectors (
+    id BIGSERIAL PRIMARY KEY,
+    host_id INTEGER REFERENCES hosts(id) NOT NULL,
+
+    -- NULL if sector isn't pinned yet
+    contract_id INTEGER REFERENCES contracts(id),
```
n8mgr: Do `host_id` and `contract_id` benefit from being in a separate table now or can they be moved into `sectors` without a problem? There's a good argument from a joins perspective, but I'm not confident it won't mess up something else.

```sql
CREATE TABLE sectors (
    id BIGSERIAL PRIMARY KEY,
    sector_root BYTEA UNIQUE NOT NULL,
     
    -- slab
    slab_id BIGINT REFERENCES slabs(id) NOT NULL,

    host_id INTEGER REFERENCES hosts(id) NOT NULL,
    contract_id INTEGER REFERENCES contracts(id), -- NULL if not pinned yet

    uploaded_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(), 

    UNIQUE(slab_id, slab_index) -- enforce one sector per index per slab
);
```

ChrisSchinnerl: Ngl I thought about that and it gave me a bad feeling but now that we don't do duplicate sectors anymore there might really not be a reason against it.

I'll update it and we'll see where that takes us.

n8mgr (approved):

PR #34, opened 2025-02-20
RFC:  Data Migration

ChrisSchinnerl:
This RFC describes the health code and migration of unhealthy slabs

### DIFF
--- a/doc/rfc/0007_data_migration.md
+++ b/doc/rfc/0007_data_migration.md
@@ -11,42 +11,19 @@ unhealthy slabs from the database and repairs them.
 
 ### Health
 
-The health of a slab is determined by the percentage of available parity shards.
-That makes it easy to compare the health of slabs of various redundancy
-settings. A slab with fewer redundancy shards will drop in health faster when a
-host goes bad than one with many and will therefore be repaired sooner.
+The health of a slab is determined by its likelihood of becoming unrecoverable.
+In `renterd` this was a relatively hard and expensive to compute metric since it
+accounts for sectors that are stored redundantly on multiple hosts. The database
+schema of `indexd` is more straightforward. Since we only ever keep a single row
+per sector in the `sectors` table of our database, we can simply determine the
+number of shards until failure by counting the result of joining `slabs.id` with
+`sectors.slab_id` and `sectors.contract_id` with `contracts.id` and filtering for
+`sectors.host_id != NULL AND contract.is_good == TRUE`. Subtracting
+`slabs.min_shards` from that number gives us the remaining shards until failure.
 
-Computing the health is achieved with the following algorithm which is
-periodically applied to all slabs that have a health_valid_until that is in the
-past:
-
-1. Retrieve all rows from `host_sectors` that are either not pinned yet
-(`contract_id` is `NULL`) or have a good contract (`is_good` is `TRUE`).
-2. Make sure they are sorted by their index within the slab and their
-`uploaded_at` time in descending order (to prioritize the more recent ones).
-3. Loop over the sectors and increment a counter for each sector that:
-    - Has a `slab_index` for which we counted a good sector yet.
-    - Has a `host_id` for which we haven't counted a sector yet.
-    - Has an IP subnet for which we haven't counted a sector yet.
-4. Compute the health of the slab using the formula:
-    `health = 100 * (good_shards - min_shards / (total_shards - min_shards))`
-5. Update the health of the slab in the database:
-    - If it is < 0 set it to -1 so that all lost slabs have the same health.
-    - If it is > 100, search for the bug that is causing it because it should never happen.
-
-With all of the above in place, the database also exposes a `slabs, nextRefresh := UnhealthySlabs(lastRepairAttemptCutoff,  limit)` method which returns up to `limit` slabs for which the following holds true:
-- The slabs haven't had a repair attempated on them since `lastRepairAttemptCutoff`.
-- The slabs are sorted from most unhealthy to least unhealthy.
-
-The point of the former is to allow the repair loop to exhaust the list of
-unhealthy slabs without getting stuck on unrepairable slabs. The repair time is
-set to NULL upon a successful repair or to the current batches start time upon a
-failed repair. A failed repair won't be attempted for another another.
-
-The point of `nextRefresh` is to provide a marker for the repair loop. If
-`UnhealthySlabs` returns fewer than `limit` slabs, the repair loop knows to
-sleep until `nextRefresh` since there won't be any new slabs to repair until the
-health is updated again.
+Fetching slabs for repair is then as simple as querying the health of all slabs,
+ordered by the number of shards until failure and limiting the results to a
+number just enough to saturate our repair loop such as `10`.
 
 ### Repair Loop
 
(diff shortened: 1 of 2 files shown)


n8mgr (dismissed):

doc/rfc/0007_data_migration.md
```diff
@@ -0,0 +1,97 @@
...
+(`contract_id` is `NULL`) or have a good contract (`is_good` is `TRUE`).
+2. Make sure they are sorted by their index within the slab and their
+`uploaded_at` time in descending order (to prioritize the more recent ones).
+3. Loop over the sectors and increment a counter for each sector that:
+    - Has a `slab_index` for which we counted a good sector yet.
+    - Has a `host_id` for which we haven't counted a sector yet.
+    - Has an IP subnet for which we haven't counted a sector yet.
+4. Compute the health of the slab using the formula:
+    `health = 100 * (good_shards - min_shards / (total_shards - min_shards))`
+5. Update the health of the slab in the database:
```
n8mgr: Do we need to compute and store this? It made sense in `renterd`, imo, because the joins were attrocious, but here we don't have the object to worry about so the query should be simpler. 

Alternatively, it might make this process simpler and more efficient to consider health of the whole contract rather than the slab: if the contract is bad, all the sectors stored in that contract are in need of migration.

n8mgr: This is half assed, but it may help to make my thoughts clearer and might overlap with ideas or expectations you had that aren't in this RFC.

```sql
CREATE TABLE host_sectors (
  sector_id BIGINT REFERENCES sectors(id),
  contract_id INT REFERENCES contracts(id),
  last_health_check TIMESTAMP WITH TIME ZONE
);

CREATE TABLE contracts (
  last_success_health_check TIMESTAMP WITH TIME ZONE,
  next_health_check TIMESTAMP WITH TIME ZONE,
  failed_health_checks
);
```

1. Grab a bunch of contracts needing a health check
```sql
SELECT c.id FROM contracts
ORDER BY next_health_check ASC
LIMIT 50;
```
2. Spawn a goroutine for each contract
3. Grab a sector without a recent check
```sql
SELECT s.root
FROM host_sectors hs 
INNER JOIN sectors s ON (hs.sector_id=s.id)
WHERE hs.contract_id=$1
ORDER BY last_health_check ASC
LIMIT 1
```
4. Call RPCVerify for the sector
5. Use the results of the RPC and whatever other metrics to determine if it's time to mark the contract `is_good=false`

The repair loop can check for any slabs pinned to a contract with is_good=false and without a recent repair attempt.

If you wanted to display the health of a slab, it's something like:
```sql
SELECT slab_id, COUNT(*) FILTER (WHERE c.is_good = false) AS bad_slabs
FROM host_sectors hs
INNER JOIN sector s 
INNER JOIN contracts c
GROUP BY slab_id
```

ChrisSchinnerl: Just some renterd history: `renterd` started without persistence too. At first we didn't have any, then the `health` on slabs and finally on `objects` as well. The queries for updating the health of slabs and objects are also separate. At first we updated all slabs and then all affected objects. So the `objects` didn't have anything to do with the introduction of the `health` on `slabs` directly. That said the query was/is still atrocious and on larger renters that have millions and tens of millions of slabs, just too slow.

It seems like what you are describing is what I'd refer to as the "data integrity checks" for which I intended a separate RFC. Essentially the process of periodically scanning contracts for lost sectors, removing those from `host_sectors` and somehow remembering that a host lost sectors. However, I didn't expect that we would start on that before the MVP is done.

My intention for this RFC was to solely determine the health of slabs and whether they need to be migrated based off of whether they are stored on good contracts and distributed across subnets. So that e.g. if 10 hosts go offline, we start migrating slabs away from their contracts because some slabs might have lost up to 50% of their parity at 10/30 EC. Not saying we need to persist the `health` for that as long as it is fast to retrieve slabs to repair.

Additional considerations which might not be perfectly outlined here but affect the complexity of determining the health of a slab:
- `host_sectors` on the same CIDR are possible and should only count as 1 good sector
- the same sector can be part of more than 1 good contract but should only count as 1 good sector
- contracts can go from good -> bad -> good (e.g. host stop accepting contracts and starts accepting again)
- due to the other reasons a sector being on a bad contract doesn't mean it requires migration
- we generally don't want to repair a slab as soon as a single sector is bad but only when a certain percentage goes bad

doc/rfc/001_db_schema.md
```diff
@@ -167,6 +168,7 @@ CREATE TABLE slabs (
     encryption_key BYTEA NOT NULL,
     health SMALLINT NOT NULL DEFAULT 100 CHECK(health >= 0 AND health <= 100), -- 0 unhealthy, 100 healthy
     health_refresh_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT '0001-01-01 00:00:00+00',
+    last_repair_attempt TIMESTAMP WITH TIME ZONE DEFAULT NULL, -- NULL if no repair has failed yet
```
peterjan: I'd avoid `NULL` and use `'0001-01-01 00:00:00+00'`

doc/rfc/0007_data_migration.md
```diff
@@ -0,0 +1,97 @@
+# Data Migration
```
peterjan, line 1: I think Slab Maintenance covers the load better. It also fits in with wallet and contract maintenance.

doc/rfc/0007_data_migration.md
n8mgr: I think we need to carefully consider the scalability of a system that expects to update millions of rows over a relatively short period of time. 

In my opinion, data integrity checks are a better fit for the MVP because data integrity should be the primary driver for repairs.

If we're thinking about a health system, there are really only three reasons to migrate, right? All related to the contract/host rather than individual slabs. If part of the contract data is unhealthy, all the slabs are naturally in need of migration. 

1. The contract is expiring soon and failing to renew
2. The host failed to provide correct data
3. The host is offline frequently

One and two are cut and dry enough that I'd feel comfortable saying once a contract is marked as unhealthy, it shouldn't come back.

It's definitely a big question that doesn't have a clear or easy answer.

> host_sectors on the same CIDR are possible and should only count as 1 good sector

If we pick separate CIDRs at upload, will the occasional CIDR shift have a significant effect to warrant a check here too? This should be rare enough that I'm not sure it's worthwhile.

> the same sector can be part of more than 1 good contract but should only count as 1 good sector

We talked about having a one-to-many relationship for sectors and contracts, but how many times has that saved data for users? It feels necessary, but is it because of historically inefficient repairs? Is it worth the added health complexity?

> we generally don't want to repair a slab as soon as a single sector is bad but only when a certain percentage goes bad

Is there history behind this decision? If it's because of cost or churn, I would argue that we don't really have good data on cost because our health algorithms have historically been overly aggressive towards pushing hosts in and out.

doc/rfc/0007_data_migration.md
```diff
@@ -0,0 +1,97 @@
...
+## Abstract
+
+Data migration is the process of keeping slab data alive on the network. If a
+slab is missing pieces, we need to download the minimum number of shards to
+restore the missing ones and then upload the missing shards to new hosts. This
+process consists mainly of 2 parts. The first one is determining the health of a
+slab while the second one is the actual repair loop which periodically fetches
+unhealthy slabs from the database and repairs them.
+
+### Health
```
n8mgr, line 12: I think we ended up deciding on simplifying this section to purely the health query on a slab

doc/rfc/001_db_schema.md
```diff
@@ -184,6 +184,7 @@ CREATE TABLE sectors (
     slab_index SMALLINT NOT NULL, -- index within corresponding slab to retrieve sectors in right order
     UNIQUE(slab_id, slab_index) -- enforce one sector per index per slab
 )
+CREATE INDEX sectors_host_id_idx ON sectors(host_id)
```
n8mgr: Foreign keys get automatic indices in postgres

doc/rfc/001_db_schema.md
```diff
@@ -166,7 +166,7 @@ CREATE TABLE slabs (
 
     digest BYTEA UNIQUE NOT NULL, -- unique identifier for the slab derived from sector roots
     encryption_key BYTEA NOT NULL,
-    last_repair_attempt TIMESTAMP WITH TIME ZONE DEFAULT NULL, -- NULL if no repair has failed yet
+    last_repair_attempt TIMESTAMP WITH TIME ZONE DEFAULT DEFAULT '0001-01-01 00:00:00+00',
```
n8mgr: Should this be not null?

n8mgr (approved):

PR #36, opened 2025-02-20
RFC: Contracts schema

ChrisSchinnerl:


doc/rfc/001_db_schema.md
```diff
@@ -152,8 +152,35 @@ CREATE TABLE host_settings (
 ```postgresql
 CREATE TABLE contracts (
   id SERIAL PRIMARY KEY
-
-  -- TODO: f/u with more fields in a separate PR
+  host_id INTEGER REFERENCES hosts(id) NOT NULL,
+  fcid BYTEA REFERENCES NATE IS A JERK FOR REVIEWING DRAFTS ON DELETE CASCADE,
```
n8mgr: 👀

n8mgr (changes requested):

doc/rfc/001_db_schema.md
ChrisSchinnerl: 67d05bbf3c35f88b82a4c19ea16d270694721395a658f0cf2d2870799c2c3c04_1-300x260

doc/rfc/001_db_schema.md
```diff
@@ -152,8 +152,34 @@ CREATE TABLE host_settings (
...
+  archived BOOLEAN NOT NULL DEFAULT FALSE, -- whether the contract is archived
+  proof_height BIGINT NOT NULL, -- start of proof window
+  expiration_height BIGINT NOT NULL, -- end of proof window
+  renewed_from BYTEA DEFAULT NULL,
+  renewed_to BYTEA DEFAULT NULL,
+  state SMALLINT NOT NULL, -- 0 = pending, 1 = active, 2 = expired
+
+  -- metrics for visualization (not ACID)
+  capacity BIGINT NOT NULL DEFAULT 0 CHECK(capacity >= size),
+  size BIGINT NOT NULL DEFAULT 0,
```
peterjan: might as well `CHECK >= 0` here

doc/rfc/001_db_schema.md
```diff
@@ -152,8 +152,34 @@ CREATE TABLE host_settings (
...
+
+  -- costs
+  contract_price DECIMAL(50, 0) NOT NULL, -- used to display cost of forming contract
+  initial_allowance DECIMAL(50, 0) NOT NULL, -- used when refreshing contract to increase budget
```
peterjan, line 184: `contract_budget` or `contract_initial_budget`? I like budget better than allowance

doc/rfc/001_db_schema.md
```diff
@@ -152,8 +152,34 @@ CREATE TABLE host_settings (
 ```postgresql
 CREATE TABLE contracts (
   id SERIAL PRIMARY KEY
-
-  -- TODO: f/u with more fields in a separate PR
+  host_id INTEGER REFERENCES hosts(id) NOT NULL,
+  contract_id BYTEA NOT NULL UNIQUE,
```
peterjan, line 169: This is a little confusing when writing queries imo. It looks like a ref to the contracts table. I'd go for `fcid` or `contract_key`.

doc/rfc/001_db_schema.md
ChrisSchinnerl: RHP4 calls it allowance which is why I decided to stay consistent here to avoid introducing too much terminology

doc/rfc/001_db_schema.md
ChrisSchinnerl: I don't think we will be confused by this considering it is on the `contracts` table and if we are the query doesn't work and we fix it. Apart from `renterd` it seems like we always use `contract_id` in our schemas.

n8mgr (approved):

doc/rfc/001_db_schema.md
```diff
@@ -152,8 +152,34 @@ CREATE TABLE host_settings (
...
+
+  -- lifetime related columns
+  archived BOOLEAN NOT NULL DEFAULT FALSE, -- whether the contract is archived
+  proof_height BIGINT NOT NULL, -- start of proof window
+  expiration_height BIGINT NOT NULL, -- end of proof window
+  renewed_from BYTEA DEFAULT NULL,
```
n8mgr: Can do `renewed_from INTEGER REFERENCES contracts(id)` and `renewed_to INTEGER REFERENCES contracts(id)`

doc/rfc/001_db_schema.md
```diff
@@ -152,8 +152,34 @@ CREATE TABLE host_settings (
...
+  renewed_to BYTEA DEFAULT NULL,
+  state SMALLINT NOT NULL, -- 0 = pending, 1 = active, 2 = expired
```
n8mgr: Need one for `rejected` and `resolved`

doc/rfc/001_db_schema.md
```diff
@@ -152,8 +152,34 @@ CREATE TABLE host_settings (
 ```postgresql
 CREATE TABLE contracts (
   id SERIAL PRIMARY KEY
-
-  -- TODO: f/u with more fields in a separate PR
+  host_id INTEGER REFERENCES hosts(id) NOT NULL,
+  contract_id BYTEA NOT NULL UNIQUE,
+
+  -- lifetime related columns
+  archived BOOLEAN NOT NULL DEFAULT FALSE, -- whether the contract is archived
```
n8mgr: I don't think we need both `archived` and `state` if we have an index on `state`

n8mgr (changes requested):

doc/rfc/001_db_schema.md
ChrisSchinnerl: Just to clarify `archived` is meant for the process of deleting a contract. e.g. if a user manually deletes it or if we hit the proof window without renewing the contract. Effectively making sure we don't use the contract anymore.

A contract could still be archived while also being `resolved` or `expired`. We could overload the `state` and add an `archived` state or something, but we'd lose some information about the contract since it might still resolve or not. Actually if it expires we probably want to know about that to unlock the funds.

n8mgr: My opinion is to not plan on deleting contracts until we have a reason to. If we can filter them performantly, there’s no need to delete them.

ChrisSchinnerl: Well maybe "delete" is the wrong word. We archive the contract so no information is lost. 

I added a new state to cover the case where we usually archive a contract to prevent usage. `awaiting proof`. `renterd` effectively does that 144 blocks before the window start since the host will refuse to revise a contract at that point.

n8mgr: Instead of another column we can just use the query, right? With an index, there's no issue. 

`SELECT * FROM contracts WHERE proof_height > 144`

There's a balance between adding additional columns for optimizing query performance and managing the overhead of static state on a table.

If you go with a static check, you lose some granularity to change the criteria for a "good contract" in the future and you need to handle additional edge cases where you need to revert the state.

n8mgr: I'm not necessarily saying this is a good example of "bad static state," depending on how complex the checks are going to get, but having a dynamic check is something to consider.

ChrisSchinnerl: I removed the state. But just so that we are on the same page, it was not about performance. I don't expect this to be slow. Just the ability to manually (e.g. click contract in UI and hit "archive") archive contracts.

n8mgr (approved):

PR #38, opened 2025-02-20
Add TestWalletAPI

peterjan:


api/api_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"go.sia.tech/coreutils/wallet"
+	"go.sia.tech/indexd/internal/testutils"
+)
+
+func TestWalletAPI(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	// create cluster without hosts
+	cluster := testutils.NewCluster(t, testutils.WithHosts(0))
```
peterjan: I had a bare-bones approach but the required setup is so similar to the cluster I figured there's no really point in not using it or trying to avoid it. I would personally refactor the cluster a bit though, maybe we should call it a `Network`..  hosts and indexers can be created on the network. Hosts can be announced, blocks can be mined on the network, but the type doesn't keep track of everything. It's just there to hook up syncers and ensure mined blocks get broadcasted successfully.

n8mgr: I would disagree -- it's not necessarily beneficial to avoid the duplication at all cost in tests, IMO. Cluster/Indexer will eventually start a bunch of background threads that may interact with the wallet in unanticipated ways and cause NDFs or unexpected races. Being in the habit of layering up makes it easier to find bugs, fix NDFs, and gives you more overall coverage of different scenarios.

Using this PR as purely an example, not necessarily saying anything should be changed, we've tested that the API works here and the `SingleAddressWallet` is relatively well tested in `coreutils`, but there's a layer between the API and the database impl with effectively zero coverage right now. That may be fixed in the future, but, to me, it's easier to start small, build up, and be thorough now than assume we'll go back and prioritize adding coverage later.

To me, for the API tests, the `Indexer` struct makes a lot of sense, there's an additional layer of standalone `SingleAddressWallet` tests missing, and `Cluster` is a very hard pill for me to swallow.

api/api_test.go
```diff
@@ -0,0 +1,95 @@
...
+		t.Fatal("expected pending transaction")
+	} else if pending[0].Type != wallet.EventTypeV2Transaction {
+		t.Fatal("unexpected transaction type", pending[0].Type)
+	} else if pending[0].ID != types.Hash256(txnID) {
+		t.Fatal("expected transaction id to match")
+	}
+
+	// mine a block and allow for it to be processed
+	cluster.Indexer.MineBlocks(t, types.Address{}, 1)
+	time.Sleep(time.Second)
```
n8mgr: You can make this more robust by checking the last indexed tip against `ChainManager.Tip()` instead of picking a random amount of time. It could even be added to `Indexer.MineBlocks`. It's basically a purpose-built `Retry`.

```go
func WaitForBlock(tb testing.TB, cm *chain.Manager, waiter interface{ Tip() (types.ChainIndex, error)}) {
	for i := 0; i < 1000; i++ {
		time.Sleep(time.Millisecond)
		tip, _ := waiter.Tip()
		if tip == cm.Tip() {
			return
		}
	}
	tb.Fatal("timed out waiting for block")
}
```

api/api_test.go
peterjan: For me it's always been a convenience trade-off. The host is (roughly) a sector store away from being a wallet and a hostless cluster is essentially an indexer, but I agree that over time it would've suffered from more and more complexity being added. I added a test wallet and removed the cluster in favour of a plain indexer. I'm probably not fully hitting the spot you wanted to hit yet, if so please elaborate a little bit more. I thought this was a pretty good middle ground though.

api/api_test.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	// create a wallet
+	w := testutils.NewWallet(t, network, genesis)
+	err = w.Connect(ctx, indexer.SyncerAddr())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// assert host wallet is empty
```
ChrisSchinnerl, line 64:
```suggestion
	// assert wallet is empty
```

internal/testutils/wallet.go
```diff
@@ -0,0 +1,96 @@
...
+	// Wallet is an ephemeral wallet that can be used for testing.
+	Wallet struct {
+		*wallet.SingleAddressWallet
+		ws *testutil.EphemeralWalletStore
+		s  *syncer.Syncer
+	}
+)
+
+// NewWallet creates a new ephemeral wallet.
+func NewWallet(t testing.TB, n *consensus.Network, genesis types.Block) *Wallet {
```
ChrisSchinnerl: I might have gone a slightly different route like this:

```go
// NewWallet creates a new ephemeral wallet.
func NewWallet(t testing.TB, cm *chain.Manager) *Wallet {
	ws := testutil.NewEphemeralWalletStore()
	w, err := wallet.NewSingleAddressWallet(types.GeneratePrivateKey(), cm, ws)
	if err != nil {
		t.Fatal(err)
	}
	t.Cleanup(func() { w.Close() })

	reorgCh := make(chan struct{}, 1)
	t.Cleanup(func() { close(reorgCh) })
	go func() {
		for range reorgCh {
			reverted, applied, err := cm.UpdatesSince(w.Tip(), 1000)
			if err != nil {
				panic(err)
			}

			if err := ws.UpdateChainState(func(tx wallet.UpdateTx) error {
				return w.UpdateChainState(tx, reverted, applied)
			}); err != nil {
				panic(err)
			}
		}
	}()
	stop := cm.OnReorg(func(index types.ChainIndex) {
		select {
		case reorgCh <- struct{}{}:
		default:
		}
	})
	t.Cleanup(stop)

	return &Wallet{
		SingleAddressWallet: w,
		store:               ws,
	}
}
```
and then make us of that. This can also be used in `NewIndexer` to construct the wallet now.

Theoretically you can also expose a convenience method on `Indexer` itself like `(I *Indexer) NewWallet` to create a wallet from the same chain manager which calls the `NewWallet` above. 

Overall it would be nice for all those little helpers we add to eventually add layer after layer until we get to the cluster. Reusing each other so that you get to pick the ideal level of abstraction for your test.

peterjan: > This can also be used in NewIndexer to construct the wallet now.

I don't think this is true because in the indexer we want one store and the indexer has a subscriber that processes all chain updates, so covering host announcements and wallet updates. 

> Theoretically you can also expose a convenience method on Indexer itself like (I *Indexer) NewWallet to create a wallet from the same chain manager which calls the NewWallet above.

This feels similar to `indexer.MineBlocks`, in my head it's not supposed to be there. I don't see why we should try and avoid having a chain manager and syncer for every test node that wants to join the party. That way we ensure we take sync delays and block propagation between syncers into account as well.

I think what I proposed earlier, switching `Cluster` for a `Network` could work well. For this PR though, I don't mind changing it but I was wondering if you feel the same way after reading the above comments.

ChrisSchinnerl: That might be true. Probably fixable by passing the store though.

The reason why reusing these components is nice is the fact that it reduces concurrency issues. If e.g. the whole cluster only uses a single chain manager, mining a block and applying it to the chain manager directly rules out any timing issues that might occur due to the syncer loops.

If you take it a step further and immediately apply every mined block of a cluster to every component that is interested in it (although I'm not sure if we can do that for all components) you'd eliminate the need to wait for blocks in our entire test suite. Ruling out what is probably the biggest source of NDFs.

peterjan: I'd like to F/U with a PR that refactors the `testutils`. It's somewhat in scope here, but not really. I'm also OOO starting tomorrow, fine to keep this open though if you want. Passing in the store and chain manager hurts when I write it out. Most importantly though I agree with what you are saying, I feel we're not 100% there yet with the types and how they interact with each other.

ChrisSchinnerl (dismissed):

n8mgr (approved):

PR #41, opened 2025-02-21
Add host scanning

peterjan:
This PR adds host scanning. It's quite big so I left a couple of TODOs. There's one missing 🧪 , which is an integration test that verifies the scan loop properly executes and (re)scans hosts.


hosts/manager.go
```diff
@@ -27,30 +51,210 @@ type (
...
+	} else if m.scanInterval == 0 {
+		return nil, fmt.Errorf("scanInterval can not be zero")
 	}
 
+	m.scheduleNextScan(0) // trigger first scan immediately
 	return m, nil
 }
 
+func (m *HostManager) scan(ctx context.Context) {
+	defer m.scheduleNextScan(m.scanFrequency) // ensure scan loop never stops
```
ChrisSchinnerl: Imo this is quite a bit harder to read and follow than
```go
for {
  doScanning()

  select {
    case m.tg.Done():
      return
    case <-time.After(interval):
  }
}
```

hosts/manager.go
```diff
@@ -27,30 +51,210 @@ type (
...
+	hosts, err := m.store.HostsForScanning(ctx)
+	if err != nil {
+		m.log.Error("failed to get hosts for scanning", zap.Error(err))
+		return
+	}
+
+	<-m.scanHosts(ctx, hosts)
+}
+
+func (m *HostManager) scanHosts(ctx context.Context, hosts []types.PublicKey) chan struct{} {
```
ChrisSchinnerl: I think we don't need to pass a `ctx` to functions that we always call with a context derived from `tg` with `context.Background` such as here.

hosts/manager.go
```diff
@@ -27,30 +51,210 @@ type (
...
+func (m *HostManager) scan(ctx context.Context) {
+	defer m.scheduleNextScan(m.scanFrequency) // ensure scan loop never stops
+
+	hosts, err := m.store.HostsForScanning(ctx)
+	if err != nil {
+		m.log.Error("failed to get hosts for scanning", zap.Error(err))
+		return
+	}
+
+	<-m.scanHosts(ctx, hosts)
```
ChrisSchinnerl: Returning a channel here is redundant if we always block on it like this. Better to wait for all goroutines to complete within `scanHosts` using a wait group.

hosts/manager.go
```diff
@@ -27,30 +51,210 @@ type (
...
+func (m *HostManager) scanHosts(ctx context.Context, hosts []types.PublicKey) chan struct{} {
+	doneChan := make(chan struct{})
+	if len(hosts) == 0 {
+		m.log.Info("scan skipped")
+		close(doneChan)
+		return doneChan
+	}
+
+	m.log.Info("scan started", zap.Int("hosts", len(hosts)))
+	c := make(chan types.PublicKey, len(hosts))
```
ChrisSchinnerl: No need for a buffered channel. First spin up the workers and then push the data into the channel.

hosts/manager.go
```diff
@@ -27,30 +51,210 @@ type (
...
+	if err != nil {
+		m.log.Error("failed to get hosts for scanning", zap.Error(err))
+		return
+	}
+
+	<-m.scanHosts(ctx, hosts)
+}
+
+func (m *HostManager) scanHosts(ctx context.Context, hosts []types.PublicKey) chan struct{} {
+	doneChan := make(chan struct{})
```
ChrisSchinnerl: No need for this channel. `scanHosts` can be blocking

hosts/manager.go
```diff
@@ -27,30 +51,210 @@ type (
...
+
+	m.log.Info("scan started", zap.Int("hosts", len(hosts)))
+	c := make(chan types.PublicKey, len(hosts))
+	for _, host := range hosts {
+		c <- host
+	}
+	close(c)
+
+	start := time.Now()
+	activeThreads := int32(m.scanThreads)
```
ChrisSchinnerl: I wouldn't even make this number configurable right away

hosts/manager.go
```diff
@@ -27,30 +51,210 @@ type (
...
+						log.Error("failed to resolve host", zap.Error(err))
+						continue
+					}
+					err = m.scanHost(ctx, host, address)
+					if err != nil {
+						log.Error("failed to scan host", zap.Error(err))
+					}
+				}
+			}
+			if atomic.AddInt32(&activeThreads, -1) == 0 {
```
ChrisSchinnerl: No need to reimplement a `WaitGroup` here. Create a `WaitGroup` add all the threads, wait for it to be done. The return.
You can even get rid of the `m.tg.Add` calls then since this will always wait for all goroutines to finish before returning.

hosts/manager.go
```diff
@@ -1,22 +1,46 @@
...
+	errHostPrivateIP = errors.New("host has a private IP")
+)
+
 type (
+	// Store is the interface for persisting host metadata.
+	Store interface {
+		AddHostScan(ctx context.Context, hk types.PublicKey, hs proto4.HostSettings, nextScan time.Time) error
+		HostsForScanning(ctx context.Context) ([]types.PublicKey, error)
+		HostAddresses(ctx context.Context, hk types.PublicKey) ([]chain.NetAddress, error)
+		UpdateHostResolvedCIDRs(ctx context.Context, hk types.PublicKey, cidrs []string) error
```
ChrisSchinnerl: I'd update the CIDRs atomically with the remaining scan results. I'd really like a full scan to be an atomic change we apply within a single txn to the database for a consistent view of our hosts.

hosts/manager.go
```diff
@@ -27,30 +51,210 @@ type (
...
+	for _, addr := range addrs {
+		if addr.Protocol == rhp4.ProtocolTCPSiaMux {
+			return addr.Address, nil
+		}
+	}
+	return "", errHostNoAddress
+}
+
+func (m *HostManager) scanHost(ctx context.Context, hk types.PublicKey, addr string) error {
+	scanCtx, cancel := context.WithTimeout(ctx, m.scanTimeout)
```
ChrisSchinnerl: 30 second constant? Really trying hard to avoid making stuff configurable this time around that will never ever be updated by users anyway.

hosts/manager.go
```diff
@@ -27,30 +51,210 @@ type (
...
+		return fmt.Errorf("failed to add host scan: %w", err)
+	}
+
+	return nil
+}
+
 // Close closes the manager.
 func (m *HostManager) Close() error {
+	m.cancelNextScan()
+	m.tg.Stop()
```
ChrisSchinnerl, line 123: Shouldn't it be enough to call `m.tg.Stop`? Our scans should be interrupted when `tg` is stopped. We just need to check if `tg` is stopped right before we persist the results to make sure we don't persist failures which are a result of the closed `ThreadGroup`.

hosts/manager.go
```diff
@@ -27,30 +51,210 @@ type (
...
+			return "", fmt.Errorf("ip %q, %w", addr.IP, errHostPrivateIP)
+		}
+	}
+
+	for _, addr := range addrs {
+		if addr.Protocol == rhp4.ProtocolTCPSiaMux {
+			return addr.Address, nil
+		}
+	}
+	return "", errHostNoAddress
```
ChrisSchinnerl: I think `resolveHost` tries to do too many things right now. Imo the only thing it should do is resolve all of the announced `chain.NetAddress`es of a host which potentially contains hostnames and then return a `[]chain.NetAddress` slice that only contains IP addresses. Private IP addresses and addresses for protocols which we don't support can be silently dropped. We don't care about those but I wouldn't punish a host for them.
Even returning an empty slice should be fine. The caller can handle that.

When we scan a host we should also probe every address and protocol. A host that is reachable via Siamux but not via Webtransport is potentially useless for clients and the scanning code is the only place where we have a chance to figure out that the host isn't reachable.

hosts/manager.go
```diff
@@ -27,30 +51,210 @@ type (
...
+	if err != nil {
+		select {
+		case <-ctx.Done():
+			return err
+		default:
+			m.log.Debug("scan failed", zap.Stringer("hk", hk), zap.Error(err))
+		}
+	}
+
+	err = m.store.AddHostScan(ctx, hk, hs, time.Now().Add(m.scanInterval))
```
ChrisSchinnerl: I'd restructure the code a bit for `scanHost` to perform the whole scan. Meaning
1. resolving addresses
2. fetching the settings over the resolved addresses
3. updating the host's CIDRs and settings

These things can be separate functions but the concept of a scan should now include all of these things.

ChrisSchinnerl (changes requested):

hosts/manager.go
n8mgr: +1 for atomicity of host updates. Consider adding a struct to hold all the fields necessary.

hosts/manager.go
n8mgr: No need for a channel at all -- push the work into the goroutine and wait for all of them to finish. You can batch the updates into a slice without using a channel.

hosts/manager.go
n8mgr: +1

hosts/manager.go
n8mgr: +1

hosts/manager.go
```diff
@@ -27,30 +51,210 @@ type (
...
+	if err != nil {
```
n8mgr: Since we're batching scans, could also batch the database update into a single transaction

hosts/manager.go
```diff
@@ -86,6 +290,60 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+
+	for _, block := range privateIPBlocks {
+		if block.Contains(ip) {
+			return true
+		}
+	}
+	return false
+}
+
 func validateAddress(na chain.NetAddress) error {
```
n8mgr, line 379: Should we add this to coreutils?

n8mgr (changes requested):
I feel like the scheduling here is much more complicated than it needs to be. Should be able to simplify, reduce some of the nesting and indirection, get rid of the cancel func on the struct, the mutex, the extra channels. 

```go
go func() {
	ctx, cancel, err := m.tg.AddContext(context.Background())
	if err != nil {
		return
	}
	defer cancel()

	var wg sync.WaitGroup
	for {
		hosts, err := m.store.HostsForScanning(ctx, 50)
		if err != nil {
			m.log.Error("failed to get hosts for scanning", zap.Error(err))
			continue
		}

		for _, host := range hosts {
			wg.Add(1)
			go func(hk types.PublicKey) {
				defer wg.Done()

				// scan
				// resolve
				// update db
			}(host)
		}

		wg.Wait()

		select {
		case <-ctx.Done():
			return
		case <-time.After(m.scanFrequency):
		default:
		}
	}
}()
```

hosts/manager.go
peterjan: resolving this for now, don't see any added value in batch updating hosts

hosts/manager.go
```diff
@@ -87,6 +150,225 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
 	return nil
 }
 
+func (m *HostManager) fetchHost(hk types.PublicKey) (*host, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
```
n8mgr: Leaking `contexts` with this double stacking. Passing in the existing threadgroup context would get rid of it. I get the desire to timeout the db query, but I'm not sure it's necessary or desirable.

hosts/manager.go
```diff
@@ -87,6 +150,225 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+	ctx, cancel = m.tg.WithContext(ctx)
```
n8mgr: threads should only be added at the top-level goroutine, not in every func

hosts/manager.go
```diff
@@ -87,6 +150,225 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+	}
+
+	return &host{
+		addresses: addrs,
+		hk:        hk,
+	}, nil
+}
+
+func (m *HostManager) hostsForScanning() ([]types.PublicKey, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
```
n8mgr: Leaking contexts with this double stacking. Passing in the existing threadgroup context would get rid of it. I get the desire to timeout the db query, but I'm not sure it's necessary or desirable.

hosts/manager.go
```diff
@@ -1,23 +1,47 @@
...
 
 const (
-	// maxAddrsPerProtocol is the maximum number of announced addresses we will
-	// track per host, per protocol
+	ipv4FilterRange = 24
+	ipv6FilterRange = 32
+
 	maxAddrsPerProtocol = 2
+
+	scanThreads = 4
```
n8mgr: Could probably up this to 10-20 or even 50.

hosts/manager.go
```diff
@@ -87,6 +150,225 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+
+				h, err := m.fetchHost(hk)
+				if err != nil {
+					logger.Error("failed to get host", zap.Error(err))
+					continue
+				}
+
+				err = m.scanHost(h)
+				if err != nil {
+					logger.Error("failed to scan host", zap.Error(err))
```
n8mgr: This should probably be debug since it's unlikely to be an actionable error.

hosts/manager.go
```diff
@@ -87,6 +150,225 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+
+	var wg sync.WaitGroup
+	for range scanThreads {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for hk := range queue {
+				logger := m.log.With(zap.Stringer("hk", hk))
+
+				h, err := m.fetchHost(hk)
```
n8mgr: hosts for scanning should return the necessary details instead of just the public key since we're immediately fetching more data from the database regardless. Being cognizant of roundtrips is important since we're dealing with a potentially remote database.

hosts/manager.go
```diff
@@ -27,31 +51,70 @@ type (
...
-	// sanity check options
 	if m.announcementMaxAge == 0 {
 		return nil, fmt.Errorf("announcementMaxAge can not be zero")
+	} else if m.scanFrequency == 0 {
+		return nil, fmt.Errorf("scanFrequency can not be zero")
+	} else if m.scanInterval == 0 {
+		return nil, fmt.Errorf("scanInterval can not be zero")
 	}
 
+	done, err := m.tg.Add()
```
n8mgr: This should really be `AddWithContext` and have that passed into `run` for interupting scans. Then instead of adding more threads at lower levels, you can reuse the existing parent context.

hosts/manager.go
```diff
@@ -87,6 +150,225 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+	if err != nil {
+		m.log.Error("failed to get hosts for scanning", zap.Error(err))
+		return
+	} else if len(hosts) == 0 {
+		m.log.Info("scan skipped")
+		return
+	}
+
+	queue := make(chan types.PublicKey)
+	defer close(queue)
```
n8mgr: The nested loop, channel, and extra goroutine at the bottom are unnecessary. Return `scanThreads` hosts in `hostsForScanning` and spin up `len(hosts)` goroutines.

```go
hosts, err := m.hostsForScanning()
if err != nil {
    m.log.Error("failed to get hosts for scanning", zap.Error(err))
    return
}

var wg sync.WaitGroup
for _, hk := range hosts {
	wg.Add(1)
	go func(ctx context.Context, hk types.PublicKey) {
		defer wg.Done()

        log := m.log.With(zap.Stringer("hk", hk))
        h, err := m.fetchHost(ctx, hk)
        if err != nil {
            log.Error("failed to get host", zap.Error(err))
        } else if err = m.scanHost(ctx, h); err != nil {
            log.Debug("failed to scan host", zap.Error(err))
        } else if err = m.updateHost(ctx, h); err != nil {
            log.Error("failed to update host", zap.Error(err))
        }
    }(ctx, hk)
}
wg.Wait()
```

Your pattern would be better with long running workers if we were trying to minimize downtime between scans, but this isn't a hot path we need maximum efficiency on. Simple concurrency will be easier to maintain and test.

hosts/manager.go
```diff
@@ -87,6 +150,225 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+	case <-m.tg.Done():
+		m.log.Info("host scans interrupted")
+	default:
+		m.log.Info("host scans finished", zap.Int("hosts", len(hosts)), zap.Duration("duration", time.Since(start)))
+	}
+}
+
+func (m *HostManager) run() {
+	for {
+		m.performHostScans()
```
n8mgr: nit: there's an extra level of indirection here that I prefer to avoid: `NewManager -> go run() -> performHostScans -> (fetchHosts, scanHosts, updateHost)`. Jumping through files to get to logic does not spark joy.

hosts/manager.go
```diff
@@ -27,31 +51,70 @@ type (
...
+		store    Store
 
+		tg  *threadgroup.ThreadGroup
 		log *zap.Logger
 	}
+
+	// host is a helper type that exposes a resolve and scan method that
+	// decorate the resolved CIDRs and host settings to then be persisted in the
+	// database
+	host struct {
```
n8mgr: Having methods that modify the internals is undesirable, imo. Prefer having a func and explicitly updating the fields.

- `func ResolveHost(resolver net.Resolver, addresses []chain.NetAddress) ([]string, error)`
- `func HostSettings(client rhp4.Transport, addresses []chain.NetAddress) (proto4.Settings, error)`

hosts/manager.go
```diff
@@ -87,6 +150,225 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+	return m.store.HostsForScanning(ctx)
+}
+
+func (m *HostManager) performHostScans() {
+	hosts, err := m.hostsForScanning()
+	if err != nil {
+		m.log.Error("failed to get hosts for scanning", zap.Error(err))
+		return
+	} else if len(hosts) == 0 {
+		m.log.Info("scan skipped")
```
n8mgr: This should be debug since it's unlikely to be user actionable. It just creates spam if our scanner happens to be ahead.

hosts/manager.go
```diff
@@ -87,6 +150,225 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+			}
+		}
+	}()
+
+	m.log.Info("scans started")
+	start := time.Now()
+	wg.Wait()
+	select {
+	case <-m.tg.Done():
+		m.log.Info("host scans interrupted")
```
n8mgr: These should be debug since they just create log spam that isn't important to users.

hosts/manager.go
```diff
@@ -87,6 +150,225 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+	default:
+		m.log.Info("host scans finished", zap.Int("hosts", len(hosts)), zap.Duration("duration", time.Since(start)))
```
n8mgr: These should be debug since they just create log spam that isn't important to users.

hosts/manager.go
```diff
@@ -87,6 +150,225 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+	ctx, cancel := context.WithTimeout(context.Background(), scanTimeout)
+	ctx, cancel = m.tg.WithContext(ctx)
+	defer cancel()
+
+	err := h.resolve(ctx, m.resolver)
+	if err != nil {
+		return fmt.Errorf("failed to resolve host, %w", err)
+	}
+
+	err = h.settings(ctx, m.client, m.log)
```
n8mgr: consider passing in log if it's important here. My gut is that since we are also logging the error higher up the stack, logging at a lower level is unnecessary.

hosts/manager.go
```diff
@@ -87,6 +150,225 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+func (h *host) settings(ctx context.Context, client RHP4Client, logger *zap.Logger) error {
+	addr := h.address()
+	if addr == "" {
+		return fmt.Errorf("no address")
+	}
+
+	var err error
+	h.hs, err = client.Settings(ctx, h.hk, addr)
+	if err != nil {
+		logger.Error("failed to get host settings", zap.Stringer("hk", h.hk), zap.String("address", addr), zap.Error(err))
```
n8mgr: I see what you're trying to do, but personally should still return an error here and handle the failed scan higher up.

hosts/manager.go
```diff
@@ -1,23 +1,47 @@
...
 	maxAddrsPerProtocol = 2
+
+	scanThreads = 4
+	scanTimeout = time.Minute
 )
 
 type (
+	// Store defines the interface for persisting host scans in the database.
+	Store interface {
+		HostsForScanning(ctx context.Context) ([]types.PublicKey, error)
```
n8mgr, line 40: This should take the number of hosts to return

persist/postgres/hosts.go
```diff
@@ -42,6 +44,66 @@ func (u *updateTx) AddHostAnnouncement(hk types.PublicKey, ha chain.V2HostAnnoun
...
 }
 
+// UpdateHost updates a host in the database, the given parameters are the result of scanning the host.
+func (s *Store) UpdateHost(ctx context.Context, hk types.PublicKey, hs proto4.HostSettings, cidrs []string, nextScan time.Time) error {
+	success := hs != (proto4.HostSettings{})
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		var hostID int64
+		err := tx.QueryRow(ctx, `UPDATE hosts SET
+total_scans = total_scans + 1,
+failed_scans = CASE WHEN $1 = FALSE THEN failed_scans + 1 ELSE failed_scans END,
```
n8mgr: Consider a separate query for failed scans over multiple CASE statements.

persist/postgres/hosts.go
```diff
@@ -42,6 +44,66 @@ func (u *updateTx) AddHostAnnouncement(hk types.PublicKey, ha chain.V2HostAnnoun
 	return nil
 }
 
+// UpdateHost updates a host in the database, the given parameters are the result of scanning the host.
+func (s *Store) UpdateHost(ctx context.Context, hk types.PublicKey, hs proto4.HostSettings, cidrs []string, nextScan time.Time) error {
+	success := hs != (proto4.HostSettings{})
```
n8mgr: You may want to consider passing/storing an `error` for failed scans, it would be a more explicit check than the zero value.

persist/postgres/init.sql
```diff
@@ -2,24 +2,50 @@ CREATE TABLE hosts (
     id SERIAL PRIMARY KEY,
     public_key BYTEA UNIQUE NOT NULL,
     total_scans INTEGER NOT NULL DEFAULT 0,
-    successful_scans INTEGER NOT NULL DEFAULT 0,
     failed_scans INTEGER NOT NULL DEFAULT 0,
     consecutive_failed_scans INTEGER NOT NULL DEFAULT 0,
     last_announcement TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT '0001-01-01 00:00:00+00',
-    last_scan_success BOOLEAN NOT NULL DEFAULT FALSE,
+    last_scan TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT '0001-01-01 00:00:00+00',
```
n8mgr: Would it be more beneficial to only update this if the scan was successful? The time since the last failed scan isn't relevant to our needs, but the time since the last successful scan could replace the global `uptime` and `downtime` when paired with `total_scans` and `failed_scans`. 

This is only somewhat relevant to this PR, so feel free to ignore, but if we are checking for `%` uptime, then we've got duplicate data `% uptime = (total_scans - failed_scans) / total_scans`. If we're checking for a maximum `downtime`, older hosts may be unnecessarily penalized. Regardless, I'm a bit concerned about churn with older hosts because downtime and failed scans accumulate forever.

hosts/manager.go
peterjan: I see how that would make sense but we don't have any other validators in `coreutils`, so I'm not sure if this should be the first one to add. It's quite a silly validator to add imo and some implementations might want to consider some protocols invalid and what not. Not really worth it imo.

n8mgr (changes requested):

peterjan:
@n8maninger thank you for the review but I'm still updating this `DRAFT` - I'll re-request a review when I'm done

hosts/manager.go
peterjan: resolving in favour of https://github.com/SiaFoundation/indexd/pull/41#discussion_r1979980303

hosts/manager.go
peterjan: I was expecting some pushback on this, and I don't mind changing it, but I do want to explain why I opted to go this route. Fetching all hosts for scanning, to then fetch host addresses separately, is a reasonable approach because you avoid creating some weird return type in the store, host keys are small, but more importantly you ensure the scanning is exhaustive. 

In your pattern for instance, if you return `HostsForScanning` with a limit, you have to be sure consecutive calls to the database don't keep returning the same hosts over and over again. You avoid this entirely if you fetch all hosts that need to be scanned. If you return only the host's public key that's not going to be an issue at any scale.

I would agree that fetching host addresses separately can be seen as problematic and can be seen as a waste of resources and simply too many round trips to the database... But is it really that big of an issue? We scan pretty infrequently, over time the host's `nextScan` should smear out so we don't have to scan all of them at the same time, we will prune offline hosts and we have a max age cap on announcements. Even with a ridiculous number of hosts the database would still shrug that off.

Fetching `hostsForScanning` with `scanThreads` as `LIMIT` to then spin up `len(hosts)` goroutines is incredibly elegant though. It's a pattern I wouldn't immediately come up with myself and when I read it, it brought a smile to my face. I'm not sure it's worth the trade-off though, introducing a limit invites too many bugs imo.

persist/postgres/init.sql
peterjan: I think `uptime` and `downtime` as intervals are useful information to have and potentially display at one point. I do somewhat share the concern with regards to churn, but I feel that's something you can solve in your business logic, rather than changing the way you keep track of these metrics.

persist/postgres/hosts.go
peterjan: I considered it but the `CASE` statements allow to easily perform the `UPDATE` and apply some extra conditions, like not setting the downtime if the first scan fails. The alternative would be to query some fields first, that would clean up the query a bit but the logic is easy enough to execute on the database imo. I'm fine with updating it though, I actually originally started writing it out as you propose.

hosts/manager.go
peterjan: resolved in favour of https://github.com/SiaFoundation/indexd/pull/41#discussion_r1980305655

hosts/manager.go
n8mgr: > You avoid this entirely if you fetch all hosts that need to be scanned.

"fetch all" or "fetch everything" is a huge red flag when thinking about rows in a database, particularly a remote database, unless there's a known cap on the maximum possible number of rows.

> If you return HostsForScanning with a limit, you have to be sure consecutive calls to the database don't keep returning the same hosts over and over again.

This is true, but not overly problematic -- `HostsForScanning` is serial and each host in the batch will update `next_scan_time` before it gets called again.

> But is it really that big of an issue?

Fetching 500 rows, no probably not; 5000 rows, starting to get there; 100,000 rows, definitely. It's an easily avoidable performance refactor that adds no complexity, imo.

n8mgr: Even if you keep the fetch all logic -- the worker pattern, nested loop, and extra concurrency bits at the bottom can still be refactored out:
```go
func (m *HostManager) performHostScans(ctx context.Context) {
	hosts, err := m.hostsForScanning(ctx)
	if err != nil {
		m.log.Error("failed to get hosts for scanning", zap.Error(err))
		return
	} else if len(hosts) == 0 {
		m.log.Info("scan skipped")
		return
	}

	var wg sync.WaitGroup
	sema := make(chan struct{}, scanThreads)
	for _, hk := range hosts {
		sema <- struct{}{}
		wg.Add(1)
		go func(ctx context.Context, hk types.PublicKey) {
			defer func() {
				<-sema
				wg.Done()
			}()

			select {
			case <-ctx.Done():
				return
			default:
			}

			logger := m.log.With(zap.Stringer("hk", hk))

			h, err := m.fetchHost(ctx, hk)
			if err != nil {
				logger.Error("failed to get host", zap.Error(err))
			} else if err = m.scanHost(ctx, h); err != nil {
				logger.Error("failed to scan host", zap.Error(err))
			} else if err = m.updateHost(ctx, h); err != nil {
				logger.Error("failed to update host", zap.Error(err))
			}
		}(ctx, hk)
	}

	wg.Wait()
}
```

peterjan: I updated the pattern. I don't mind updating the `LIMIT` thing either but I personally really like the fact that host scanning is 100% certain to be exhaustive. We randomize next scan with a +/- 6 hour offset to smear out scans (as per the spec) so I don't see us ever running into a performance issue fetching all host keys for scanning. I'll wait for Chris to chime in, fine with updating it of course if you both feel it's a worse solution!

hosts/manager.go
```diff
@@ -87,6 +175,258 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
 	return nil
 }
 
+func (m *HostManager) hostsForScanning(ctx context.Context) ([]types.PublicKey, error) {
+	ctx, cancel := context.WithTimeout(ctx, time.Minute)
```
ChrisSchinnerl: Seems a bit excessive to define a timeout for talking to our trusted database. I don't think this will ever end up blocking indefinitely.

hosts/manager.go
```diff
@@ -87,6 +175,258 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+			logger := m.log.With(zap.Stringer("hk", hk))
+			if networks, hs, err := m.performHostScan(ctx, hk); errors.Is(err, context.Canceled) {
+				return
+			} else if err != nil {
+				logger.Error("failed to perform host scan", zap.Error(err))
+			} else if err := m.persistHostScan(ctx, hk, networks, hs); err != nil {
+				logger.Error("failed to persist host scan", zap.Error(err))
+			}
+
+			// TODO: perform host checks
```
ChrisSchinnerl: I think we can remove this TODO here and move it to where we fetch the hosts in the database. It should be possible to perform all checks live now as we fetch hosts from the database. That way they are always up-to-date.

hosts/manager.go
```diff
@@ -87,6 +175,258 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+		m.log.Debug("host scans finished", zap.Int("hosts", len(hosts)), zap.Duration("duration", time.Since(start)))
+	}
+}
+
+// FetchSettings uses the given client to fetch the settings of the host with
+// given public key. It only returns the settings if the host is available on
+// every address. The only error this function returns is [context.Canceled],
+// other errors that occur during the resolving and parsing are debug logged but
+// otherwise ignored.
+func FetchSettings(ctx context.Context, client RHP4Client, hk types.PublicKey, addresses []chain.NetAddress, log *zap.Logger) (proto4.HostSettings, error) {
```
ChrisSchinnerl: Do these functions need to be exposed? Might be better for testing to change `manager_test.go`'s package to `hosts` and mock the database rather than exposing all these functions.

Then we can unit test the manager better here and keep the interface tests for the database in the `postgres` package.

e.g. it would become quite easy to manipulate the `consecutiveScanFailures` for testing and see if the `nextScan` time is updated correctly in the mock.

hosts/manager_test.go
```diff
@@ -16,13 +19,50 @@ import (
...
+
+	hs, ok := c.settings[addr]
+	if ok {
+		return hs, nil
+	}
+	return proto4.HostSettings{}, errors.New("") // mock host being unavailable on unknown address
+}
+
 func TestHostManager(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
```
ChrisSchinnerl: I'd avoid this in testing. The tests have a timeout already and print a nice stack trace.

hosts/manager.go
```diff
@@ -87,6 +175,258 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+	var hs proto4.HostSettings
+	for _, addr := range addresses {
+		if addr.Protocol == quic.Protocol {
```
ChrisSchinnerl: I'd either `!= siamux.Protocol` here or `== siamux.Protocol` and wrap the whole thing.
Right now we'll try the siamux client even if `addr.Protocol` is invalid for some reason.

hosts/manager.go
```diff
@@ -87,6 +175,258 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+			continue // TODO: add QUIC support to client
+		}
+
+		if settings, err := client.Settings(ctx, hk, addr.Address); errors.Is(err, context.Canceled) {
+			return proto4.HostSettings{}, err
+		} else if err != nil {
+			log.Debug("failed to get host settings", zap.String("address", addr.Address), zap.Error(err))
+			hs = proto4.HostSettings{} // require host to be available on all addresses
+		} else {
+			hs = settings
```
ChrisSchinnerl: Hm this is a bit annoying. Theoretically, knowing the indexer, a host could cheat us by providing good settings via the siamux and bad settings over quic. So the indexer thinks the host is great but the clients get horrible settings over quic.

I think the easiest way to avoid this for now is returning a random set of settings.

persist/postgres/init.sql
ChrisSchinnerl: I'm definitely all for removing `uptime` and `downtime` if we don't need them right now. But I'm not sure I'm getting the `paired with total_scans and failed_scans` part.

I think our condition for whether a host has had a week or so of uptime could be replaced with checking whether we have scanned the host 10 times already + the ratio of scans.

Not sure where the last successful scan comes into play.

persist/postgres/init.sql
```diff
@@ -2,24 +2,50 @@ CREATE TABLE hosts (
     id SERIAL PRIMARY KEY,
     public_key BYTEA UNIQUE NOT NULL,
     total_scans INTEGER NOT NULL DEFAULT 0,
-    successful_scans INTEGER NOT NULL DEFAULT 0,
     failed_scans INTEGER NOT NULL DEFAULT 0,
     consecutive_failed_scans INTEGER NOT NULL DEFAULT 0,
```
ChrisSchinnerl, line 6: Isn't this obsolete now with you keeping it in memory?

persist/postgres/init.sql
```diff
@@ -2,24 +2,50 @@ CREATE TABLE hosts (
...
     next_scan TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT '0001-01-01 00:00:00+00',
     uptime INTERVAL NOT NULL DEFAULT '0 seconds',
     downtime INTERVAL NOT NULL DEFAULT '0 seconds'
 );
+CREATE INDEX idx_hosts_next_scan_idx ON hosts(next_scan);
 
 CREATE TABLE host_addresses (
     id SERIAL PRIMARY KEY,
-    host_id INTEGER REFERENCES hosts(id) ON DELETE CASCADE,
+    host_id INTEGER NOT NULL REFERENCES hosts(id) ON DELETE CASCADE,
```
ChrisSchinnerl, line 16: might require updating the schema too

ChrisSchinnerl (changes requested):

persist/postgres/init.sql
peterjan: I removed `last_scan_success` because it was redundant seeing as a successful last scan means `consecutive_failed_scans` is 0. I do think `consecutive_failed_scans` will end up being useful for pruning hosts. 

edit: going to resolve in favour of https://github.com/SiaFoundation/indexd/pull/41/files#r1980023533

persist/postgres/init.sql
peterjan: Ok, I removed `uptime` and `downtime`, that also means we don't need `last_scan` anymore. I kept `consecutive_scan_failures` to be able to tell whether the last scan succeeded, and also to be able to roughly tell how long a host has been down for to potentially prune it.

hosts/manager.go
```diff
@@ -87,6 +175,257 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+	if err != nil {
+		return nil, proto4.HostSettings{}, fmt.Errorf("failed to get host addresses, %w", err)
+	}
+
+	logger := m.log.With(zap.Stringer("hk", hk))
+
+	var networks []string
+	addrs, networks, err = resolveHost(ctx, m.resolver, addrs, logger)
+	if err != nil {
+		return nil, proto4.HostSettings{}, err
```
n8mgr: add error context

hosts/manager.go
```diff
@@ -87,6 +175,257 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+	}
+
+	settings, err := fetchSettings(ctx, m.client, hk, addrs, logger)
+	if err != nil {
+		return nil, proto4.HostSettings{}, err
```
n8mgr: add error context

hosts/manager.go
```diff
@@ -87,6 +175,257 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+	}
+
+	return networks, settings, nil
+}
+
+func (m *HostManager) persistHostScan(ctx context.Context, hk types.PublicKey, networks []string, settings proto4.HostSettings) error {
+	var consecScanFailures int
```
n8mgr: It feels very odd to calculate the next scan time based on the in-memory value rather than the db column.

hosts/manager.go
```diff
@@ -87,6 +175,257 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+
+	hosts, err := m.store.HostsForScanning(ctx)
+	if err != nil {
+		m.log.Error("failed to get hosts for scanning", zap.Error(err))
+		return
+	} else if len(hosts) == 0 {
+		m.log.Debug("scan skipped")
+		return
+	} else {
+		m.log.Debug("scan started")
```
n8mgr: Extra nesting is unnecessary

hosts/manager.go
```diff
@@ -87,6 +175,257 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+		}(ctx, hk)
+	}
+
+	wg.Wait()
+
+	select {
+	case <-ctx.Done():
+		m.log.Debug("scan cancelled")
+	default:
+		m.log.Debug("host scans finished", zap.Int("hosts", len(hosts)), zap.Duration("duration", time.Since(start)))
```
n8mgr: Even if the function was cancelled, this information is still valuable since some scans may have completed. I'd recommend ditching the `select` and logging this line.

hosts/manager.go
```diff
@@ -87,6 +175,257 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+			settings = append(settings, hs)
+		}
+		// TODO: add support for QUIC
+	}
+
+	select {
+	case <-ctx.Done():
+		return proto4.HostSettings{}, ctx.Err()
+	default:
+	}
```
n8mgr: This is racy and I don't think it's worth cancelling here. The code below isn't net or IO bound and will complete nearly instantly.

hosts/manager.go
```diff
@@ -87,6 +175,257 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+
+	// return a random setting to avoid the host cheating us by providing
+	// different settings over different addresses or protocols
+	for _, hs := range settings {
```
n8mgr: Isn't this effectively `settings[0]`? Adding a `if len(settings) > 0` is significantly clearer.

Regarding the comment: I don't think grabbing a random one provides any significant guarantees or benefits when dealing with cheating hosts. Returning the first will have the same guarantee the host will actually honor anything it returns.

n8mgr (changes requested):

hosts/manager.go
peterjan: Oof that was dumb, I had it in a `map` and then made it a slice and didn't revisit this line. Still returning a random setting for the time being. Wouldn't this at least deter hosts from trying to cheat? It's harmless imo but also something that's unlikely we'll ever run into.

hosts/manager.go
peterjan: This boils down to returning custom host types from the store again. I personally didn't really see an issue with basing the back off algo on an in-memory value that resets when you restar the indexer. I can agree it's odd but is it worth fetching that information from the database? I'll check what @ChrisSchinnerl thinks about this, don't mind changing this but generally when I go and expose custom types it gets flagged in reviews.

**edit: doing host checks made me realise we actually need `last_successful_scan` and we can drop the consecutive failures column. I'll make that change in a separate PR**

hosts/manager.go
ChrisSchinnerl: It's not foolproof but it prevents the obvious way to cheat. Considering that we'll usually get about 2 siamux and 2 quic settings we'd at least have a chance of 25% -50% to get served the bad settings, if any, rather than 0%.

It raises the bar from "I can consistently hand out bad settings without the indexer knowing" to "I need to at least figure out whether my peer is an indexer before handing out bad settings to not lose customers"

hosts/manager.go
ChrisSchinnerl: Tbh I was confused about an in-memory value as well. Seems like the in-memory tracking adds more complexity than the persisted one since we already have a ton of other fields to update upon scanning.

hosts/manager.go
```diff
@@ -27,34 +60,89 @@ type (
...
+		scanFrequency:      time.Hour,
+		scanInterval:       time.Hour * 24,
+
+		resolver: &net.Resolver{},
+		client:   rhp.New(),
+		store:    store,
+		tg:       threadgroup.New(),
+		log:      zap.NewNop(),
+
+		consecutiveScanFailures: make(map[types.PublicKey]int),
```
ChrisSchinnerl: Considering that two of us thought it was kinda odd I'd remove the in-memory tracking.

persist/postgres/hosts_test.go
ChrisSchinnerl, line 1: I think `HostAddresses` and `HostsForScanning` might still be missing tests here

persist/postgres/hosts.go
```diff
@@ -17,6 +19,14 @@ var ErrHostNotFound = errors.New("host not found")
 
 type dbHost struct {
 	id int64
+
+	totalScans             int
+	failedScans            int
+	consecutiveFailedScans int
+	nextScan               time.Time
+
+	networks []string
```
ChrisSchinnerl: I'd move all these fields into the `hosts.Host` type. After all it's all information we are interested in as users. Then you can populate `networks` the same way you do with `addresses`. 

The more `dbHost` diverges from just being a `hosts.Host` with an ID, the nastier things get.
e.g. this `networks` field is already only used for testing. The production code uses the `dbHost` type but just ignores the `networks` field.

persist/postgres/hosts_test.go
```diff
@@ -0,0 +1,240 @@
...
+		t.Fatal(err)
+	} else if len(h.networks) != 1 {
+		t.Fatal("unexpected networks", h.networks)
+	} else if h.networks[0] != networks[0] {
+		t.Fatal("unexpected network", h.networks[0])
+	}
+}
+
+// rawHost returns the host for given public key
+func (s *Store) rawHost(ctx context.Context, hk types.PublicKey) (dbHost, error) {
```
ChrisSchinnerl: I'd move this to `hosts.go` and call it `Host` while also returning a `hosts.Host`. We are going to expose fetching a single host through the API anyway.

ChrisSchinnerl (changes requested):

hosts/manager.go
peterjan: It makes more sense to add `last_successful_scan` and remove `consecutive_failed_scans` in the database and keep this in-memory value. If you want to calculate uptime % you can do so with the number of failed and successful scans, and if you want to know how long a host has been offline for you can use  `last_successful_scan` and avoid doing some weird calculation that has to take into account your scan interval. Also, we will soon have scan triggers that will likely persist the scan results, so if we ever run into a situation where that trigger fires a lot, and we calculate the downtime by extrapolating the number of failed scans and the scan interval, you would possible get a very skewed result.

hosts/manager.go
n8mgr: A cheating host could still identify an indexer by its RPC activity and hand out settings based on that. I'm not convinced this has any benefit at all to justify. You just attack it slightly differently with the exact same result. A more reasonable mitigation would be to have the indexer send signed prices to clients since it has to keep them updated regardless. The indexer already knows the host is willing to honor that price table because it's also using it itself.

I would avoid trying to build mitigations around hosts with multiple net addresses acting as distinct entities.

hosts/manager.go
n8mgr: If you don't use the stored value old hosts can get re-added to the scan loop. i.e. if a host has 8 consecutive failed scans, we calculate next_scan correctly. When the service is rebooted and the host fails its next scan, it will be will be calculated as if the host had never failed a scan.

peterjan: I don't see why it matters that hosts are re-added to the scan loop if the exp. backoff is quite aggressive and the scans are infrequent. We prune the hosts if they're offline for x amount of time. I don't mind updating it if you both feel the same way but I have a hard time understanding why keeping track of `consecutive_failed_scans` to multiply it by a scan frequency to determine how long the host has been offline for, to then also have to fetch host key + consecutive failed scans, is better than keeping it in memory and persisting the exact time a scan was successful so you can see exactly how long its been since the host was online

edit: I'll change it, exposing all those fields on the `Host` type changed things a lot for me, that means I don't have to add a custom type to return the host addresses and the consecutive failures. I also like the store interface a lot better now, it's feels very clean for me. Happy with the PR the way it is now.

hosts/manager.go
ChrisSchinnerl: Oh I'd definitely return the prices as well. At the very least, even without the client refetching them from the indexer, it gives them a ballpark of how expensive that host is supposed to be when fetching their own prices.

I think though that adding a single shuffle to raise the complexity from simply announcing bad prices to performing RPC analysis is a worthwhile tradeoff. After all that's all we can really do in distributed systems. Making attacks more expensive to perform. Be it only accepting hosts after a certain age, randomizing scan intervals or limiting the size of a request we are willing to read from the wire. None of those things are foolproof and only serve making attacks/analysis harder to perform.

hosts/manager.go
```diff
@@ -87,6 +202,180 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+// fetchSettings uses the given client to fetch the settings of the host with
+// given public key. It only returns the settings if the host is available on
+// every address. The only error this function returns is [context.Canceled],
+// other errors that occur during the resolving and parsing are debug logged but
+// otherwise ignored.
+func fetchSettings(ctx context.Context, client RHP4Client, hk types.PublicKey, addresses []chain.NetAddress, log *zap.Logger) (proto4.HostSettings, error) {
+	var settings []proto4.HostSettings
+	for _, addr := range addresses {
+		if addr.Protocol == siamux.Protocol {
+			hs, err := client.Settings(ctx, hk, addr.Address)
```
n8mgr, line 258: nit: wondering if we want to use the transport pool for scanning. I think it makes sense for repairs and pinning. For scanning though, we're opening multiple conns per host. We should probably close them immediately rather than trying to keep them alive.

hosts/host.go
```diff
@@ -1,17 +1,26 @@
...
 	"time"
 
+	proto4 "go.sia.tech/core/rhp/v4"
 	"go.sia.tech/core/types"
 	"go.sia.tech/coreutils/chain"
 )
 
 type (
 	// Host is a host on the network.
 	Host struct {
```
n8mgr, line 14: nit: could be moved into manager.go since it's such a short file alredy

hosts/manager.go
```diff
@@ -87,6 +202,180 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+		go func(ctx context.Context, hk types.PublicKey) {
+			defer func() {
+				<-sema
+				wg.Done()
+			}()
+
+			if _, err := m.ScanHost(ctx, hk); errors.Is(err, context.Canceled) {
+				return
+			} else if err != nil {
+				m.log.Error("failed to perform host scan", zap.Stringer("hk", hk), zap.Error(err))
```
n8mgr, line 239: nit: I would probably debug here too since it's unlikely to be user-actionable.

n8mgr (approved):

hosts/manager_test.go
```diff
@@ -34,153 +89,236 @@ func TestHostManager(t *testing.T) {
...
+	} else if len(networks) != 1 {
+		t.Fatal("unexpected", len(networks))
 	}
 }
+
+func testMuxAddr(addr string) chain.NetAddress {
+	return chain.NetAddress{Protocol: siamux.Protocol, Address: addr}
+}
+
+func cancelledCtx() context.Context {
```
ChrisSchinnerl, line 320: (nit) slightly better might be
var cancelledCtx = func() context.Context { ... }

hosts/manager_test.go
```diff
@@ -34,153 +89,236 @@ func TestHostManager(t *testing.T) {
...
+	} else if len(addrs) != 0 {
+		t.Fatal("unexpected", len(addrs))
 	}
-	h1Addr = addresses(h1.PublicKey(), indexed)
-	if len(h1Addr) != 1 {
-		t.Fatal("unexpected", len(h1Addr))
-	} else if h1Addr[0].Address != "[::]:4848" || h1Addr[0].Protocol != quic.Protocol {
-		t.Fatalf("unexpected address %+v", h1Addr[0])
+
+	// assert networks are parsed and added
```
ChrisSchinnerl, line 305: added?

ChrisSchinnerl (approved):

PR #42, opened 2025-02-24
Add ConsensusNode type and remove Retry

ChrisSchinnerl:
This PR is based on https://github.com/SiaFoundation/indexd/pull/38 to demonstrate how it can potentially clean up our testing.

The idea here is to create a `ConsensusNode` type which is the source of truth for our tests when it comes to consensus. It can be used to mine blocks and will blockingly update all components that need to be subscribed to consensus.

This effectively eliminates the source of 90% of our NDFs while keeping our tests fast and avoiding a `Retry` altogether.

If we ever want to do a full stack integration test which relies only on the `Subscriber`s spawned goroutine, we can disable the manual sync by adding a config option to the `ConsensusNode` which prevents the implicit syncing.

### DIFF
--- a/internal/testutils/consensusnode.go
+++ b/internal/testutils/consensusnode.go
@@ -1,0 +1,68 @@
+package testutils
+
+import (
+	"testing"
+	"time"
+
+	"go.sia.tech/core/consensus"
+	"go.sia.tech/core/types"
+	"go.sia.tech/coreutils"
+	"go.sia.tech/coreutils/chain"
+	"go.sia.tech/coreutils/testutil"
+	"go.uber.org/zap"
+)
+
+// ConsensusNode is a helper type that serves as the source of truth for
+// anything consensus-related during testing. It can be used to create test
+// types such as the Indexer or Wallet and will guarantee that a call to
+// `MineBlock` won't return before all created components are done subscribing
+// to the newly mined blocks.
+type ConsensusNode struct {
+	tb      testing.TB
+	cm      *chain.Manager
+	genesis types.Block
+	network *consensus.Network
+
+	syncFns []func()
+}
+
+// NewConsensusNode creates a new node ready for testing. It will mine enough
+// blocks to reach the V2 require height before returning.
+func NewConsensusNode(tb testing.TB, log *zap.Logger) *ConsensusNode {
+	network, genesis := testutil.V2Network()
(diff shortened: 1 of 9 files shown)


internal/testutils/indexer.go
```diff
@@ -42,18 +40,12 @@ type Indexer struct {
 
 // NewIndexer creates a new indexer for testing that is automatically closed up
 // after the test is finished.
-func NewIndexer(t testing.TB, n *consensus.Network, genesis types.Block, log *zap.Logger) *Indexer {
+func (c *ConsensusNode) NewIndexer(t testing.TB, log *zap.Logger) *Indexer {
```
n8mgr: I don't feel too strongly, but `func NewIndexer(tb testing.TB, c *ConsensusNode, log *zap.Logger) *Indexer` feels more "correct". It's a bit weird that way though since that means the constructor needs to essentially hook itself into the internals of the `ConsensusNode`.

internal/testutils/wallet.go
```diff
@@ -1,96 +1,41 @@
...
-		syncer.WithSendBlocksTimeout(2*time.Second),
-		syncer.WithRPCTimeout(2*time.Second),
-		syncer.WithSyncInterval(100*time.Millisecond),
-	)
-	t.Cleanup(func() { s.Close() })
-	go s.Run()
-
+// NewWallet creates a new SingleAddressWallet for testing which is connected to
+// all the other components created via the ConsensusNode.
+func (c *ConsensusNode) NewWallet() *wallet.SingleAddressWallet {
```
n8mgr: Ditto

internal/testutils/consensusnode.go
```diff
@@ -0,0 +1,68 @@
...
+	"go.uber.org/zap"
+)
+
+// ConsensusNode is a helper type that serves as the source of truth for
+// anything consensus-related during testing. It can be used to create test
+// types such as the Indexer or Wallet and will guarantee that a call to
+// `MineBlock` won't return before all created components are done subscribing
+// to the newly mined blocks.
+type ConsensusNode struct {
+	tb      testing.TB
```
n8mgr: Should avoid keeping references to any lifecycle-oriented types. This goes for things like context too.

n8mgr (changes requested):

n8mgr (approved):

PR #44, opened 2025-02-26
Add contracts persistence 

ChrisSchinnerl:
This PR implements basic persistence for contracts which includes

- adding formed contracts
- adding renewed contracts
- fetching a contract by id
- marking a contract as good or bad


persist/postgres/contracts.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// Contracts queries the contracts in the database. By default, only active
+// contracts are returned.
+func (s *Store) Contracts(queryOpts ...ContractQueryOpt) ([]api.Contract, error) {
+	opts := defaultContractQueryOpts
+	for _, opt := range queryOpts {
+		opt(&opts)
+	}
+	panic("not implemented")
```
ChrisSchinnerl, line 143: leaving this for a f/u since this is already 500+ lines of code

persist/postgres/contracts.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// SetContractGood updates the "good" column of a contract.
+func (s *Store) SetContractGood(contractID types.FileContractID, good bool) error {
```
n8mgr: In what scenarios would you want a contract to go from bad->good? In my mind, it's better to only enable good->bad so we can avoid flip-flopping on decisions. If we have a faulty heuristic flipping them to bad, we can fix the heuristic.

api/types.go
```diff
@@ -26,6 +26,63 @@ type (
 	}
 )
 
+// The following consts define the various states a contract can be in.
+const (
```
n8mgr: Shouldn't all of this be in a "contracts" package? API should be purely for request and response types, not internals.

persist/postgres/types.go
```diff
@@ -57,6 +59,52 @@ func (ci *sqlChainIndex) Scan(src any) error {
 	}
 }
 
+type sqlContractState api.ContractState
+
+func (s sqlContractState) Value() (driver.Value, error) {
```
n8mgr, line 64: Base type should either be an iota or stored as a string.

persist/postgres/types.go
```diff
@@ -178,6 +226,8 @@ func (h *sqlHash256) Scan(src any) error {
 		}
 		copy(h[:], src)
 		return nil
+	case nil:
```
n8mgr: Implicit nil handling is similar to generic encode imo. Better to add a nullable helper

persist/postgres/contracts.go
ChrisSchinnerl: I'm thinking of scenarios where the host is temporarily bad causing the contract to be bad. e.g. the operator setting new gouging settings, figuring out they are too low, raising them again but now good contracts are bad. However not marking them as bad in the first place would lead us to not migrate data away from contracts that might remain bad until the proof window. 

Or the host temporarily not accepting contracts, being too expensive etc. Which would lose the host potentially all of its business even though they only had bad settings for a couple of hours.

doc/rfc/001_db_schema.md
```diff
@@ -164,16 +164,16 @@ CREATE TABLE host_settings (
...
-  contract_id BYTEA NOT NULL UNIQUE,
+  contract_id BYTEA NOT NULL UNIQUE DEFERRABLE,
 
   -- lifetime related columns
   proof_height BIGINT NOT NULL, -- start of proof window
   expiration_height BIGINT NOT NULL, -- end of proof window
-  renewed_from INTEGER REFERENCES contracts(id),
-  renewed_to INTEGER REFERENCES contracts(id),
-  state SMALLINT NOT NULL, -- 0 = 'pending', 1 = 'active', 2 = 'resolved', 3 = 'expired', 4 = 'rejected'
+  renewed_from INTEGER REFERENCES contracts(id) UNIQUE DEFERRABLE DEFAULT NULL,
```
n8mgr: Don't need to set a default when it's `NULL`

n8mgr (dismissed):
I'm going to approve, but drop the `DEFAULT NULL` before merging.

PR #53, opened 2025-03-03
Subscribe contracts to chain state

ChrisSchinnerl:
This PR adds a contract manager, hooks it up to our test code and the indexer and calls its update method from the chain subscriber.

Since this has grown quite large again I left some todos for follow ups.

### DIFF
--- a/contracts/chain.go
+++ b/contracts/chain.go
@@ -2,7 +2,6 @@ package contracts
 
 import (
 	"fmt"
-	"time"
 
 	"go.sia.tech/core/consensus"
 	"go.sia.tech/core/types"
@@ -15,7 +14,6 @@ type (
 	// chain update in the database.
 	UpdateTx interface {
 		IsKnownContract(contractID types.FileContractID) (bool, error)
-		RejectContracts(time.Duration) error
 		UpdateContractElement(fce types.V2FileContractElement) error
 		UpdateContractState(contractID types.FileContractID, state ContractState) error
 	}
@@ -43,19 +41,19 @@ func (tx *updateTx) IsKnownContract(fcid types.FileContractID) (bool, error) {
 // UpdateChainState state updates the contracts' state in the database and
 // broadcasts revisions for failed expired contracts.
 func (m *ContractManager) UpdateChainState(tx UpdateTx, reverted []chain.RevertUpdate, applied []chain.ApplyUpdate) error {
-	cTx := &updateTx{
+	uTx := &updateTx{
 		UpdateTx: tx,
 	}
 
 	for _, cru := range reverted {
-		err := m.revertChainUpdate(cTx, cru)
+		err := m.revertChainUpdate(uTx, cru)
 		if err != nil {
 			return fmt.Errorf("failed to revert chain update: %w", err)
 		}
 	}
 
 	for _, cau := range applied {
-		err := m.applyChainUpdate(cTx, cau)
+		err := m.applyChainUpdate(uTx, cau)
 		if err != nil {
 			return fmt.Errorf("failed to apply chain update: %w", err)
 		}
@@ -91,23 +89,26 @@ func (m *ContractManager) applyChainUpdate(tx *updateTx, cau chain.ApplyUpdate) 
 
 func (m *ContractManager) applyContractDiff(tx *updateTx, diff consensus.V2FileContractElementDiff) error {
 	// update contract state
-	var state ContractState
-	if resolution := diff.Resolution; resolution != nil {
-		state = ContractStateResolved
-	} else if diff.Created {
-		state = ContractStateActive
+	if diff.Resolution != nil || diff.Created {
+		var state ContractState
+		switch {
+		case diff.Resolution != nil:
+			state = ContractStateResolved
+		case diff.Created:
+			state = ContractStateActive
+		default:
+			panic("unknown state") // unreachable
+		}
+		if err := tx.UpdateContractState(diff.V2FileContractElement.ID, state); err != nil {
+			return fmt.Errorf("failed to update contract state for %v: %w", diff.V2FileContractElement.ID, err)
+		}
+		m.log.Info("contract state changed", zap.Stringer("contractID", diff.V2FileContractElement.ID),
+			zap.Stringer("state", state))
 	}
-	if err := tx.UpdateContractState(diff.V2FileContractElement.ID, state); err != nil {
-		return fmt.Errorf("failed to update contract state for new contract: %w", err)
-	}
-	m.log.Info("contract state changed", zap.Stringer("contractID", diff.V2FileContractElement.ID),
-		zap.Stringer("state", state))
 
 	// update contract elements
-	var fce types.V2FileContractElement
-	if rev, ok := diff.V2RevisionElement(); !ok {
-		fce = diff.V2FileContractElement
-	} else {
+	fce := diff.V2FileContractElement
+	if rev, ok := diff.V2RevisionElement(); ok {
 		fce = rev
 	}
 	if err := tx.UpdateContractElement(fce); err != nil {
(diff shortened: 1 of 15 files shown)


contracts/manager.go
```diff
@@ -0,0 +1,51 @@
...
+
+type (
+	// ContractManagerOpt is a functional option for the ContractManager.
+	ContractManagerOpt func(*ContractManager)
+
+	// ContractManager manages the host announcements.
+	ContractManager struct {
+		log *zap.Logger
+
+		contractRejectBuffer           time.Duration
```
n8mgr, line 17: Should this be blocks? It's extremely unlikely, but 6h could still only be one block

persist/postgres/init.sql
```diff
@@ -70,6 +70,7 @@ CREATE TABLE contracts (
   contract_id BYTEA NOT NULL UNIQUE DEFERRABLE,
 
   -- lifetime related columns
+  formation_time TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
```
n8mgr: Should this be height?

persist/postgres/types.go
```diff
@@ -293,6 +267,40 @@ func (pk *sqlPublicKey) Scan(src interface{}) error {
...
+	var buf bytes.Buffer
+	e := types.NewEncoder(&buf)
+	s.inner.EncodeTo(e)
+	if err := e.Flush(); err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
+
+func asSiaEncoded[T encoderToDecoderFrom](t T) siaEncoded[T] {
```
n8mgr: This makes me feel pretty similarly to the generic encoder func. It's helpful, but makes it easier to introduce subtle bugs. Man I miss traits...

contracts/manager.go
ChrisSchinnerl: I was going back and forth on this and `formation_time` tbh. On one hand blocks might be slow but on the other one they might be super fast as well.
So 6 hours (or maybe more?) seemed very predictable and consistent which was nice.

persist/postgres/init.sql
ChrisSchinnerl: I picked time here because height felt like a lie. We add the contract right after forming it but before it is mined. So this seemed more accurate.

contracts/chain.go
```diff
@@ -0,0 +1,160 @@
...
+		return false, fmt.Errorf("failed to determine whether contract is known: %w", err)
+	}
+	tx.knownContracts[fcid] = known
+	return known, nil
+}
+
+// UpdateChainState state updates the contracts' state in the database and
+// broadcasts revisions for failed expired contracts.
+func (m *ContractManager) UpdateChainState(tx UpdateTx, reverted []chain.RevertUpdate, applied []chain.ApplyUpdate) error {
+	cTx := &updateTx{
```
peterjan: I don't think we can do `cTx` 😅 . I would never `nit` on a variable name but this is too confusing.

contracts/chain.go
```diff
@@ -0,0 +1,160 @@
...
+	// TODO: prune expired contracts 'expiredContractPruneBuffer' blocks after
+	// we begin broadcasting resolutions
+
+	return nil
+}
+
+func (m *ContractManager) applyContractDiff(tx *updateTx, diff consensus.V2FileContractElementDiff) error {
+	// update contract state
+	var state ContractState
+	if resolution := diff.Resolution; resolution != nil {
```
peterjan: probably a leftover here, you're not using the `resolution` variable

I would personally do something like this:

```go
var state ContractState = ContractStatePending // default state
switch {
case diff.Resolution != nil:
	state = ContractStateResolved
case diff.Created:
	state = ContractStateActive
}
```

contracts/chain.go
```diff
@@ -0,0 +1,160 @@
...
+		state = ContractStateResolved
+	} else if diff.Created {
+		state = ContractStateActive
+	}
+	if err := tx.UpdateContractState(diff.V2FileContractElement.ID, state); err != nil {
+		return fmt.Errorf("failed to update contract state for new contract: %w", err)
```
peterjan:
```suggestion
		return fmt.Errorf("failed to update contract state %v: %w", diff.V2FileContractElement.ID, err)
```

contracts/chain.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	m.log.Info("contract state changed", zap.Stringer("contractID", diff.V2FileContractElement.ID),
+		zap.Stringer("state", state))
+
+	// update contract elements
+	var fce types.V2FileContractElement
```
peterjan: `nit`

```go
fce := diff.V2FileContractElement
if rev, ok := diff.V2RevisionElement(); ok {
    fce = rev
}
```

contracts/chain.go
```diff
@@ -0,0 +1,160 @@
...
+		if err := m.revertContractDiff(tx, diff); err != nil {
+			return fmt.Errorf("failed to apply contract diff: %w", err)
+		}
+	}
+	return nil
+}
+
+func (m *ContractManager) revertContractDiff(tx *updateTx, diff consensus.V2FileContractElementDiff) error {
+	// update contract state
+	var state ContractState
```
peterjan:
```go
var state ContractState = ContractStatePending // default
switch {
case diff.Resolution != nil:
	state = ContractStateActive
case diff.Created:
	state = ContractStatePending
}
```

contracts/chain.go
```diff
@@ -0,0 +1,160 @@
...
+	if diff.Created {
+		state = ContractStatePending
+	} else if resolution := diff.Resolution; resolution != nil {
+		state = ContractStateActive
+	}
+	if err := tx.UpdateContractState(diff.V2FileContractElement.ID, state); err != nil {
+		return fmt.Errorf("failed to update contract state for new contract: %w", err)
```
peterjan:
```suggestion
		return fmt.Errorf("failed to update contract state %v: %w", diff.V2FileContractElement.ID, err)
```

persist/postgres/contracts_test.go
```diff
@@ -235,3 +235,142 @@ func TestSetContractGood(t *testing.T) {
...
+		return tx.AddHostAnnouncement(hk, chain.V2HostAnnouncement{}, time.Now())
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// create a contract
+	contractID := types.FileContractID{1, 2, 3}
+
+	updateState := func(state contracts.ContractState) {
```
peterjan, line 346: missing `t.Helper()`

persist/postgres/init.sql
peterjan: Huge `nit` but are we doing `_time` instead of `_at`? Fine with either as long as we're consistent. This is the first though, just wanted to mention it.

contracts/chain_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+	// mock the update tx
+	mock := newMockUpdateTx()
+	mock.AddContract(fce)
+	updateTx := &updateTx{
+		UpdateTx:       mock,
+		knownContracts: make(map[types.FileContractID]bool),
+	}
+
+	// helper to apply/revert diff
```
peterjan, line 99: missing `t.Helper()` twice

persist/postgres/contracts_test.go
```diff
@@ -235,3 +235,142 @@ func TestSetContractGood(t *testing.T) {
...
+		ele, err := store.FileContractElement(fce.ID)
+		if err != nil {
+			t.Fatal(err)
+		} else if !reflect.DeepEqual(ele, fce) {
+			t.Fatalf("mismatch: \n%v+\n%v+", fce, ele)
+		}
+	}
+
+	// contract shouldn't exist
+	isKnownContract := func() (known bool) {
```
peterjan: `nit` `assertKnownContract` that takes a bool would make the test look really clean

contracts/chain.go
```diff
@@ -0,0 +1,160 @@
...
+	"go.sia.tech/coreutils/wallet"
+	"go.uber.org/zap"
+)
+
+type (
+	// UpdateTx defines what the contract manager needs to atomically process a
+	// chain update in the database.
+	UpdateTx interface {
+		IsKnownContract(contractID types.FileContractID) (bool, error)
+		RejectContracts(time.Duration) error
```
peterjan: `nit` might as well leave this out for the time being

persist/postgres/contracts.go
```diff
@@ -154,6 +156,47 @@ func (s *Store) SetContractBad(contractID types.FileContractID) error {
...
+
+func (tx *updateTx) RejectContracts(duration time.Duration) error {
+	panic("not implemented")
+}
+
+func (tx *updateTx) UpdateContractElement(fce types.V2FileContractElement) error {
+	_, err := tx.tx.Exec(tx.ctx, `
+INSERT INTO contract_elements (contract_id, contract, leaf_index, merkle_proof)
+VALUES (
+  (SELECT id FROM contracts WHERE contracts.contract_id = $1),
```
peterjan:
```suggestion
  (SELECT id FROM contracts WHERE contract_id = $1),
```

peterjan: I would just return `err` here because we're repeating all our db errors again higher up. I think perhaps a good rule is to never provide the error context in the database code but rather provide context higher up.

peterjan (commented):
Looks great! I got nothing but small nits I'm afraid. One thing we could consider is modelling the contract lifecycle as a state machine. We've never done that, but it would be the perfect candidate. There's definitely value in asserting a contract's state can never make illegal state transitions.

contracts/chain.go
ChrisSchinnerl: Pretty good catch. The default state was actually a bug. I changed it and added more testing

persist/postgres/init.sql
ChrisSchinnerl: I guess we don't really need _time at all? After all we call timestamp stuff like `first_seen`, `next_scan`, `last_connect` so we can probably remove the `_time` here altogether.

persist/postgres/contracts.go
ChrisSchinnerl: I don't think we can have this as a general rule since some database code potentially fails at multiple different tasks. e.g. `AddRenewedContract`. But we can probably omit it where the context essentially matches the function name like here.

peterjan (dismissed):

persist/postgres/wallet.go
```diff
@@ -166,7 +166,7 @@ func (u *updateTx) WalletApplyIndex(index types.ChainIndex, created, spent []typ
 	if len(events) > 0 {
 		for _, e := range events {
 			if res, err := u.tx.Exec(u.ctx, `INSERT INTO wallet_events (chain_index, maturity_height, event_id, event_type, event_data) VALUES ($1, $2, $3, $4, $5)`,
-				sqlChainIndex(e.Index),
+				(*sqlChainIndex)(&e.Index),
```
n8mgr: This shouldn't need to be changed, right?

ChrisSchinnerl: Nope, just a leftover from the refactor

peterjan (approved):
approving this evil PR

PR #55, opened 2025-03-05
RFC: API Authentication

### LINKED ISSUE #16: Add RFCs
This issue serves as a collection of RFCs which should write before actually implementing the corresponding feature. If any RFCs come to mind that should be added to this list please add them.

- [x] Database schema and migration
- [x] (User) Account management
- [x] (Ephemeral) Account management (registering, refills)
- [x] API authentication
  - [ ] Basic Authentication (root access) 
  - [x] For apps (e.g. url signing)
- [x] Pinning
- [x] Price Pinning
- [x] Data migration
- [x] Pruning
- [x] Host scanning, scoring and selection
- [x] Contract maintenance and lifecycle
- [x] Gouging Checks
- [x] Data Integrity checks

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/indexd/issues/16

### DIFF
--- a/doc/rfc/011_api_authentication.md
+++ b/doc/rfc/011_api_authentication.md
@@ -1,0 +1,42 @@
+# API Authentication
+
+## Abstract
+
+The `indexer` serves an http API that allows clients to interact with it. There
+are two different types of clients, the admin, who gets full access to the API
+including the UI, and users, who can only access the API relevant to pinning
+slabs. For additional security, these APIs are served on separate ports.
+
+## Admin API
+
+The admin API is protected via http basic authentication and should never be
+exposed on the public internet. Instead, tools such as SSH tunnels, VPNs or
+existing acess control systems such as Authelia should be used to avoid
+reinventing the wheel in the `indexer`.
+
+For example, SiaHub will probably access the admin API for registering a new
+account for an app after autenticating the user. That authentication will be
+implemented within SiaHub which will have access to the admin API via some sort
+of virtual network over docker, K8s or similar.
+
+## User API
+
+The user API is a bit more tricky. Users or rather their apps need to be able to
+authenticate via a private key corresponding to a previously registered public
+key. This public key serves for identifying the app as well as the account the
+`indexer` funds for the user with its hosts.
+
+### Signed URLs
+
+TODO: write section
+
+NOTE: This section is heavily inspired by v4 signatures used by AWS or actually
+Google Cloud's documentation on it. For more information see the
+[overview](https://cloud.google.com/storage/docs/access-control/signed-urls) and
+[manual
+signing](https://cloud.google.com/storage/docs/access-control/signing-urls-manually)
+pages of their docs.
+
+### Discussion
+
+For extra security it
--- a/doc/rfc/011_api_authentication.md
+++ b/doc/rfc/011_api_authentication.md
@@ -34,21 +34,18 @@ To create a signed URL, follow the steps below:
 
 ```
 HOST: <hostname> // e.g. indexer.sia.tech
-ValidFrom: <unix timestamp>
-Validity: <seconds>
+X-SiaIdx-ValidUntil: <unix timestamp>
 ```
 
 - `Host`: Makes sure a request for indexer.sia.tech can't be reused for indexer.thirdparty.tech
-- `X-SiaIdx-ValidFrom`: The time at which the request becomes usable in Unix timestamp format
-- `X-SiaIdx-Validity`: The time after which the request expires in seconds (e.g. 10)
+- `X-SiaIdx-ValidUntil`: The time at which the request expires in Unix timestamp format
 
 2. Sign the payload using ED25519
 
 3. Construct the signed URL by attaching the query parameters:
 - `X-SiaIdx-Credential`: The public key used to verify the signature
 - `X-SiaIdx-Signature`: The signature from step 2
-- `X-SiaIdx-ValidFrom`: The time at which the request becomes usable in Unix timestamp format
-- `X-SiaIdx-Validity`: The time after which the request expires in seconds
+- `X-SiaIdx-ValidUntil`: The time at which the request expires in Unix timestamp format
 
 #### Trade-offs:
 
--- a/doc/rfc/011_api_authentication.md
+++ b/doc/rfc/011_api_authentication.md
@@ -30,62 +30,35 @@ key. This public key serves for identifying the app as well as the account the
 
 To create a signed URL, follow the steps below:
 
-1. Construct the canonical request like so
+1. Construct the payload to sign using the following format
 
 ```
-HTTP_VERB\n
-PATH_TO_RESOURCE\n
-CANONICAL_QUERY_STRING\n
-CANONICAL_HEADERS\n
-
-SIGNED_HEADERS
+HOST: <hostname> // e.g. indexer.sia.tech
+ValidFrom: <unix timestamp>
+Validity: <seconds>
 ```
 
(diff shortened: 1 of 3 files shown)


n8mgr:
👀

doc/rfc/011_api_authentication.md
```diff
@@ -0,0 +1,91 @@
...
+CANONICAL_HEADERS\n
+
+SIGNED_HEADERS
+```
+
+- `HTTP_VERB`: The HTTP verb like `GET`, `POST`, etc.
+- `PATH_TO_RESOURCE`: The path to the resource like `/api/slab/pin`
+- `CANONICAL_QUERY_STRING`: The query string sorted by name using a lexicographical sort by code point value, separated by '&' or a \n if there are no query parameters.
+- `CANONICAL_HEADERS`: One line per header, lowercase name for each header, sorted like the query strings, duplicate headers collapsed into comma-separated list, folding whitespaces or newlines replaced with single space
+- `SIGNED_HEADERS`: List of lowercase header names to be signed, separated by semicolon (TODO: not sure if that is actually necessary considering that we have the CANONICAL_HEADERS)
```
ChrisSchinnerl: This is something I don't get. But maybe one of the reviewers does.

Definitely makes sense to send the `X-SiaIdx-SignedHeaders` for the `indexer` to know which headers are part of the canonical request, but I don't get why the actual canonical request needs both `CANONICAL_HEADERS` as well as `SIGNED_HEADERS`. That seems like redundant information to me.

doc/rfc/011_api_authentication.md
```diff
@@ -0,0 +1,91 @@
...
+`host`: The hostname the request gets sent to like `indexer.sia.tech`
+`x-siaidx-` prefixed headers: All potential `indexer`-specific headers to be consumed by the API
+
+
+2. Construct the string to sign:
+
+NOTE: AWS creates this string from the signing algorithm, the date (same as
+`X-SiaIdx-Date`), the credential scope and the hashed request. It is not quite
+clear why but it seems like we can simplify this by just using the canonical
+request directly for signing.
```
ChrisSchinnerl: If you check out the provided links, you'll see that the string-to-sign is actually a middle layer that contains the hash of the canonical request and additional information such as credential scope and the same Date as the query strings.
The latter is again redundant info in my opinion and the former not useful to us.

doc/rfc/011_api_authentication.md
```diff
@@ -0,0 +1,91 @@
...
+The user API is a bit more tricky. Users or rather their apps need to be able to
+authenticate via a private key corresponding to a previously registered public
+key. This public key serves for identifying the app as well as the account the
+`indexer` funds for the user with its hosts.
+
+### Signed URLs
+
+To create a signed URL, follow the steps below:
+
+1. Construct the canonical request like so
```
n8mgr: I'm not super thrilled this is based off the AWS signature scheme. Feels like there's a simpler implementation somewhere to accomplish effectively the same.

doc/rfc/011_api_authentication.md
```diff
@@ -0,0 +1,91 @@
...
+reinventing the wheel in the `indexer`.
+
+For example, SiaHub will probably access the admin API for registering a new
+account for an app after autenticating the user. That authentication will be
+implemented within SiaHub which will have access to the admin API via some sort
+of virtual network over docker, K8s or similar.
+
+## User API
+
+The user API is a bit more tricky. Users or rather their apps need to be able to
```
peterjan:
```suggestion
The user API is a bit more tricky. Users, or rather their apps, need to be able to
```

doc/rfc/011_api_authentication.md
peterjan: Definitely seems redundant. I'm not sure why AWS did that but it's probably either simply for clarity or some obscure security related thing. I tried but I could not find many threads online that ask the same question, I also wouldn't trust an answer from chat gpt on this 😛 .

doc/rfc/011_api_authentication.md
ChrisSchinnerl: So the only way I think we can get rid of more complexity is by assuming that clients always use https and we don't have to worry about tampering requests.

Then we can probably get away with only signing the data and validity for the canonical request. e.g.
```
Host: indexer.sia.tech
ValidFrom: foo
ValidDuration: bar
```
Sign that, add those two as query parameters to the request. 
That way we get no protection against tampering or adding headers and query params but that might be ok.
The `Host` at least makes sure a indexer can't MITM you by proxying your request to another indexer if you reuse your keys between different indexers.

peterjan: We probably want to add a nonce for replay protection? An attacker could use a compromised, but otherwise valid and authenticated, request more than once.

ChrisSchinnerl: That's effectively a feature. Because for a nonce to protect against that you need it to be generated by the server which requires a handshake.
So the compromise is to make the expiry as short as possible. e.g. 10 seconds, rely on https being secure and the malicious actor neither having control over the indexer nor the client.

peterjan (dismissed):

peterjan (dismissed):

doc/rfc/011_api_authentication.md
```diff
@@ -0,0 +1,64 @@
...
+account the `indexer` funds for the user with its hosts.
+
+### Signed URLs
+
+To create a signed URL, follow the steps below:
+
+1. Construct the payload to sign using the following format
+
+```
+HOST: <hostname> // e.g. indexer.sia.tech
```
n8mgr, line 36: Should this be hostname or the URL with path and query string? Definitely agree we don't need to sign the headers individually since we don't use headers for anything.

ChrisSchinnerl: Just the hostname. We could include the path and query string at the cost of slightly more complexity since we need to deterministically construct the canonical query string.
But then an attacker could also only replay the exact operations we have already performed.

Do you want me to add that?

doc/rfc/011_api_authentication.md
```diff
@@ -0,0 +1,64 @@
...
+- `X-SiaIdx-ValidFrom`: The time at which the request becomes usable in Unix timestamp format
+- `X-SiaIdx-Validity`: The time after which the request expires in seconds (e.g. 10)
+
+2. Sign the payload using ED25519
+
+3. Construct the signed URL by attaching the query parameters:
+- `X-SiaIdx-Credential`: The public key used to verify the signature
+- `X-SiaIdx-Signature`: The signature from step 2
+- `X-SiaIdx-ValidFrom`: The time at which the request becomes usable in Unix timestamp format
+- `X-SiaIdx-Validity`: The time after which the request expires in seconds
```
lukechampine: how do you feel about replacing `ValidFrom` and `Validity` with `ValidUntil`?

doc/rfc/011_api_authentication.md
lukechampine: imo the point of this scheme is that HTTPS handles all of this. It guarantees privacy, i.e. only the server sees your path and query params -- and the server can't replay our requests to a different server (and has no incentive to replay attack *itself*).

Honestly the whole thing isn't much different from BasicAuth + HTTPS 🤔

ChrisSchinnerl: Indeed. Only real difference being the expiry.

doc/rfc/011_api_authentication.md
ChrisSchinnerl: Yeah I think that makes sense

peterjan (approved):

lukechampine (approved):

PR #56, opened 2025-03-07
Add QUIC support to RHP4 client

peterjan:
This PR adds QUIC support to our RHP4 client (and test host). Keeping our transport pool ended up being ~~slightly~~ pretty messy because we have the `transport` do the `Dial`. This means that if we want to pass in client options, like disabling the cert. verification in tests, we'd have to drag the options all the way down with us or refactor the transport pool. On top of that we have to keep track of siamux and quic transports separately. Because we already [commented on it here](https://github.com/SiaFoundation/indexd/pull/25#issuecomment-2654123583), I figured it made sense to propose to remove the complexity instead of refactoring it.


internal/rhp/v4/client.go
ChrisSchinnerl, line 1: I wouldn't add quic yet. We haven't tested if it even works the way we imagine it to yet, probably need to build a certificate service to get all hosts on board and overall probably won't even need it before the first client apps start implementing it.

When we do add it I would also use the coreutils lib directly rather than adding it here. We will only ever use it for scanning which means we will never reuse any connections.

peterjan: Oh wow ok. I figured we were pretty close to fully supporting both. If we're this far off, do we really need to add QUIC support in the scanner at all? It feels a little weird using a client for the siamux and go another route for connecting to the host over QUIC. It's much cleaner if the manager gets injected something that allows it to seamlessly connect to hosts over both protocols.

ChrisSchinnerl: I'd say we cross that bridge when we come to it. For now we can just give the scanner an interface it can use to mock the mux in testing and once we have quic support we worry about integrating that.

PR #57, opened 2025-03-07
Update contract element proofs 

ChrisSchinnerl:
This PR adds updating of contract element proofs via a `updateContractElementProofs` function which is unit tested using a mocked store and proof updater.
Whether the function is called as part of a consensus update will become apparent when we have integration tests actually forming contracts.

### DIFF
--- a/contracts/chain.go
+++ b/contracts/chain.go
@@ -6,6 +6,7 @@ import (
 	"go.sia.tech/core/consensus"
 	"go.sia.tech/core/types"
 	"go.sia.tech/coreutils/chain"
+	"go.sia.tech/coreutils/wallet"
 	"go.uber.org/zap"
 )
 
@@ -13,8 +14,9 @@ type (
 	// UpdateTx defines what the contract manager needs to atomically process a
 	// chain update in the database.
 	UpdateTx interface {
+		ContractElements() ([]types.V2FileContractElement, error)
 		IsKnownContract(contractID types.FileContractID) (bool, error)
-		UpdateContractElement(fce types.V2FileContractElement) error
+		UpdateContractElements(fces ...types.V2FileContractElement) error
 		UpdateContractState(contractID types.FileContractID, state ContractState) error
 	}
 
@@ -59,8 +61,6 @@ func (m *ContractManager) UpdateChainState(tx UpdateTx, reverted []chain.RevertU
 		}
 	}
 
-	// TODO: update file contract element proofs
-
 	// TODO: reject all contracts that have been pending for more than 'contractRejectBuffer'
 
 	// TODO: broadcast resolutions for expired contracts
@@ -84,7 +84,9 @@ func (m *ContractManager) applyChainUpdate(tx *updateTx, cau chain.ApplyUpdate) 
 			return fmt.Errorf("failed to apply contract diff: %w", err)
 		}
 	}
-	return nil
+
+	// update state element proofs
+	return updateContractElementProofs(tx, cau)
 }
 
 func (m *ContractManager) applyContractDiff(tx *updateTx, diff consensus.V2FileContractElementDiff) error {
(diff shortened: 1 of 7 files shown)


contracts/chain.go
```diff
@@ -6,15 +6,17 @@ import (
...
 )
 
 type (
 	// UpdateTx defines what the contract manager needs to atomically process a
 	// chain update in the database.
 	UpdateTx interface {
+		ContractElements() ([]types.V2FileContractElement, error)
 		IsKnownContract(contractID types.FileContractID) (bool, error)
-		UpdateContractElement(fce types.V2FileContractElement) error
+		UpdateContractElements(fces ...types.V2FileContractElement) error
```
n8mgr, line 21: This doesn't look like its ever used with more than one element. Could be reverted unless you have specific plans for it.

persist/postgres/contracts.go
```diff
@@ -154,6 +154,26 @@ func (s *Store) SetContractBad(contractID types.FileContractID) error {
...
+INNER JOIN contracts c ON fces.contract_id = c.id
+`)
+	if err != nil {
+		return nil, err
+	}
+	var fces []types.V2FileContractElement
+	for rows.Next() {
+		fce, err := scanContractElement(rows)
+		if err != nil {
+			return nil, err
```
n8mgr: add error context

persist/postgres/contracts.go
```diff
@@ -164,15 +184,20 @@ func (tx *updateTx) IsKnownContract(contractID types.FileContractID) (bool, erro
...
+		_, err := tx.tx.Exec(tx.ctx, `
 INSERT INTO contract_elements (contract_id, contract, leaf_index, merkle_proof)
 VALUES (
   (SELECT id FROM contracts WHERE contract_id = $1),
   $2, $3, $4
 ) ON CONFLICT (contract_id) DO UPDATE SET contract = EXCLUDED.contract, leaf_index = EXCLUDED.leaf_index, merkle_proof = EXCLUDED.merkle_proof
 `, sqlHash256(fce.ID), (*sqlFileContract)(&fce.V2FileContract), fce.StateElement.LeafIndex, sqlMerkleProof(fce.StateElement.MerkleProof))
-	return err
+		if err != nil {
+			return err
```
n8mgr: Might be useful to have the parent ID in the context.

n8mgr (changes requested):

contracts/chain.go
peterjan: It's used [here](https://github.com/SiaFoundation/indexd/pull/57/files#diff-153e46308df095ee60b2297b930dde10e74f9cc6491493e94797990c263a548aR174)

contracts/chain_test.go
```diff
@@ -10,20 +10,43 @@ import (
...
+	updateFn func(*types.StateElement)
+}
+
+func (u *mockProofUpdater) UpdateElementProof(stateElement *types.StateElement) {
+	u.updateFn(stateElement)
+}
+
+// newMockProofUpdater creates a mock implementation of ProofUpdater with a custom
+// update function.
+func newMockProofUpdater(updateFn func(*types.StateElement)) wallet.ProofUpdater {
```
peterjan: this is not used and I would remove it instead of using it in the spots we need the mock proof updater

persist/postgres/contracts.go
```diff
@@ -164,15 +184,20 @@ func (tx *updateTx) IsKnownContract(contractID types.FileContractID) (bool, erro
...
 }
 
-func (tx *updateTx) UpdateContractElement(fce types.V2FileContractElement) error {
-	_, err := tx.tx.Exec(tx.ctx, `
+func (tx *updateTx) UpdateContractElements(fces ...types.V2FileContractElement) error {
+	for _, fce := range fces {
+		_, err := tx.tx.Exec(tx.ctx, `
 INSERT INTO contract_elements (contract_id, contract, leaf_index, merkle_proof)
 VALUES (
   (SELECT id FROM contracts WHERE contract_id = $1),
```
peterjan, line 199: it's a little awkward but we could get rid of this sub select if you return a map in `ContractElements` that exposes the `contract_id`

ChrisSchinnerl: Exposing database ids outside of the persist package is definitely a bit too nasty for my taste ^^

peterjan (dismissed):

persist/postgres/contracts.go
peterjan: Yeah it's pretty horrible

peterjan (approved):

n8mgr (approved):

PR #59, opened 2025-03-10
Add host pruning

peterjan:


hosts/manager.go
```diff
@@ -25,6 +25,10 @@ const (
 	ipv4FilterRange = 24
 	ipv6FilterRange = 32
 
+	pruneFrequency                  = time.Hour * 24
+	pruneMinConsecutiveScanFailures = 10
+	pruneMinDowntime                = time.Hour * 24 * 365 // 1 year
```
peterjan, line 30: One year is what it is in the spec but that seems excessive. What about 3 months?

hosts/manager.go
```diff
@@ -101,7 +106,23 @@ func NewManager(store Store, opts ...Option) (*HostManager, error) {
 	if err != nil {
 		return nil, err
 	}
-	go func() {
+	go func(ctx context.Context, cancel context.CancelFunc) {
```
n8mgr: We don't really need a separate goroutine for pruning. The query should be fairly performant and I don't think it will cause enough IO to need to wait 24 hours.

```go
go func() {
for {
    m.pruneHosts(ctx)
    m.scanHosts(ctx)
    select {
    case <-ctx.Done():
        return
    case <-time.After(min(pruneFrequency,scanFrequency))
    }
}
}()
```

hosts/manager.go
ChrisSchinnerl: I wouldn't change any of the constants we already signed off on unless there is a good reason to do so. In this case we probably wanted it to also align with the 1 year cutoff for host announcements.

hosts/manager.go
ChrisSchinnerl: I agree that any I/O happening in parallel is just a chance for the database to lock up. I'd just propose a minor change here

```go
go func() {

pruneTicker := ...
scanTicker := ...

for {
    select {
    case <-pruneTicker.C:
        m.pruneHosts()
    case <-scanTicker.C:
        m.scanHosts()
    case <-ctx.Done():
        return
    }
}
}()
```

That way it is still single-threaded but every operation occurs at their respective interval.
If we use `min` it feels like one of the vars is obsolete and we should condense them into a single one in the first place.

persist/postgres/hosts.go
```diff
@@ -145,6 +145,25 @@ func (s *Store) HostsForScanning(ctx context.Context) ([]types.PublicKey, error)
...
 }
 
+// PruneHosts removes hosts that have not been successfully scanned since the
+// given cutoff time and have failed scans consecutively for at least
+// minConsecutiveFailedScans times.
+func (s *Store) PruneHosts(ctx context.Context, minLastSuccessfulScan time.Time, minConsecutiveFailedScans int) (int64, error) {
+	var n int64
+	if err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		// TODO: the good = TRUE condition should probably be extended or altered because that field is not updated on renewal
+		res, err := tx.Exec(ctx, `DELETE FROM hosts WHERE last_successful_scan <= $1 AND consecutive_failed_scans >= $2 AND NOT EXISTS (SELECT * FROM contracts WHERE host_id = hosts.id AND good = TRUE)`, minLastSuccessfulScan, minConsecutiveFailedScans)
```
ChrisSchinnerl: This will fail since we don't delete cascade contracts. Assuming we have an offline host with a bad contract.

Considering that we slightly diverged from the spec by never actually deleting contracts, we probably need to update the inner part of the query to `SELECT 1 FROM contracts WHERE host_id = hosts.id` to straight up avoid pruning any host we have or had a contract with.

For deleting hosts we had contracts with we'll need to provide tools for users to manually delete contracts or hosts.

ChrisSchinnerl (changes requested):

hosts/manager.go
peterjan: I wanted to avoid using tickers. I definitely agree with the `min` making the separate intervals obsolete. I do think it makes sense to keep both instead of condensing them into one so I updated the code to use tickers instead.

ChrisSchinnerl (approved):

n8mgr (approved):

PR #61, opened 2025-03-11
Host Scanning F/Us

peterjan:
I resolved 5 nits to tackle as F/U's but ended up not doing all of them.

These are done:
- [x] https://github.com/SiaFoundation/indexd/pull/41#discussion_r1988687657
- [x] https://github.com/SiaFoundation/indexd/pull/41#discussion_r1988198970
- [x] https://github.com/SiaFoundation/indexd/pull/41#discussion_r1988684992
- [x] https://github.com/SiaFoundation/indexd/pull/41#discussion_r1988196200

This one is not:

- https://github.com/SiaFoundation/indexd/pull/41#discussion_r1988200253
Don't mind making this a `.Debug` and I usually agree with these, but in this case we should never see an error there. All host related errors (think network I/O) as well as `context.Cancelled` are handled, so it's really only if the database is not reachable or some other very exotic event, which is why I'd prefer to keep it as an `error`.


ChrisSchinnerl (commented):
> I was going to do this one but not using the Client is kind of annoying.

Removing the client shouldn't really affect the host manager too much right? We effectively need to pass a `Client` interface with a `Settings` method.
Then we pass a minimal wrapper around the siamux which implements `Settings` by dialing a stream and calling `RPCSettings` on it. 

Later, when we add quic support we can extend the client with a `SettingsQuic` interface method which is also implemented by the wrapper.

hosts/scanner.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+type scanner struct{}
+
+// Settings executes the RPCSettings RPC on the host.
+func (c *scanner) Settings(ctx context.Context, hk types.PublicKey, addr string) (hs proto4.HostSettings, _ error) {
+	t, err := siamux.Dial(ctx, addr, hk)
+	if err != nil {
+		return proto4.HostSettings{}, fmt.Errorf("failed to upgrade connection: %w", err)
+	}
```
ChrisSchinnerl, line 20: Missing a `defer t.Close()`.

hosts/scanner.go
```diff
@@ -0,0 +1,23 @@
...
+	proto4 "go.sia.tech/core/rhp/v4"
+	"go.sia.tech/core/types"
+	"go.sia.tech/coreutils/rhp/v4"
+	"go.sia.tech/coreutils/rhp/v4/siamux"
+)
+
+type scanner struct{}
+
+// Settings executes the RPCSettings RPC on the host.
+func (c *scanner) Settings(ctx context.Context, hk types.PublicKey, addr string) (hs proto4.HostSettings, _ error) {
```
ChrisSchinnerl:
```suggestion
func (c *scanner) Settings(ctx context.Context, hk types.PublicKey, addr string) (proto4.HostSettings, error) {
```

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

n8mgr (approved):

PR #62, opened 2025-03-11
Add host checks

peterjan:
This PR adds a usability field on the `Host` type and populates it on the fly when fetching hosts from the database.


doc/rfc/001_db_schema.md
```diff
@@ -39,6 +39,11 @@ CREATE TABLE global_settings (
     max_storage_price NUMERIC(50,0), -- hastings / byte / block
     max_ingress_price NUMERIC(50,0), -- hastings / byte
     max_egress_price NUMERIC(50,0) -- hastings / byte
+
+    -- host checks
```
peterjan, line 49: I wasn't too sure what to do about these. The `chain_tip_height` is used to check whether the host is not too far behind, it gets set by the subscriber every time we ask the chain manager's tip. I'm well aware we could provide these parameters to the store but if we really want to perform the host checks in the database I think it's kind of nice to have all ingredients in the database. 

I wasn't a big supporter of doing the host checks on the database, but I warmed up to it. I would probably still prefer doing it in Go though. I don't see the big deal with performing these checks on the database, but maybe I'm missing some obvious advantage of being able to do that?

Instead of asking I figured I might as well write it out to show you how readable the queries ended up being.

doc/rfc/001_db_schema.md
```diff
@@ -114,6 +119,7 @@ CREATE TABLE hosts (
     total_scans INTEGER NOT NULL DEFAULT 0,
     failed_scans INTEGER NOT NULL DEFAULT 0,
     consecutive_failed_scans INTEGER NOT NULL DEFAULT 0,
+    last_failed_scan TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT '0001-01-01 00:00:00+00',
```
peterjan, line 124: I'm realising that I could've gone back to tracking `uptime` and `downtime`. What do you guys think? I'm sure we can somehow avoid keeping track of this somewhat redundant info but I don't see the point if you have to extrapolate it with some kind of `[num_scans * scan_frequency]` interval that only makes things more complex. Two timestamps per host are not that expensive.

persist/postgres/hosts.go
```diff
@@ -14,14 +14,33 @@ import (
...
+	// not being found.
+	ErrHostNotFound = errors.New("host not found")
+)
 
-type dbHost struct {
-	id int64
-	hosts.Host
-}
+var (
+	unusableNoSettings         = "Host has no settings"
```
n8mgr: Should these be errors instead of strings?

persist/postgres/hosts.go
```diff
@@ -226,33 +202,185 @@ DO UPDATE SET
...
+		hs.storage_price,
+		hs.ingress_price,
+		hs.egress_price,
+		hs.free_sector_price,
+		hs.tip_height,
+		hs.valid_until,
+
+		array_remove(ARRAY[
+			CASE WHEN ((h.total_scans - h.failed_scans) * 100.0 / NULLIF(h.total_scans, 0) < 90) OR h.last_failed_scan + INTERVAL '1 week' > NOW() THEN '`+unusableUptime+`' ELSE NULL END,
+			CASE WHEN hs.max_contract_duration < settings.contract_period THEN '`+unusableMaxDuration+`' ELSE NULL END,
```
n8mgr: It's clean enough, but would honestly rather keep this in Go. In my experience array and json ops can be slow on even a moderately sized table without proper indexing.

persist/postgres/hosts.go
```diff
@@ -226,33 +202,185 @@ DO UPDATE SET
 	})
 }
 
-func queryHosts(ctx context.Context, tx *txn, offset, limit int) ([]dbHost, error) {
-	rows, err := tx.Query(ctx, `SELECT id, public_key, last_announcement, total_scans, failed_scans, last_successful_scan, next_scan  FROM hosts LIMIT $1 OFFSET $2`, limit, offset)
+func queryHosts(ctx context.Context, tx *txn, hk *types.PublicKey, offset, limit int) ([]hosts.Host, error) {
```
n8mgr: nit: this doesn't really need to be a pointer. Potentially saves an alloc, but tend to avoid this except for very large ones.

persist/postgres/hosts.go
```diff
@@ -47,73 +66,30 @@ func (u *updateTx) AddHostAnnouncement(hk types.PublicKey, ha chain.V2HostAnnoun
...
-				return err
-			}
-
-			// TODO: perform host checks
-
-			hosts = append(hosts, h.Host)
-		}
-		return nil
+	if err := s.transaction(ctx, func(ctx context.Context, tx *txn) (err error) {
+		hosts, err = queryHosts(ctx, tx, nil, offset, limit)
```
n8mgr: This should really be a separate function and query. It's clever and the duplication is unfortunate, but trying to maintain and optimize it for both cases will be more difficult.

If you want to deduplicate, provide a `scanHost` func that deduplicates the logic part.

persist/postgres/hosts.go
```diff
@@ -226,33 +202,185 @@ DO UPDATE SET
...
+
+	if hk != nil {
+		cond = `WHERE h.public_key = $1`
+		args = append(args, sqlPublicKey(*hk))
+	} else {
+		cond = `LIMIT $1 OFFSET $2`
+		args = append(args, limit, offset)
+	}
+
+	query := fmt.Sprintf(`
```
n8mgr: Definitely want to avoid string concatenation in queries. It's probably fine since it's not arbitrary user data, but it sets a bad precedent and isn't necessary.

persist/postgres/hosts.go
```diff
@@ -226,33 +202,185 @@ DO UPDATE SET
...
+			CASE WHEN hs.contract_price > settings.one_sc THEN '`+unusableContractPrice+`' ELSE NULL END,
+			CASE WHEN hs.collateral < settings.min_collateral OR hs.collateral < 2 * hs.storage_price THEN '`+unusableCollateral+`' ELSE NULL END,
+			CASE WHEN hs.storage_price > settings.max_storage_price THEN '`+unusableStoragePrice+`' ELSE NULL END,
+			CASE WHEN hs.ingress_price > settings.max_ingress_price THEN '`+unusableIngressPrice+`' ELSE NULL END,
+			CASE WHEN hs.egress_price > settings.max_egress_price THEN '`+unusableEgressPrice+`' ELSE NULL END,
+			CASE WHEN hs.free_sector_price > settings.one_sc / settings.one_tb THEN '`+unusubleFreeSectorPrice+`' ELSE NULL END,
+			CASE WHEN hs.tip_height + 144 < settings.chain_tip_height THEN '`+unusableTipHeight+`' ELSE NULL END
+		], NULL) as usability_reasons
+	FROM hosts h
+	LEFT JOIN host_settings hs ON hs.host_id = h.id
```
n8mgr: Disappointing that we need to left join here. Might want to combine `hosts` and `host_settings` in a follow up.

n8mgr (changes requested):

doc/rfc/001_db_schema.md
n8mgr: Neither one is great, imo.

Cumulative uptime and downtime are bad for established hosts. A host that is 5 years old may have a cumulative downtime of 1440 hours of downtime and would be considered bad even though it's spread out over 60 months vs a host that is 3 months old.

Calculating percentage based on total and failed scans is somewhat better, but older hosts are able to get away with more failed scans before being considered bad.

I'm fine with going back to a duration, but really feel like either way we need a sliding window for the uptime check to be useful and not cause unnecessary churn.

persist/postgres/init.sql
```diff
@@ -4,6 +4,7 @@ CREATE TABLE hosts (
     total_scans INTEGER NOT NULL DEFAULT 0,
     failed_scans INTEGER NOT NULL DEFAULT 0,
     consecutive_failed_scans INTEGER NOT NULL DEFAULT 0,
+    last_failed_scan TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT '0001-01-01 00:00:00+00',
```
n8mgr, line 6: Is this necessary since we already have the opposite `last_successful_scan`?

doc/rfc/001_db_schema.md
peterjan: I'm not a fan of cumulative uptime and downtime. We should probably use some sort of decay to ensure old hosts aren't invincible, but also at the same time to stabilize new hosts as quickly as possible.

persist/postgres/init.sql
peterjan: Having both makes it incredibly easy to check for recent downtime and uptime. We need recent uptime for host checks and we need recent downtime for host pruning. It's the opposite but you can't go from one to the either right.

Going to resolve this in favour of https://github.com/SiaFoundation/indexd/pull/62#discussion_r1990576331. It's essentially all the same problem.

persist/postgres/hosts.go
peterjan: Usually they would be errors I guess but we don't have to expose them yet and I don't really want to deal with errors unless I have to. I made it constants for now.

persist/postgres/hosts.go
peterjan: I merged settings into hosts in https://github.com/SiaFoundation/indexd/pull/62 and added a `scanHost` there. If I rebase this PR I'll re-evaluate and see if I can live with `scanHost`.

persist/postgres/hosts.go
peterjan: > I wasn't a big supporter of doing the host checks on the database, but I warmed up to it. I would probably still prefer doing it in Go though. I don't see the big deal with performing these checks on the database, but maybe I'm missing some obvious advantage of being able to do that?

I'm asking about that here. I do feel it's clean enough though. When I started out I didn't expect to end up with something this concise and tight. I assumed you and Chris had a discussion where doing it in the database would give us some advantage later down the line.

@ChrisSchinnerl I'll wait on your input to go ahead with these changes.

doc/rfc/001_db_schema.md
n8mgr: +1 to doing it in Go

peterjan: Ok, I'm going ahead with moving the checks to Go and I'll remove the `chain_tip_height` from the settings.

doc/rfc/001_db_schema.md
ChrisSchinnerl: Regardless of whether we do percentage of uptime/downtime or successfull/failed scans the implications are the same right? Older hosts get away with more failures.

Regardless of what we choose though, compressing these values should be easy enough. e.g. let's say a host has 1000 scans total. 900 good, 100 bad. 
If we want to keep it to at most 1000 scans, we could computer the ratio (in this case 90%), apply it to 900 total scans and update the good scans to 900*0.9 and the bad ones to 900*0.1.
That way we keep the ratio but compress the scans to 900. Same can be done with actual uptime and downtime.

That way hosts benefit from having a lot of uptime up to 900 scans total which seems fair enough.

doc/rfc/001_db_schema.md
ChrisSchinnerl: I'd probably drop the tip height check from the query anyway since it's special and it's probably not even safe / might cause unnecessary churn. Rather the check probably only needs to be performed right before pinning slabs on fresh settings. We should have the luxury of fetching new prices before starting to pin slabs. 

My biggest issue with moving it out of the database is our ability to filter usable hosts. In `renterd` we have 3 spots I think where we filter hosts by their usability.
1. the UI, since people want to filter by good/bad hosts which includes pagination
2. when forming contracts we only fetch good hosts
3. when handing out usable hosts to the worker

The way we fixed this is by periodically updating columns on hosts to indicate whether they are good and then filter by them. But from a previous PR I know that Nate is opposed to fields that require updating.

So unless we are 100% confident we won't ever feel the need to filter hosts by whether they are passing or failing any particular checks, I'd try and keep the ones we are interested in as part of the query and move the ones we don't want, such as the tip height check to where they are needed.

persist/postgres/hosts.go
```diff
@@ -240,32 +221,153 @@ WHERE public_key = $18 RETURNING id`,
...
+			CASE WHEN settings_valid_until < NOW() + INTERVAL '1 hour' THEN '`+unusablePriceValidity+`' ELSE NULL END,
+			CASE WHEN settings_accepting_contracts IS FALSE THEN '`+unusableAcceptingContracts+`' ELSE NULL END,
+			CASE WHEN settings_contract_price > globals.one_sc THEN '`+unusableContractPrice+`' ELSE NULL END,
+			CASE WHEN settings_collateral < globals.min_collateral OR settings_collateral < 2 * settings_storage_price THEN '`+unusableCollateral+`' ELSE NULL END,
+			CASE WHEN settings_storage_price > globals.max_storage_price THEN '`+unusableStoragePrice+`' ELSE NULL END,
+			CASE WHEN settings_ingress_price > globals.max_ingress_price THEN '`+unusableIngressPrice+`' ELSE NULL END,
+			CASE WHEN settings_egress_price > globals.max_egress_price THEN '`+unusableEgressPrice+`' ELSE NULL END,
+			CASE WHEN settings_free_sector_price > globals.one_sc / globals.one_tb THEN '`+unusubleFreeSectorPrice+`' ELSE NULL END,
+			CASE WHEN settings_tip_height + 144 < globals.chain_tip_height THEN '`+unusableTipHeight+`' ELSE NULL END
+		], NULL) as usability_reasons
```
ChrisSchinnerl: I don't think we need the reasons in the query. Should be good enough to return bools and have a struct of booleans on the host

doc/rfc/001_db_schema.md
```diff
@@ -39,6 +39,11 @@ CREATE TABLE global_settings (
     max_storage_price NUMERIC(50,0), -- hastings / byte / block
     max_ingress_price NUMERIC(50,0), -- hastings / byte
     max_egress_price NUMERIC(50,0) -- hastings / byte
+
+    -- host checks
+    chain_tip_height INTEGER NOT NULL DEFAULT 0, -- chain tip height of the last block
```
ChrisSchinnerl: This would be redundant. We could split up `last_scanned_index` like I did in my contract proof updating PR.

doc/rfc/001_db_schema.md
n8mgr: I'm against fields that require updating in two cases:
1. There's more than one location flipping a boolean based on unknown criteria
2. It causes a lot of IO or churn for little benefit

ChrisSchinnerl: Ok so to make that decision easier based on those two criteria here is a list of all the checks we perform with some additional info:

- `Updated By`: the components that affect the outcome of the check
- `Relevant to`: the components that need to consume the result of the check
- `Makes host bad`: whether that check failing causes a host to be temporarily/permanently considered bad

In my head these are also the checks we add to the `Host` type as a struct of booleans. Regardless of whether the checks are fields in the db that get updated or fetched dynamically. So that we can then have methods on the `Host` such as `GoodForAccountFunding`, `GoodForContractFormation` etc.

| Check  | Updated by | Relevant to | Makes host bad |
| ------------- | ------------- | ------------| -----------------|
| uptime  | recording host scan  | contract maintenance | yes |
| max contract duration  | recording host scan, user config  | contract maintenance | yes |
| max collateral  | recording host scan  | contract maintenance | yes |
| version  | recording host scan, user | contract maintenance  | yes |
| price validity  | recording host scan  | contract maintenance | yes
| accepting contracts  | recording host scan  | contract maintenance | yes |
| contract price  | recording host scan, fetching prices, user, price pinning  | contract maintenance, migrations | yes |
| collateral | recording host scan, fetching prices, user, price pinning | contract maintenance, migrations  | yes |
| storage price  | recording host scan, fetching prices, user, price pinning  | contract maintenance, migrations | yes |
| ingress price  | recording host scan, fetching prices, user, price pinning  | contract maintenance, migrations | yes |
| egress price  | recording host scan, fetching prices, user, price pinning  | contract maintenance, migrations | yes |
| free sector price  | recording host scan, fetching prices | contract maintenance, migrations | yes |
| tip height  | nothing  | slab pinning | no |

Looking at the table it seems like we can pretty cleanly split the checks into 2 groups and ignore the tip height for now.
1. The first 6 rows which are only needed in contract maintenance. So we can essentially do what we do in `renterd` which is updating all hosts once per iteration of the maintenance.
2. The gouging checks which are affected by scans, updating prices, manual user configuration and price pinning changes. For those it probably makes sense to perform the checks in the query to avoid duplicating logic.

Then we should end up with only a single spot that updates each check, allowing us to nicely filter, paginate etc. to our heart's content.

peterjan: That overview is helpful 👍 . If we can drop the tip height, which was the thing that was bothering me most, why would we do both? We could re-add uptime and downtime and apply a decay. That allows us to perform all checks on the database right? I guess I don't see the reason why we would do what we do in `renterd` for the first 6 rows of the table... Those aren't complex enough to warrant an updated field, aside from maybe the uptime check but we can solve that by having a decayed field we update on scan.

persist/postgres/hosts.go
ChrisSchinnerl: Imo the `Host` should just have a struct with one boolean per check that we scan these into to keep the query clean. The checks should be simple enough to not require an array of reasons attached to the host.

doc/rfc/001_db_schema.md
ChrisSchinnerl: That's fine by me. Obviously it would be great to have a single source of truth when it comes to these checks without updating any fields at all.

doc/rfc/001_db_schema.md
peterjan: I added an extra field called `uptime_ema`, which essentially tracks the host's _recent_ uptime performance. Every time we scan the host we update the field and apply a time-based decay factor. I think it's important to make it time based because we have scan triggers and we wouldn't want to skew the stats only because you scanned 1000 times in short succession... It makes new hosts more robust and old hosts less invincible. When a host's uptime is low, it takes a while for the host to reach 90% uptime again, which imo is actually a nice feature. I can write some more tests around or play with the parameters (e.g. 85% uptime, or diff alpha) of course, wanted to run it by you guys first. I definitely like how we do the checks on the database now.

persist/postgres/hosts.go
```diff
@@ -240,32 +246,129 @@ WHERE public_key = $18 RETURNING id`,
...
+		last_failed_scan, last_successful_scan, next_scan,
+		settings_protocol_version, settings_release, settings_wallet_address,
+		settings_accepting_contracts, settings_max_collateral, settings_max_contract_duration,
+		settings_remaining_storage, settings_total_storage, settings_contract_price,
+		settings_collateral, settings_storage_price, settings_ingress_price,
+		settings_egress_price, settings_free_sector_price, settings_tip_height, settings_valid_until,
+		total_scans - failed_scans > 0 as scanned,
+		(get_byte(settings_protocol_version, 0) << 16) + (get_byte(settings_protocol_version, 1) << 8) + (get_byte(settings_protocol_version, 2)) as settings_version
+	FROM hosts
+	WHERE public_key = $1 OR $1 IS NULL
```
peterjan: @n8maninger I feel this is pretty clean, it allows us to achieve full deduplication and have the host query in one spot for both `Host` and `Hosts`. So I ended up getting rid of `scanHost` and calling `queryHosts` with a NULL host key.

n8mgr: Definitely cleaner, but I'm still against it. I strongly believe focusing on deduplication for database queries ends up hurting maintainability and performance in the long run. However, if I'm the only one that feels this way, I'm not going to fight too hard on this one. 

I haven't actually looked at the query plan, but the `IS NULL` override is definitely something that I would double check doesn't alter it in subtle and annoying ways, like triggering a full table scan for no reason. Using a CTE for `hosts` is bound to reduce performance and is probably not necessary as well. Biggest thing with SQL queries is does the query plan look sane. If it doesn't, it has to be changed.

ChrisSchinnerl: I definitely would have jumped through hoops to deduplicate a lot of database code in the past but I think now I'm also leaning towards duplicating code rather than finding workarounds to merge both queries.

persist/postgres/types.go
```diff
@@ -305,10 +305,13 @@ func (np *sqlNetworkProtocol) Scan(src interface{}) error {
 type sqlPublicKey types.PublicKey
 
 func (pk sqlPublicKey) Value() (driver.Value, error) {
+	if pk == (sqlPublicKey{}) {
```
ChrisSchinnerl: Feels a bit odd that for things like block ids we desperately try to avoid NULL in our database by specifying a default value and here we turn the default into a NULL on purpose.

Especially since we now produce database entries in `Value` that `Scan` can't actually decode which destroys the symmetry between these functions. Almost like being able to json encode a type but not being able to json decode it afterwards.

persist/postgres/init.sql
```diff
@@ -1,9 +1,11 @@
 CREATE TABLE hosts (
     id SERIAL PRIMARY KEY,
     public_key BYTEA UNIQUE NOT NULL CHECK (LENGTH(public_key) = 32),
+    uptime_ema DOUBLE PRECISION NOT NULL DEFAULT 1,
```
ChrisSchinnerl: Could check for this to be between 0 and 1.

persist/postgres/hosts.go
```diff
@@ -14,14 +14,15 @@ import (
 	"go.sia.tech/indexd/hosts"
 )
 
-// ErrHostNotFound is returned by database operations that fail due to a host
-// not being found.
-var ErrHostNotFound = errors.New("host not found")
+const (
+	uptimeHalfLife = 60 * 60 * 24 * 6 // 6 weeks
```
ChrisSchinnerl: I'd leave a comment here giving a few more details on what it means for this to be 6 weeks with an example. e.g. what the uptime of a host looks like that originally had 100% uptime and then went offline for 6 weeks and maybe also how long it takes to reach 90% uptime on a host like that.

doc/rfc/001_db_schema.md
ChrisSchinnerl: I'd call this `recent_uptime`. `ema` isn't very descriptive.

persist/postgres/hosts.go
peterjan: OK, duplicating... (to be fair, the Scan Value comment got me)

persist/postgres/hosts.go
```diff
@@ -14,9 +14,22 @@ import (
...
+	// 50%. This decay mechanism emphasizes recent scan results, giving new
+	// hosts a fairer evaluation and ensuring that long-established hosts do not
+	// maintain high uptime scores solely based on older data.
+	uptimeHalfLife = 60 * 60 * 24 * 7 * 4 // 1 month
+)
+
+var (
+	// ErrHostNotFound is returned by database operations that fail due to a host
+	// not being found.
+	ErrHostNotFound = errors.New("host not found")
```
n8mgr: This should be exposed in the `hosts` package

persist/postgres/hosts.go
```diff
@@ -297,34 +311,143 @@ WHERE public_key = $18 RETURNING id`,
 	})
 }
 
-func queryHosts(ctx context.Context, tx *txn, offset, limit int) ([]dbHost, error) {
+func queryHost(ctx context.Context, tx *txn, hk types.PublicKey) (hosts.Host, error) {
```
n8mgr: This can be inlined now

persist/postgres/hosts.go
```diff
@@ -297,34 +311,143 @@ WHERE public_key = $18 RETURNING id`,
...
+		return hosts.Host{}, fmt.Errorf("failed to query host addresses: %w", err)
+	}
+	host.Networks, err = queryHostNetworks(ctx, tx, host.id)
+	if err != nil {
+		return hosts.Host{}, fmt.Errorf("failed to query host networks: %w", err)
+	}
+	return host.Host, nil
+}
+
+func queryHosts(ctx context.Context, tx *txn, offset, limit int) ([]hosts.Host, error) {
```
n8mgr: This can be inlined now

persist/postgres/hosts.go
```diff
@@ -82,33 +122,88 @@ WHERE hosts.public_key = $1`, sqlPublicKey(hk)))
...
+		settings_collateral, settings_storage_price, settings_ingress_price,
+		settings_egress_price, settings_free_sector_price, settings_tip_height, settings_valid_until,
+		GREATEST(last_failed_scan, last_successful_scan) != '0001-01-01 00:00:00+00'::timestamptz as scanned,
+		(get_byte(settings_protocol_version, 0) << 16) + (get_byte(settings_protocol_version, 1) << 8) + (get_byte(settings_protocol_version, 2)) as settings_version
+	FROM hosts
+	LEFT JOIN hosts_blocklist hb ON hosts.public_key = hb.public_key
+	LIMIT $1 OFFSET $2
+) SELECT 
+ 	hosts.*,
+	scanned AND recent_uptime > 0.9 AND last_failed_scan < NOW() - INTERVAL '1 week',
```
ChrisSchinnerl: Doesn't that immediately make a host bad upon a single failed scan? I guess without an actual `uptime` this becomes trickier to achieve but the original intention was for a host to have been online for at least 1 week but not for a host that has been around for longer than that to immediately go bad upon a single failed scan. 

I think there are effectively 2 ways around this:

1. add yet another field like `first_successful_scan` to check if that was more than 1 week in the past
2. insert hosts with a pessimistic `recent_uptime` of 0.80 and let it grow to 0.9 over time knowing that this takes about 2 weeks

The first option is simple and straightforward but adds another field. The latter one is quite elegant but we need to get the numbers right which I think you already confirm in your test.

It seems we also need testing to make sure a single scan doesn't ruin a host.

ChrisSchinnerl (changes requested):

persist/postgres/hosts.go
peterjan: edit: Re-reading your comment made me realise I missed `More than 1 week of uptime`. I have always read that as, if the host hasn't been around for a week, it goes bad. I never interpreted that as: "we wait for a full week of uptime before using the host". 

We could indeed default the uptime to a value that's a little lower than the 90% requirement. What numbers are you thinking of?

ChrisSchinnerl: Yeah that's what I was getting at with `2`. Only relying on `recent_uptime`. I'd keep 1 week as the goal for constant uptime before using a host. 
Maybe we can also adjust the decay overall to take at least 1 week to go from 100% uptime to 90%. To be a bit more generous.

peterjan: Ok it roughly comes out at those numbers. Check out `TestHostsRecentUptime`, the assertions there I feel closely match the goal here.

persist/postgres/hosts_test.go
```diff
@@ -194,6 +198,260 @@ func TestHost(t *testing.T) {
...
+
+	// assert host is usable
+	if h, err := db.Host(context.Background(), hk); err != nil {
+		t.Fatal(err)
+	} else if !h.Usability.Usable() {
+		t.Fatal("expected host to be usable")
+	}
+}
+
+func TestHostsRecentUptime(t *testing.T) {
```
ChrisSchinnerl, line 375: Overall this PR looks pretty good now but this test might require a bit more tweaking.

e.g. calling `simulateScan(true, "7 days")` yields quite different results from calling `simulateScan(true, "1 days")` 7 times due to the decay. So to better reflect real-world usage, I think we need to add a loop to `simulateScan` which actually simulates `7` individual scans of 24 hours when passing 7 days.

persist/postgres/hosts_test.go
```diff
@@ -194,6 +198,260 @@ func TestHost(t *testing.T) {
...
+	}
+
+	// assert default uptime
+	if uptime() != .895 {
+		t.Fatal("unexpected", uptime())
+	}
+
+	// assert one week of uptime passes the uptime check
+	simulateScan(true, "7 days")
+	if uptime() < .9 {
```
ChrisSchinnerl, line 428: We also need to check if `uptime() >= 0.9` after 6 days to be sure the host doesn't become good after the first day or so.

ChrisSchinnerl (changes requested):

persist/postgres/hosts_test.go
peterjan: Yeah that's what I originally had actually. I reworked it to more easily play around with the half life in relation to the default uptime etc. I'll revert the test closer to what it was initially.

n8mgr (dismissed):

ChrisSchinnerl (approved):

PR #65, opened 2025-03-12
Broadcast contract resolutions of expired contracts

ChrisSchinnerl:


contracts/chain.go
```diff
@@ -2,19 +2,25 @@ package contracts
...
+	"go.sia.tech/coreutils/wallet"
 	"go.uber.org/zap"
 )
 
 type (
 	// UpdateTx defines what the contract manager needs to atomically process a
 	// chain update in the database.
 	UpdateTx interface {
+		ContractElements() ([]types.V2FileContractElement, error)
+		ContractElementsForBroadcast(maxBlocksSinceExpiry uint64) ([]types.V2FileContractElement, error)
```
n8mgr: This doesn't need to be atomic with the other chain index things. You can move this out of the `UpdateTx` and handle it after the commit. Check out `hostd`'s subscriber as an example for what I mean. There are effectively two stages to each update:

1. Atomically update the database with all the chain changes.
2. Handle everything where atomicity isn't required, like broadcasting revisions.

```go
ContractManager interface {
    // UpdateChainState atomically updates the blockchain state of the
    // contract manager.
    UpdateChainState(tx contracts.UpdateStateTx, reverted []chain.RevertUpdate, applied []chain.ApplyUpdate) error
    // ProcessActions is called after the chain state is updated to
    // trigger additional actions related to contract lifecycle management.
    // This operation should not assume that it will be called for every
    // block. During initial syncing, it will be called at most once per batch.
    ProcessActions(index types.ChainIndex) error
}
```

I found this to be a pretty logical separation and helped reduce some concurrency issues with transaction broadcasts in `hostd`

contracts/chain.go
```diff
@@ -155,8 +167,73 @@ func (m *ContractManager) revertContractDiff(tx *updateTx, diff consensus.V2File
...
 	return nil
 }
+
+func (m *ContractManager) broadcastExpiredContracts(tx UpdateTx) error {
+	expiredFCEs, err := tx.ContractElementsForBroadcast(m.expiredContractBroadcastBuffer)
+	if err != nil {
+		return fmt.Errorf("failed to get expired contracts for broadcast: %w", err)
+	}
+	for _, fce := range expiredFCEs {
+		go func(fce types.V2FileContractElement) {
```
n8mgr: Spawning a goroutine and funding a transaction per expiration is a concurrency nightmare. Could also save some miner fees by batching up to 50 expire resolutions together.

contracts/chain.go
```diff
@@ -155,8 +167,73 @@ func (m *ContractManager) revertContractDiff(tx *updateTx, diff consensus.V2File
...
+					strings.Contains(err.Error(), "not present in the accumulator")) {
+				m.w.ReleaseInputs(nil, []types.V2Transaction{txn})
+				m.log.Debug("failed to broadcast contract expiration txn", zap.Error(err))
+				return
+			} else if err != nil {
+				m.log.Error("failed to broadcast contract expiration txn", zap.Error(err))
+				m.w.ReleaseInputs(nil, []types.V2Transaction{txn})
+				return
+			}
+			m.s.BroadcastV2TransactionSet(basis, []types.V2Transaction{txn})
```
n8mgr: this should probably be the only part in a goroutine

n8mgr (changes requested):

contracts/chain.go
ChrisSchinnerl: It was necessary due to the ongoing transaction but if we separate it out we don't need it anymore. 

The comment about batching is true but I have held off on that since it increases complexity for an edge case that the renter will probably never run into. At this point we have already given the host 144 or so blocks to resolve the contract. So this should only happen if more than 1 host decided to not unlock their funds for 2 contracts which happen to expire at the same height. Which imo is unlikely enough to not handle it for now.

contracts/chain.go
ChrisSchinnerl: Makes sense. I added it with 2 slight differences

1. instead of passing the chain index it uses the stored one
2. it is not called after every batch but only after being completely caught up. Seems to me that this is good enough for renting since all the maintenance we want to perform is best performed when caught up anyway.

subscriber/subscriber.go
```diff
@@ -189,6 +190,11 @@ func (s *Subscriber) Sync() error {
 		}
 	}
 
+	// post-sync actions
+	if err := s.contracts.ProcessActions(); err != nil {
+		s.log.Error("ContractManager.ProcessActions failed", zap.Error(err))
```
n8mgr: nit: use a named log

contracts/chain.go
```diff
@@ -169,6 +187,49 @@ func (m *ContractManager) revertContractDiff(tx *updateTx, diff consensus.V2File
...
+		if err != nil {
+			m.log.Error("failed to fund contract expiration txn", zap.Error(err))
+			continue
+		}
+		m.w.SignV2Inputs(&txn, toSign)
+
+		// verify txn and broadcast it
+		_, err = m.cm.AddV2PoolTransactions(basis, []types.V2Transaction{txn})
+		if err != nil &&
+			(strings.Contains(err.Error(), "has already been resolved") ||
```
peterjan: `nit` I'd move this into `if err != nil {}` to make it more clear you're only checking these to adjust the log level

contracts/chain.go
```diff
@@ -42,6 +44,26 @@ func (tx *updateTx) IsKnownContract(fcid types.FileContractID) (bool, error) {
...
+	ctx, cancel, err := m.tg.AddContext(context.Background())
+	if err != nil {
+		return err
+	}
+	defer cancel()
+
+	// broadcast resolutions for expired contracts
+	// 'expiredContractBroadcastBuffer' blocks after their window end to give
+	// hosts a chance to do it themselves before we do it
+	if err := m.broadcastExpiredContracts(ctx); err != nil {
```
peterjan: check for `[context.Canceled]` here to avoid unnecessarily logging these errors on shutdown

contracts/chain_test.go
```diff
@@ -234,3 +292,45 @@ func TestUpdateContractElementProofs(t *testing.T) {
...
+		V2FileContract: types.V2FileContract{
+			ExpirationHeight: 100,
+		},
+	}
+
+	// broadcast when no contract should be broadcasted
+	if err := contracts.broadcastExpiredContracts(context.Background()); err != nil {
+		t.Fatal(err)
+	}
+	time.Sleep(100 * time.Millisecond)
```
peterjan: Hm, should `broadcastExpiredContracts` have a waitgroup and wait on the goroutines it launches to effectively broadcast the txn set? I'm fine with either but we'd avoid the nasty sleep here, we'd be made aware if `BroadcastV2TransactionSet` on the syncer hangs for some reason (because we would add a sane timeout to the context passed to `broadcastExpiredContracts`). A context on the syncer would've been nice here.

persist/postgres/contracts.go
```diff
@@ -144,6 +144,40 @@ func (s *Store) Contracts(queryOpts ...ContractQueryOpt) ([]contracts.Contract,
 	panic("not implemented")
 }
 
+// ContractElementsForBroadcast returns the contract elements of contracts that
+// have been expired for at least 'maxBlocksSinceExpiry' blocks.
+func (s *Store) ContractElementsForBroadcast(ctx context.Context, maxBlocksSinceExpiry uint64) ([]types.V2FileContractElement, error) {
```
peterjan, line 149: `joke(?):`  `expirationHeightLeeway` 👀

persist/postgres/init.sql
```diff
@@ -83,7 +83,10 @@ CREATE TABLE wallet_siacoin_elements (
 CREATE TABLE global_settings (
     id INTEGER PRIMARY KEY NOT NULL DEFAULT 0 CHECK (id = 0), -- enforce a single row
     db_version INTEGER NOT NULL, -- used for migrations
-    last_scanned_index BYTEA CHECK (LENGTH(last_scanned_index) = 8+32) -- chain index of the last scanned block
+
+    -- chain index of the last scanned block
+    scanned_height BIGINT NOT NULL DEFAULT 0 CHECK(scanned_height >= 0),
+    scanned_block_id BYTEA CHECK (LENGTH(scanned_block_id) = 32)
```
peterjan: We can default to `'\x0000000000000000000000000000000000000000000000000000000000000000'::bytea` and make it `NOT NULL` and avoid `asNullable`

contracts/chain_test.go
peterjan: Definitely leaning towards making `BroadcastV2TransactionSet` blocking.

persist/postgres/contracts.go
ChrisSchinnerl: No more leeways 😅

contracts/chain.go
```diff
@@ -169,6 +188,48 @@ func (m *ContractManager) revertContractDiff(tx *updateTx, diff consensus.V2File
...
+		// fund and sign txn
+		basis, toSign, err := m.w.FundV2Transaction(&txn, txn.MinerFee, true)
+		if err != nil {
+			m.log.Error("failed to fund contract expiration txn", zap.Error(err))
+			continue
+		}
+		m.w.SignV2Inputs(&txn, toSign)
+
+		// verify txn and broadcast it
+		_, err = m.cm.AddV2PoolTransactions(basis, []types.V2Transaction{txn})
```
n8mgr: Need to get the unconfirmed parents before attempting to add it to the pool or broadcasting.

n8mgr (changes requested):

subscriber/subscriber.go
```diff
@@ -26,6 +26,7 @@ type (
 
 	// ContractManager manages contract state.
 	ContractManager interface {
+		ProcessActions() error
```
peterjan: Since it's exposed, I feel it should take a context.

internal/testutils/indexer.go
```diff
@@ -43,6 +43,22 @@ func NewIndexer(t testing.TB, c *ConsensusNode, log *zap.Logger) *Indexer {
...
+	// peers will reject us if our hostname is empty or unspecified, so use loopback
+	s := syncer.New(syncerListener, c.cm, testutil.NewEphemeralPeerStore(), gateway.Header{
+		GenesisID:  c.genesis.ID(),
+		UniqueID:   gateway.GenerateUniqueID(),
+		NetAddress: syncerListener.Addr().String(),
+	},
+		syncer.WithSendBlocksTimeout(2*time.Second),
+		syncer.WithRPCTimeout(2*time.Second),
+	)
+	go s.Run()
```
peterjan, line 60: is it worth changing the syncer one last time to get rid of the `go s.Run()` you have to call on the caller side manually

subscriber/subscriber.go
```diff
@@ -198,6 +199,11 @@ func (s *Subscriber) Sync(ctx context.Context) error {
 		}
 	}
 
+	// post-sync actions
+	if err := s.contracts.ProcessActions(ctx); err != nil {
```
peterjan, line 203: If the context gets cancelled we'll break out of the sync loop and this'll post process the actions and immediately return with `context.Cancelled` which we'll error log... we should probably `return` in the for loop or ignore the error. Before this change we also debug log we're synced right before shutting down apparenlty.

internal/testutils/indexer.go
ChrisSchinnerl: If we ever decide to tag v1.0 I think we could sneak it in, but by itself it's just not worth the time it takes to change it and then fix multiple broken repos.

n8mgr (approved):

PR #66, opened 2025-03-12
Add host blocklist

peterjan:


api/api.go
```diff
@@ -25,6 +25,8 @@ type (
 	Store interface {
 		Host(ctx context.Context, hk types.PublicKey) (hosts.Host, error)
 		Hosts(ctx context.Context, offset, limit int) ([]hosts.Host, error)
+		HostsBlocklist(ctx context.Context, offset, limit int) ([]types.PublicKey, error)
+		UpdateHostsBlocklist(ctx context.Context, add, remove []types.PublicKey, clear bool) error
```
ChrisSchinnerl: I'd get rid of `clear` for now. If we ever need that behavior I'd add a separate `SetHostBlocklist`.
Combining `remove` and `clear` has always been a bit awkward in `renterd`.

99% of the time you just want to click a single host in the UI and block it

persist/postgres/hosts.go
```diff
@@ -50,7 +50,7 @@ func (s *Store) Host(ctx context.Context, hk types.PublicKey) (hosts.Host, error
 	var host hosts.Host
 	if err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
 		var hostID int64
-		err := tx.QueryRow(ctx, `SELECT id, public_key, last_announcement, total_scans, failed_scans, consecutive_failed_scans, last_successful_scan, next_scan FROM hosts WHERE public_key = $1`, sqlPublicKey(hk)).Scan(
+		err := tx.QueryRow(ctx, `SELECT id, public_key, last_announcement, total_scans, failed_scans, consecutive_failed_scans, last_successful_scan, next_scan FROM hosts WHERE public_key = $1 AND NOT EXISTS (SELECT 1 FROM hosts_blocklist WHERE hosts_blocklist.public_key = $1) `, sqlPublicKey(hk)).Scan(
```
ChrisSchinnerl: It should still be possible to fetch a blocked host. It should just be marked as blocked

persist/postgres/hosts.go
```diff
@@ -121,6 +121,37 @@ func (s *Store) Hosts(ctx context.Context, offset, limit int) ([]hosts.Host, err
...
+	var blocklist []types.PublicKey
+	if err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		rows, err := tx.Query(ctx, `SELECT public_key FROM hosts_blocklist LIMIT $1 OFFSET $2`, limit, offset)
+		if err != nil {
+			return fmt.Errorf("failed to query hosts blocklist: %w", err)
+		}
+		defer rows.Close()
+
+		for rows.Next() {
+			var hk sqlPublicKey
```
ChrisSchinnerl: Usually we declare the proper type and cast in the call to `Scan`.

persist/postgres/hosts.go
```diff
@@ -226,8 +257,42 @@ DO UPDATE SET
...
+			}
+		}
+
+		return nil
+	})
+}
+
 func queryHosts(ctx context.Context, tx *txn, offset, limit int) ([]dbHost, error) {
-	rows, err := tx.Query(ctx, `SELECT id, public_key, last_announcement, total_scans, failed_scans, last_successful_scan, next_scan  FROM hosts LIMIT $1 OFFSET $2`, limit, offset)
+	rows, err := tx.Query(ctx, `SELECT id, public_key, last_announcement, total_scans, failed_scans, last_successful_scan, next_scan  FROM hosts WHERE NOT EXISTS (SELECT 1 FROM hosts_blocklist WHERE hosts_blocklist.public_key = hosts.public_key) LIMIT $1 OFFSET $2`, limit, offset)
```
ChrisSchinnerl: I'd still return blocked hosts. Just mark them as blocked. If we end up needing it we can later add filter options

ChrisSchinnerl (changes requested):
Looking good overall but the `Host` and `Hosts` method shouldn't be opinionated about what hosts to query (exclude blocked hosts) and instead the `hosts` type should indicate whether the host is blocked.

api/api.go
peterjan: Hm are you sure? It costs us nothing, it behaves quite intuitively and you'll be wishing you had it if you have to manually remove a large number of hosts from the blocklist. I don't see any reason to add a separate `SetHostBlocklist` since it's just clear+add.

n8mgr: Honestly would separate `Add` and `Remove` and not worry about `Clear` or `Set`. It aligns best with a RESTful API and Clear/Set are pretty niche. More predictable, imo.

```
[GET] /api/hosts/blocklist?limit=50&offset=0
[PUT] /api/hosts/blocklist { "keys": [] }
[DELETE] /api/hosts/blocklist/:key
```

persist/postgres/hosts.go
```diff
@@ -50,7 +50,7 @@ func (s *Store) Host(ctx context.Context, hk types.PublicKey) (hosts.Host, error
 	var host hosts.Host
 	if err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
 		var hostID int64
-		err := tx.QueryRow(ctx, `SELECT id, public_key, last_announcement, total_scans, failed_scans, consecutive_failed_scans, last_successful_scan, next_scan FROM hosts WHERE public_key = $1`, sqlPublicKey(hk)).Scan(
+		err := tx.QueryRow(ctx, `SELECT id, hosts.public_key, last_announcement, total_scans, failed_scans, consecutive_failed_scans, last_successful_scan, next_scan, EXISTS (SELECT 1 FROM hosts_blocklist WHERE public_key = $1) AS blocked FROM hosts WHERE hosts.public_key = $1`, sqlPublicKey(hk)).Scan(
```
n8mgr: Can do this with a LEFT JOIN. Might be more performant than a sub query exists.

```sql
SELECT id, hosts.public_key, last_announcement, total_scans, failed_scans, consecutive_failed_scans, last_successful_scan, next_scan, hb IS NOT NULL AS blocked 
FROM hosts h
LEFT JOIN hosts_blocklist hb ON hb.public_key = h.public_key
WHERE hosts.public_key = $1
```

persist/postgres/hosts.go
```diff
@@ -226,8 +258,39 @@ DO UPDATE SET
 	})
 }
 
+// UpdateHostsBlocklist adds or removes hosts from the blocklist. If clear is
+// true, the blocklist is cleared before adding the given hosts. If a host is
+// both in the add and remove lists, it will be added.
+func (s *Store) UpdateHostsBlocklist(ctx context.Context, add, remove []types.PublicKey, clear bool) error {
```
n8mgr: I don't think it makes sense to combine this logic. Separate `Add` and `Remove` would be best.

persist/postgres/hosts.go
peterjan: I changed it from a `LEFT JOIN` to a subquery exist because I figure it would be faster when you have a large blocklist, and the performance would be roughly equal when either only hosts table is large or both are small. I changed it to use a `LEFT JOIN` instead, when hosts is fully done I'll follow up with a benchmark that queries hosts.

api/api.go
```diff
@@ -25,6 +25,9 @@ type (
 	Store interface {
 		Host(ctx context.Context, hk types.PublicKey) (hosts.Host, error)
 		Hosts(ctx context.Context, offset, limit int) ([]hosts.Host, error)
+		HostsBlocklist(ctx context.Context, offset, limit int) ([]types.PublicKey, error)
+		HostsBlocklistAdd(ctx context.Context, hks []types.PublicKey) error
+		HostsBlocklistRemove(ctx context.Context, hk types.PublicKey) error
```
ChrisSchinnerl: (nit) Ngl I don't love using unnatural sounding names for functions apart from the API handlers just for them to be grouped together. Practically, when coding, the IDE is going to group together all functions if you type "host" which have that word in it.
So it should be fine to name them things like `BlockedHosts`, `BlockHost` and `UnblockHost`.

persist/postgres/hosts.go
```diff
@@ -51,12 +51,13 @@ func (s *Store) Host(ctx context.Context, hk types.PublicKey) (hosts.Host, error
...
 SELECT
-	id, public_key, last_announcement, total_scans, failed_scans, consecutive_failed_scans, last_successful_scan, next_scan,
+	id, hosts.public_key, last_announcement, hb.public_key IS NOT NULL AS blocked,
+	total_scans, failed_scans, consecutive_failed_scans, last_successful_scan, next_scan,
 	settings_protocol_version, settings_release, settings_wallet_address, settings_accepting_contracts,
 	settings_max_collateral, settings_max_contract_duration, settings_remaining_storage, settings_total_storage,
 	settings_contract_price, settings_collateral, settings_storage_price, settings_ingress_price, settings_egress_price,
 	settings_free_sector_price, settings_tip_height, settings_valid_until
-FROM hosts WHERE public_key = $1`, sqlPublicKey(hk)))
+FROM hosts LEFT JOIN hosts_blocklist hb ON hosts.public_key = hb.public_key WHERE hosts.public_key = $1`, sqlPublicKey(hk)))
```
ChrisSchinnerl: (uber nit) move the `WHERE to the next line

persist/postgres/hosts_test.go
```diff
@@ -75,6 +76,83 @@ func TestAddHostAnnouncement(t *testing.T) {
...
+		}
+		hks, err := db.HostsBlocklist(context.Background(), 0, 10)
+		if err != nil {
+			t.Fatal(err)
+		} else if slices.Contains(hks, hk) != blocked {
+			t.Fatal("expected host to be in blocklist")
+		}
+	}
+
+	updateBlocklist := func(add, remove types.PublicKey) {
```
ChrisSchinnerl: I'd split this up to make the test more readable

api/api.go
peterjan: I went back and forth on this myself. I agree for the `Store` but what about the `Client`?  I did it for the client mostly, where we already have things like `WalletSendSiacoins` for instance, the names were OK enough to make it consistent throughout and use the same names in the store.

edit: updated store, kept them in the client...

ChrisSchinnerl: The client can match its API but there is no need for other interfaces to match the client imo.

ChrisSchinnerl (approved):

n8mgr (approved):

PR #69, opened 2025-03-13
Prune expired contracts

ChrisSchinnerl:
This PR adds pruning of expired contracts to the contract manager.

### DIFF
--- a/contracts/chain.go
+++ b/contracts/chain.go
@@ -62,6 +62,13 @@ func (m *ContractManager) ProcessActions() error {
 	if err := m.broadcastExpiredContracts(ctx); err != nil && !errors.Is(err, context.Canceled) {
 		return fmt.Errorf("failed to broadcast expired contracts: %w", err)
 	}
+
+	// prune expired contracts 'expiredContractPruneBuffer' blocks after
+	// we begin broadcasting resolutions
+	if err := m.store.PruneExpiredContractElements(ctx, m.expiredContractBroadcastBuffer+m.expiredContractPruneBuffer); err != nil {
+		return fmt.Errorf("failed to prune expired contracts: %w", err)
+	}
+
 	return nil
 }
 
@@ -91,9 +98,6 @@ func (m *ContractManager) UpdateChainState(tx UpdateTx, reverted []chain.RevertU
 	if err := tx.RejectPendingContracts(maxFormation); err != nil {
 		return fmt.Errorf("failed to reject pending contracts: %w", err)
 	}
-
-	// TODO: prune expired contracts 'expiredContractPruneBuffer' blocks after
-	// we begin broadcasting resolutions
 
 	return nil
 }
--- a/contracts/chain.go
+++ b/contracts/chain.go
@@ -20,7 +20,6 @@ type (
 	UpdateTx interface {
 		ContractElements() ([]types.V2FileContractElement, error)
 		IsKnownContract(contractID types.FileContractID) (bool, error)
-		RejectPendingContracts(maxFormation time.Time) error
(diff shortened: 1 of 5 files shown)


contracts/chain.go
```diff
@@ -55,12 +54,25 @@ func (m *ContractManager) ProcessActions() error {
...
 	// broadcast resolutions for expired contracts
 	// 'expiredContractBroadcastBuffer' blocks after their window end to give
 	// hosts a chance to do it themselves before we do it
 	if err := m.broadcastExpiredContracts(ctx); err != nil {
 		return fmt.Errorf("failed to broadcast expired contracts: %w", err)
 	}
+
+	// prune expired contracts 'expiredContractPruneBuffer' blocks after
+	// we begin broadcasting resolutions
+	if err := m.store.PruneExpiredContractElements(ctx, m.expiredContractBroadcastBuffer+m.expiredContractPruneBuffer); err != nil {
```
peterjan, line 73: should it be `m.broadcastResolutionPruneBuffer` 🤔  it's a little odd `expiredContractPruneBuffer` magically includes the broadcast buffer

persist/postgres/contracts.go
```diff
@@ -219,10 +235,14 @@ func (tx *updateTx) IsKnownContract(contractID types.FileContractID) (bool, erro
...
 
-func (tx *updateTx) RejectPendingContracts(maxFormation time.Time) error {
-	_, err := tx.tx.Exec(tx.ctx, `UPDATE contracts SET state = $1 WHERE state = $2 AND formation < $3`,
-		sqlContractState(contracts.ContractStateRejected), sqlContractState(contracts.ContractStatePending), maxFormation)
-	return err
+// RejectPendingContracts marks all contracts as rejected that are currently
+// pending and have a formation height older than 'maxFormation'.
+func (s *Store) RejectPendingContracts(ctx context.Context, maxFormation time.Time) error {
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `UPDATE contracts SET state = $1 WHERE state = $2 AND formation < $3`,
```
peterjan, line 242: `nit` `<=`

persist/postgres/contracts_test.go
```diff
@@ -86,6 +87,106 @@ func TestContractElementsForBroadcast(t *testing.T) {
...
+		_ = store.UpdateChainState(context.Background(), func(tx subscriber.UpdateTx) error {
+			fces, err := tx.ContractElements()
+			if err != nil {
+				t.Fatal(err)
+			} else if len(fces) != len(contractIDs) {
+				t.Fatalf("expected %d contracts, got %d", len(contractIDs), len(fces))
+			}
+		outer:
+			for _, c := range fces {
+				for _, id := range contractIDs {
```
peterjan: use `slices.Contains`

n8mgr (approved):

persist/postgres/contracts.go
ChrisSchinnerl: It's supposed to reject contracts with a formation height older than the max so I think `<` is appropriate. It's also just a bit odd to use `>=` on timestamps with microsecond accuracy since the equal case will probably never be hit. Almost like checking high precision floats for equality.

contracts/chain.go
ChrisSchinnerl: I'm not sure that makes it any more obvious 😅  It's a buffer for pruning expired contracts. If we ever make that configurable will just add a good docstring to the option function.

peterjan (approved):

n8mgr (approved):

PR #70, opened 2025-03-14
Contract Maintenance Loop

ChrisSchinnerl:
This PR does 2 things

1. it outlines the contract maintenance loop in the contract manager with TODO to then implement in subsequent PRs
2. implements maintenance settings in the store


cmd/indexd/run.go
```diff
@@ -107,6 +107,7 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
 	if err != nil {
 		return fmt.Errorf("failed to create contracts manager: %w", err)
 	}
+	go contracts.Run()
```
n8mgr: nit: I really dislike having to call a separate run function. 9 times out of 10 it gets forgotten. The initializer should spin up the goroutine.

ChrisSchinnerl: Yeah I'm also not a huge fan. I broke it up into a private and public initializer instead where only the public one spins up the goroutine. To make sure we can create a manager without running the background tasks for unit testing.

It's called `newContractManager` to avoid the confusing naming lint error you get when calling it the same as the exposed one.

contracts/manager.go
```diff
@@ -71,6 +101,22 @@ func WithLogger(l *zap.Logger) ContractManagerOpt {
...
 func NewManager(chainManager ChainManager, store Store, syncer Syncer, wallet Wallet, opts ...ContractManagerOpt) (*ContractManager, error) {
+	cm, err := newContractManager(chainManager, store, syncer, wallet, opts...)
+	if err != nil {
+		return nil, err
+	}
+	go cm.run()
+	return cm, nil
+}
+
+// Close closes the contract manager, terminates any background tasks and waits
```
peterjan: `nit` 
```suggestion
// Close terminates any background tasks and waits
```

contracts/manager.go
```diff
@@ -84,16 +130,98 @@ func NewManager(chainManager ChainManager, store Store, syncer Syncer, wallet Wa
...
 	return cm, nil
 }
 
-// Close closes the contract manager, terminates any background tasks and waits
-// for them to exit.
-func (cm *ContractManager) Close() error {
-	cm.tg.Stop()
+// Run starts the contract manager's background tasks.
+func (cm *ContractManager) run() {
+	ctx, cancel, err := cm.tg.AddContext(context.Background())
```
peterjan: I think it's cleaner to add it to the threadgroup at the caller. It's the caller that's calling it in a goroutine so he should handle it properly.

```go
ctx, cancel := tg.AddContext(context.Background())
go func() {
    defer cancel()
    go run(ctx)
}
```

contracts/manager.go
```diff
@@ -84,16 +130,98 @@ func NewManager(chainManager ChainManager, store Store, syncer Syncer, wallet Wa
...
+
+	// block until we are online and the consensus is synced
+	log := cm.log.Named("maintenance")
+	if !cm.blockUntilReady(log) {
+		return // shutdown
+	}
+
+	for {
+		select {
+		case <-cm.tg.Done():
```
peterjan: This could then be `ctx.Done()` which is also slightly cleaner imo.

contracts/manager.go
```diff
@@ -84,16 +130,98 @@ func NewManager(chainManager ChainManager, store Store, syncer Syncer, wallet Wa
...
+			return
+		case <-time.After(cm.maintenanceFrequency):
+		}
+
+		if err := cm.performContractMaintenance(ctx); err != nil {
```
peterjan, line 163: We should consider applying a sane timeout in every iteration of the loop because we're essentially passing it the background context here and this can potentially hang until we close the manager.

persist/postgres/init.sql
```diff
@@ -86,7 +86,13 @@ CREATE TABLE global_settings (
 
     -- chain index of the last scanned block
     scanned_height BIGINT NOT NULL DEFAULT 0 CHECK(scanned_height >= 0),
-    scanned_block_id BYTEA NOT NULL DEFAULT '\x0000000000000000000000000000000000000000000000000000000000000000'::bytea CHECK (LENGTH(scanned_block_id) = 32)
+    scanned_block_id BYTEA NOT NULL DEFAULT '\x0000000000000000000000000000000000000000000000000000000000000000'::bytea CHECK (LENGTH(scanned_block_id) = 32),
+
+    -- contract manager settings
+    contracts_maintenance_enabled BOOLEAN NOT NULL DEFAULT FALSE,
+    contracts_wanted INTEGER NOT NULL DEFAULT 50 CHECK(contracts_wanted > 0), -- number of contracts to maintain
```
peterjan, line 98:
```suggestion
    contracts_wanted INTEGER NOT NULL DEFAULT 50 CHECK(contracts_wanted >= 0), -- number of contracts to maintain
```

contracts/manager.go
ChrisSchinnerl: This will be done on a per-job-basis. I don't want account funding to never happen just because we got stuck on contract maintenance until we ran into the timeout.

persist/postgres/init.sql
ChrisSchinnerl: Imo setting contracts to 0 is illegal. If you don't want contracts, you'd disable the maintenance altogether. If the maintenance isn't disabled, this should be > 0.

peterjan: fair enough

contracts/manager.go
```diff
@@ -84,16 +135,93 @@ func NewManager(chainManager ChainManager, store Store, syncer Syncer, wallet Wa
...
+
+func (cm *ContractManager) blockUntilReady(log *zap.Logger) bool {
+	var once sync.Once
+	for {
+		select {
+		case <-cm.tg.Done():
+			return false
+		case <-time.After(time.Second):
+		}
+		if len(cm.s.Peers()) > 0 && time.Since(cm.cm.TipState().PrevTimestamps[0]) < 3*time.Hour {
```
n8mgr: checking len(peers) is annoying for cluster in `renterd` and I feel the goal is effectively handled by the timestamp check.

persist/postgres/init.sql
```diff
@@ -86,7 +86,13 @@ CREATE TABLE global_settings (
 
     -- chain index of the last scanned block
     scanned_height BIGINT NOT NULL DEFAULT 0 CHECK(scanned_height >= 0),
-    scanned_block_id BYTEA NOT NULL DEFAULT '\x0000000000000000000000000000000000000000000000000000000000000000'::bytea CHECK (LENGTH(scanned_block_id) = 32)
+    scanned_block_id BYTEA NOT NULL DEFAULT '\x0000000000000000000000000000000000000000000000000000000000000000'::bytea CHECK (LENGTH(scanned_block_id) = 32),
+
+    -- contract manager settings
+    contracts_maintenance_enabled BOOLEAN NOT NULL DEFAULT FALSE,
```
n8mgr, line 97: we've had this discussion a few times, but I'm still not convinced disabling maintenance is desirable.

contracts/manager.go
ChrisSchinnerl: Ok I think we can get rid of it here but the scanner should probably still verify we are online before recording a failed scan to avoid a connection dropping to ruin our hosts' uptimes.

persist/postgres/init.sql
ChrisSchinnerl: Yeah but I still think it's good UX to give people a chance to accept the terms at which the software spends their money before actually doing so. Especially when we provide defaults for everything now.

There is also no way to easily stop using the software without that option considering that it will just keep renewing your contracts until you don't have enough funds left for even a single one 😅

ChrisSchinnerl:
This PR blew up a bit cause https://github.com/SiaFoundation/indexd/pull/69/files was accidentally merged into this branch

n8mgr (dismissed):

contracts/chain.go
```diff
@@ -50,12 +49,30 @@ func (tx *updateTx) IsKnownContract(fcid types.FileContractID) (bool, error) {
 // not guaranteed to be called on every update but will eventually be called
 // after applying all batches of a sync.
 func (m *ContractManager) ProcessActions(ctx context.Context) error {
+	ctx, cancel, err := m.tg.AddContext(context.Background())
```
peterjan: you should pass the `ctx` here

contracts/chain_test.go
```diff
@@ -313,26 +339,35 @@ func TestBroadcastExpiredContracts(t *testing.T) {
...
 	}
 
+	// assert asserts the number of txns in the mocked pool, the number of
+	// broadcasted transactions in the mocked syncer, the number of resolutions
+	// in the latest broadcasted transactions and the contract elements in the
+	// store.
+	assert := func(poolTxns, broadcastedTxns, resolutions, pruneCalls, rejectCalls int) {
+		t.Helper()
+		if len(cmMock.V2PoolTransactions()) != poolTxns {
+			t.Fatalf("expected 0 contract in tpool, got %v", len(cmMock.tpool))
```
peterjan: you're passing in the number of expected elements, shouldn't fix it at 0 (x3)

peterjan (approved):

PR #72, opened 2025-03-16
Subscriber fixes

peterjan:
I was updating the subscriber at one point and then reverted it again, I want to keep these small changes though. Using `AddContext` makes it consistent with the host manager and it's just cleaner to pass a context to `Sync`

### DIFF
--- a/subscriber/subscriber.go
+++ b/subscriber/subscriber.go
@@ -115,12 +115,15 @@ func New(cm ChainManager, hm HostManager, contracts ContractManager, wm WalletMa
 	}
 	go func() {
 		defer cancel()
-		for range reorgCh {
+		for {
 			select {
 			case <-reorgCh:
-				if err := s.Sync(ctx); err != nil && !errors.Is(err, context.Canceled) {
+				err := s.Sync(ctx)
+				if err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, threadgroup.ErrClosed) {
 					s.log.Panic("failed to sync database", zap.Error(err))
 				}
+			case <-ctx.Done():
+				return
 			}
 		}
 	}()
@@ -133,6 +136,12 @@ func New(cm ChainManager, hm HostManager, contracts ContractManager, wm WalletMa
 // to manually call this since the Subscriber will do that itself but it can be
 // used to guarantee the subscriber is synced at a given point in time.
 func (s *Subscriber) Sync(ctx context.Context) error {
+	ctx, cancel, err := s.tg.AddContext(ctx)
+	if err != nil {
+		return err
+	}
+	defer cancel()
+
 	s.syncMu.Lock()
 	defer s.syncMu.Unlock()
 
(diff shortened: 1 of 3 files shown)


subscriber/subscriber.go
```diff
@@ -108,35 +109,34 @@ func New(cm ChainManager, hm HostManager, contracts ContractManager, wm WalletMa
...
-	done, err := s.tg.Add()
+	ctx, cancel, err := s.tg.AddContext(context.Background())
 	if err != nil {
-		panic(err)
+		return nil, err
 	}
 	go func() {
-		defer done()
+		defer cancel()
 		for range reorgCh {
```
n8mgr: This loop is wrong, right? It requires a double send on `reorgCh`. Not really relevant, but I'd be fine fixing it while we're in here. If not it can be a follow-up.

Should be
```go
for {
    select {
    case <-reorgCh:
        // sync
    case <-ctx.Done():
        return
    }
}

File: subscriber/subscriber.go [hunk 2]
```diff
@@ -157,8 +157,8 @@ func (s *Subscriber) Sync() error {
 			break
 		}
 
-		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
-		if err := s.store.UpdateChainState(ctx, func(tx UpdateTx) error {
+		updateCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+		if err := s.store.UpdateChainState(updateCtx, func(tx UpdateTx) error {
```
n8mgr: nit: can this be moved out of the `if err :=`? Multi-line if statements are a personal pet peeve.

```go
updateCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
err := s.store.UpdateChainState(updateCtx, func(tx UpdateTx) error { })
cancel()
if err != nil {
    
}
```

subscriber/subscriber.go
```diff
@@ -173,7 +173,7 @@ func (s *Subscriber) Sync() error {
 				index = rus[len(rus)-1].State.Index
 			}
 
-			if err := tx.UpdateLastScannedIndex(ctx, aus[len(aus)-1].State.Index); err != nil {
+			if err := tx.UpdateLastScannedIndex(updateCtx, aus[len(aus)-1].State.Index); err != nil {
```
n8mgr: Should use `index` here

subscriber/subscriber.go
```diff
@@ -108,35 +109,34 @@ func New(cm ChainManager, hm HostManager, contracts ContractManager, wm WalletMa
...
 	reorgCh <- struct{}{} // trigger initial sync
-	return s
+	return s, nil
 }
 
 // Sync syncs the subscriber with the chain manager. It's usually not necessary
 // to manually call this since the Subscriber will do that itself but it can be
 // used to guarantee the subscriber is synced at a given point in time.
-func (s *Subscriber) Sync() error {
+func (s *Subscriber) Sync(ctx context.Context) error {
```
n8mgr, line 138: Should probably add `s.tg.AddContext` here since we do want to wait for external calls to `Sync` to return before returning from `Close`.

subscriber/subscriber.go
peterjan: I have a hard time with error checks not immediately following `err :=`

subscriber/subscriber.go
peterjan: jesus

n8mgr (approved):

ChrisSchinnerl (approved):

PR #73, opened 2025-03-17
Implement contract formations

ChrisSchinnerl:
This PR adds contract formations to the contract maintenance loop


contracts/form_test.go
```diff
@@ -0,0 +1,244 @@
...
+		return proto.HostSettings{}, nil
+	}
+	s.store.UpdateHostSettings(hk, settings)
+	return settings, nil
+}
+
+// TestPerformContractFormationWithoutContracts tests the
+// performContractFormation method assuming that we don't have any contracts
+// yet.
+func TestPerformContractFormationWithoutContracts(t *testing.T) {
```
ChrisSchinnerl, line 91: I'll f/u with another test which tests the edge cases around existing contracts. But this PR seems to be getting slightly out of hand already.

contracts/contract.go
```diff
@@ -23,6 +23,37 @@ var (
 	ErrNotFound = errors.New("contract not found")
 )
 
+type (
+	// ContractQueryOpts contains the options for querying contracts.
+	ContractQueryOpts struct {
```
n8mgr: This shouldn't need to be exposed since we're going with the option pattern

contracts/manager.go
```diff
@@ -7,10 +7,20 @@ import (
...
 	"go.sia.tech/coreutils/syncer"
 	"go.sia.tech/coreutils/threadgroup"
+	"go.sia.tech/indexd/hosts"
 	"go.uber.org/zap"
+	"lukechampine.com/frand"
+)
+
+const (
+	dialTimeout         = 10 * time.Second
+	minRemainingStorage = 10e9 / uint64(proto.SectorSize) // 10GB
```
peterjan: we're pretty inconsistent with GB vs GiB, e.g. in the gouging checks I purposefully use TB and not TiB, `proto.SectorSize` is 4MiB so we should probably make a rule and try to be as consistent as possible

persist/postgres/contracts.go
```diff
@@ -126,7 +91,7 @@ func (s *Store) Contract(ctx context.Context, contractID types.FileContractID) (
 	var contract contracts.Contract
 	if err := s.transaction(ctx, func(ctx context.Context, tx *txn) (err error) {
 		contract, err = scanContract(tx.QueryRow(ctx, `
-SELECT c.contract_id, c.formation, h.public_key, c.proof_height, c.expiration_height, c_from.contract_id, c_to.contract_id, c.state, c.capacity, c.size, c.contract_price, c.initial_allowance, c.miner_fee, c.good, c.append_sector_spending, c.free_sector_spending, c.fund_account_spending, c.sector_roots_spending
+SELECT c.contract_id, c.formation, h.public_key, c.proof_height, c.expiration_height, c_from.contract_id, c_to.contract_id, c.state, c.capacity, c.size, c.contract_price, c.initial_allowance, c.miner_fee, c.good, c.append_sector_spending, c.free_sector_spending, c.fund_account_spending, c.sector_roots_spending, c.total_collateral
```
peterjan: `nit` could we select it right after `miner_fee` and update the scan helper, it's neat to maintain field order

contracts/manager.go
```diff
@@ -115,16 +146,21 @@ func NewManager(chainManager ChainManager, store Store, syncer Syncer, wallet Wa
...
 
-func newContractManager(chainManager ChainManager, store Store, syncer Syncer, wallet Wallet, opts ...ContractManagerOpt) *ContractManager {
+func newContractManager(renterKey types.PublicKey, chainManager ChainManager, contractor Contractor, scanner Scanner, store Store, syncer Syncer, wallet Wallet, opts ...ContractManagerOpt) *ContractManager {
 	cm := &ContractManager{
 		cm: chainManager,
 		s:  syncer,
 		w:  wallet,
 
-		store: store,
+		cf:        contractor,
```
peterjan: it might make sense to abstract broadcasting a transaction and limit the amount of dependencies we tack on the manager. It also makes it clear what dependencies are needed for what functionality. Having a broadcaster would allow us to remove the syncer and the wallet.

contracts/manager.go
```diff
@@ -115,16 +146,21 @@ func NewManager(chainManager ChainManager, store Store, syncer Syncer, wallet Wa
 	return cm, nil
 }
 
-func newContractManager(chainManager ChainManager, store Store, syncer Syncer, wallet Wallet, opts ...ContractManagerOpt) *ContractManager {
+func newContractManager(renterKey types.PublicKey, chainManager ChainManager, contractor Contractor, scanner Scanner, store Store, syncer Syncer, wallet Wallet, opts ...ContractManagerOpt) *ContractManager {
```
peterjan: We might as well call it `HostsManager` instead of calling it a scanner. To me the contract manager should be the `contractor`. The manager should expose those methods so it becomes a more useful component.

peterjan: Should we consider having a `TransactionPool` that 's made up of the chain manager, the syncer and the wallet and that exposes a simple `BroadcastTransaction`. We're always doing the same thing and we're injecting 3 things to accomplish one goal.

contracts/form.go
```diff
@@ -0,0 +1,241 @@
...
+	proto "go.sia.tech/core/rhp/v4"
+	"go.sia.tech/core/types"
+	"go.sia.tech/coreutils/chain"
+	"go.sia.tech/coreutils/rhp/v4"
+	"go.sia.tech/coreutils/rhp/v4/siamux"
+	"go.sia.tech/indexd/hosts"
+	"go.uber.org/zap"
+)
+
+const sectorsPerGB = uint64(1<<30) / proto.SectorSize
```
peterjan: `sectorsPerGiB`

contracts/form.go
```diff
@@ -0,0 +1,241 @@
...
+		return rhp.RPCFormContractResult{}, fmt.Errorf("failed to form contract: %w", err)
+	}
+
+	return res, nil
+}
+
+// performContractFormation makes sure that we have at least 'wanted' good
+// contracts with good hosts in unique CIDRs.
+func (cm *ContractManager) performContractFormation(ctx context.Context, period uint64, wanted uint, log *zap.Logger) error {
+	formationLog := log.Named("formation")
```
peterjan, line 92: I don't mind named loggers but we should analyse the output of contract maintenance and then really scrutinise the nesting, the metadata and the log messages. The `renterd` logging is very hard to get through.

contracts/form.go
```diff
@@ -0,0 +1,241 @@
...
+	activeContracts, err := cm.store.Contracts(ctx, WithRevisable(true))
+	if err != nil {
+		return fmt.Errorf("failed to fetch active contracts: %w", err)
+	}
+
+	// helpers for CIDR check
+	usedCidrs := make(map[string]struct{})
```
peterjan: Might map it to the host key. It could be useful to log what hosts have overlapping CIDrs.

contracts/form.go
```diff
@@ -0,0 +1,241 @@
...
+		for _, cidr := range host.Networks {
+			if _, known := usedCidrs[cidr.IP.String()]; known {
+				return true
+			}
+		}
+		return false
+	}
+
+	// helper to check if a host is good to form a contract with
+	checkHost := func(host hosts.Host, log *zap.Logger) bool {
```
peterjan: `nit` goodForFormation

contracts/form.go
```diff
@@ -0,0 +1,241 @@
...
+		// host checks
+		host, err := cm.store.Host(ctx, contract.HostKey)
+		if err != nil {
+			contractLog.Error("failed to fetch host for contract", zap.Error(err))
+			continue
+		} else if !checkHost(host, contractLog) {
+			continue
+		}
+
+		// contract checks
```
peterjan: Does it make sense to move this up? We potentially have a lot of host checks and logs to then find the contract is no good. Perhaps we should filter the active contracts by the contract checks, and only perform the host checks on contracts that match all criteria.

contracts/form.go
```diff
@@ -0,0 +1,241 @@
...
+			contractLog.Debug("ignore contract since the host won't put more collateral into it", zap.Stringer("maxCollateral", host.Settings.MaxCollateral), zap.Stringer("usedCollateral", contract.UsedCollateral))
+			continue
+		}
+
+		// contract is good
+		addHost(host)
+	}
+
+	// fetch all hosts
+	hosts, err := cm.store.Hosts(ctx, 0, -1)
```
peterjan: This is definitely a little odd. This might be a large number of hosts, I feel we should limit this somehow or perhaps even expose a method to get `n` random hosts from the store. That would also get rid of `shuffle` and make your tests deterministic because we could mock that method on the store.

contracts/form.go
```diff
@@ -0,0 +1,241 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to fetch hosts to form contracts with: %w", err)
+	}
+
+	// randomize their order to avoid preferring any host
+	cm.shuffle(len(hosts), func(i, j int) { hosts[i], hosts[j] = hosts[j], hosts[i] })
+
+	for i := range hosts {
+		formationLog := formationLog.With(zap.Stringer("hostKey", hosts[i].PublicKey))
```
peterjan: This is fine but at the same time it's confusing and it invites bugs imo. We already had this exact bug in a pretty important spot in the renterd codebase. Perhaps rename it to `hostLog`? 

On the topic of loggers, we keep passing them around but we could also define some logger constants and create the log at the beginning of the method. At the very least it would be immediately obvious what loggers we have and potentially highlight when the named logger is redundant or unused.

```go
const logMaintenance = "maintenance"

func (m Manager) Foo() {
   log := m.log.named(logMaintenance)
}

func (m Manager) Bar() {
   log := m.log.named(logMaintenance)
}
```

contracts/form.go
```diff
@@ -0,0 +1,241 @@
...
+
+		// scan host before forming a contract and fetch it from the database to
```
peterjan: I know I said this is how I thought it was going to look, but seeing it now it doesn't make a lot of sense for the contract manager to have to deal with this stuff. I think the scanner should be called the host manager and either the host manager should expose a method to get a scanned host, or `ScanHost` should return the `Host` and not the settings. I didn't do that initially because the spec said the host manager should expose a scan host and I'm reusing that method in the loop, but this `scanCtx` here and not using the host settings but fetching the host directly from the store afterwards is pretty gnarly and is making me reconsider.

contracts/form.go
```diff
@@ -0,0 +1,241 @@
...
+		hostAddr := host.SiamuxAddr()
+
+		allowance, collateral := initialContractFunding(host.Settings.Prices, period)
+		formationCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+		res, err := cm.cf.FormContract(formationCtx, host.PublicKey, hostAddr, host.Settings, proto.RPCFormContractParams{
+			RenterPublicKey: cm.renterKey,
+			RenterAddress:   cm.w.Address(),
+			Allowance:       allowance,
+			Collateral:      collateral,
+			ProofHeight:     cm.cm.TipState().Index.Height + period,
```
peterjan, line 208: `cm.cm` is not great

contracts/manager.go
peterjan: I kind of get the feeling that `ContractFormer` is going to be refactored into a `Contractor` that does everything?

contracts/manager.go
ChrisSchinnerl: Not sure I follow. You want to merge the `syncer` and the `wallet` into a `broadcaster`? 
In theory that works but it makes it more annoying to pass in our types since they no longer implement the interfaces. e.g. the `Synced` won't be a `Broadcaster`. So we end up with quite a bit more code than just the one extra argument we need to pass.

contracts/manager.go
ChrisSchinnerl: Yeah it will end up with form, renew and refresh to avoid creating 3 interfaces which are quite similar.

ChrisSchinnerl: I'm not sure the added code to glue the `SingleAddressWallet`, `ChainManager` and `Synced` together warrants a slightly cleaner function signature.

contracts/form.go
ChrisSchinnerl: The logging during the test seems pretty sane to me. But we can obviously take some time when we start running it to revise our logging strategy.

contracts/form.go
ChrisSchinnerl: Changed it to `isGood`. Should be better than `checkHost` but I don't think we need the additional context since we are already in `performContractFormation`.

contracts/form.go
ChrisSchinnerl: I think checking the host first makes sense. If a host is bad, the contract is always bad. Eventually we might also want to cache these checks since one host can potentially have many contracts. We also can't perform all contract checks without fetching the host.

contracts/form.go
ChrisSchinnerl: I thought the same when implementing the maintenance in renterd but it has been doing fine fetching all hosts for a long time now. It works and we avoid a nasty shuffle in the database code.

contracts/form.go
ChrisSchinnerl: I don't think it's an issue here since it's shadowed with `:=` and in `renterd` it was `=` but I changed it to `hostLog`

contracts/form.go
ChrisSchinnerl: It's not too bad but we could refactor it in a f/u

contracts/form.go
ChrisSchinnerl: No, but also not that big of a deal if it let's us keep consistent naming 🤷

contracts/manager.go
ChrisSchinnerl: I changed those to GiB and TiB since we perform them on contract directly. Overall we can't be consistent I think. e.g. `protocol.SectorSize` can never be `4MB` but the pinned settings should be in `TB` since they are user-facing. 
On the other hand the raw values we use for gouging checks should be `TiB` and so on again since we directly compare them to the host's settings. 

Overall a good rule is probably user-facing == TB and internal == TiB

contracts/form.go
peterjan: It's just a little `ORDER BY RANDOM()` but up to you

peterjan (dismissed):

contracts/form.go
n8mgr: I agree with PJ that it's better to be efficient about it now and it shouldn't be that bad.

n8mgr (approved):

PR #75, opened 2025-03-18
Make timestamp columns nullable

peterjan:
Using `0001-01-01 00:00:00+00` as `NULL` time to avoid nullable columns in our database has a little annoying quirk where scanning those null times into a `time.Time` returns `0001-01-01 00:17:30 +0017` (for me). This has to do with LMT time and historical time zone rules since that date. This means that `IsZero()` currently returns false, which we definitely don't want.

We have a couple of options, I explored one or two but landed on making timestamps nullable in the database. An alternative approach would be to use a more recent timestamp as `NULL` value, like `1970-01-01 00:00:00` and have a custom sql type for Time that uses the scanner interface to convert `time.Time{}` to our `NULL` value. Having nullable columns has it drawbacks, but it's also more natural than comparing things to `0001-01-01 00:00:00+00` in queries.


ChrisSchinnerl (commented):
I think relying on NULL makes sense. Not sure why we ended up aggressively avoiding it with default values but NULL is the perfect sentinel value for something that just hasn't happened yet.

peterjan:
> I think relying on NULL makes sense. Not sure why we ended up aggressively avoiding it with default values but NULL is the perfect sentinel value for something that just hasn't happened yet.

Aggressively avoiding `NULL` values is my doing. It's a little annoying to have `NULL` values in the settings because I have to either `COALESCE` or have a weird extra bool check. With the columns as they are now though, having both `last_failed_scan` and `last_successful_scan` that's less of a nuisance.

ChrisSchinnerl (approved):

n8mgr (approved):

PR #76, opened 2025-03-19
Add explorer API

peterjan:
This PR adds the explorer API in preparation of price pinning. 

Small notes:
- `TestExplorerAPI` doesn't include a test to check for "explorer disabled", verified it manually, wanted to avoid overcomplicating the indexer test config
- config has `Enabled` instead of `Disable`, I feel this is more consistent with the config and it avoids `if !cfg.Explorer.disable`

### DIFF
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -1,7 +1,6 @@
 package explorer
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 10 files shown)


explorer/explorer.go
```diff
@@ -0,0 +1,81 @@
...
+	err = makeRequest(ctx, http.MethodGet, fmt.Sprintf("%s/exchange-rate/siacoin/%s", e.url, currency), nil, &rate)
+	return
+}
+
+func drainAndClose(r io.ReadCloser) {
+	io.Copy(io.Discard, io.LimitReader(r, 1024*1024))
+	r.Close()
+}
+
+func makeRequest(ctx context.Context, method, url string, requestBody, response any) error {
```
ChrisSchinnerl: `make` sounds a bit odd to me. Like we create something but not execute it.

How about calling this `request`?

ChrisSchinnerl (dismissed):

explorer/explorer.go
```diff
@@ -0,0 +1,81 @@
...
+	err = request(ctx, http.MethodGet, fmt.Sprintf("%s/exchange-rate/siacoin/%s", e.url, currency), nil, &rate)
+	return
+}
+
+func drainAndClose(r io.ReadCloser) {
+	io.Copy(io.Discard, io.LimitReader(r, 1024*1024))
+	r.Close()
+}
+
+func request(ctx context.Context, method, url string, requestBody, response any) error {
```
n8mgr: nit: this can be inlined. It's only called in one location and we have no intention of adding more explorer functionality any time soon.

n8mgr (dismissed):

ChrisSchinnerl (approved):

PR #78, opened 2025-03-20
Add pin manager

### LINKED ISSUE #49: Price pinning
Gouging settings should be pinnable to fiat currencies

peterjan:
This PR adds a pin manager. I'm going to F/U with settings endpoints.

Closes https://github.com/SiaFoundation/indexd/issues/49


persist/postgres/settings.go
```diff
@@ -4,8 +4,56 @@ import (
 	"context"
 
 	"go.sia.tech/core/types"
+	"go.sia.tech/indexd/settings"
 )
 
+// PinnedSettings returns the pinned settings.
+func (s *Store) PinnedSettings(ctx context.Context) (ps settings.PinnedSettings, err error) {
+	err = s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		query := `SELECT TRIM(pinned_currency), pinned_min_collateral, pinned_max_storage_price, pinned_max_ingress_price, pinned_max_egress_price FROM global_settings`
```
n8mgr: Really no need for TRIM in the query. Trim it before it’s added to the database.

peterjan: The default value of '' is three spaces because it's a fixed size `CHAR[3]`. I changed it to a `VARCHAR[3]` which shouldn't have that behaviour, I tried that but it still did and then I decided to just `TRIM` it. I'll double check that tomorrow though because perhaps I didn't test it properly. I only found out because it showed up in a test.

peterjan: There's a slight I updated it in the doc instead of the schema :D

settings/manager.go
ChrisSchinnerl, line 1: My preference would be for this to be a `PriceManager` rather than a generic `SettingsManager`. A `SettingsManager` feels like a global `types` package which is not very intuitive to use.

e.g. I'd expect to be able to set contract maintenance related settings on the `ContractManager`, scanning related settings on the `HostManager` and so on. Effectively putting the settings where they are consumed.
The fact that it's all stored in the same table doesn't need to be reflected in the structure of the managers.

settings/manager.go
```diff
@@ -0,0 +1,320 @@
...
+		tg:  threadgroup.New(),
+		log: zap.NewNop(),
+	}
+	for _, opt := range opts {
+		opt(sm)
+	}
+
+	if sm.updatePriceFrequency > sm.rateWindow {
+		return nil, errors.New("price update frequency exceeds rate window")
+	}
```
ChrisSchinnerl: I think you might need to check if `updatePriceFrequency` is 0 here to avoid a division by 0.

settings/manager.go
```diff
@@ -0,0 +1,320 @@
...
+	// update average
+	sm.average = avg
+	return shouldUpdate
+}
+
+// convertCurrencyToSC converts a value in an external currency and an exchange
+// rate to Siacoins.
+func convertCurrencyToSC(target decimal.Decimal, rate decimal.Decimal) (types.Currency, error) {
+	if rate.IsZero() {
+		return types.Currency{}, nil
```
ChrisSchinnerl: This should definitely error out imo. If the rate is 0, something is wrong and we probably don't want to update anything.

settings/manager.go
```diff
@@ -0,0 +1,320 @@
...
+			return fmt.Errorf("failed to convert MaxStoragePrice price, %w", err)
+		}
+		prices.MaxStoragePrice = value.Div64(oneTB)
+	}
+	if pins.MinCollateral.Enabled() {
+		value, err := convertCurrencyToSC(decimal.NewFromFloat(float64(pins.MinCollateral)), decimal.NewFromFloat(rate))
+		if err != nil {
+			return fmt.Errorf("failed to convert MinCollateral price, %w", err)
+		}
+		prices.MinCollateral = value
```
ChrisSchinnerl: This is missing the conversion from TB to byte.

Looking at the checks in the db it seems they are incorrect as well. `MinCollateral` should be checked against `Collateral`. Not `MaxCollateral`.

settings/manager_test.go
```diff
@@ -0,0 +1,261 @@
...
+	"lukechampine.com/frand"
+)
+
+var (
+	// testPriceSettings is a set of random prices used in testing
+	testPriceSettings = PriceSettings{
+		MaxEgressPrice:  types.NewCurrency64(frand.Uint64n(1e6)),
+		MaxIngressPrice: types.NewCurrency64(frand.Uint64n(1e6)),
+		MaxStoragePrice: types.NewCurrency64(frand.Uint64n(1e6)),
+		MinCollateral:   types.NewCurrency64(frand.Uint64n(1e6)),
```
ChrisSchinnerl: In this case random values hurt more than they help imo.
Instead I'd set these to realistic real-world values.

settings/manager_test.go
```diff
@@ -0,0 +1,261 @@
...
+	time.Sleep(time.Second)
+
+	// check that the price have been updated
+	prices, _ = s.PriceSettings(context.Background())
+	if err := checkPrices(prices, pins, 1.2); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestConvertConvertCurrencyToSC(t *testing.T) {
```
ChrisSchinnerl: Double convert

ChrisSchinnerl (changes requested):

hosts/host.go
```diff
@@ -50,6 +50,16 @@ type (
 		EgressPrice     bool `json:"egressPrice"`
 		FreeSectorPrice bool `json:"freeSectorPrice"`
 	}
+
+	// UsabilitySettings contains the settings that are used to check if a host
+	// is usable.
+	UsabilitySettings struct {
```
peterjan, line 56: avoiding `hosts.Settings` or `HostSettings` here

edit: I guess `MaintenanceSettings` are an option to be consistent with contracts but kind of prefer this

persist/postgres/hosts_test.go
```diff
@@ -724,6 +733,42 @@ func TestUpdateHost(t *testing.T) {
 	}
 }
 
+func TestUpdateUsabilitySettings(t *testing.T) {
```
ChrisSchinnerl, line 735: I'd either move this to `settings_test.go` or `TestPinnedSettings` to `contracts_test.go`.

persist/postgres/hosts_test.go
```diff
@@ -218,28 +219,36 @@ func TestHostChecks(t *testing.T) {
...
+	if err := db.UpdateUsabilitySettings(context.Background(), hosts.UsabilitySettings{
+		MaxEgressPrice:     settingMaxEgressPrice,
+		MaxIngressPrice:    settingMaxIngressPrice,
+		MaxStoragePrice:    settingMaxStoragePrice,
+		MinCollateral:      settingMinCollataral,
+		MinProtocolVersion: [3]uint8{1, 0, 0},
+	}); err != nil {
+		t.Fatal(err)
+	}
+	if err := db.UpdateMaintenanceSettings(context.Background(), contracts.MaintenanceSettings{
```
ChrisSchinnerl, line 246: I think this one is missing a test still. The usability and pin settings already got one.

persist/postgres/hosts.go
```diff
@@ -160,15 +160,15 @@ WITH globals AS (
  	hosts.*,
 	recent_uptime > 0.9,
 	has_settings AND settings_max_contract_duration >= globals.contracts_period,
-	has_settings AND settings_max_collateral > globals.min_collateral AND settings_max_collateral >= settings_collateral * globals.one_tb * globals.contracts_period,
-	has_settings AND settings_version >= globals.min_version,
+	has_settings AND settings_max_collateral > globals.hosts_min_collateral AND settings_max_collateral >= settings_collateral * globals.one_tb * globals.contracts_period,
```
ChrisSchinnerl:
```suggestion
	has_settings AND settings_max_collateral >= settings_collateral * globals.one_tb * globals.contracts_period,
```

persist/postgres/hosts.go
```diff
@@ -86,15 +86,15 @@ WITH globals AS (
 	hosts.*,
 	recent_uptime > 0.9,
 	has_settings AND settings_max_contract_duration >= globals.contracts_period,
-	has_settings AND settings_max_collateral > globals.min_collateral AND settings_max_collateral >= settings_collateral * globals.one_tb * globals.contracts_period,
-	has_settings AND settings_version >= globals.min_version,
+	has_settings AND settings_max_collateral > globals.hosts_min_collateral AND settings_max_collateral >= settings_collateral * globals.one_tb * globals.contracts_period,
```
ChrisSchinnerl:
```suggestion
	has_settings settings_max_collateral >= settings_collateral * globals.one_tb * globals.contracts_period,
```

ChrisSchinnerl (approved):

ChrisSchinnerl (changes requested):
Little leftover in the query

persist/postgres/hosts_test.go
peterjan: I added that in https://github.com/SiaFoundation/indexd/pull/80, this PR didn't need the `MaintenanceSettings` counterpart yet. That F/U PR has both fetch and update and adds the test.

persist/postgres/hosts_test.go
peterjan: Wouldn't you rather keep it consistent? `UsabilitySettings` is used by hosts manager, `PinnedSettings` is used by the pins manager. There is no settings manager. I considered housing pinned settings somewhere but pinned settings are not necessarily limited to host settings, any price setting can be pinned.

ChrisSchinnerl: Yeah that's why I commented but I guess the actual problem here is `settings.go` being called `settings.go` rather than `pins.go`. Which makes it seems inconsistent.t

peterjan: Oh yeah but I realised that already and fixed it in https://github.com/SiaFoundation/indexd/pull/80/files as well 😁

persist/postgres/hosts_test.go
```diff
@@ -218,28 +219,36 @@ func TestHostChecks(t *testing.T) {
 		} else if f.Kind() != reflect.Bool {
 			t.Fatalf("field '%s' is not a boolean", check)
 		} else if !f.Bool() {
+			t.Log(h.Usability)
```
n8mgr: left over?

n8mgr (dismissed):

persist/postgres/hosts_test.go
peterjan: dammit

n8mgr (dismissed):

ChrisSchinnerl (dismissed):

ChrisSchinnerl (approved):

PR #79, opened 2025-03-20
Add offline check to host scans

### LINKED ISSUE #77: Don't punish hosts upon failing a scan when the indexer is offline
To do so we should do the following after a failed scan to determine whether we are online.

1. check if len(syncer.Peers()) > 0
2. if that doesn't work ping 3 well-known websites until the first one responds successfully (cloudflare, google, amazon) as a fallback

If this happens we need to interrupt the scan and wait until we are online again before scheduling the next batch

peterjan:
This is not really what the issue describes, but it just works ™. I don't see why the syncer peers check needs be included in the online check. It also complicates the dependency injection part of it. If we're offline I stop scanning too.

Fixes https://github.com/SiaFoundation/indexd/issues/77

### DIFF
--- a/hosts/manager.go
+++ b/hosts/manager.go
@@ -13,6 +13,7 @@ import (
 	"go.sia.tech/coreutils/chain"
 	"go.sia.tech/coreutils/rhp/v4/quic"
 	"go.sia.tech/coreutils/rhp/v4/siamux"
+	"go.sia.tech/coreutils/syncer"
 	"go.sia.tech/coreutils/threadgroup"
 	"go.uber.org/zap"
 	"lukechampine.com/frand"
@@ -46,20 +47,20 @@ type (
 		scanFrequency      time.Duration
 		scanInterval       time.Duration
 
-		pinger   Pinger
-		resolver Resolver
-		scanner  Scanner
-		store    Store
+		onlineChecker OnlineChecker
+		resolver      Resolver
+		scanner       Scanner
+		store         Store
 
 		tg  *threadgroup.ThreadGroup
 		log *zap.Logger
 	}
 
-	// Pinger defines an interface to check whether the indexer is online. It's
+	// OnlineChecker defines an interface to check whether the indexer is online. It's
 	// used to ensure hosts aren't punished for failing a scan if the indexer is
 	// offline.
-	Pinger interface {
-		Online() bool
+	OnlineChecker interface {
+		IsOnline() bool
 	}
 
 	// Resolver defines an interface to resolve hostnames.
@@ -81,6 +82,11 @@ type (
 		UpdateHost(ctx context.Context, hk types.PublicKey, networks []net.IPNet, hs proto4.HostSettings, scanSucceeded bool, nextScan time.Time) error
 	}
 
+	// Syncer defines an interface that exposes the Peers method.
+	Syncer interface {
+		Peers() []*syncer.Peer
+	}
+
 	// UpdateTx defines what the host manager needs to atomically process a
 	// chain update in the database.
 	UpdateTx interface {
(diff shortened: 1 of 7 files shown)


hosts/scanner.go
```diff
@@ -22,3 +26,26 @@ func (c *scanner) Settings(ctx context.Context, hk types.PublicKey, addr string)
 
 	return rhp.RPCSettings(ctx, t)
 }
+
+var fallbackSites = []string{
+	"https://1.1.1.1", // cloudflare
```
n8mgr: Don’t think we need the extra overhead of tls/https. Just call net.Dial with a very short timeout

hosts/scanner.go
```diff
@@ -22,3 +26,26 @@ func (c *scanner) Settings(ctx context.Context, hk types.PublicKey, addr string)
...
 	return rhp.RPCSettings(ctx, t)
 }
+
+var fallbackSites = []string{
+	"1.1.1.1:80", // Cloudflare
+	"www.google.com:80",
+	"www.amazon.com:80",
+}
+
+type pinger struct{}
```
ChrisSchinnerl: Could test that this actually works by running a test against a `httptest` server where the first site fails and the next one is the local server which doesn't fail.

hosts/manager.go
```diff
@@ -347,6 +364,9 @@ func resolveHost(ctx context.Context, resolver Resolver, addresses []chain.NetAd
 		if errors.Is(err, context.Canceled) {
 			return nil, nil, err
 		} else if err != nil {
+			if !pinger.Online() {
```
ChrisSchinnerl: The point of also checking the peers we are already connected to is to provide a primarily decentralized solution to the issue that works instantly since it's an in-memory lookup.

So you don't end up potentially pinging Google 1000 times in a row in case you have a bad batch of hosts and instead are immediately able to tell that a host is bad.

hosts/manager.go
```diff
@@ -333,7 +350,7 @@ func calculateNextScanTime(lastScan time.Time, success bool, consecScanFailures
 // and/or private addresses, and a list of networks in CIDR notation. The only
 // error this function returns is [context.Canceled], other errors that occur
 // during the resolving and parsing are debug logged but otherwise ignored.
-func resolveHost(ctx context.Context, resolver Resolver, addresses []chain.NetAddress, log *zap.Logger) ([]chain.NetAddress, []net.IPNet, error) {
+func resolveHost(ctx context.Context, pinger Pinger, resolver Resolver, addresses []chain.NetAddress, log *zap.Logger) ([]chain.NetAddress, []net.IPNet, error) {
```
ChrisSchinnerl: I'd put this into `ScanHost` instead doing something like 
```
if !success && !pinger.Online {
  return proto4.HostSettings{}, errNodeOffline
}
```

Resolving a host can work just because your OS caches the DNS lookup. If you do it in `ScanHost` it is guaranteed to cover all host interactions.

hosts/scanner.go
```diff
@@ -22,3 +26,26 @@ func (c *scanner) Settings(ctx context.Context, hk types.PublicKey, addr string)
 
 	return rhp.RPCSettings(ctx, t)
 }
+
+var fallbackSites = []string{
+	"1.1.1.1:80", // Cloudflare
+	"www.google.com:80",
+	"www.amazon.com:80",
+}
```
ChrisSchinnerl: I'd dial `443` with `net.Dial`. Afaik it doesn't perform a SSL handshake so we don't get the overhead but we at least ping something that's not a deprecated port returning permanent redirects.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (dismissed):

n8mgr (dismissed):

ChrisSchinnerl (approved):

PR #81, opened 2025-03-24
Add Accounts API

peterjan:
waiting on 
- https://github.com/SiaFoundation/indexd/pull/96
- https://github.com/SiaFoundation/indexd/pull/97


persist/postgres/accounts.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// HostAccountsForFunding returns up to limit accounts for the given host key
+// that are due for funding. This method also ensures that there is a host
+// account for the given host key and every account in the database.
+func (s *Store) HostAccountsForFunding(ctx context.Context, hk types.PublicKey, limit int) ([]accounts.HostAccount, error) {
+	var accs []accounts.HostAccount
+	if err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		if _, err := tx.Exec(ctx, `
+INSERT INTO account_hosts (account_id, host_id)
```
peterjan: I'm going back and forth on this one. I think it's probably a good idea to ensure we don't leave out any accounts that need funding. So this might be a good spot to ensure they exist. This 'll essentially create them on the fly when needed, which is also nice. It goes against accessors making changes to the state a bit but maybe this is a good exception case. 

What's a little silly is that the account manager almost becomes obsolete. The only responsibility it has currently is to decide on what the next fund time should be upon success/failure. We might as well give that to the contract manager. Not sure it makes sense to have an account manager at all if it's the contract manager that's instigating everything and calling the shots.

peterjan: I went back. The `limit` got me in the end. I also moved everything around so most of the accounts logic is in the account manager. The contract manager only knows about hosts and contracts, that used to be different.

persist/postgres/accounts.go
```diff
@@ -0,0 +1,114 @@
...
+		} else if res.RowsAffected() == 0 {
+			return accounts.ErrExists
+		}
+		return nil
+	})
+}
+
+// EnsureAccounts ensures that there is a host account for the given host key
+// and every account in the database.
+func (s *Store) EnsureAccounts(ctx context.Context, hk types.PublicKey) error {
```
ChrisSchinnerl: This shouldn't be necessary. `AccountsForFunding` can theoretically always return missing accounts. As soon as we need to rely on something being called in certain key spots or periodically, things will become inconsistent.

persist/postgres/accounts.go
```diff
@@ -0,0 +1,114 @@
...
+func (s *Store) AccountsForFunding(ctx context.Context, hk types.PublicKey, limit int) ([]accounts.Account, error) {
+	var accs []accounts.Account
+	if err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		rows, err := tx.Query(ctx, `
+SELECT a.public_key, h.public_key, ea.consecutive_failed_funds, ea.next_fund
+FROM account_hosts ea
+INNER JOIN hosts h ON h.public_key = $1
+INNER JOIN accounts a ON a.public_key = ea.account_id
+WHERE next_fund <= NOW() ORDER BY next_fund ASC`)
+		if err != nil {
```
ChrisSchinnerl: Limit isn't used

persist/postgres/accounts.go
```diff
@@ -0,0 +1,114 @@
...
+		return rows.Err()
+	}); err != nil {
+		return nil, err
+	}
+
+	return accs, nil
+}
+
+// CreateAccount inserts a new account in the database with given account key.
+func (s *Store) CreateAccount(ctx context.Context, ak types.PublicKey) error {
```
ChrisSchinnerl: Would call this `AddAccount` to be consistent with other methods where we add stuff like host announcements and contracts.

persist/postgres/accounts.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+		return rows.Err()
+	}); err != nil {
+		return nil, err
+	}
+	return accs, nil
+}
+
+// UpdateAccounts updates the given host accounts in the database.
+func (s *Store) UpdateAccounts(ctx context.Context, accounts []accounts.Account) error {
```
ChrisSchinnerl: `UpdateAccounts` and `AccountsForFunding` are probably going be likely candidates for slow queries over large tables so I'd add a benchmark for each of them.

These should be reasonably fast even if we have a million accounts registered over 1000 hosts.

contracts/manager.go
```diff
@@ -230,6 +250,73 @@ func (cm *ContractManager) blockUntilReady(log *zap.Logger) bool {
...
+		case sema <- struct{}{}:
+		}
+
+		wg.Add(1)
+		go func(i int, hk types.PublicKey, contracts []types.FileContractID, log *zap.Logger) {
+			defer func() {
+				<-sema
+				wg.Done()
+			}()
+
```
ChrisSchinnerl: Might want to give each host 1 minute for funding to prevent this from blocking forever.

accounts/manager.go
```diff
@@ -0,0 +1,165 @@
...
+			return totalUsage, fmt.Errorf("failed to update accounts: %w", err)
+		}
+	}
+
+	return totalUsage, nil
+}
+
+// FundingSignalChan returns a channel that delivers a tick every time a new
+// account is registered or account funding is due.
+func (m *AccountManager) FundingSignalChan() <-chan struct{} {
```
ChrisSchinnerl: This isn't nice. The caller is in charge of determining the interval at which call this. I think as a start it should be fine to not make any exceptions for account funding and just perform a full maintenance every 20 minutes or so.

accounts/manager.go
```diff
@@ -0,0 +1,165 @@
...
+	"time"
+
+	proto "go.sia.tech/core/rhp/v4"
+	"go.sia.tech/core/types"
+	"go.sia.tech/indexd/hosts"
+	"go.uber.org/zap"
+)
+
+const (
+	accountFundBatch          = 1e3
```
ChrisSchinnerl: Unless it's too much for the database I'd make the batch size the max size we allow in coreutils

accounts/manager.go
```diff
@@ -0,0 +1,165 @@
...
+	accountFundInterval       = time.Hour
+	accountExpBackoffMaxHours = 128
```
ChrisSchinnerl: Should be minutes according to the spec. Because we want to retry sooner than the 1 hour retry interval if we failed to fund the account to make up for it.

ChrisSchinnerl (changes requested):
This PR does way more than it should. I'd say just the persistence code will probably make for a 400-500 lines PR all by itself with testing and benchmarks.

Then we can review the account funder separately and finally think about what we want to expose in the API.

persist/postgres/accounts.go
peterjan: I had that originally but I moved it out because of the `limit` parameter. It's definitely a trade off but figured it might be fine because it's the account manager that's responsible. I don't moving it back though, was on the fence and flip flopped twice already.

edit: getting out of dev standup Chris, we do want to split this up right? If we do ever get 1M accounts, we're inserting 1M records here for every batch...

accounts/manager.go
peterjan: I was expecting a comment on this, I don't think it's too bad though? We want the contract maintenance loop and the account funding to not necessarily happen at the same frequency. The most important thing though is that we want to trigger account funding when a new account is added, so there must be a signal somewhere.

contracts/manager.go
peterjan: I apply a pessimistic 5 minute timeout in the accounts manager. Guess it's better at the caller site though, will update.

cmd/indexd/run.go
```diff
@@ -105,8 +106,12 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
 	}
 	defer hm.Close()
 
+	funder := accounts.NewFunder(cm, contracts.NewSigner(walletKey), types.Siacoins(1))
```
ChrisSchinnerl:
```suggestion
	funder := accounts.NewFunder(cm, wm, types.Siacoins(1))
```

internal/testutils/indexer.go
```diff
@@ -71,8 +72,11 @@ func NewIndexer(t testing.TB, c *ConsensusNode, log *zap.Logger) *Indexer {
 		t.Fatalf("failed to create host manager: %v", err)
 	}
 
+	funder := accounts.NewFunder(c.cm, contracts.NewSigner(walletKey), types.Siacoins(1))
```
ChrisSchinnerl:
```suggestion
	funder := accounts.NewFunder(c.cm, wm, types.Siacoins(1))
```

contracts/signer.go
ChrisSchinnerl, line 1: Shouldn't need this wrapper since `SingleAddressWallet` implements `SignHash`.

ChrisSchinnerl (changes requested):

api/api.go
```diff
@@ -90,6 +101,10 @@ func NewServer(chain ChainManager, syncer Syncer, wallet Wallet, store Store, op
 	return jape.Mux(map[string]jape.Handler{
 		"GET /state": a.handleGETState,
 
+		// accounts endpoints
+		"GET  /accounts":            a.handleGETAccounts,
+		"POST /account/:accountkey": a.handlePOSTAccount,
```
peterjan: add a `DELETE`?

ChrisSchinnerl: Yeah why not. Probably won't need it before we add a UI but it should be a tiny addition

ChrisSchinnerl (approved):

n8mgr (approved):

PR #83, opened 2025-03-26
Block bad hosts and contracts

ChrisSchinnerl:
This PR implements blocking bad hosts and their contracts as part of the contract maintenance.

Considering that it is now a lot harder for a host to become bad, we decided that marking them bad should be something that is not done automatically.

As a result, this PR uses the same mechanism to block a host that would usually be used by a user to block a manually host. This has the nice side-effect of these hosts showing up in a potential view within the UI where users can then manually unblock them if their usability changed to "good". Rather than doing so automatically.

Thanks to the way we perform host checks we can also easily filter hosts for blocked hosts which are usable. To indicate to users that there are blocked hosts which are good again without the use of a churn alert.


peterjan:
I'm sure you talked this over with Nate already but to me it feels a bit odd we're using the blocklist in this fashion. Especially since we allow the user to unblock a host manually, we would be setting those contracts to `good=TRUE`. More importantly though, you mention that hosts showing up on the blocklist if they become unusable is a good thing. I would be pretty confused if I block host x and y, and all of a sudden host z shows up in that list. Being able to unblock a host that became usable from the UI is now a hard task, because you're essentially asking the user to remember every host he ever blocked, for whatever reason that may be, because if he would go off of the usability, he might mistakingly unblock a host he previously blocked for a totally different reason. Could you elaborate a bit on why you both felt this approach had nothing but upsides?

ChrisSchinnerl:
> I'm sure you talked this over with Nate already but to me it feels a bit odd we're using the blocklist in this fashion. Especially since we allow the user to unblock a host manually, we would be setting those contracts to `good=TRUE`. More importantly though, you mention that hosts showing up on the blocklist if they become unusable is a good thing. I would be pretty confused if I block host x and y, and all of a sudden host z shows up in that list. Being able to unblock a host that became usable from the UI is now a hard task, because you're essentially asking the user to remember every host he ever blocked, for whatever reason that may be, because if he would go off of the usability, he might mistakingly unblock a host he previously blocked for a totally different reason. Could you elaborate a bit on why you both felt this approach had nothing but upsides?

Yeah we talked about this. Hosts are not expected to be blocked a lot anymore so there shouldn't be too many hosts on that list. 
The idea being that eventually we add a "reason" that you can filter by which makes it easy to navigate the list and maybe a whitelist at some point that prevents hosts from being marked as bad in the first place. To give users full control.

Since this PR is still pretty short I could add the reason right away which should clear things up. Do you agree?

peterjan:
> > I'm sure you talked this over with Nate already but to me it feels a bit odd we're using the blocklist in this fashion. Especially since we allow the user to unblock a host manually, we would be setting those contracts to `good=TRUE`. More importantly though, you mention that hosts showing up on the blocklist if they become unusable is a good thing. I would be pretty confused if I block host x and y, and all of a sudden host z shows up in that list. Being able to unblock a host that became usable from the UI is now a hard task, because you're essentially asking the user to remember every host he ever blocked, for whatever reason that may be, because if he would go off of the usability, he might mistakingly unblock a host he previously blocked for a totally different reason. Could you elaborate a bit on why you both felt this approach had nothing but upsides?
> 
> Yeah we talked about this. Hosts are not expected to be blocked a lot anymore so there shouldn't be too many hosts on that list. The idea being that eventually we add a "reason" that you can filter by which makes it easy to navigate the list and maybe a whitelist at some point that prevents hosts from being marked as bad in the first place. To give users full control.
> 
> Since this PR is still pretty short I could add the reason right away which should clear things up. Do you agree?

I still think it's a little odd to interweave the concept of a blocklist with host/contract usability. I'm ready to be proven wrong on that though! We never got it right in `renterd` either. Adding a reason is a reasonable fix for the UX issue.

The PR description mentions that hosts shouldn't be blocked automatically, even though the maintenance loop does that. Is that a leftover? Don't mind either way, just asking. That leaves the setting contracts to `good=TRUE` part when we unblock them. Doesn't it make sense to perform host checks at that point and only flip the ones for which the host is still usable? Technically the PR looks OK, just wanted to ask this last question.

ChrisSchinnerl:
> > > I'm sure you talked this over with Nate already but to me it feels a bit odd we're using the blocklist in this fashion. Especially since we allow the user to unblock a host manually, we would be setting those contracts to `good=TRUE`. More importantly though, you mention that hosts showing up on the blocklist if they become unusable is a good thing. I would be pretty confused if I block host x and y, and all of a sudden host z shows up in that list. Being able to unblock a host that became usable from the UI is now a hard task, because you're essentially asking the user to remember every host he ever blocked, for whatever reason that may be, because if he would go off of the usability, he might mistakingly unblock a host he previously blocked for a totally different reason. Could you elaborate a bit on why you both felt this approach had nothing but upsides?
> > 
> > 
> > Yeah we talked about this. Hosts are not expected to be blocked a lot anymore so there shouldn't be too many hosts on that list. The idea being that eventually we add a "reason" that you can filter by which makes it easy to navigate the list and maybe a whitelist at some point that prevents hosts from being marked as bad in the first place. To give users full control.
> > Since this PR is still pretty short I could add the reason right away which should clear things up. Do you agree?
> 
> I still think it's a little odd to interweave the concept of a blocklist with host/contract usability. I'm ready to be proven wrong on that though! We never got it right in `renterd` either. Adding a reason is a reasonable fix for the UX issue.
> 
> The PR description mentions that hosts shouldn't be blocked automatically, even though the maintenance loop does that. Is that a leftover? Don't mind either way, just asking. That leaves the setting contracts to `good=TRUE` part when we unblock them. Doesn't it make sense to perform host checks at that point and only flip the ones for which the host is still usable? Technically the PR looks OK, just wanted to ask this last question.

We do block them automatically. We just don't unblock them automatically. That's what the user has to do manually. Which is also the reason for marking contracts as good again. The only reason for a contract to be bad apart from not renewing in time is a bad host. So it's fine to mark a contract as good when a user explicitly wants to unblock a host. Worst thing that can happen is that it gets marked bad again on the next maintenance.

contracts/manager.go
```diff
@@ -230,6 +234,45 @@ func (cm *ContractManager) blockUntilReady(log *zap.Logger) bool {
...
+		return fmt.Errorf("failed to fetch active contracts: %w", err)
+	}
+
+	log := cm.log.Named("blockhosts")
+
+	for _, c := range contracts {
+		contractLog := log.With(zap.Stringer("contractID", c.ID), zap.Stringer("hostKey", c.HostKey))
+
+		// fetch the host for the contract
+		host, err := cm.store.Host(ctx, c.HostKey)
```
n8mgr: Would be better to introduce a query to specifically grab a bad but not already blocked host.

peterjan (dismissed):

api/endpoints.go
```diff
@@ -119,11 +119,14 @@ func (a *api) handleGETHostsBlocklist(jc jape.Context) {
 }
 
 func (a *api) handlePUTHostsBlocklist(jc jape.Context) {
-	var hks []types.PublicKey
-	if jc.Decode(&hks) != nil {
+	var hosts HostsBlocklistRequest
+	if jc.Decode(&hosts) != nil {
+		return
+	} else if len(hosts.Reason) > 50 {
```
n8mgr: Pretty arbitrary number of characters. If a user wants to leave out the reason, I think that's fine. Alternatively, consider just `hosts.Reason == ""`

n8mgr (changes requested):
>

api/endpoints.go
ChrisSchinnerl: I thought 50 is generous but still reasonably alright to visualize in a UI. But yeah we can remove the check altogether.

contracts/manager.go
ChrisSchinnerl: We could do that but keep in mind that we are also only interested in hosts that we have active contracts with.
So we'd need to fetch hosts that are bad, not blocked and have a contract with a proof height below the current block height.

Regardless of what we settle on, I started a PR to add the query options to the `Hosts` since that is something we need anyway. https://github.com/SiaFoundation/indexd/pull/94

n8mgr: I think being more specific here is best, particularly since we might have more than one contract per host now. Best to filter in the db over grabbing a list of things then filtering in Go whenever possible.

ChrisSchinnerl: Ok added all the filters in the mentioned PR. Up to you whether you want to approve this and have me use that filter in the other PR or finish the other PR first and then rebase this one.

persist/postgres/hosts.go
```diff
@@ -272,26 +272,51 @@ func (s *Store) BlockedHosts(ctx context.Context, offset, limit int) ([]types.Pu
 	return blocklist, nil
 }
 
-// BlockHosts adds the given host keys to the blocklist.
-func (s *Store) BlockHosts(ctx context.Context, hks []types.PublicKey) error {
+// BlockHosts adds the given host keys to the blocklist and marks all of its
+// contracts as bad.
+// If a host is already on the blocklist, the reason remains unchanged to
+// preserve the original reason for blocking.
+func (s *Store) BlockHosts(ctx context.Context, hks []types.PublicKey, reason string) error {
```
peterjan, line 279: I wouldn't allow an empty string

api/endpoints.go
```diff
@@ -143,11 +143,11 @@ func (a *api) handleGETHostsBlocklist(jc jape.Context) {
...
 
 func (a *api) handlePUTHostsBlocklist(jc jape.Context) {
-	var hks []types.PublicKey
-	if jc.Decode(&hks) != nil {
+	var hosts HostsBlocklistRequest
+	if jc.Decode(&hosts) != nil {
 		return
 	}
-	jc.Check("failed to add host keys to blocklist", a.store.BlockHosts(jc.Request.Context(), hks))
+	jc.Check("failed to add host keys to blocklist", a.store.BlockHosts(jc.Request.Context(), hosts.HostKeys, hosts.Reason))
```
peterjan, line 150: check `hosts.Reason` and 400 if it's empty?

persist/postgres/hosts.go
ChrisSchinnerl: I think it's fine not to provide one and Nate didn't mind either https://github.com/SiaFoundation/indexd/pull/83#discussion_r2022118759

peterjan: Ok. I think it'll be annoying because we'll end up using it in the UI. But fine to allow empty.

peterjan (approved):

n8mgr (approved):

PR #87, opened 2025-03-27
Contract Renewals

ChrisSchinnerl:


contracts/renew.go
```diff
@@ -0,0 +1,105 @@
...
+
+		res, err := cm.contractor.RenewContract(ctx, contract.HostKey, host.SiamuxAddr(), host.Settings, contract.ID, cm.cm.TipState().Index.Height+renewWindow)
+		if err != nil {
+			contractLog.Debug("failed to renew", zap.Error(err))
+			continue
+		}
+		renewed := res.Contract
+		minerFee := res.RenewalSet.Transactions[len(res.RenewalSet.Transactions)-1].MinerFee
+
+		err = cm.store.AddRenewedContract(ctx, contract.ID, renewed.ID, renewed.Revision.ProofHeight, renewed.Revision.ExpirationHeight, host.Settings.Prices.ContractPrice, renewed.Revision.RenterOutput.Value, minerFee, renewed.Revision.TotalCollateral)
```
n8mgr, line 97: nit: almost out of scope for this PR, but large number of params increases the chance of bugs -- especially with easily swappable things like proof and expiration height next to each other. Would be nicer to pass a struct in with the necessary fields.

n8mgr (approved):

contracts/renew.go
ChrisSchinnerl: Addressed in #89

contracts/renew_test.go
```diff
@@ -0,0 +1,179 @@
...
+	// first one is good with a good contract and a bad one
+	good := goodHost(1)
+	scanner.settings[good.PublicKey] = goodSettings
+	formContract(types.FileContractID{1}, good.PublicKey, true)  // will renew
+	formContract(types.FileContractID{2}, good.PublicKey, false) // won't renew
+
+	// second one is bad since it's not accepting contracts with a good contract
+	bad := goodHost(2)
+	bad.Usability.AcceptingContracts = false
+	scanner.settings[bad.PublicKey] = goodSettings
```
peterjan, line 108: I'm sure it's OK but this was surprising?

peterjan (approved):

contracts/renew_test.go
ChrisSchinnerl: Yeah I guess it's a bit odd. In the mock the scanner results don't affect the usability. Which is something we could probably refine a bit. I'll see about that in a f/u.

PR #88, opened 2025-03-27
Accounts Persistence

peterjan:


persist/postgres/bench_test.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+// BenchmarkAccountsForFunding is a benchmark to ensure the performance of
+// AccountsForFunding, which ensures a host account exists for the given host
+// and every account in the database and then goes on to return all accounts
+// with a next scan in the past.
+//
+// M1 Max | 1_000 accounts | 17.46 ms/op
+// M1 Max | 10_000 accounts | 106.27 ms/op
+func BenchmarkAccountsForFunding(b *testing.B) {
```
peterjan: Performance isn't great here. Especially with the `limit` parameter I feel we should ensure hosts outside of fetching accounts for funding. I tested it and it's ~70% faster. Of course you'd pay the price for the inserts somewhere.

When I read the spec I was wondering whether we could get away without having a join table between our `hosts` and `accounts` because it's bound to give issues at some point.

persist/postgres/init.sql
```diff
@@ -28,6 +33,15 @@ CREATE TABLE hosts (
 );
 CREATE INDEX idx_hosts_next_scan_idx ON hosts(next_scan);
 
+CREATE TABLE account_hosts (
+    account_id INTEGER NOT NULL REFERENCES accounts(id) ON DELETE CASCADE,
```
peterjan, line 37: I'm wondering if we should reference the `account_key` and host `public_key` directly. If we make a combined primary key we could avoid the join on both tables when updating these accounts. I'm not sure though how much performance gain we would get from that. Do you feel that's worth trying out and benchmarking?

persist/postgres/accounts.go
```diff
@@ -0,0 +1,120 @@
...
+func (s *Store) AccountsForFunding(ctx context.Context, hk types.PublicKey, limit int) ([]accounts.Account, error) {
+	if limit < 0 {
+		return nil, errors.New("limit can not be negative")
+	} else if limit == 0 {
+		return nil, nil
+	}
+
+	var accs []accounts.Account
+	if err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `
```
n8mgr: You don't need to create the account link here:
```sql
SELECT a.public_key,
       COALESCE(ea.consecutive_failed_funds, 0),
       COALESCE(ea.next_fund, NOW())
FROM accounts a
LEFT JOIN LATERAL (
  SELECT *
  FROM account_hosts ah
  JOIN hosts h ON h.id = ah.host_id
  WHERE ah.account_id = a.id
    AND h.public_key=$1
) ea ON true
WHERE ea.id IS NULL OR ea.next_fund <= NOW()
ORDER BY ea.next_fund ASC NULLS FIRST
LIMIT $2;
```

You will need to change the `UPDATE` query later to `INSERT ON CONFLICT DO UPDATE`. Gets you significantly better performance in the hot path since it's only doing an insert against the batch rather than every account in the database. When scanning it's safe to populate `HostKey: hk` rather than scanning it from the db when it is guaranteed to match.

n8mgr (changes requested):

accounts/account.go
```diff
@@ -0,0 +1,25 @@
...
+
+var (
+	// ErrExists is returned by database operations that fail due to an account
+	// already existing.
+	ErrExists = errors.New("account already exists")
+)
+
+type (
+	// Account represents an ephemeral account on a host.
+	Account struct {
```
ChrisSchinnerl: I think we might get into confusing naming territory if we call this `Account`. Down the road we are going to use accounts for authentication, we are going to back up private keys and slabs will be scoped to accounts.

This type mostly refers to the ephemeral account on a specific host. So maybe `HostAccount` would be a better name here.

persist/postgres/bench_test.go
ChrisSchinnerl: The `accounts` table and join table reflect very different things. One of them is a user/app account which will be used for other purposes down the road. The join table is effectively a list of ephemeral accounts. I don't think we can easily merge them.

That a readonly operation is faster than one that includes insertions isn't surprising. But I don't think it needs to include insertions at all. Nate's example reflects what I had in mind when I commented on the ensure method on the other PR.

persist/postgres/bench_test.go
```diff
@@ -0,0 +1,123 @@
...
+	var accs []accounts.Account
+	rows, err := store.pool.Query(context.Background(), `
+SELECT a.public_key, h.public_key, ea.consecutive_failed_funds, ea.next_fund
+FROM account_hosts ea
+INNER JOIN hosts h ON h.id = ea.host_id
+INNER JOIN accounts a ON a.id = ea.account_id`)
+	if err != nil {
+		b.Fatal(err)
+	}
+	defer rows.Close()
```
ChrisSchinnerl: Not sure about everything `Close` does on rows under the hood but I'd still close them before the timer is reset and the actual benchmark starts.

persist/postgres/bench_test.go
```diff
@@ -0,0 +1,123 @@
...
+		if err != nil {
+			b.Fatal(err)
+		}
+	}
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
+		hk := types.GeneratePrivateKey().PublicKey()
+		_, err := store.pool.Exec(context.Background(), `INSERT INTO hosts (public_key, last_announcement) VALUES ($1, NOW());`, sqlPublicKey(hk))
```
ChrisSchinnerl: `AccountsForFunding` is primarily a readonly operation so I'd keep the number of hosts static. That way we benchmark how many times we can fetch account for funding on a steady host and user base rather than having an increasing host base.

Meaning I'd create 1000 hosts before the loop and maybe register 100,000 user accounts also before the loop.
Then I'd make sure the join table is populated. Finally run `AccountsForFunding` in a loop with 10k accounts per iteration. The host we use could just be `hosts[i % 1000]` or something to not repeat the same query over and over.

accounts/account.go
peterjan: I flip-flopped between `Account` and `HostAccount`. I landed on `Account` to keep it short and sweet but I'll revert it. 

Edit: I remember why I reverted back to `Account`. It's because I felt `HostAccount` would need to be reflected in the database table, but I definitely wanted `account_hosts` because I think the opposite makes less sense, also because we have `account_slabs`, so I just called it `Account` instead. For the record, I would still prefer calling it `Account` and live with the confusing name territory between Account and the public keys in the accounts table.

persist/postgres/accounts.go
peterjan: I didn't consider inserting a batch on update! Thanks!

persist/postgres/bench_test.go
peterjan: I didn't think to insert the batch on update and select the host accounts virtually. 

Yeah I know the accounts and ephemeral accounts are very different, I just felt that since we no longer track the balance, we **might** get away with not tracking anything at all and just do a best effort when funding hosts and keep a `next_fund` on the account itself. Obviously that's pretty out there, I think in practice we might get away with that though 🤷‍♂️ .

persist/postgres/accounts.go
n8mgr: Have not been able to get rid of the SEQ SCAN on the accounts table. Without some sort of filter we end up scanning every account row. Definitely not scalable. Tempting to turn this query into a materialized view that we refresh periodically in the background. Still not ideal.

peterjan: I took a stab at it as well. I updated the performance of `UpdateHostAccounts` quite a lot by doing a batch update. I think it's fine because we have use a batch size of 1000 which won't change any time soon. Unfortunately it's not fast enough to allow us to insert every ephemeral account while doing benchmarks.

I updated `HostAccountsForFunding` and split the query into fetching `NULL` and fetching ephemeral accounts separately. The downside of splitting up the query is that benchmarking it is quite annoying because you'd have to insert all of the ephemeral accounts aside from half a batch size to ensure the accounts cover the entire query. That's not an option so I resorted to passing in a parameter, definitely not ideal but at least it allows us to benchmark both parts of the query.

I was thinking this might be a really good fit for redis (using `SETEX` and/or `TTL`s), similar to a materialized view but performance would likely be better. Definitely not ideal though :/ not sure if we can get out of going the materialized view route. The benchmark will always be a little meaningless if we can't populate the entire join table in a reasonable amount of time.

persist/postgres/bench_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+	// run benchmark for different number of accounts
+	for _, numAccounts := range []int{10_000, 100_000, 1_000_000} {
+		var once sync.Once
+		var hosts []types.PublicKey
+
+		b.Run(fmt.Sprintf("%d_accounts", numAccounts), func(b *testing.B) {
+			// prepare the database once, this ensures that we don't have to
+			// recreate the setup for different values of b.N
+			once.Do(func() {
```
ChrisSchinnerl: You shouldn't need `once` if you use `b.Run` since anything outside of `b.Run` should only run once. The parent isn't measured as soon as you call `b.Run`.

persist/postgres/bench_test.go
```diff
@@ -0,0 +1,214 @@
...
+				// fetch accounts without account_host entry
+				accounts, err := store.HostAccountsForFunding(context.Background(), hosts[i%numHosts], batchSize, true)
+				if err != nil {
+					b.Fatal(err)
+				} else if len(accounts) != batchSize {
+					b.Fatalf("expected %d accounts, got %d", batchSize, len(accounts))
+				}
+
+				// fetch accounts with account_host entry
+				accounts, err = store.HostAccountsForFunding(context.Background(), hosts[i%numHosts], batchSize, false)
```
ChrisSchinnerl: This seems odd. You are benchmarking 3 queries now instead of 2 right? First call runs both and the second one only the second query.

persist/postgres/accounts.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+		return nil
+	})
+}
+
+// HostAccountsForFunding returns up to limit accounts for the given host key
+// that are due for funding. If allowNew is true, it will first return accounts
+// for which there are currently no account host entries, this is currently only
+// used in benchmarking and should always be true in production.
+func (s *Store) HostAccountsForFunding(ctx context.Context, hk types.PublicKey, limit int, allowNew bool) ([]accounts.HostAccount, error) {
```
ChrisSchinnerl: Nicer way would be splitting this up into 2 functions each executing a single query and this method passing the same db txn to both of them.
Then you can benchmark them individually a lot more cleanly.

n8mgr: Agree on separating these into two funcs and also dropping the `allowNew` param. A param on a published method that is always supposed to be true in prod is an odd design.

persist/postgres/accounts_test.go
```diff
@@ -0,0 +1,238 @@
...
+		t.Fatal(err)
+	}
+
+	// add an account
+	ak := types.GeneratePrivateKey().PublicKey()
+	if err := store.AddAccount(context.Background(), ak); err != nil {
+		t.Fatal(err)
+	}
+
+	// fetch accounts for funding to ensure host accounts are created
```
ChrisSchinnerl: outdated comment

persist/postgres/accounts.go
```diff
@@ -0,0 +1,200 @@
...
+		a.id AS account_id,
+		h.id AS host_id,
+		vals.consecutive_failed_funds,
+		vals.next_fund
+	FROM (VALUES %s) AS vals(account_pubkey, host_pubkey, consecutive_failed_funds, next_fund)
+	JOIN accounts a ON a.public_key = vals.account_pubkey
+	JOIN hosts h ON h.public_key = vals.host_pubkey
+)
+INSERT INTO account_hosts (account_id, host_id, consecutive_failed_funds, next_fund)
+SELECT account_id, host_id, consecutive_failed_funds, next_fund FROM cte
```
ChrisSchinnerl: I don't think we need the `cte` here? This could be a `FROM ( Select...)`.

persist/postgres/accounts.go
```diff
@@ -0,0 +1,200 @@
...
+
+	return accs, nil
+}
+
+// UpdateHostAccounts updates the given host accounts in the database.
+func (s *Store) UpdateHostAccounts(ctx context.Context, accounts []accounts.HostAccount) error {
+	if len(accounts) == 0 {
+		return nil
+	} else if len(accounts) > 1000 {
+		panic("too many accounts to update") // developer error
```
ChrisSchinnerl: No harm in returning an error here.

persist/postgres/accounts.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		vals := make([]string, 0, len(accounts))
+		args := make([]any, 0, len(accounts)*4)
+		for i, account := range accounts {
```
ChrisSchinnerl: Given our experience with batching and indices within batches I have to ask. Was the speedup significant? Because if not this really invites errors.

persist/postgres/bench_test.go
```diff
@@ -0,0 +1,178 @@
...
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
+		accounts, err := store.HostAccountsForFunding(context.Background(), hosts[i%numHosts], batchSize)
+		if err != nil {
+			b.Fatal(err)
+		}
+		b.StartTimer()
+
+		err = store.UpdateHostAccounts(context.Background(), accounts)
```
ChrisSchinnerl, line 178: If `b.N` > `numHosts` this will no longer insert accounts right? Just update them. Might be good to now if we actually run into that by maybe checking for that condition and failing the benchmark if so.

persist/postgres/accounts.go
peterjan: Roughly 20 times faster

ChrisSchinnerl (approved):

persist/postgres/accounts.go
```diff
@@ -0,0 +1,200 @@
...
+		accs = append(accs, acc)
+	}
+	if err := rows.Err(); err != nil {
+		return nil, err
+	}
+
+	return accs, nil
+}
+
+func (s *Store) existingHostAccountsForFunding(ctx context.Context, tx *txn, hk types.PublicKey, hostID int64, limit int) ([]accounts.HostAccount, error) {
```
n8mgr, line 173: nit: there's no need for these to be methods on Store

n8mgr (approved):

persist/postgres/accounts.go
peterjan: I'll move them in a F/U

PR #89, opened 2025-03-27
Add `used_collateral` to database

ChrisSchinnerl:
Adds the `used_collateral` field to the database. Needed to determine whether a contract is running low on collateral or whether it already uses more collateral than the host's most recent `MaxCollateral` settings.

### DIFF
--- a/persist/postgres/contracts.go
+++ b/persist/postgres/contracts.go
@@ -37,7 +37,7 @@ func (s *Store) AddFormedContract(ctx context.Context, contractID types.FileCont
 // - Duplicate the existing contract/row and point the copy to the original
 // - Update a potential row that referenced the existing contract in renewed_to to point to the new row
 // - Overwrite the existing contract to match the renewed contract
-func (s *Store) AddRenewedContract(ctx context.Context, renewedFrom, renewedTo types.FileContractID, proofHeight, expirationHeight uint64, contractPrice, allowance, minerFee, usedCollateral, totalCollateral types.Currency) error {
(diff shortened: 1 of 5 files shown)


persist/postgres/contracts.go
```diff
@@ -37,7 +37,7 @@ func (s *Store) AddFormedContract(ctx context.Context, contractID types.FileCont
 // - Duplicate the existing contract/row and point the copy to the original
 // - Update a potential row that referenced the existing contract in renewed_to to point to the new row
 // - Overwrite the existing contract to match the renewed contract
-func (s *Store) AddRenewedContract(ctx context.Context, renewedFrom, renewedTo types.FileContractID, proofHeight, expirationHeight uint64, contractPrice, allowance, minerFee, totalCollateral types.Currency) error {
```
n8mgr, line 40: nit: consider changing this to take a struct as a parameter over adding additional params

ChrisSchinnerl: I honestly did but what's kept me from it is the fact that structs are not exhaustive. My personal preference is for the build to break when a function needs more arguments and doesn't allow default values. 
But honestly I ran into both issues. Wrong order of params as well as missing ones so it probably doesn't matter.

Added params for `AddRenewedContract`. Will f/u with `AddFormedContract` since it's not directly related to this PR.

peterjan: Having a struct as a parameter is definitely neat and I'm usually in favour of doing that, however I think at a certain point you should explode it into separate arguments. The store would be the level at which you explode it into parameters imo. But I don't mind! Just wanted to `nit` I guess

peterjan (approved):

PR #91, opened 2025-03-31
Add refreshing contracts to contract maintenance

ChrisSchinnerl:


contracts/contract.go
```diff
@@ -128,12 +128,25 @@ func (c Contract) GoodForUpload(maxCollateral types.Currency) bool {
 		return false
 	} else if c.Size >= maxContractSize {
 		return false
-	} else if c.UsedCollateral.Cmp(maxCollateral) > 0 {
+	} else if c.TotalCollateral.Cmp(maxCollateral) >= 0 {
```
n8mgr: Why is this a check for `GoodForUpload`? Feels like it should be: 
```
c.Good && c.Size < maxContractSize && renterAllowance > 1 sector append cost && hostCollateral > 1 sector append cost
```

contracts/manager.go
```diff
@@ -296,7 +297,8 @@ func (cm *ContractManager) performContractMaintenance(ctx context.Context, log *
 		return fmt.Errorf("failed to renew contracts: %w", err)
 	}
 
-	// TODO: Refresh any good contracts that are either out of collateral or funds
+	// refresh any good contracts that are either out of collateral or funds
+	cm.performContractRefreshes(ctx, log)
```
n8mgr: Either check the error or don't return an error. Even if it's 

```
if err != nil {
    log.Debug('')
}
```

contracts/manager.go
```diff
@@ -296,7 +297,8 @@ func (cm *ContractManager) performContractMaintenance(ctx context.Context, log *
...
 
 	// mark any contracts too close to their expiration height as bad
 	cm.store.MarkUnrenewableContractsBad(ctx, blockHeight+settings.RenewWindow/2)
```
n8mgr: Out of scope for this PR, but ditto

contracts/refresh.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (cm *ContractManager) performContractRefreshes(ctx context.Context, log *zap.Logger) error {
+	renewalLog := log.Named("refresh")
+	contracts, err := cm.store.Contracts(ctx, WithGood(true), WithRevisable(true))
+	if err != nil {
+		return fmt.Errorf("failed to fetch contracts for refreshing: %w", err)
+	}
+
+	for _, contract := range contracts {
+		if !contract.Good {
```
n8mgr: Could be combined into `NeedsRefresh`

n8mgr (changes requested):

contracts/manager.go
ChrisSchinnerl: hah completely missed that those return errors

contracts/contract.go
ChrisSchinnerl: Hm yeah I guess `UsedCollateral` made more sense. The goal was to cover the edge case where you shouldn't upload to a contract anymore that a host won't be willing to renew since it has too much collateral in it already.

I aded the other checks as well. I thought it might be good enough to have the `OutOfX` checks but if you really manage to drain a contract down from 10% to 1 sector it's probably time to form a new contract and replace it.

n8mgr: I guess that's true. If the host's max collateral went under the existing amount for some reason, we really shouldn't continue uploading to it.

n8mgr (dismissed):
> I took a stab at it as well. I updated the performance of `UpdateHostAccounts` quite a lot by doing a batch update. I think it's fine because we have use a batch size of 1000 which won't change any time soon. Unfortunately it's not fast enough to allow us to insert every ephemeral account while doing benchmarks.
> 
> I updated `HostAccountsForFunding` and split the query into fetching `NULL` and fetching ephemeral accounts separately. The downside of splitting up the query is that benchmarking it is quite annoying because you'd have to insert all of the ephemeral accounts aside from half a batch size to ensure the accounts cover the entire query. That's not an option so I resorted to passing in a parameter, definitely not ideal but at least it allows us to benchmark both parts of the query.
> 
> I was thinking this might be a really good fit for redis (using `SETEX` and/or `TTL`s), similar to a materialized view but performance would likely be better. Definitely not ideal though :/ not sure if we can get out of going the materialized view route. The benchmark will always be a little meaningless if we can't populate the entire join table in a reasonable amount of time.

PR #92, opened 2025-03-31
Config command

ChrisSchinnerl:
This PR adds a `indexd config` command plus a few minor changes to make `indeed` convenient to run from the CLI since we are soon going to be able to launch indexd on Zen to form and maintain a few contracts.

### DIFF
--- a/cmd/indexd/config.go
+++ b/cmd/indexd/config.go
@@ -14,7 +14,7 @@ import (
 	"go.sia.tech/core/types"
 	"go.sia.tech/coreutils/wallet"
 	"golang.org/x/term"
-	"gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v3"
 )
 
 func runConfigCmd(fp string) {
--- a/cmd/indexd/config.go
+++ b/cmd/indexd/config.go
@@ -1,0 +1,326 @@
+package main
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"net"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"strings"
+
+	"go.sia.tech/core/types"
+	"go.sia.tech/coreutils/wallet"
+	"golang.org/x/term"
+	"gopkg.in/yaml.v2"
+)
+
+func runConfigCmd(fp string) {
+	fmt.Println("indexd Configuration Wizard")
+	fmt.Println("This wizard will help you configure indexd for the first time.")
+	fmt.Println("You can always change these settings with the config command or by editing the config file.")
+
+	if fp == "" {
+		fp = configPath()
+	}
+	fp, err := filepath.Abs(fp)
+	checkFatalError("failed to get absolute path of config file", err)
+
+	fmt.Println("")
+	fmt.Printf("Config Location %q\n", fp)
+
+	if _, err := os.Stat(fp); err == nil {
+		if !promptYesNo(fmt.Sprintf("%q already exists. Would you like to overwrite it?", fp)) {
+			return
+		}
+	} else if !errors.Is(err, os.ErrNotExist) {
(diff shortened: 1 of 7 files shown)


cmd/indexd/main.go
```diff
@@ -190,17 +203,20 @@ func main() {
 	rootCmd := flagg.Root
 	rootCmd.Usage = flagg.SimpleUsage(rootCmd, ``)
 	rootCmd.StringVar(&cfg.Directory, "dir", cfg.Directory, "directory to store indexd metadata in")
-	rootCmd.StringVar(&cfg.HTTP.Address, "http", cfg.HTTP.Address, "address to serve API on")
+	rootCmd.StringVar(&cfg.AdminAPI.Address, "adminAPI", cfg.AdminAPI.Address, "address to serve admin API on")
```
n8mgr: nit: I'd probably go `-api.admin` and `-api.app`. camel case CLI flags feel weird to me.

cmd/indexd/main.go
```diff
@@ -31,9 +32,12 @@ const (
 
 var cfg = config.Config{
 	Directory: os.Getenv(dataDirEnvVar), // default to env variable
-	HTTP: config.HTTP{
-		Address:  "127.0.0.1:8080",
-		Password: "changeme_please",
+	AdminAPI: config.AdminAPI{
+		Address:  "127.0.0.1:9980",
+		Password: "",
```
n8mgr: nit: should allow this to be set via ENV `INDEXED_ADMIN_PASSWORD` or something

cmd/indexd/main.go
```diff
@@ -233,12 +249,23 @@ func main() {
...
+
+		runConfigCmd(configPath)
 	case rootCmd:
 		if len(cmd.Args()) != 0 {
 			cmd.Usage()
 			return
 		}
 
+		if cfg.AdminAPI.Password == "" {
+			setAPIPassword()
```
n8mgr: Better to return an error. Prompting with stdin has always been odd, imo.

config/config.go
```diff
@@ -11,66 +11,74 @@ import (
...
+		Address string `yaml:"address"`
 	}
 
 	// Syncer contains the configuration for the p2p syncer.
 	Syncer struct {
-		Address    string   `yaml:"address,omitempty"`
-		Bootstrap  bool     `yaml:"bootstrap,omitempty"`
-		EnableUPnP bool     `yaml:"enableUPnP,omitempty"`
-		Peers      []string `yaml:"peers,omitempty"`
+		Address    string   `yaml:"address"`
```
n8mgr, line 29: Was there a reason to remove `omitempty`?

cmd/indexd/config.go
```diff
@@ -0,0 +1,326 @@
...
+	"os"
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"strings"
+
+	"go.sia.tech/core/types"
+	"go.sia.tech/coreutils/wallet"
+	"golang.org/x/term"
+	"gopkg.in/yaml.v2"
```
n8mgr: yaml.v3

n8mgr (changes requested):

config/config.go
peterjan: I'm guessing it serves a documentation purpose after running the `config` command?

peterjan (approved):

config/config.go
ChrisSchinnerl: The trigger was that the config file got written to disk without `Level` for the loggers. Because `info` is actually the default value and that got omitted. 
So I went through the fields and since we set non-default values for most fields I removed it. Imo it's pretty nice to have all the available options generated since it can save you a trip to the docs.

n8mgr (approved):

PR #95, opened 2025-04-01
Upgrade golangci.yml and go version

peterjan:
I had to upgrade `golangci-lint` locally and noticed they updated their YAML. They have a handy migrate tool though, so this is the result of running that tool. I noticed in our go.mod our go language version differs from our toolchain version so figured we might as well update it to go 1.24.1, `vaultd` does it too.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,150 +1,111 @@
-# Based off of the example file at https://github.com/golangci/golangci-lint
-
-# options for analysis running
+version: "2"
 run:
-  # default concurrency is a available CPU number
   concurrency: 4
-
-  # timeout for analysis, e.g. 30s, 5m, default is 1m
-  timeout: 600s
-
-  # exit code when at least one issue was found, default is 1
   issues-exit-code: 1
-
-  # include test files or not, default is true
   tests: true
-
-  # list of build tags, all linters use it. Default is empty list.
-  build-tags: []
-
-# output configuration options
-output:
-  # print lines of code with issue, default is true
-  print-issued-lines: true
-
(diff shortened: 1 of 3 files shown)


ChrisSchinnerl (changes requested):
Failing CI

peterjan:
> Failing CI

bad! that's a bad CI [insert-cartman-gif]

.golangci.yml
```diff
@@ -1,150 +1,111 @@
...
-  fast: false
+  default: none
   enable:
-    - tagliatelle
     - gocritic
-    - gofmt
-    - revive
     - govet
     - misspell
-    - typecheck
```
ChrisSchinnerl, line 130: Is this gone now?

peterjan: Yeah I noticed that too, there's a couple of nuances actually, `gofmt` is another one (although that one is added to the formatters below). When I tried re-adding `typecheck` running `golangci-lint run` returns the following error:

```
Error: can't load config: typecheck is not a linter, it cannot be enabled or disabled
Failed executing command with error: can't load config: typecheck is not a linter, it cannot be enabled or disabled
```

It's also mentioned in their migration guide: https://golangci-lint.run/product/migration-guide/#typecheck.

When I saw that I was kind of dreading going through all of them separately. If you run their `migrate` command it lists meaningful changes and prints warnings. Ours had only one actionable item in the output which was about timeouts. I then figured it was safe to rely on their migrate command and re-added our `timeout`. In v2 there's no timeouts by default.

n8mgr (approved):

ChrisSchinnerl (approved):

PR #96, opened 2025-04-02
Add AccountManager

peterjan:


accounts/manager.go
```diff
@@ -0,0 +1,117 @@
...
+	"time"
+
+	proto "go.sia.tech/core/rhp/v4"
+	"go.sia.tech/core/types"
+	"go.sia.tech/indexd/hosts"
+	"go.uber.org/zap"
+)
+
+const (
+	accountFundBatch            = 1000 // equals max batch size in replenish RPC
```
ChrisSchinnerl: I'd reuse `MaxAccountBatchSize` from `core` rather than having a bunch of `1000` constants over the codebase with comments explaining that they should be the same as the RPC's limit.

accounts/manager.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// Close closes the AccountManager.
+func (m *AccountManager) Close() error {
+	return nil
+}
+
+// FundAccounts attempts to fund all accounts for the given host key. It does so
+// using the provided contract IDs, which are used in the order they're given.
+// It returns a breakdown of the total costs incurred by the funding to allow
```
ChrisSchinnerl: Returning a single `Usage` won't allow us to track spending on a per-contract basis right? 

It's probably fine for this PR to not return a `Usage` at all and then follow up with a PR that adds tracking. Every time we revise a contract we want to make sure we update all of the contracts fields in the store to match the latest revision again. Spending, revision number, size, root, used collateral, remaining allowance etc.
So the easiest approach might be to actually update the store right after using the contract rather than returning the usage to the contract maintenance. That way we don't need to pass as much state around.

accounts/manager.go
```diff
@@ -0,0 +1,117 @@
...
+// tracking contract spending.
+func (m *AccountManager) FundAccounts(ctx context.Context, hk types.PublicKey, contractIDs []types.FileContractID, log *zap.Logger) (proto.Usage, error) {
+	host, err := m.store.Host(ctx, hk)
+	if err != nil {
+		return proto.Usage{}, fmt.Errorf("failed to fetch host: %w", err)
+	}
```
ChrisSchinnerl, line 84: It might be worthwhile to just pass the host to `FundAccounts` rather than fetching it from the store. The caller needs to fetch the hosts anyway so we can avoid doing this here.

We should probably also sanity check that the host is both usable and not blocked here.

accounts/manager.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+
+		totalUsage = totalUsage.Add(result.Usage)
+		for i, funded := range result.Funded {
+			if funded {
+				accounts[i].ConsecutiveFailedFunds = 0
+				accounts[i].NextFund = time.Now().Add(accountFundInterval)
+			} else {
+				accounts[i].ConsecutiveFailedFunds++
+				accounts[i].NextFund = time.Now().Add(time.Duration(min(math.Pow(2, float64(accounts[i].ConsecutiveFailedFunds)), accountExpBackoffMaxMinutes)) * time.Minute)
```
ChrisSchinnerl: This would be a good candidate for a tiny helper function with a table test.

accounts/funder.go
```diff
@@ -0,0 +1,142 @@
...
+	// FundResult is the result of a funding operation.
+	FundResult struct {
+		Funded []bool
+		Usage  proto.Usage
+	}
+)
+
+func newFundResult(accounts []HostAccount) FundResult {
+	return FundResult{
+		Funded: make([]bool, len(accounts)),
```
ChrisSchinnerl: This should be able to just be a single number right? Assuming you are starting from the beginning of the `accounts` slice and work your way through in batches until funding fails, there will never be any gaps with a bunch of `false` in this slice.

accounts/funder.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		log.Debug("failed to dial host", zap.Error(err))
+		return result, nil
+	}
+	defer t.Close()
+
+	// prepare unfunded indices
+	unfunded := make([]int, 0, len(accounts))
+	for i := range accounts {
+		unfunded = append(unfunded, i)
```
ChrisSchinnerl: Related to the question before. Why isn't this just tracked using an offset that indicates the accounts that haven't been funded yet? Batches will never partially fail right? So you can increment a single offset to now what batch comes next.

accounts/funder.go
```diff
@@ -0,0 +1,142 @@
...
+			continue
+		} else if !rev.Revisable {
+			contractLog.Debug("contract is not revisable") // sanity check
+			continue
+		} else if rev.Contract.RenterOutput.Value.IsZero() {
+			contractLog.Debug("contract is out of funds")
+			continue
+		}
+		balance := rev.Contract.RenterOutput.Value
+
```
ChrisSchinnerl: You will need to also adjust your batch size using the remaining balance of the contract and the target.

e.g. if your batch size would be 1000 accounts with a target of 1SC but you only got 500SC left in the contract, you should do a batch of 500 instead of 1000 first. If you got 400 SC left afterwards you can do another batch of 400 accounts and so on. 
Otherwise you are bound to run into issues for contracts with less than 1000SC of remaining balance. Especially since we might run into an edge case where the contract doesn't get refilled because it got plenty of funds left but we still fail to fund accounts.

accounts/funder.go
```diff
@@ -0,0 +1,142 @@
...
+		// fetch the latest revision, check it's revisable and has money
+		rev, err := rhp.RPCLatestRevision(ctx, t, fcid)
+		if err != nil {
+			contractLog.Debug("failed to fetch latest revision", zap.Error(err))
+			continue
+		} else if !rev.Revisable {
+			contractLog.Debug("contract is not revisable") // sanity check
+			continue
+		} else if rev.Contract.RenterOutput.Value.IsZero() {
```
ChrisSchinnerl: I'd ignore a contract as soon as it has a balance < target. Because at that point it is a gamble whether the one account we want to fund can actually be funded. 

Instead we should make sure that the contract manager is aware of the target balance we use for funding and update `OutOfFunds` to cause the contract to get refreshed when that happens.

accounts/funder.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+
+		// update unfunded accounts
+		unfunded = unfunded[len(batch):]
+		if len(unfunded) == 0 {
+			break
+		}
+	}
+
+	return result, nil
```
ChrisSchinnerl: `FundAccount` has a lot of logic in it with a bunch of batching and potential edge cases. I'd add a unit test for it to make sure we don't run into any issues there.

ChrisSchinnerl (changes requested):

accounts/funder.go
peterjan: Reminder to F/U with check to `OutOfFunds`

accounts/manager.go
```diff
@@ -0,0 +1,125 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to fetch accounts for funding: %w", err)
+		} else if len(accounts) < accountFundBatch {
+			exhausted = true
+		}
+		if len(accounts) == 0 {
+			break
+		}
+
+		n, err := m.funder.FundAccounts(ctx, host, accounts, contractIDs, log)
```
peterjan: I considered manipulating the `accounts` slice with the funding results. I would usually not do that but here it kind of makes sense. On the other hand, I don't want to put too much responsibility in the `AccountFunder` and instead restrict it to interacting with the host. When we add usage tracking, I plan on changing this to `n, usages := m.funder.FundAccounts` and get rid of the error (which is always `nil` anyway). 

I went back and fourth on this a bit. I originally wanted to reintroduce the `FundResult` struct and have it contain the updated accounts and the usages, but it's a little awkward when we fail to dial. I think I prefer returning `funded, usages` instead and deal with it on the caller side.

accounts/funder.go
peterjan: Well I was, but I made it shorter.

n8mgr (commented):
~~Not sure I see a reason for a specific package since it has to be tightly coupled to the contract manager regardless~~

n8mgr (dismissed):

accounts/funder.go
ChrisSchinnerl: Doesn't matter really but this check should now be redundant.

accounts/funder.go
```diff
@@ -0,0 +1,137 @@
...
+			continue
+		} else if rev.Contract.RenterOutput.Value.IsZero() {
+			contractLog.Debug("contract is out of funds")
+			continue
+		} else if rev.Contract.RenterOutput.Value.Cmp(f.target) < 0 {
+			contractLog.Debug("contract has insufficient funds")
+			continue
+		}
+
+		// prepare batch
```
ChrisSchinnerl, line 115: I think we might be missing an inner loop here? Right now we fund batch 1 with contract 1, batch 2 with contract 2 and so on and stop when we either have funded all batches or run out of contracts.
Rather than funding as many batches as possible with contract 1 and only then moving on to contract 2.

To make sure that's not happening I'd add another test that requires multiple batches to fund accounts.
With 3 contracts where the first contract is good enough to fill the first couple of accounts, the second one is good for the remaining accounts and the third one isn't used.

ChrisSchinnerl (changes requested):

accounts/funder.go
peterjan: Yeah left it on purpose actually but will remove it.

accounts/funder.go
peterjan: Hm I was actually calling `FundAccounts` with a single batch. I'll rewrite it and add the loop.

accounts/funder.go
```diff
@@ -96,42 +105,39 @@ func (f *Funder) FundAccounts(ctx context.Context, host hosts.Host, accounts []H
 		} else if !rev.Revisable {
 			contractLog.Debug("contract is not revisable") // sanity check
 			continue
-		} else if rev.Contract.RenterOutput.Value.IsZero() {
-			contractLog.Debug("contract is out of funds")
-			continue
 		} else if rev.Contract.RenterOutput.Value.Cmp(f.target) < 0 {
 			contractLog.Debug("contract has insufficient funds")
 			continue
```
ChrisSchinnerl, line 112: should probably also increment `drained` for these 2 cases here when the contract isn't revisable or out of funds. Not just when it's out of funds after a successful replenish.

peterjan: ffs

accounts/funder_test.go
```diff
@@ -0,0 +1,141 @@
...
+	proto "go.sia.tech/core/rhp/v4"
+	"go.sia.tech/core/types"
+	"go.sia.tech/coreutils/rhp/v4"
+	"go.sia.tech/indexd/hosts"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zaptest/observer"
+)
+
+var badContractId = types.FileContractID{1, 1, 1}
```
ChrisSchinnerl, line 20: (nit) Instead of a sentinel value like this which imo is a bit too easy to accidentally pick in a test, why not use a non-existing contract id in the test? That will also cause `RPCReplenishAccounts` to fail.

ChrisSchinnerl (dismissed):

accounts/funder_test.go
peterjan: Because it would fail fetching the revision 🤷‍♂️ , never get to replinishing.

ChrisSchinnerl (approved):

accounts/funder.go
```diff
@@ -0,0 +1,145 @@
...
+func (f *Funder) FundAccounts(ctx context.Context, host hosts.Host, accounts []HostAccount, contractIDs []types.FileContractID, log *zap.Logger) (funded int, drained int, _ error) {
+	// sanity check
+	if len(accounts) > proto.MaxAccountBatchSize {
+		return 0, 0, errors.New("too many accounts") // developer error
+	}
+
+	// dial host
+	dialCtx, cancel := context.WithTimeout(ctx, dialTimeout)
+	t, err := f.host.Dial(dialCtx, host.SiamuxAddr(), host.PublicKey)
+	cancel()
```
n8mgr, line 83: nit: any reason this isn't `defer cancel()`? At first I thought there might be another cancellation somewhere, but there's not.

n8mgr (approved):

accounts/funder.go
peterjan: I'll take this with me in the next one. I sometimes do it when I have a separate context with a timeout to clearly show it only concerns the dial. It's in my head though there's no real reason for it, I like the way it looks visually.

PR #97, opened 2025-04-02
PerformAccountFunding

peterjan:


contracts/manager.go
```diff
@@ -230,6 +244,74 @@ func (cm *ContractManager) blockUntilReady(log *zap.Logger) bool {
 	}
 }
 
+func (cm *ContractManager) performAccountFunding(ctx context.Context, log *zap.Logger) error {
+	start := time.Now()
+	log = log.Named("accounts")
+
+	// fetch all good contracts
+	contracts, err := cm.store.Contracts(ctx, WithRevisable(true), WithGood(true))
```
n8mgr: It's worth adding a query to return a good contract per host, sorted by remaining_allowance, now that we track it in #96. It's an extremely easy query. Continuing to extend the generic query and filter on the Go side is becoming more trouble than it's worth. Our decisions around where to filter results—Postgres vs. Go—have been pretty inconsistent.

n8mgr (changes requested):

contracts/manager.go
```diff
@@ -270,6 +284,64 @@ func (cm *ContractManager) blockBadHosts(ctx context.Context) error {
...
+	// fund accounts on all hosts
+	var wg sync.WaitGroup
+	sema := make(chan struct{}, fundThreads)
+	defer close(sema)
+
+	for hk, contractIDs := range hkToContractIDs {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case sema <- struct{}{}:
```
ChrisSchinnerl: This needs to be moved down right before the `go func`. Otherwise we are potentially deadlocking due to the `continue`s where we don't receive from the `sema`.

contracts/manager.go
```diff
@@ -58,6 +67,7 @@ type (
 		AddRenewedContract(ctx context.Context, params AddRenewedContractParams) error
 		ContractElementsForBroadcast(ctx context.Context, maxBlocksSinceExpiry uint64) ([]types.V2FileContractElement, error)
 		Contracts(ctx context.Context, queryOpts ...ContractQueryOpt) ([]Contract, error)
+		ContractsForFunding(ctx context.Context) (map[types.PublicKey][]types.FileContractID, error)
```
ChrisSchinnerl: Not sure how Nate feels about this but I might be leaning slightly towards making the public key an argument and only returning contracts for a given host. To avoid fetching a very large batch of information every time on indexers with possibly hundreds of hosts.

The `WithActiveContracts` filter on the `Hosts` function could be used to first return all hosts with contracts.

contracts/fund_test.go
```diff
@@ -0,0 +1,120 @@
...
+	store.hosts[hk2] = hosts.Host{
+		PublicKey: hk2,
+		Usability: goodUsability,
+	}
+	store.contracts = append(store.contracts, Contract{
+		ID:                 types.FileContractID{3},
+		HostKey:            hk2,
+		RemainingAllowance: types.Siacoins(1),
+	})
+
```
ChrisSchinnerl, line 96: I'd add 2 more hosts. One being blocked and one being unusable to make sure we don't fund for those.

ChrisSchinnerl (changes requested):

contracts/manager.go
peterjan: Urgh, nice catch. Introduced it when I moved from `hk` -> `host`.

contracts/manager.go
peterjan: Will wait on Nate's input before making this change, I like the suggestion, on the other hand it's kind of nice to have the entire filter in one query.

cmd/indexd/run.go
```diff
@@ -106,7 +106,7 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
 	defer hm.Close()
 
 	cf := contracts.NewContractor(cm, wm, walletKey)
-	contracts, err := contracts.NewManager(walletKey.PublicKey(), cm, cf, hm, store, s, wm, contracts.WithLogger(log.Named("contracts")))
+	contracts, err := contracts.NewManager(walletKey.PublicKey(), nil, cm, cf, hm, store, s, wm, contracts.WithLogger(log.Named("contracts")))
```
ChrisSchinnerl: Can create an account manager now right?

contracts/manager.go
```diff
@@ -270,6 +284,70 @@ func (cm *ContractManager) blockBadHosts(ctx context.Context) error {
...
+		hosts.WithUsable(true),
+		hosts.WithBlocked(false),
+		hosts.WithActiveContracts(true),
+	}
+
+	var offset int
+	var exhausted bool
+	for !exhausted {
+		if err := ctx.Err(); err != nil {
+			break
```
ChrisSchinnerl: If you already use `Err` you might as well do `for !exhausted && ctx.Err() == nil`

contracts/manager.go
```diff
@@ -270,6 +284,70 @@ func (cm *ContractManager) blockBadHosts(ctx context.Context) error {
...
+		}
+
+		// fetch hosts
+		hostss, err := cm.store.Hosts(ctx, offset, fundThreads, opts...)
```
ChrisSchinnerl: (nit) For those of us no familiar with Parseltongue I'd probably go with `hostsToFund`, `toFund`, `fundableHosts` etc. to avoid the package name clash.

ChrisSchinnerl (commented):
Left a few nits but looks good overall

contracts/manager.go
peterjan: haha

ChrisSchinnerl (approved):

ChrisSchinnerl:
@n8maninger is this ready to go?

peterjan:
> @n8maninger is this ready to go?

@n8maninger possible this notification slipped through the cracks?

n8mgr (approved):

PR #98, opened 2025-04-03
Sync revision state in contract maintenance 

ChrisSchinnerl:
The point of this is to make sure our perception of a contract is the same as the host's before we start maintenance. That way we can base our checks off the values we have stored in our database without having to worry about them being outdated.


contracts/chain.go
```diff
@@ -36,7 +36,7 @@ func (tx *updateTx) IsKnownContract(fcid types.FileContractID) (bool, error) {
 	if found {
 		return known, nil
 	}
-	known, err := tx.IsKnownContract(fcid)
+	known, err := tx.UpdateTx.IsKnownContract(fcid)
```
n8mgr, line 39: One reason I really hate embedding

contracts/manager.go
```diff
@@ -309,6 +316,72 @@ func (cm *ContractManager) performContractMaintenance(ctx context.Context, log *
...
+	log := cm.log.Named("syncRevisionState")
+
+	sema := make(chan struct{}, 50)
+	defer close(sema)
+
+	var wg sync.WaitGroup
+	for _, contract := range contracts {
+		select {
+		case <-ctx.Done():
+			break
```
n8mgr: need to break the loop not the switch

n8mgr (changes requested):

peterjan (dismissed):
If we revise our contracts in a synchronous manner, we should be able to update this state when we revise the contract and be in sync at all times. (edit: I obviously approve _after_ tackling Nate's comments)

contracts/manager.go
```diff
@@ -309,6 +316,72 @@ func (cm *ContractManager) performContractMaintenance(ctx context.Context, log *
...
+		}
+		wg.Add(1)
+		go func(contract Contract) {
+			defer func() {
+				<-sema
+				wg.Done()
+			}()
+			contractLog := log.With(zap.Stringer("contractID", contract.ID), zap.Stringer("hostKey", contract.HostKey))
+
+			host, err := cm.store.Host(ctx, contract.HostKey)
```
peterjan, line 346: I know we once removed the host info from the contract metadata in `renterd` so I'm not saying we should tightly couple those, but it's still quite expensive to fetch all contract and all host metadata to get the host's key and siamux address and the file contract ID.

contracts/manager.go
ChrisSchinnerl: I really hate this in Go^^

contracts/chain.go
ChrisSchinnerl: Yeah. Only upside is that it's not sneaky cause it crashes immediately when you run it. We just don't have integration tests yet.

contracts/manager.go
ChrisSchinnerl: I really want to avoid prematurely optimizing things this time around. If fetching a full host takes more than a ms we are probably already doing something wrong and for contract maintenance which isn't that time sensitive it's likely to be fine.
If it actually is an issue once we start testing this e2e we can pretty easily start swapping out specific queries for custom ones that only return partial data.

n8mgr (approved):

contracts/manager.go
```diff
@@ -328,10 +328,11 @@ func (cm *ContractManager) syncRevisionState(ctx context.Context) error {
 	defer close(sema)
 
 	var wg sync.WaitGroup
+LOOP:
 	for _, contract := range contracts {
 		select {
 		case <-ctx.Done():
-			break
+			break LOOP
```
peterjan, line 335: much cleaner imo:

```go
for _, contract := range contracts {
	if ctx.Err() != nil {
		break
	}
```

ChrisSchinnerl: It's functionally different though and ignores the semaphore. Sending to the semaphore needs to be aborted too when the context is closed. 
When you need both the error check and the select it ends up a lot nastier again.

peterjan (approved):

PR #99, opened 2025-04-03
Add database code for remaining contract allowance

ChrisSchinnerl:
Seems like that field was added but not persisted yet

### DIFF
--- a/persist/postgres/contracts_test.go
+++ b/persist/postgres/contracts_test.go
@@ -223,10 +223,11 @@ func TestFormRenewContract(t *testing.T) {
 		ExpirationHeight: 200,
 		State:            contracts.ContractStatePending,
 
-		ContractPrice:    types.Siacoins(1),
-		InitialAllowance: types.Siacoins(2),
-		MinerFee:         types.Siacoins(3),
-		TotalCollateral:  types.Siacoins(4),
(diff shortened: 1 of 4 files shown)


persist/postgres/contracts_test.go
```diff
@@ -223,10 +223,11 @@ func TestFormRenewContract(t *testing.T) {
 		ExpirationHeight: 200,
 		State:            contracts.ContractStatePending,
 
-		ContractPrice:    types.Siacoins(1),
-		InitialAllowance: types.Siacoins(2),
-		MinerFee:         types.Siacoins(3),
-		TotalCollateral:  types.Siacoins(4),
+		ContractPrice:      types.Siacoins(1),
+		InitialAllowance:   types.Siacoins(2),
+		RemainingAllowance: types.Siacoins(2),
```
peterjan, line 228: this triggers me, should be `types.Siacoins(3),` (and so on)

n8mgr (approved):

peterjan (approved):

persist/postgres/contracts_test.go
ChrisSchinnerl: Then the test will break cause they both need to be the same ^^

PR #103, opened 2025-04-07
RFC: Metadata Store

### LINKED ISSUE #100: RFC: Object store
Write an RFC for an object store that allows for storing arbitrary encrypted metadata under a key. That store can then be used to store metadata about slabs making up an object and serves as a backup solution for object metadata. 

Ideally the store is also tamper resistant. e.g. by making the key a signature of the value.

ChrisSchinnerl:
Adds the RFC for a generic, encrypted and tamper-resistant metadata store apps can use for backing up object metadata.

Closes https://github.com/SiaFoundation/indexd/issues/100

### DIFF
--- a/doc/rfc/001_db_schema.md
+++ b/doc/rfc/001_db_schema.md
@@ -222,7 +222,9 @@ CREATE TABLE contract_elements (
 ```postgresql
 CREATE TABLE accounts (
     id SERIAL PRIMARY KEY,
-    public_key BYTEA UNIQUE NOT NULL CHECK (LENGTH(public_key) = 32)
+    public_key BYTEA UNIQUE NOT NULL CHECK (LENGTH(public_key) = 32),
+    max_metadata_size BIGINT NOT NULL DEFAULT 1000000000000 CHECK (metadata_size <= max_metadata_size), -- 1GB of metadata
+    metadata_size BIGINT NOT NULL DEFAULT 0 -- sum of size of stored metadata (updated on insert and delete)
 );
 
 CREATE TABLE account_slabs (
@@ -276,4 +278,16 @@ CREATE INDEX sectors_contract_id_uploaded_at_idx ON host_sectors(contract_id, up
 
 -- index over contract_id and next_integrity_check since we only check pinned sectors
 CREATE INDEX sectors_contract_id_next_integrity_check_idx ON host_sectors(contract_id, next_integrity_check ASC)
+```
+
+### 2.7 Metadata
+
+```postgresql
+CREATE TABLE metadata (
+    id BIGSERIAL PRIMARY KEY,
+
+    account_id INTEGER REFERENCES accounts(id) NOT NULL,
+    key BYTEA NOT NULL,
+)
+CREATE UNIQUE INDEX metadata_account_id_key ON metadata(account_id, key)
 ```
--- a/doc/rfc/001_db_schema.md
+++ b/doc/rfc/001_db_schema.md
@@ -289,5 +289,5 @@ CREATE TABLE metadata (
     account_id INTEGER REFERENCES accounts(id) NOT NULL,
     key BYTEA NOT NULL
 );
-CREATE UNIQUE INDEX metadata_account_id_key ON metadata(account_id, key)
+CREATE UNIQUE INDEX metadata_account_id_key_idx ON metadata(account_id, key)
 ```
--- a/doc/rfc/001_db_schema.md
+++ b/doc/rfc/001_db_schema.md
@@ -223,7 +223,7 @@ CREATE TABLE contract_elements (
 CREATE TABLE accounts (
     id SERIAL PRIMARY KEY,
     public_key BYTEA UNIQUE NOT NULL CHECK (LENGTH(public_key) = 32),
-    max_metadata_size BIGINT NOT NULL DEFAULT 1000000000000 CHECK (metadata_size <= max_metadata_size), -- 1GB of metadata
+    max_metadata_size BIGINT NOT NULL DEFAULT 1000000000 CHECK (metadata_size <= max_metadata_size), -- 1GB of metadata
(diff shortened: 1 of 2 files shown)


doc/rfc/012_metadata_store.md
```diff
@@ -0,0 +1,81 @@
...
+
+- Store a value for a given 32-byte key (SHA-256 HMAC)
+- Retrieve a value for a given 32-byte key
+- Fetch all available keys with offset+limit
+
+While not important for the indexer implementation since it can't verify it, the
+key is intended to be a fingerprint of the value that only the client can
+create. To achieve that, we use an HMAC over the encrypted value.
+
+### HMAC Creation
```
ChrisSchinnerl, line 31: I'm not sure if this is necessary or if we can use the 16 byte poly1305 tag instead of the 32 byte HMAC.

doc/rfc/012_metadata_store.md
```diff
@@ -0,0 +1,81 @@
...
+
+Clients already use ED25519 keys for authentication. To avoid increasing the
+complexity of using indexd, we try to keep the number of keys a client
+application needs to manage to one. To achieve that, we derive the key we use
+for the HMAC from the ED25519 private key since it's not considered safe to
+reuse the same key material for different cryptographic algorithms directly.
+
+For that, we use the key derivation function HKDF which is specifically made for
```
ChrisSchinnerl, line 39: In theory a simple append+hash might also work but doing some research it seems like it's generally considered better to use a proper KDF to derive keys rather than just hashing. The opinion seems to be that while not proven flawed, something like HKDF provides better guarantees. There is also a reference to this in the HKDF paper.

doc/rfc/001_db_schema.md
```diff
@@ -277,3 +279,15 @@ CREATE INDEX sectors_contract_id_uploaded_at_idx ON host_sectors(contract_id, up
...
+### 2.7 Metadata
+
+```postgresql
+CREATE TABLE metadata (
+    id BIGSERIAL PRIMARY KEY,
+
+    account_id INTEGER REFERENCES accounts(id) NOT NULL,
+    key BYTEA NOT NULL
+);
+CREATE UNIQUE INDEX metadata_account_id_key ON metadata(account_id, key)
```
peterjan:
```suggestion
CREATE UNIQUE INDEX metadata_account_id_idx ON metadata(account_id, key)
```

doc/rfc/012_metadata_store.md
```diff
@@ -0,0 +1,81 @@
...
+back up metadata on top pinning slabs, there is no way for client applications
+to recover information about what slabs make up an object or additional metadata
+such as the object's content type.
+
+To add that behavior, the indexer offers a generic metadata store which holds
+client-side encrypted, tamper-resistant data. To maximize privacy, there is no
+relation between these blobs of metadata and slabs that the indexer knows of.
+This severely limits the capability of the store as the indexer can't perform
+searches on this data, but the client can add any necessary complexity
+client-side while the metadata store serves as a pure backup mechanism.
```
peterjan:
```suggestion
client-side while the metadata store serves purely as a backup mechanism.
```

doc/rfc/012_metadata_store.md
```diff
@@ -0,0 +1,81 @@
...
+domain-specific keys. The salt/domain should be the string `MD-HMAC-SECRET`. We
+then use the resulting key to create the HMAC over the encrypted value to store
+in the metadata store.
+
+### Value Encryption
+
+Similar to the HMAC creation, we derive another key for the encryption of the
+value with HKDF as well as a random nonce. This time we use the salt
+`MD-ENCRYPTION-SECRET`. Then we use CHACHA20-Poly1305 to encrypt the plaintext
+value and append the nonce to it. The nonce is appended since it mustn't be
```
peterjan: Could add a small section about the nonce (12-byte nonce size, explicitly mention reuse is insecure)

peterjan (commented):
LGTM

doc/rfc/001_db_schema.md
```diff
@@ -277,3 +279,15 @@ CREATE INDEX sectors_contract_id_uploaded_at_idx ON host_sectors(contract_id, up
 -- index over contract_id and next_integrity_check since we only check pinned sectors
 CREATE INDEX sectors_contract_id_next_integrity_check_idx ON host_sectors(contract_id, next_integrity_check ASC)
 ```
+
+### 2.7 Metadata
+
+```postgresql
+CREATE TABLE metadata (
```
n8mgr, line 286: Would it be better to enforce some structure here to make abuse less likely over enforcing a max value size per account?

```psql
CREATE TABLE objects (
    id BIGSERIAL PRIMARY KEY,
    encrypted_key BYTEA UNIQUE NOT NULL CHECK(LENGTH=1000),
    encrypted_size BYTEA NOT NULL CHECK(LENGTH(encrypted_size) = 28)
);

CREATE TABLE object_slabs (
    object_id BIGINT REFERENCES objects(id) NOT NULL,
    encrypted_slab_id BYTEA NOT NULL CHECK(LENGTH(encrypted_slab_id) = 56)
);
```

ChrisSchinnerl: I guess it depends on a) if we consider this enough and b) what we want to offer.

Ad a) More structure makes it harder to abuse but theoretically you can still encode raw data as encrypted slab ids. So we might need to limit the number of objects/slabs a user can store anyway.

Ad b) Making it generic is nice because it's less complexity on our end and allows for use-cases that we can't think of yet. Such as supporting generic metadata in an S3 app or setting custom hashes in an rclone-like app.

n8mgr: Fair 👍🏻

doc/rfc/012_metadata_store.md
ChrisSchinnerl: updated it

peterjan (approved):

n8mgr (approved):

PR #104, opened 2025-04-07
Add API authentication via account

peterjan:


cmd/indexd/run.go
```diff
@@ -140,9 +139,14 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
...
 
+	hostname, _, _ := net.SplitHostPort(cfg.ApplicationAPI.Address)
+	if hostname == "" {
+		hostname = "localhost"
+	}
+
 	web := http.Server{
 		Handler: webRouter{
-			api: jape.BasicAuth(cfg.AdminAPI.Password)(api.NewServer(cm, s, wm, store, apiOpts...)),
+			api: api.Auth(cfg.AdminAPI.Password, hostname, store)(api.NewServer(cm, s, wm, store, apiOpts...)),
```
ChrisSchinnerl: This is not the application API. It's the admin API which should not have signed URL authentication.

cmd/indexd/run.go
```diff
@@ -140,9 +139,14 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
 	}
 	defer pm.Close()
 
+	hostname, _, _ := net.SplitHostPort(cfg.ApplicationAPI.Address)
```
ChrisSchinnerl: The hostname needs to be a separate config value. `ApplicationAPI` is the address that the application API's tcp listener listens on.

api/auth.go
```diff
@@ -0,0 +1,127 @@
...
+				if err != nil {
+					httpWriteError(w, err.Error(), http.StatusUnauthorized)
+					return
+				}
+				hash := requestHash(hostname, ts)
+				if !pk.VerifyHash(hash, sig) {
+					httpWriteError(w, "invalid signature", http.StatusUnauthorized)
+					return
+				}
+				h.ServeHTTP(w, req)
```
ChrisSchinnerl: This is missing something. When we use signed URL authentication, we want to pass the account key on to the handler. So that the handler can then use that.

So application API endpoints will probably look like `func (a *appAPI) handlePinSlab(jc jape.Context, account SomeType) {`  and then you know what account to pass to the store for pinning that slab.

api/auth.go
```diff
@@ -0,0 +1,127 @@
...
+					return
+				}
+
+				ts, err := parseValidUntil(req)
+				if err != nil {
+					httpWriteError(w, err.Error(), http.StatusUnauthorized)
+					return
+				}
+
+				pk, err := parseCredential(req, store)
```
ChrisSchinnerl: Might want to do this last since verifying the request is the cheaper operation compared to db I/O.

api/auth.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// Auth returns a middleware that authenticates requests using either basic auth
+// or signed URLs. If the request is a signed URL, it verifies the signature and
+// checks if the account exists. If the request is not a signed URL, it uses
+// basic auth with the provided password.
+func Auth(password, hostname string, store AccountStore) func(http.Handler) http.Handler {
+	return func(h http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			if isSignedRequest(req) {
```
ChrisSchinnerl: shouldn't be an issue to assume this is always meant to be a signed request and notify the caller about missing fields on the request

cmd/indexd/run.go
peterjan: Perhaps we should consider using the `/admin` prefix and keep the `/api` prefix for the application API?

cmd/indexd/run.go
peterjan: Yeah figured. I had that at one point but figured we might be able to extract it from the `address`.

api/auth.go
peterjan: That's a good use case for `context.WithValue` imo. I'd usually avoid that at all times but in this case it's perfect for it.

api/auth.go
```diff
@@ -0,0 +1,137 @@
...
+			if err != nil {
+				httpWriteError(w, err.Error(), http.StatusInternalServerError)
+				return
+			} else if !known {
+				httpWriteError(w, "unknown account", http.StatusUnauthorized)
+				return
+			}
+
+			// attach the account key to the context
+			ctx := context.WithValue(req.Context(), accountKey, pk)
```
n8mgr: You can get rid of this pretty easily. We aren't really crossing API boundaries to where we *need* the context value and it's definitely preferable to avoid it since it's so brittle. Side benefit of using the jape helpers instead of recreating them.

```go
type authedHandler func(jape.Context, types.PublicKey)

var (
	ErrSignatureExpired = errors.New("signature expired")
	ErrInvalidSignature = errors.New("invalid signature")
	ErrUnknownAccount   = errors.New("unknown account")
	ErrInternalError    = errors.New("internal error")
)

// checkSignedURLAuth returns a middleware that authenticates requests using signed
// URLs. It verifies the signature to ensure the request was signed with the
// given hostname and a valid until timestamp. It will also check if the account
// key exists in the store.
func checkSignedURLAuth(jc jape.Context, hostname string, store AccountStore) (types.PublicKey, bool) {
	req := jc.Request
	// validate presence of required parameters
	if !isSignedRequest(req) {
		err := fmt.Errorf("missing required query parameters: %q, %q, %q", queryParamCredential, queryParamSignature, queryParamValidUntil)
		jc.Error(err, http.StatusUnauthorized)
		return types.PublicKey{}, false
	}

	// extract query string parameters
	ts, err := parseValidUntil(req)
	if err != nil {
		jc.Error(err, http.StatusUnauthorized)
		return types.PublicKey{}, false
	}
	pk, err := parseCredential(req)
	if err != nil {
		jc.Error(err, http.StatusUnauthorized)
		return types.PublicKey{}, false
	}
	sig, err := parseSignature(req)
	if err != nil {
		jc.Error(err, http.StatusUnauthorized)
		return types.PublicKey{}, false
	}

	// check for expiration and verify the signature
	if ts.Before(time.Now().UTC()) {

	} else if !pk.VerifyHash(requestHash(hostname, ts), sig) {
		jc.Error(ErrInvalidSignature, http.StatusUnauthorized)
		return types.PublicKey{}, false
	}

	// check if the account exists
	known, err := store.HasAccount(req.Context(), pk)
	if err != nil {
		jc.Error(ErrInternalError, http.StatusInternalServerError)
		return types.PublicKey{}, false
	} else if !known {
		jc.Error(ErrUnknownAccount, http.StatusUnauthorized)
		return types.PublicKey{}, false
	}
	return pk, true
}

func wrapSignedAPI(hostname string, store AccountStore, wrapped map[string]authedHandler) http.Handler {
	handlers := make(map[string]jape.Handler)
	for path, handler := range wrapped {
		handlers[path] = func(jc jape.Context) {
			pk, ok := checkSignedURLAuth(jc, hostname, store)
			if !ok {
				return
			}
			handler(jc, pk)
		}
	}
	return jape.Mux(handlers)
}

func NewSignedAPI(hostname string, store AccountStore) http.Handler {
	return wrapSignedAPI(hostname, store, map[string]authedHandler{
		"/test": func(jc jape.Context, pk types.PublicKey) {
			jc.Encode(pk)
		},
	})
}
```

api/auth.go
```diff
@@ -0,0 +1,137 @@
...
+// to a known account.
+type AccountStore interface {
+	HasAccount(ctx context.Context, ak types.PublicKey) (bool, error)
+}
+
+// SignedURLAuth returns a middleware that authenticates requests using signed
+// URLs. It verifies the signature to ensure the request was signed with the
+// given hostname and a valid until timestamp. It will also check if the account
+// key exists in the store.
+func SignedURLAuth(hostname string, store AccountStore) func(http.Handler) http.Handler {
```
n8mgr: Mostly an observation, but I'm not sure why we started doing this optional auth middleware with all our apps over having the initializer wrap it.

n8mgr (changes requested):

api/auth.go
```diff
@@ -0,0 +1,137 @@
...
+
+	"go.sia.tech/core/types"
+)
+
+type ctxKey string
+
+const (
+	accountKey ctxKey = "accountKey"
+
+	queryParamCredential = "X-SiaIdx-Credential"
```
n8mgr: `X-` was a prefix for HTTP headers, not query parameters. Looked at the RFC again and found `X-` was deprecated in 2012, so we should probably drop it https://datatracker.ietf.org/doc/html/rfc6648.

api/auth.go
ChrisSchinnerl: My memory might be wrong here but I feel like the jape analyzer didn't like that.

api/auth.go
ChrisSchinnerl: I agree. Definitely would avoid the context here since it just adds one more option for us to miss attaching a value somewhere that we have to check for.

cmd/indexd/run.go
ChrisSchinnerl: They won't use the same TCP port so in theory the admin api can have `/` for the UI and `/api` for the API while the application API uses `/` since it won't serve a UI.

Maybe that's where the confusion comes from. The fact that we don't have the second http server for the application routes yet. The idea being that the admin API listens on `127.0.0.1:9980` and is only reachable locally or via ssh tunnel and the application API is exposed via `:9981` or something. Used ports are up for discussion obviously.

ChrisSchinnerl (changes requested):

api/auth.go
n8mgr: Don't use the jape analyzer? 🤷🏻

peterjan: The jape analyzer catches things quite frequently to be honest. I think, but I have yet to try it, that wrapping the handler the way you propose also makes the routes undetectable, causing the analyzer to error out when checking the client. Checking that now.

api/auth.go
```diff
@@ -0,0 +1,137 @@
...
+			}
+
+			// check if the account exists
+			known, err := store.HasAccount(req.Context(), pk)
+			if err != nil {
+				httpWriteError(w, err.Error(), http.StatusInternalServerError)
+				return
+			} else if !known {
+				httpWriteError(w, "unknown account", http.StatusUnauthorized)
+				return
```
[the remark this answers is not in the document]
peterjan: I was actually a little surprised the both of you didn't comment on this. I think I left it in my PR descr at  one point or a comment. I thought about it and went back and fourth but in our case I think it doesn't really matter?

api/auth.go
peterjan: lol 2012, nice that you double checked! 💪

api/auth.go
n8mgr: A short test for each method on the API client works similarly without requiring us to work around it due to false positives.

api/auth.go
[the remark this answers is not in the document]
n8mgr: Copilot commented on it, so didn’t feel I needed to. I’d probably do a generic error message for all the auth responses, but I don’t think it matters too much since it’s a key pair not something recoverable like an email or phone

api/auth.go
```diff
@@ -0,0 +1,156 @@
...
+		return types.PublicKey{}, false
+	}
+
+	// check for expiration and verify the signature
+	if ts.Before(time.Now().UTC()) {
+		jc.Error(ErrUnauthorized, http.StatusUnauthorized)
+		return types.PublicKey{}, false
+	} else if !pk.VerifyHash(requestHash(hostname, ts), sig) {
+		jc.Error(ErrUnauthorized, http.StatusUnauthorized)
+		return types.PublicKey{}, false
```
ChrisSchinnerl, line 85: There is no harm in telling the user that the request is expired or the signature is invalid. That's information they could easily verify themselves.

api/auth_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	// assert unauthorized if no auth was set
+	status, _ := doRequest(func(req *http.Request) {})
+	if status != http.StatusUnauthorized {
+		t.Fatal("unexpected", status)
+	}
+
+	// assert unauthorized if 'SiaIdx-ValidUntil' is invalid
+	status, errorMsg := doRequest(func(req *http.Request) {
+		val := url.Values{}
```
ChrisSchinnerl: (nit) I'd have a `goodValues` helper and then only swap out the value you want to corrupt to make it super obvious that the other ones are good and safe a few lines of code.

ChrisSchinnerl (changes requested):

api/auth.go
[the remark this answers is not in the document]
ChrisSchinnerl: I'd rather return useful errors than useless obfuscation. There is no information to gain for an attacker just because we return "invalid signature" or "account not found". The upside of being able to debug things is huge though. Similar to how hostd makes things very hard with its "internal error".

The only way for an attacker to figure out if an account exists is by knowing the private key already. Otherwise they get "signature is invalid" which is something they already know if they don't have the private key.

api/auth.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func requestHash(hostname string, validUntil time.Time) types.Hash256 {
+	h := types.NewHasher()
+	h.E.Write([]byte(hostname))
+	h.E.WriteUint64(uint64(validUntil.Unix()))
+	return h.Sum()
+}
+
+func httpWriteError(w http.ResponseWriter, msg string, statusCode int) {
```
n8mgr: Is this still used?

ChrisSchinnerl: Only in one spot within a test. So we can probably get rid of it here and inline it

ChrisSchinnerl (dismissed):

ChrisSchinnerl (approved):

n8mgr (approved):

PR #106, opened 2025-04-08
Add BenchmarkHosts

### LINKED ISSUE #85: Benchmark host queries
Some of the common host queries we need are pretty involved and include CTEs which are often times not very performant. We should keep benchmarks around to make sure they are fast and don't degrade over time.

- [x] Host - repeatedly select a host from a large host table
- [x] Hosts - repeatedly select batches of various sizes (e.g. 10, 100, 1000) from a large host table
- [x] UpdateHost - repeatedly update a host with scans

I'd probably go a bit overboard with the number of hosts in the benchmark and run them against tables with 100K hosts or so.

It might also be good idea to see if we can run the benchmarks with some latency to compare localhost performance with local network performance or slower. e.g. via `tc` or some go library even.

peterjan:
closes https://github.com/SiaFoundation/indexd/issues/85

### DIFF
--- a/persist/postgres/bench_test.go
+++ b/persist/postgres/bench_test.go
@@ -3,14 +3,131 @@ package postgres
 import (
 	"context"
 	"fmt"
+	"net"
 	"strings"
 	"testing"
+	"time"
 
+	proto "go.sia.tech/core/rhp/v4"
 	"go.sia.tech/core/types"
 	"go.sia.tech/indexd/accounts"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest"
+	"lukechampine.com/frand"
 )
+
+// BenchmarkHosts is a set of benchmarks that verify the performance of the host
+// methods in the store that use common table expressions.
+//
+// M1 Max | Host       | 3 ms/op
+// M1 Max | Hosts_10   | 21 ms/op
+// M1 Max | Hosts_100  | 150 ms/op
+// M1 Max | Hosts_1000 | 1.5 s/op
+// M1 Max | UpdateHost | 2 ms/op
+func BenchmarkHosts(b *testing.B) {
+	// define parameters
+	const (
+		numHosts     = 10_000
+		numBlocklist = 1000
+	)
+
+	// prepare test variables
+	networks := []net.IPNet{
+		{IP: net.IPv4(1, 2, 3, 4), Mask: net.CIDRMask(32, 32)},
+		{IP: net.IPv4(2, 3, 4, 5), Mask: net.CIDRMask(32, 32)},
+	}
+
+	// prepare database
+	hosts := make([]types.PublicKey, numHosts)
+	store := initPostgres(b, zap.NewNop())
+	if err := store.transaction(context.Background(), func(ctx context.Context, tx *txn) error {
+		for i := range numHosts {
+			var hostID int64
+			hk := types.GeneratePrivateKey().PublicKey()
+			err := tx.QueryRow(context.Background(), `INSERT INTO hosts (public_key, last_announcement) VALUES ($1, NOW()) RETURNING id;`, sqlPublicKey(hk)).Scan(&hostID)
+			if err != nil {
+				return err
+			}
+			hosts[i] = hk
+
+			_, err = tx.Exec(ctx, `INSERT INTO host_resolved_cidrs (host_id, cidr) VALUES ($1, $2), ($1, $3)`, hostID, networks[0].String(), networks[1].String())
+			if err != nil {
+				return err
+			}
+		}
+
+		// we LEFT JOIN the blocklist so we populate it with random entries
+		for range numBlocklist {
+			_, err := tx.Exec(ctx, "INSERT INTO hosts_blocklist (public_key, reason) VALUES ($1, 'none') ON CONFLICT (public_key) DO NOTHING", sqlPublicKey(types.GeneratePrivateKey().PublicKey()))
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	}); err != nil {
+		b.Fatal(err)
+	}
+
+	b.Run("Host", func(b *testing.B) {
+		var i int
+		for b.Loop() {
+			_, err := store.Host(context.Background(), hosts[i%numHosts])
+			if err != nil {
+				b.Fatal(err)
+			}
+			i++
+		}
+	})
+
+	for _, limit := range []int{10, 100, 1000} {
+		b.Run(fmt.Sprintf("Hosts_%d", limit), func(b *testing.B) {
+			for b.Loop() {
+				offset := frand.Intn(numHosts - limit)
+				hosts, err := store.Hosts(context.Background(), offset, limit)
+				if err != nil {
+					b.Fatal(err)
+				} else if len(hosts) != limit {
+					b.Fatalf("expected %d hosts, got %d", limit, len(hosts)) // sanity check
+				}
+			}
+		})
+	}
+
+	b.Run("UpdateHost", func(b *testing.B) {
+		ts := time.Now()
+		hs := proto.HostSettings{
+			ProtocolVersion:     [3]uint8{1, 2, 3},
+			Release:             b.Name(),
+			WalletAddress:       types.Address{1},
+			AcceptingContracts:  true,
+			MaxCollateral:       types.NewCurrency64(frand.Uint64n(1e6)),
+			MaxContractDuration: frand.Uint64n(1e6),
+			RemainingStorage:    frand.Uint64n(1e6),
+			TotalStorage:        frand.Uint64n(1e6),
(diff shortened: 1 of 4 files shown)


persist/postgres/bench_test.go
```diff
@@ -3,15 +3,132 @@ package postgres
...
+	"lukechampine.com/frand"
 )
 
+// BenchmarkHosts is a set of benchmarks that verify the performance of the host
+// methods in the store that use common table expressions.
+//
+// M1 Max | Host       | 2.7ms/op
+// M1 Max | Hosts_10   | 21ms/op
+// M1 Max | Hosts_100  | 150ms/op
+// M1 Max | Hosts_1000 | 1.5s/op
```
peterjan: @ChrisSchinnerl  damn 👀 and this is only for 10k hosts, if I do 100k (as the issue suggests) it's just unworkably slow (~8s). I'll look into optimizing the query tomorrow

persist/postgres/hosts.go
```diff
@@ -485,48 +476,96 @@ WHERE hosts.id = computed.id RETURNING hosts.id`,
...
+			args = append(args, hosts[i].id)
+			parts = append(parts, fmt.Sprintf("$%d", i+1))
 		}
-		addresses = append(addresses, na)
-	}
+		placeHolders := fmt.Sprintf("(%s)", strings.Join(parts, ","))
 
-	if err := rows.Err(); err != nil {
-		return nil, err
+		rows, err := tx.Query(ctx, `SELECT host_id, net_address, protocol FROM host_addresses WHERE host_id IN `+placeHolders, args...)
```
n8mgr: You should be able to use a slice natively here. You probably need to look at the pgx documentation for how to actually do it, but this might work:

```sql
SELECT host_id, net_address, protocol 
FROM host_addresses 
WHERE host_id = ANY($1)
```

Unlike SQLite, string concatenation for slices is mostly unnecessary with the Postgres driver and should be avoided when possible.

peterjan: Postgres is sweet

ChrisSchinnerl (dismissed):

persist/postgres/hosts.go
```diff
@@ -485,48 +475,88 @@ WHERE hosts.id = computed.id RETURNING hosts.id`,
...
 	}
-	defer rows.Close()
 
-	var addresses []chain.NetAddress
-	for rows.Next() {
-		var na chain.NetAddress
-		if err := rows.Scan(&na.Address, (*sqlNetworkProtocol)(&na.Protocol)); err != nil {
-			return nil, fmt.Errorf("failed to scan host address: %w", err)
+	const batchSize = 1000
+	for len(hosts) > 0 {
```
n8mgr: I don't think we should worry about batching in these decorate methods. If we're planning on fetching more than 1000 hosts in a single call, we should refactor that. Setting a limit somewhere between 250-500 per call to `Hosts` and returning an error if it goes over would be very reasonable, imo.

persist/postgres/hosts.go
```diff
@@ -485,48 +475,88 @@ WHERE hosts.id = computed.id RETURNING hosts.id`,
...
+		batch := min(len(hosts), batchSize)
+
+		var hostIds []any
```
n8mgr: nit: prefer `[]int64` over `[]any`

persist/postgres/hosts.go
peterjan: It's too big of a speedup not to imo.

This is the result with un-batching only networks, un-batching host addresses is another 2x iirc
```
| Benchmark                     | N   | Time per op     |
|------------------------------|-----|------------------|
| BenchmarkHosts/Host-10       | 692 | 1.85 ms          |
| BenchmarkHosts/Hosts_10-10   | 100 | 11.07 ms         |
| BenchmarkHosts/Hosts_100-10  |  24 | 48.83 ms         |
| BenchmarkHosts/Hosts_1000-10 |   3 | 393.76 ms        |
```

With batching it scales so well

```
| Benchmark                     | N   | Time per op     |
|------------------------------|-----|------------------|
| BenchmarkHosts/Host-10       | 692 | 1.85 ms          |
| BenchmarkHosts/Hosts_10-10   | 100 | 7.1 ms         |
| BenchmarkHosts/Hosts_100-10  |  24 | 10.5 ms         |
| BenchmarkHosts/Hosts_1000-10 |   3 | 13 ms        |
```

I didn't like the placeholder string, but now that I know we can use `id = ANY($1)` .. I don't really think it's that big of a deal now.

ChrisSchinnerl: I also don't mind the batching as much with the new syntax but I am surprised about how bad it scales. 100 hosts is not that many and going from 10.5ms with batching to 48ms is quite the hit.

peterjan: Yeah that surprised me as well. Perhaps it has to do with the fact that it's (potentially) querying multiple rows per host and the cursors are expensive. Do we want to dig deeper on that? I'll leave it open for a bit, see if Nate is happy to leave in the batching.

ChrisSchinnerl: I'd maybe check the query plan and see if it changes drastically between queries with multiple hosts and single ones. Happened to me recently when picking a larger limit for a query.

peterjan: I can't really check the query plan for that right? The difference between batching and not batching is that we fetch the hosts first and then decorate the host networks and addresses in a loop separately. Both tables have the appropriate index on `host_id` so the difference must come from fetching both cursors per host `n` times...

n8mgr: When I said remove batching I meant drop the `for len(hosts) > 0`. That shouldn’t alter the benchmarks significantly.

n8mgr: Effectively this:

```go
func decorateHostAddresses(ctx context.Context, tx *txn, hosts ...*dbHost) {
	hostIDs := make([]int64, 0, len(hosts))
	idToIdx := make(map[int64]int64, len(hosts))
	for i := range hosts {
		idToIdx[hosts[i].id] = int64(i)
		hostIDs = append(hostIDs, hosts[i].id)
	}

	rows, err := tx.Query(ctx, `SELECT host_id, net_address, protocol FROM host_addresses WHERE host_id = ANY($1)`, hostIDs)
	if err != nil {
		return fmt.Errorf("failed to query host addresses: %w", err)
	}
	defer rows.Close()

	for rows.Next() {
		var hostID int64
		var na chain.NetAddress
		if err := rows.Scan(&hostID, &na.Address, (*sqlNetworkProtocol)(&na.Protocol)); err != nil {
			return fmt.Errorf("failed to scan host address: %w", err)
		}
		hosts[idToIdx[hostID]].Addresses = append(hosts[idToIdx[hostID]].Addresses, na)
	}
	if err := rows.Err(); err != nil {
		return err
	}
	return nil
}
```

```
goos: darwin
goarch: arm64
cpu: Apple M4 Max
BenchmarkHosts/Host-16         	    2302	    541562 ns/op	   11677 B/op	     190 allocs/op
BenchmarkHosts/Hosts_10-16     	     382	   3519796 ns/op	   68563 B/op	    1090 allocs/op
BenchmarkHosts/Hosts_100-16    	     320	   3622489 ns/op	  589027 B/op	    9905 allocs/op
BenchmarkHosts/Hosts_1000-16   	     172	   6948900 ns/op	 6132914 B/op	   98062 allocs/op
BenchmarkHosts/UpdateHost-16   	    2088	    571803 ns/op	    3496 B/op	     149 allocs/op
```

Since we're hitting a networked DB, each query has latency and data transfer costs. We shouldn't be pulling more than a few hundred records at a time unless there's a need, especially in hot paths.

If we're not planning to support >1000 hosts in a single call, the outer `for` loop and slicing adds complexity without real benefit. If that changes down the line, we can reintroduce batching as needed.

persist/postgres/hosts.go
```diff
@@ -205,31 +202,24 @@ WHERE
...
+			dbHosts = append(dbHosts, &host)
 		}
 		if err := rows.Err(); err != nil {
 			return err
 		} else if len(dbHosts) == 0 {
 			return nil
 		}
 
-		// decorate host addresses and networks
+		decorateHostAddresses(ctx, tx, dbHosts...)
```
n8mgr: missing error check

persist/postgres/hosts.go
```diff
@@ -205,31 +202,24 @@ WHERE
...
+		decorateHostNetworks(ctx, tx, dbHosts...)
```
n8mgr: ditto

n8mgr (changes requested):

persist/postgres/hosts.go
peterjan: makes sense!

ChrisSchinnerl (approved):

n8mgr (approved):

PR #107, opened 2025-04-08
Add Slab pinning to database

ChrisSchinnerl:


persist/postgres/sectors.go
```diff
@@ -0,0 +1,209 @@
...
+	err = tx.QueryRow(ctx, `
+		INSERT INTO slabs (account_id, digest, encryption_key, min_shards)
+		VALUES ($1, $2, $3, $4)
+		RETURNING id
+		`, accountID, sqlHash256(digest), sqlHash256(slab.EncryptionKey), slab.MinShards).Scan(&slabID)
+	if err != nil {
+		return SlabID{}, err
+	}
+
+	// insert slab's sectors in a single batch
```
ChrisSchinnerl: Batching sectors was quite the speedup. I also tried batching both slabs and sectors but the speedup was minor by comparison at vastly more complex code.

Using a CTE I was also able to combine the slab insert and the sector batch insert into a single query. Which was also just slightly faster.

doc/rfc/001_db_schema.md
```diff
@@ -248,33 +248,36 @@ CREATE INDEX account_hosts_host_id_next_fund_idx ON account_hosts (host_id, next
 CREATE TABLE slabs (
     id BIGSERIAL PRIMARY KEY, -- internal db id
 
-    digest BYTEA UNIQUE NOT NULL, -- unique identifier for the slab derived from sector roots
+    account_id SERIAL REFERENCES accounts(id), -- account that owns slab
```
n8mgr: Since we want to support sharing, would it be better to deduplicate with an `account_slabs` table? It is an extra join, but probably worth not also duplicating the slab sectors for every account.

doc/rfc/001_db_schema.md
```diff
@@ -248,33 +248,36 @@ CREATE INDEX account_hosts_host_id_next_fund_idx ON account_hosts (host_id, next
...
+    digest BYTEA NOT NULL, -- unique identifier for the slab derived from sector roots
+    UNIQUE(account_id, digest), -- deduplicate slabs per account
+
     encryption_key BYTEA NOT NULL,
     last_repair_attempt TIMESTAMP WITH TIME ZONE,
     min_shards SMALLINT NOT NULL CHECK(min_shards > 0)
-)
+);
+CREATE INDEX slabs_digest ON slabs(digest);
```
peterjan: `slabs_digest_idx`

persist/postgres/init.sql
```diff
@@ -174,3 +174,39 @@ CREATE TABLE contract_elements (
...
+
+    account_id SERIAL REFERENCES accounts(id), -- account that owns slab
+    digest BYTEA NOT NULL, -- unique identifier for the slab derived from sector roots
+    UNIQUE(account_id, digest), -- deduplicate slabs per account
+
+    encryption_key BYTEA NOT NULL,
+    last_repair_attempt TIMESTAMP WITH TIME ZONE,
+    min_shards SMALLINT NOT NULL CHECK(min_shards > 0)
+);
+CREATE INDEX slabs_digest ON slabs(digest);
```
peterjan: `slabs_digest_idx`

persist/postgres/init.sql
```diff
@@ -174,3 +174,39 @@ CREATE TABLE contract_elements (
     leaf_index INTEGER NOT NULL,
     merkle_proof BYTEA NOT NULL
 );
+
+CREATE TABLE slabs (
+    id BIGSERIAL PRIMARY KEY, -- internal db id
+
+    account_id SERIAL REFERENCES accounts(id), -- account that owns slab
+    digest BYTEA NOT NULL, -- unique identifier for the slab derived from sector roots
```
peterjan: I assume the `digest` is always a 32-byte hash? Could add length check here

persist/postgres/init.sql
```diff
@@ -174,3 +174,39 @@ CREATE TABLE contract_elements (
...
+    host_id INTEGER REFERENCES hosts(id), -- host that stores sector
+    contract_id INTEGER REFERENCES contracts(id) DEFAULT NULL, -- null if not pinned
+    uploaded_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(), -- allow sorting by upload time
+
+    -- slab
+    slab_id BIGINT REFERENCES slabs(id) NOT NULL,
+    slab_index SMALLINT NOT NULL, -- index within corresponding slab to retrieve sectors in right order
+
+    -- data integrity
+    next_integrity_check TIMESTAMP WITH TIME ZONE,
```
peterjan, line 209: We could easily make this `NOT NULL`, it's nice to not have to care when writing queries.

persist/postgres/sectors.go
```diff
@@ -0,0 +1,209 @@
...
+	// ErrSlabExists is returned when a slab already exists in the database.
+	ErrSlabExists = errors.New("slab already exists")
+
+	// ErrSlabNotFound is returned when a slab is not found in the database.
+	ErrSlabNotFound = errors.New("slab not found")
+)
+
+type (
+	// SlabID is the ID of a slab derived from the slab's sectors' roots.
+	SlabID types.Hash256
```
peterjan: We might as well call it `SlabDigest` here too?

persist/postgres/sectors.go
```diff
@@ -0,0 +1,209 @@
...
+		`, accountID, sqlHash256(digest), sqlHash256(slab.EncryptionKey), slab.MinShards).Scan(&slabID)
+	if err != nil {
+		return SlabID{}, err
+	}
+
+	// insert slab's sectors in a single batch
+	var placeholders []string
+	var args []any
+	for i, sector := range slab.Sectors {
+		placeholders = append(placeholders, fmt.Sprintf("($%d, (SELECT id FROM hosts WHERE public_key = $%d), $%d, $%d)", i*4+1, i*4+2, i*4+3, i*4+4))
```
peterjan: Wondering whether fetching host IDs separately would speed it up even more.

peterjan (commented):
🔥 left some small remarks but looking pretty solid

doc/rfc/001_db_schema.md
ChrisSchinnerl: Yeah definitely considered that. I think there is (or was) a line about deduplication in the RFC but since the RFC didn't have the join table I thought maybe we actively decided against it. Might have been the host_sectors table though.

I can change that since it shouldn't affect the actual health code which is what we were most worried about if I remember correctly.

ChrisSchinnerl: Oh wait while typing it up I noticed something. What happens if a user pins a slab that is already pinned by a different user using different hosts for the sectors?

If we overwrite the sectors, then one user gets to overwrite the sectors another user picked.
If we don't, then a user doesn't get to choose the hosts they uploaded to which might also be odd.
e.g. users might pick specific hosts because they are close or something.

persist/postgres/init.sql
ChrisSchinnerl: I'm leaving that up to the PR that adds the queries. Slapping a `DEFAULT NOW() + X` here right now will probably end up duplicating logic.

persist/postgres/sectors.go
ChrisSchinnerl: We could but this will be the identifier for slabs outside the database code and I think calling it `SlabID` avoids confusion about how to obtain it.

persist/postgres/sectors.go
ChrisSchinnerl: It's negligible. When fetching them one-by-one and then caching them in a map it's actually significantly slower.
When fetching them all at once and then caching them it's slower for the single slab case and about the same for the others.
Pretty sure that within the same transaction, Postgres just caches the pages of the host table it has already accessed.

peterjan (dismissed):
good to merge imo, and F/U with https://github.com/SiaFoundation/indexd/pull/107/files#r2036881027

doc/rfc/001_db_schema.md
n8mgr: Once we repair user host selection is moot anyway, so we should go ahead and be more efficient

persist/postgres/sectors.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	values := strings.Join(placeholders, ",")
+
+	_, err = tx.Exec(ctx, fmt.Sprintf(`
+		INSERT INTO sectors (sector_root, host_id, slab_id, slab_index)
+		VALUES %s
+	`, values), args...)
+	if err != nil {
+		return SlabID{}, fmt.Errorf("failed to insert sectors: %w", err)
+	}
```
n8mgr: You can use `pgx.Batch` to do this without the string manipulation. I'd check the pgx docs whenever you think string concatenation or manipulation is necessary. There are a bunch of neat helpers in there.

```suggestion
	// insert slab's sectors in a single batch
	batch := &pgx.Batch{}
	for i, sector := range slab.Sectors {
		batch.Queue(`INSERT INTO sectors (sector_root, host_id, slab_id, slab_index) VALUES ($1, (SELECT id FROM hosts WHERE public_key = $2), $3, $4)`, sqlHash256(sector.Root), sqlPublicKey(sector.HostKey), slabID, i)
	}
	if err = tx.Tx.SendBatch(ctx, batch).Close(); err != nil {
		return SlabID{}, fmt.Errorf("failed to insert sectors: %w", err)
	}
```

```
BenchmarkSlabs/PinSlabs-40MiB-16         	    1016	   1158407 ns/op	36207.52 MB/s	   42679 B/op	     494 allocs/op
BenchmarkSlabs/PinSlabs-400MiB-16        	     175	   6775070 ns/op	6190.79 MB/s	  406659 B/op	    4666 allocs/op
BenchmarkSlabs/PinSlabs-4GiB-16          	      15	  72486375 ns/op	 578.63 MB/s	 4039413 B/op	   46346 allocs/op
BenchmarkSlabs/Slabs-40MiB-16            	    1930	    599919 ns/op	69914.48 MB/s	   17469 B/op	     358 allocs/op
BenchmarkSlabs/Slabs-400MiB-16           	     417	   2794410 ns/op	15009.62 MB/s	  159717 B/op	    3427 allocs/op
BenchmarkSlabs/Slabs-4GiB-16             	      46	  26040158 ns/op	1610.71 MB/s	 1581744 B/op	   34120 allocs/op
```

n8mgr (changes requested):

persist/postgres/sectors.go
```diff
@@ -0,0 +1,234 @@
...
+	"errors"
+	"fmt"
+	"strings"
+
+	proto "go.sia.tech/core/rhp/v4"
+	"go.sia.tech/core/types"
+	"go.sia.tech/indexd/accounts"
+)
+
+var (
```
n8mgr: None of this should realistically be in the store impl.

ChrisSchinnerl: Yeah the next PR (the one with the slab manager) already moves this. Somehow my brain felt like this might just be easier to look at in the review as long as we don't need it in the other package 😅

persist/postgres/sectors.go
ChrisSchinnerl: Damn this is the game changer of game changers

n8mgr (approved):

PR #109, opened 2025-04-11
Add slab manager

ChrisSchinnerl:
This PR outlines the slab manager and hooks it up to the `PinSlabs` and `Slabs` store methods


slabs/manager.go
```diff
@@ -0,0 +1,111 @@
...
+			}
+		}
+	}()
+
+	return m, nil
+}
+
+// Close closes the manager.
+func (m *SlabManager) Close() error {
+	m.tg.Stop()
```
peterjan, line 91: Fine for now but we'll probably need a way to cleanly exit out of integrity checks and slab migrations if this is called. The way it's implemented now this is blocking on both those actions which might take a really long time obviously.

slabs/manager.go
```diff
@@ -0,0 +1,111 @@
...
+	// SlabManager is responsible for managing slabs, including pinning them,
+	// checking their integrity on the network and migrating their sectors if
+	// necessary.
+	SlabManager struct {
+		store Store
+		tg    *threadgroup.ThreadGroup
+		log   *zap.Logger
+	}
+
+	// Store defines an interface to fetch hosts that need to be scanned and
```
peterjan: copy pasta

slabs/manager.go
```diff
@@ -0,0 +1,111 @@
...
+type Option func(*SlabManager)
+
+// WithLogger sets the logger for the SlabManager.
+func WithLogger(l *zap.Logger) Option {
+	return func(m *SlabManager) {
+		m.log = l
+	}
+}
+
+// NewManager creates a new host manager.
```
peterjan: copy pasta

slabs/manager.go
ChrisSchinnerl: Yeah `performIntegrityChecks` and `performSlabMigrations` will need to interrupt whatever they are doing when we implement them.

slabs/slabs.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+)
+
+// String implements the Stringer interface for SlabID.
+func (s SlabID) String() string {
+	return types.Hash256(s).String()
+}
+
+// PinSlabs pins the given slabs and associates them with the given account.
+func (m *SlabManager) PinSlabs(ctx context.Context, account proto.Account, slabs []SlabPinParams) ([]SlabID, error) {
```
n8mgr: Should we ever expect to pin more than a single slab at a time? Optimal uploads would probably look like uploading a few slabs in parallel then each thread calls `POST /api/pin` when complete. You could also do resumable uploads by deriving a key, calculating the slab ID ahead of time, and skipping existing slabs. From a repair perspective, we're always working with a single slab.

ChrisSchinnerl: My guess was that as a dev you might appreciate pinning your uploads to be atomic for a batch of slabs to avoid the cleanup when you abort the upload.

peterjan: I'm siding with Nate on this one. Definitely see your point too but doing one slab at a time feels cleaner. Aborting (and potentially at one point even resuming uploads) feels like something we can always add later and when we do we can add stuff on top to improve the developer experience.

ChrisSchinnerl: We are not going to introduce the concept of uploads to the indexer  but I guess it doesn't matter. If we do individual slabs, clients just need to track the state of each one in their database rather than the group.

ChrisSchinnerl: If you also want to get rid of the array in the database code just let me know and I'll f/u with it. That would help me avoid conflicts on dependent PRs.

n8mgr: The db code should probably match the manager method

peterjan (approved):

slabs/slabs.go
ChrisSchinnerl: Ok will add a dedicated PR for that asap.

PR #110, opened 2025-04-14
Add SectorsForIntegrityCheck to database

ChrisSchinnerl:
For fetching the sectors for a given host to perform integrity checks on

### DIFF
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -5,98 +5,28 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
-	"strings"
 
+	"github.com/jackc/pgx/v5"
 	proto "go.sia.tech/core/rhp/v4"
 	"go.sia.tech/core/types"
 	"go.sia.tech/indexd/accounts"
+	"go.sia.tech/indexd/slabs"
 )
-
-var (
-	// ErrSlabExists is returned when a slab already exists in the database.
-	ErrSlabExists = errors.New("slab already exists")
-
-	// ErrSlabNotFound is returned when a slab is not found in the database.
-	ErrSlabNotFound = errors.New("slab not found")
-)
-
-type (
-	// SlabID is the ID of a slab derived from the slab's sectors' roots.
-	SlabID types.Hash256
-
-	// Sector is a 4MiB sector stored on a host.
-	Sector struct {
-		// Root is the sector root of a 4MiB sector
-		Root types.Hash256 `json:"root"`
-
-		// ContractID is the ID of the contract the sector is pinned to.
-		// 'nil' if the sector isn't pinned.
-		ContractID *types.FileContractID `json:"contractID,omitempty"`
-
-		// HostKey is the public key of the host that stores the sector data.
-		// 'nil' if a host lost the sector and the sector requires migration
-		HostKey *types.PublicKey `json:"host,omitempty"`
-	}
-
-	// Slab is a group of sectors that is encrypted, erasure-coded and uploaded
-	// to hosts.
-	Slab struct {
-		ID            SlabID   `json:"id"`
-		EncryptionKey [32]byte `json:"encryptionKey"`
-		MinShards     uint     `json:"minShards"`
-		Sectors       []Sector `json:"sectors"`
-	}
-
-	// SectorPinParams describes an uploaded sector to be pinned.
-	SectorPinParams struct {
-		Root    types.Hash256   `json:"root"`
-		HostKey types.PublicKey `json:"hostKey"`
-	}
-
-	// SlabPinParams is the input to PinSlabs
-	SlabPinParams struct {
-		EncryptionKey [32]byte          `json:"encryptionKey"`
-		MinShards     uint              `json:"minShards"`
-		Sectors       []SectorPinParams `json:"sectors"`
-	}
-)
-
-// String implements the Stringer interface for SlabID.
-func (s SlabID) String() string {
-	return types.Hash256(s).String()
-}
-
-// Digest creates a unique digest for the slab to be pinned by SlabPinParams. It
-// is important, that the same params always result in the same hash since we
-// deduplicate slabs using it. So if one user makes the mistake of pinning a
-// slab with a different encryption key, this shouldn't prevent other users from
-// pinning the same slab with the correct key.
-func (s SlabPinParams) Digest() (SlabID, error) {
-	hasher := types.NewHasher()
-	hasher.E.WriteUint64(uint64(s.MinShards))
-	hasher.E.Write(s.EncryptionKey[:])
-	for _, sector := range s.Sectors {
-		if _, err := hasher.E.Write(sector.Root[:]); err != nil {
-			return SlabID{}, fmt.Errorf("failed to write sector root to hasher: %w", err)
-		}
-	}
-	return SlabID(hasher.Sum()), nil
-}
 
 // PinSlabs adds slabs to the database for pinning. The slabs are associated
 // with the provided account.
-func (s *Store) PinSlabs(ctx context.Context, account proto.Account, slabs []SlabPinParams) ([]SlabID, error) {
-	if len(slabs) == 0 {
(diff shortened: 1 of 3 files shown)


persist/postgres/sectors.go
```diff
@@ -6,16 +6,50 @@ import (
...
 
+// SectorsForIntegrityCheck returns up to `limit` sectors that are due for an
+// integrity check.
+func (s *Store) SectorsForIntegrityCheck(ctx context.Context, hostKey types.PublicKey, limit int) ([]types.Hash256, error) {
+	var sectors []types.Hash256
+	err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		rows, err := tx.Query(ctx, `
+			WITH hid AS (
+				SELECT id FROM hosts WHERE public_key = $1
+			)
```
ChrisSchinnerl: This was quite surprising. I had an inner join here but that was only fast for a limit of `10` and `100` but `1000` suddenly took 50ms or so.

Looking at the query plan, they were the same for smaller limits but larger ones suddenly started spinning up 2 workers and then hash joining and sorting the results. Leading to a huge performance loss.

So then I tried 2 queries to fetch the host id first and suddenly large limits were fast as well while overall the performance was slightly lower due to the additional roundtrip so the CTE sped things up a bit more.

It seems another big gain came from this query using the composite index on host_id and next_integrity_check now rather than using the host_id index on the join and then the next_integrity_check index on the materialized result.

doc/rfc/001_db_schema.md
```diff
@@ -242,16 +242,20 @@ CREATE INDEX account_hosts_host_id_next_fund_idx ON account_hosts (host_id, next
...
     encryption_key BYTEA NOT NULL,
     last_repair_attempt TIMESTAMP WITH TIME ZONE,
     min_shards SMALLINT NOT NULL CHECK(min_shards > 0)
 );
 CREATE INDEX slabs_digest_idx ON slabs(digest);
 
+CREATE TABLE account_slabs (
+    account_id INTEGER REFERENCES accounts(id) NOT NULL, -- account that owns slab
+    slab_id BIGSERIAL REFERENCES slabs(id) NOT NULL,
+    CONSTRAINT account_slabs_pk PRIMARY KEY (account_id, slab_id)
```
n8mgr: not: this can be simplified to `PRIMARY KEY (account_id, slab_id)`

persist/postgres/init.sql
```diff
@@ -178,16 +178,20 @@ CREATE TABLE contract_elements (
...
     encryption_key BYTEA NOT NULL,
     last_repair_attempt TIMESTAMP WITH TIME ZONE,
     min_shards SMALLINT NOT NULL CHECK(min_shards > 0)
 );
 CREATE INDEX slabs_digest_idx ON slabs(digest);
 
+CREATE TABLE account_slabs (
+    account_id INTEGER REFERENCES accounts(id) NOT NULL, -- account that owns slab
+    slab_id BIGSERIAL REFERENCES slabs(id) NOT NULL,
+    CONSTRAINT account_slabs_pk PRIMARY KEY (account_id, slab_id)
```
n8mgr: nit: this can be simplified to `PRIMARY KEY (account_id, slab_id)`

n8mgr (approved):

persist/postgres/sectors.go
```diff
@@ -104,48 +139,55 @@ func (s *Store) Slabs(ctx context.Context, accountID proto.Account, slabIDs []sl
...
-		INSERT INTO slabs (account_id, digest, encryption_key, min_shards)
-		VALUES ($1, $2, $3, $4)
+		INSERT INTO slabs (digest, encryption_key, min_shards)
+		VALUES ($1, $2, $3)
+		ON CONFLICT (digest) DO NOTHING
 		RETURNING id
-		`, accountID, sqlHash256(digest), sqlHash256(slab.EncryptionKey), slab.MinShards).Scan(&slabID)
+		`, sqlHash256(digest), sqlHash256(slab.EncryptionKey), slab.MinShards).Scan(&slabID)
+	if errors.Is(err, sql.ErrNoRows) {
+		// slab already exists, fetch it's slab id
```
peterjan:
```suggestion
		// slab already exists, fetch its slab id
```

persist/postgres/sectors.go
```diff
@@ -104,48 +139,55 @@ func (s *Store) Slabs(ctx context.Context, accountID proto.Account, slabIDs []sl
...
 	err = tx.QueryRow(ctx, `
-		INSERT INTO slabs (account_id, digest, encryption_key, min_shards)
-		VALUES ($1, $2, $3, $4)
+		INSERT INTO slabs (digest, encryption_key, min_shards)
+		VALUES ($1, $2, $3)
+		ON CONFLICT (digest) DO NOTHING
 		RETURNING id
-		`, accountID, sqlHash256(digest), sqlHash256(slab.EncryptionKey), slab.MinShards).Scan(&slabID)
+		`, sqlHash256(digest), sqlHash256(slab.EncryptionKey), slab.MinShards).Scan(&slabID)
+	if errors.Is(err, sql.ErrNoRows) {
```
peterjan: you could use a CTE to combine the insert and fetch

```go
WITH insert AS (
    INSERT INTO slabs (digest, encryption_key, min_shards)
    VALUES ($1, $2, $3)
    ON CONFLICT (digest) DO NOTHING
    RETURNING id
)
SELECT id FROM insert
UNION ALL
SELECT id FROM slabs WHERE digest = $1
LIMIT 1;
```

persist/postgres/sectors.go
```diff
@@ -104,48 +139,55 @@ func (s *Store) Slabs(ctx context.Context, accountID proto.Account, slabIDs []sl
...
+	_, err = tx.Exec(ctx, `
+		INSERT INTO account_slabs (account_id, slab_id) VALUES ($1, $2)
+		ON CONFLICT (account_id, slab_id) DO NOTHING
+	`, accountID, slabID)
+	if err != nil {
+		return slabs.SlabID{}, fmt.Errorf("failed to insert slab into account_slabs: %w", err)
+	}
+
+	// if the slab already existed, we don't need to insert the sectors
+	if existingSlab {
```
peterjan: Oh I guess you'd lose out on this info if you combine the insert and fetch.
Don't we ideally want to overwrite the sectors here instead of skipping?

persist/postgres/sectors_test.go
```diff
@@ -269,3 +393,81 @@ func BenchmarkSlabs(b *testing.B) {
 		runSlabsBenchmark(b, 100)
 	})
 }
+
+// BenchmarkSectorsForIntegrityCheck benchmarks SectorsForIntegrityCheck
+//
+//	CPU  | BatchSize |	  Count  |     Time/op     |   Throughput
+//
+// M2 Pro |    10 	  |   2857   |    0.380024 ms  |  110369.50 MB/s
```
peterjan: `nit` reformat

slabs/slabs.go
```diff
@@ -62,9 +61,26 @@ func (s SlabID) String() string {
 	return types.Hash256(s).String()
 }
 
+// Digest creates a unique digest for the slab to be pinned by SlabPinParams. It
+// is important, that the same params always result in the same hash since we
+// deduplicate slabs using it. So if one user makes the mistake of pinning a
+// slab with a different encryption key, this shouldn't prevent other users from
+// pinning the same slab with the correct key.
+func (s SlabPinParams) Digest() (SlabID, error) {
```
peterjan: `nit` write a small unit test?

persist/postgres/sectors.go
ChrisSchinnerl: Yeah I considered that but I didn't want to be too fancy considering I'd also need to find a new solution for figuring out whether a slab already exists.

slabs/slabs.go
ChrisSchinnerl: The test `TestPinSlabs` already covers that by manually building the ID but I can extract that into a dedicated test

persist/postgres/sectors.go
ChrisSchinnerl: Overwriting them means that you can effectively wipe the sectors someone else pinned with bad ones. If we already have sector info that we repair and integrity check, it's safer to keep that.

PR #111, opened 2025-04-14
Add alerts manager

peterjan:


alerts/alerts.go
```diff
@@ -0,0 +1,220 @@
...
+	// A Manager manages the indexer's alerts.
+	Manager struct {
+		log *zap.Logger
+
+		mu     sync.Mutex
+		alerts map[types.Hash256]Alert
+	}
+
+	// ManagerOption is a functional option for the alert manager.
+	ManagerOption func(*Manager) error
```
n8mgr: No need to return an error. I'm also not sure adding a log is worth it. The debug logs aren't really helpful.

n8mgr (changes requested):

alerts/alerts.go
peterjan: I kind of liked the debug logs because we will have fewer alerts now and we still keep them in memory. Debug logging them would allow us to go back in time in case we want to for debugging purposes.

ChrisSchinnerl (approved):

n8mgr (approved):

PR #114, opened 2025-04-15
Add ContractsForFunding to database

peterjan:


persist/postgres/contracts.go
```diff
@@ -118,9 +118,35 @@ func (s *Store) Contracts(ctx context.Context, queryOpts ...contracts.ContractQu
 
 // ContractsForFunding returns a list of contracts for the given host key that
 // are good for funding ephemeral accounts with. The contracts are sorted by the
-// remaining allowance in descending fashion.
+// remaining allowance in descending fashion. The number of contracts returned
+// is limited to 50.
```
peterjan: I considered limiting it to 10.

ChrisSchinnerl (dismissed):

persist/postgres/init.sql
```diff
@@ -166,6 +166,7 @@ CREATE TABLE contracts (
   sector_roots_spending DECIMAL(50, 0) NOT NULL DEFAULT 0
 );
 CREATE INDEX contracts_state_formation_idx ON contracts(state, formation); -- for rejecting expired contracts
+CREATE INDEX contracts_host_id_remaining_allowance_idx ON contracts (host_id, remaining_allowance DESC) WHERE good = true AND remaining_allowance > 0 AND state <= 1; -- for fetching contracts for funding
```
peterjan, line 170: Several places where we can benefit from a partial index. Perhaps we should add a new rule or convention where we suffix `parial_idx` instead of `_idx`

persist/postgres/contracts.go
n8mgr: nit: should pass in limit and let the caller decide to be consistent with the other store methods.

n8mgr (dismissed):

contracts/fund_test.go
```diff
@@ -46,6 +46,9 @@ func (s *storeMock) ContractsForFunding(_ context.Context, hk types.PublicKey) (
 	for i, c := range contracts {
 		out[i] = c.ID
 	}
+	if len(out) > limit {
+		out = out[:limit]
+	}
```
ChrisSchinnerl, line 54: This method is a recipe for NDFs since we are not deep copying the contracts initially  😅

n8mgr (dismissed):

ChrisSchinnerl (dismissed):

ChrisSchinnerl (approved):

PR #115, opened 2025-04-15
Add store.Contracts

peterjan:
This PR implements the `Contracts` method on the store. In the process I found that it's really quite expensive to get the `renewed_from` and `renewed_to` values when hydrating a `Contract`. I adjusted the schema to reference the `contract_id` directly and ended up throwing away quite some complexity. This also eliminates some joins in `sectors` and `contract_elements` table. I look forward to all the comments 😅 personally I'm starting to lean more towards using our unique keys as primary keys


persist/postgres/bench_test.go
```diff
@@ -8,10 +8,109 @@ import (
...
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest"
+	"lukechampine.com/frand"
 )
 
+// BenchmarkContracts is a benchmark to ensure the performance of
+// Contract and Contracts.
+//
+// M1 Max | Contract | 1.1ms/op
+// M1 Max | Contracts (default) | 188ms/op
```
peterjan: still looking into these, `renewed_from` and `renewed_to` are quite annoying

contracts/renew.go
```diff
@@ -50,7 +51,7 @@ func (cf *contractor) RenewContract(ctx context.Context, hk types.PublicKey, add
 
 func (cm *ContractManager) performContractRenewals(ctx context.Context, renewWindow uint64, log *zap.Logger) error {
 	renewalLog := log.Named("renewal")
-	contracts, err := cm.store.Contracts(ctx, WithGood(true), WithRevisable(true))
+	contracts, err := cm.store.Contracts(ctx, 0, math.MaxInt64, WithGood(true), WithRevisable(true)) // TODO: page through contracts and refresh in parallell
```
peterjan, line 54: I plan to F/U with handling `offset, limit` pending the approval of this PR.

contracts/contract.go
```diff
@@ -25,40 +25,28 @@ var (
...
-	ContractQueryOpt func(*contractQueryOpts)
-)
-
-var (
-	optTrue = true
-
-	// DefaultContractQueryOpts are the default options applied when querying
-	// contracts. By default, only contracts that are revisable will be
-	// returned.
-	DefaultContractQueryOpts = contractQueryOpts{
```
peterjan: I got rid of these default options. It's a little weird because by not exposing `contractQueryOpts` you can never leave `Revisable` nil. Most importantly though I found that we specify the options in every single location that calls `Contracts`, so why even have a default? Having a default also makes it unclear whether options that are passed in are applied to the default or to nil opts.

contracts/manager.go
```diff
@@ -67,7 +68,7 @@ type (
 		AddFormedContract(ctx context.Context, contractID types.FileContractID, hostKey types.PublicKey, proofHeight, expirationHeight uint64, contractPrice, allowance, minerFee, totalCollateral types.Currency) error
 		AddRenewedContract(ctx context.Context, params AddRenewedContractParams) error
 		ContractElementsForBroadcast(ctx context.Context, maxBlocksSinceExpiry uint64) ([]types.V2FileContractElement, error)
-		Contracts(ctx context.Context, queryOpts ...ContractQueryOpt) ([]Contract, error)
+		Contracts(ctx context.Context, offset, limit int, queryOpts ...ContractQueryOpt) ([]Contract, error)
```
peterjan, line 71: This is a little annoying but I do think we should do it and be consistent with these type of methods. It already pays off in `syncRevisionState` where we can now easily apply the `Nate-pattern`

ChrisSchinnerl: What is the `Nate-pattern` 😅 

It does feel a bit off. Especially having to now pass `math.MaxInt64` which makes it awkward to use.

doc/rfc/001_db_schema.md
```diff
@@ -173,16 +173,15 @@ CREATE INDEX hosts_blocklist_reason_idx ON hosts_blocklist (reason);
...
-  contract_id BYTEA NOT NULL UNIQUE DEFERRABLE,
 
   -- lifetime related columns
   formation TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
   proof_height BIGINT NOT NULL, -- start of proof window
   expiration_height BIGINT NOT NULL, -- end of proof window
-  renewed_from INTEGER REFERENCES contracts(id) UNIQUE DEFERRABLE,
-  renewed_to INTEGER REFERENCES contracts(id) UNIQUE DEFERRABLE,
+  renewed_from BYTEA REFERENCES contracts(contract_id),
+  renewed_to BYTEA REFERENCES contracts(contract_id),
```
ChrisSchinnerl: Should still be `UNIQUE` right?

doc/rfc/001_db_schema.md
```diff
@@ -173,16 +173,15 @@ CREATE INDEX hosts_blocklist_reason_idx ON hosts_blocklist (reason);
 
 ```postgresql
 CREATE TABLE contracts (
-  id SERIAL PRIMARY KEY,
+  contract_id BYTEA PRIMARY KEY CHECK (LENGTH(contract_id) = 32),
```
ChrisSchinnerl: I'd stick to our pattern of keeping the primary key a serial since this doesn't really help us with faster decoration of `renewed_to` and `renewed_from` while every sector references the contract it is pinned to. 

An indexer with 1PiB of uploaded sectors will have a few hundred million rows in their sectors table each containing 32 bytes now instead of 4 and maybe another 32 bytes instead of 4 for the index we are building.

ChrisSchinnerl (changes requested):

doc/rfc/001_db_schema.md
n8mgr: Using the natural key for `renewed_from`, `renewed_to` and `contract_elements` is an easy win. 

1PiB of sectors would only be 8GB of foreign keys. Worst case we're talking ~20GB of extra space including indices to change the sectors table. It's not that bad. I personally prefer natural keys, but I don't feel strongly enough to die on that hill. Especially since it would likely result in performance degradation on other joins.

contracts/manager.go
peterjan: All the uses of `math.MaxInt64` are labeled with a TODO to F/U with. Wanted to keep the PR clean and there's no point in tackling all those call sites in the same PR. The `Nate-pattern` is where we fetch `limit` items and then spin that number of goroutines, instead of defining the number of threads. If you look at `syncRevisionState`, you'll see it's easily converted to using a limit and processing the contracts in batches instead.

doc/rfc/001_db_schema.md
peterjan: I got excited about removing the `LEFT JOIN` on contracts in the slabs query. I've reverted the schema to only keep the changes necessary to improve the performance of the contract queries.

ChrisSchinnerl: Definitely don't mind for `renewed_from` and `renewed_to`. `contract_elements` I'm indifferent about since it's only accessed very infrequently and is probably out of scope for this PR anyway.

For `contracts.id` I'd need to see some benchmark results not just for the `Contracts` query but also queries related to fetching sectors such as `UnhealthySlabs` over a database containing at least 1PB of sectors.

Storing 8GB is not big of a deal but browsing an 8x larger b-tree on super low granular data (hundreds of millions of sectors referencing the same 100 contracts) might just end up being annoying.

peterjan: Current version of the PR should be good though right? I think it's fine to keep the serial as PK on contracts for now, as you proposed? We can explore `contracts.id` when `UnhealthySlabs` is merged or has been benchmarked already.

persist/postgres/contracts.go
```diff
@@ -32,70 +32,44 @@ func (s *Store) AddFormedContract(ctx context.Context, contractID types.FileCont
...
-) RETURNING id
-`, existingID).Scan(&newID); err != nil {
-			return fmt.Errorf("failed to copy renewed contract: %w", err)
-		}
-
-		// update a potential row that renewed to the existing contract
-		_, err := tx.Exec(context.Background(), `UPDATE contracts SET renewed_to = $1 WHERE renewed_to = $2 AND id != $1`, newID, existingID)
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `
+INSERT INTO contracts(contract_id, host_id, formation, proof_height, expiration_height, renewed_from, capacity, size, contract_price, initial_allowance, remaining_allowance, miner_fee, used_collateral, total_collateral)
```
ChrisSchinnerl: Doesn't this break our contract <-> sector links? We intentionally made sure to keep the same `id` and update the row instead of inserting a new `id` for the refreshed contract to avoid that.

ChrisSchinnerl (changes requested):

persist/postgres/contracts.go
peterjan: 🤦‍♂️ can't believe I forgot about that, I had a feeling but there wasn't any test failure so figured it was fine. I'll add a test and fix it.

persist/postgres/contracts.go
```diff
@@ -32,70 +32,132 @@ func (s *Store) AddFormedContract(ctx context.Context, contractID types.FileCont
...
-		_, err := tx.Exec(context.Background(), `UPDATE contracts SET renewed_to = $1 WHERE renewed_to = $2 AND id != $1`, newID, existingID)
-		if err != nil {
-			return fmt.Errorf("failed to update renewed_to: %w", err)
-		}
-
-		// update the existing row to match the new contract
-		resp, err := tx.Exec(ctx, `
-UPDATE contracts SET contract_id = $1, formation = NOW(), proof_height = $2, expiration_height = $3, renewed_from = $4, renewed_to = NULL, state = 0, capacity = CASE WHEN $2 = contracts.proof_height THEN contracts.capacity ELSE contracts.size END, contract_price = $5, initial_allowance = $6, remaining_allowance = $6, miner_fee = $7, used_collateral = $8, total_collateral = $9, good = TRUE, append_sector_spending = 0, free_sector_spending = 0, fund_account_spending = 0, sector_roots_spending = 0
-WHERE id = $10`, sqlHash256(params.RenewedTo), params.ProofHeight, params.ExpirationHeight, newID, sqlCurrency(params.ContractPrice), sqlCurrency(params.Allowance), sqlCurrency(params.MinerFee), sqlCurrency(params.UsedCollateral), sqlCurrency(params.TotalCollateral), existingID)
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
```
peterjan, line 45: I ran into quite a pickle here... foreign keys cannot reference deferrable unique constraints. I tried a lot of stuff. I tried setting values to `NULL` and then updating them, I tried changing the order of insert vs update, I tried adding a separate column for the sector references... I think this is the only viable solution.

I think maybe with some clean formatting, maybe better names for the CTEs, and of course a little benchmark to see if this is not horribly slow, it's an ok solution. It's actually quite descriptive and shows what's happening a lot more. We also don't have to defer the evaluation of constraints any longer. I personally don't mind every field on a separate line, but don't mind reformatting either.

It took me quite a long time to realise I had to `SELECT` from the `updated` and return `renewed.*` there. That's the key, otherwise this doesn't work and you still get a constraint violation. It needs the dependency to guarantee CTE execution order.

n8mgr: This is slightly better, but honestly would rather add a mapping table. Gets rid of the old update madness we used to have and I'm not confident `contract_elements` are handled correctly in this PR or master.

n8mgr: Effectively this:
```sql
CREATE TABLE contracts (
  id BYTEA UNIQUE NOT NULL CHECK (LENGTH(id) = 32), -- unique identifier for the contract
  host_id INTEGER REFERENCES hosts(id) NOT NULL,

  -- lifetime related columns
  formation TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
  proof_height BIGINT NOT NULL, -- start of proof window
  expiration_height BIGINT NOT NULL, -- end of proof window
  renewed_from INTEGER UNIQUE REFERENCES contracts(id),
  renewed_to INTEGER UNIQUE REFERENCES contracts(id),
  revision_number INTEGER NOT NULL DEFAULT 0 CHECK(revision_number >= 0),
  state SMALLINT NOT NULL DEFAULT 0, -- 0 = 'pending', 1 = 'active', 2 = 'resolved', 3 = 'expired', 4 = 'rejected'

  -- metrics for visualization (not ACID)
  capacity BIGINT NOT NULL DEFAULT 0 CHECK(capacity >= size),
  size BIGINT NOT NULL DEFAULT 0 CHECK(size >= 0),

  -- costs
  contract_price DECIMAL(50, 0) NOT NULL, -- used to display cost of forming contract
  initial_allowance DECIMAL(50, 0) NOT NULL, -- used when refreshing contract to increase budget
  remaining_allowance DECIMAL(50, 0) NOT NULL DEFAULT 0 CHECK(remaining_allowance >= 0), -- remaining allowance
  miner_fee DECIMAL(50, 0) NOT NULL, -- miner fee added when forming/renewing contract
  used_collateral DECIMAL(50, 0) NOT NULL DEFAULT 0 CHECK(used_collateral <= total_collateral), -- collateral (allocated)
  total_collateral DECIMAL(50, 0) NOT NULL, -- total collateral (allocated+unallocated)

  -- contract state
  good BOOLEAN NOT NULL DEFAULT TRUE,

  -- spending (not ACID)
  append_sector_spending DECIMAL(50, 0) NOT NULL DEFAULT 0,
  free_sector_spending DECIMAL(50, 0) NOT NULL DEFAULT 0,
  fund_account_spending DECIMAL(50, 0) NOT NULL DEFAULT 0,
  sector_roots_spending DECIMAL(50, 0) NOT NULL DEFAULT 0
);
CREATE INDEX contracts_state_formation_idx ON contracts(state, formation); -- for rejecting expired contracts
CREATE INDEX contracts_host_id_remaining_allowance_idx ON contracts (host_id, remaining_allowance DESC) WHERE good = true AND remaining_allowance > 0 AND state <= 1; -- for fetching contracts for funding

CREATE TABLE contract_elements (
    id SERIAL PRIMARY KEY,
    contract_id BYTEA UNIQUE NOT NULL REFERENCES contracts(id) ON DELETE CASCADE,
    contract BYTEA NOT NULL,
    leaf_index INTEGER NOT NULL,
    merkle_proof BYTEA NOT NULL
);

CREATE TABLE contract_sectors_map (
    id SERIAL PRIMARY KEY,
    contract_id BYTEA REFERENCES contracts(id) NOT NULL,
);
CREATE INDEX contract_sectors_map_contract_id_idx ON contract_sectors_map(contract_id);

CREATE TABLE sectors (
    id BIGSERIAL PRIMARY KEY,
    sector_root BYTEA NOT NULL,

    -- uploading
    host_id INTEGER REFERENCES hosts(id), -- host that stores sector
    contract_sectors_map_id INTEGER REFERENCES contract_sectors_map(id) DEFAULT NULL, -- null if not pinned
    uploaded_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(), -- allow sorting by upload time

    -- slab
    slab_id BIGINT REFERENCES slabs(id) NOT NULL,

    -- data integrity
    next_integrity_check TIMESTAMP WITH TIME ZONE NOT NULL,
    consecutive_failed_checks SMALLINT NOT NULL DEFAULT 0
);
-- quick lookup of sectors to pin prioritized by upload time
CREATE INDEX sectors_contract_id_uploaded_at_idx ON sectors(contract_sectors_map_id, uploaded_at ASC);
```

Gets rid of deferrables, CTEs, and renewing a contract becomes as simple as inserting the new row, setting the renewed_to field on the existing contract, then updating `contract_sectors_map` to point to the new ID. No need to be extra careful about contract foreign keys.

peterjan: Considered this but figured the CTE would be the desired/better approach. @ChrisSchinnerl OK for you to add the extra mapping table? Is there something special about `contract_elements`? I figured they just need to be inserted/updated on consensus changes but don't have to be interacted with when we renew a contract in the database. 

edit: I personally wouldn't mind keeping the CTE here by the way

ChrisSchinnerl: Yeah I don't mind. 

The `contract_elements` row for the old contract needs to still be linked to the old contract after the renewal. The new contract won't have one right after being renewed but we will add it once it is mined in the subscription code.

I think the code we had so far might have failed to do so and in the edge case of a renewal never getting mined, might have led to the indexer being unable to broadcast a resolution for the old contract.

peterjan: Ok, updated to use a mapping table. I insert an entry into the mapping table when forming a contract, and then we update it on renewals. I extend the renew tests to check for an update, but didn't involve any sectors yet since we don't form those links yet.

persist/postgres/contracts.go
```diff
@@ -67,9 +67,11 @@ INSERT INTO contracts(contract_id, host_id, formation, proof_height, expiration_
 			return fmt.Errorf("expected 1 row to be affected, got %d", res.RowsAffected())
 		}
 
-		_, err = tx.Exec(ctx, `UPDATE contract_sectors_map SET contract_id = $1 WHERE contract_id = $2`, sqlHash256(params.RenewedTo), sqlHash256(params.RenewedFrom))
+		res, err = tx.Exec(ctx, `UPDATE contract_sectors_map SET contract_id = $1 WHERE contract_id = $2`, sqlHash256(params.RenewedTo), sqlHash256(params.RenewedFrom))
 		if err != nil {
 			return fmt.Errorf("failed to update contract sectors map: %w", err)
+		} else if res.RowsAffected() != 1 {
+			return fmt.Errorf("expected 1 row to be affected, got %d", res.RowsAffected())
```
n8mgr: nit: this can never be > 1, no need to return the number of rows.

n8mgr (dismissed):

persist/postgres/contracts.go
peterjan: you always approve with a `nit` - I love it

persist/postgres/sectors_test.go
```diff
@@ -794,9 +807,9 @@ func BenchmarkSectorsForIntegrityCheck(b *testing.B) {
 
 // BenchmarkUnhealthySlab benchmarks UnhealthySlab
 //
-//	CPU   |	  Count  |     Time/op
+//	CPU   |	  Count  |   Time/op
 //
-// M2 Pro |   1282   |    0.618084 ms
+// M1 Max |   782   |    3.5 ms
```
peterjan, line 799: good thing we merged the benchmark minutes before merging this in, mind rerunning it on your hardware @ChrisSchinnerl

peterjan: hmm, not sure if it's that much different from the query plan in `master` - going to check that now
```
Limit  (cost=0.87..1.35 rows=1 width=8) (actual time=156.760..156.761 rows=1 loops=1)
  ->  Nested Loop  (cost=0.87..507609.72 rows=1063396 width=8) (actual time=156.759..156.760 rows=1 loops=1)
        ->  Merge Right Join  (cost=0.58..177313.97 rows=1063518 width=8) (actual time=156.715..156.716 rows=1 loops=1)
              Merge Cond: (csm.id = sectors.contract_sectors_map_id)
              Filter: (
                        (sectors.contract_sectors_map_id IS NOT NULL AND NOT contracts.good)
                        OR (sectors.host_id IS NULL)
                      )
              Rows Removed by Filter: 393210
              ->  Nested Loop Left Join  (cost=0.15..2934.36 rows=1270 width=5) (actual time=0.055..0.063 rows=2 loops=1)
                    Join Filter: (csm.contract_id = contracts.contract_id)
                    Rows Removed by Join Filter: 1
                    ->  Index Scan using contract_sectors_map_pkey on contract_sectors_map csm  
                          (cost=0.15..67.20 rows=1270 width=36) (actual time=0.036..0.037 rows=2 loops=1)
                    ->  Materialize  (cost=0.00..12.25 rows=150 width=33) (actual time=0.009..0.010 rows=2 loops=2)
                          ->  Seq Scan on contracts  
                                (cost=0.00..11.50 rows=150 width=33) (actual time=0.015..0.016 rows=2 loops=1)
              ->  Index Scan using sectors_contract_sectors_map_id_uploaded_at_idx on sectors  
                    (cost=0.43..147789.16 rows=2126982 width=16) (actual time=0.026..127.885 rows=393211 loops=1)
        ->  Index Only Scan using slabs_id_last_repair_attempt_idx on slabs  
              (cost=0.29..0.31 rows=1 width=8) (actual time=0.040..0.040 rows=1 loops=1)
              Index Cond: ((id = sectors.slab_id) AND (last_repair_attempt <= now()))
              Heap Fetches: 0

Planning Time: 1.984 ms  
Execution Time: 156.810 ms
```

persist/postgres/sectors_test.go
```diff
@@ -265,6 +265,15 @@ func TestPinSlabs(t *testing.T) {
 func TestUnhealthySlabs(t *testing.T) {
 	store := initPostgres(t, zaptest.NewLogger(t).Named("postgres"))
 
+	getMapID := func(fcid types.FileContractID) (mapID int64) {
```
ChrisSchinnerl: Don't need this. It's always 1 as before when we used `contract_id = 1`.

persist/postgres/sectors_test.go
```diff
@@ -398,6 +407,10 @@ func TestUnpinnedSectors(t *testing.T) {
 	if err := store.AddFormedContract(context.Background(), types.FileContractID(hk), hk, 100, 200, types.Siacoins(1), types.Siacoins(1), types.Siacoins(1), types.Siacoins(1)); err != nil {
 		t.Fatal(err)
 	}
+	var mapID int64
```
ChrisSchinnerl: Don't need this. Can still use `&one`.

persist/postgres/sectors_test.go
```diff
@@ -805,6 +818,15 @@ func BenchmarkUnhealthySlab(b *testing.B) {
 		b.Fatal("failed to add account:", err)
 	}
 
+	getMapID := func(fcid types.FileContractID) (mapID int64) {
```
ChrisSchinnerl: Same here. The map id should match the contract id so we can continue to use 1 and 2.

persist/postgres/contracts.go
```diff
@@ -28,74 +28,63 @@ func (s *Store) AddFormedContract(ctx context.Context, contractID types.FileCont
...
-INSERT INTO contracts (host_id, contract_id, proof_height, expiration_height, renewed_from, renewed_to, state, capacity, size, contract_price, initial_allowance, remaining_allowance, miner_fee, used_collateral, total_collateral, good, append_sector_spending, free_sector_spending, fund_account_spending, sector_roots_spending) (
-	SELECT host_id, contract_id, proof_height, expiration_height, renewed_from, $1, state, capacity, size, contract_price, initial_allowance, remaining_allowance, miner_fee, used_collateral, total_collateral, good, append_sector_spending, free_sector_spending, fund_account_spending, sector_roots_spending
-	FROM contracts
-	WHERE contracts.id = $1
-) RETURNING id
-`, existingID).Scan(&newID); err != nil {
-			return fmt.Errorf("failed to copy renewed contract: %w", err)
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `
+INSERT INTO contracts(contract_id, host_id, formation, proof_height, expiration_height, renewed_from, capacity, size, contract_price, initial_allowance, remaining_allowance, miner_fee, used_collateral, total_collateral)
```
ChrisSchinnerl: `contract_id` and `hostd_id` are in reverse order compared to `AddFormedContract`
`formation` doesn't need to be set anymore since the default is `NOW()` anyway.

persist/postgres/sectors_test.go
peterjan:
```
➜  indexd git:(master) 
cpu: Apple M1 Max
BenchmarkUnhealthySlab-10            870           3970135 ns/op            2461 B/op         32 allocs/op
PASS
ok      go.sia.tech/indexd/persist/postgres     82.339s
```

vs 

```
➜  indexd git:(pj/contracts) 
BenchmarkUnhealthySlab-10            660           3358117 ns/op            2493 B/op         32 allocs/op
PASS
ok      go.sia.tech/indexd/persist/postgres     79.155s
```

Is it at all possible the 0.6ms in `master` is incorrect? Looking at the query plan I don't immediately see what causes it to be ~6 times slower.

persist/postgres/sectors_test.go
peterjan: I don't like relying on fixed database ids but ok.

persist/postgres/sectors_test.go
peterjan: Ok this commit claimed a speedup I don't think was ever the case: https://github.com/SiaFoundation/indexd/pull/119/commits/02fd69297043f93b31c884205cad1b91d8b78f95

```
➜  indexd git:(02fd692) 
BenchmarkUnhealthySlab-10            859           4249602 ns/op            2463 B/op         32 allocs/op
PASS
```

or perhaps it was the case at one point but it definitely no longer is the case. I tried getting it there but I don't see how we can. I tried using a CTE to limit the slabs for repair first, but everything I tried ended up being slower. I think it's fine to merge the 3.5ms as-is.

ChrisSchinnerl (approved):

PR #117, opened 2025-04-16
Add PinSectors to database

ChrisSchinnerl:
To mark sectors as pinned after adding them to a contract


n8mgr (dismissed):

persist/postgres/sectors.go
```diff
@@ -139,6 +139,31 @@ func (s *Store) Slabs(ctx context.Context, accountID proto.Account, slabIDs []sl
...
+	sqlRoots := make([]sqlHash256, len(roots))
+	for i, root := range roots {
+		sqlRoots[i] = sqlHash256(root)
+	}
+
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `
+			UPDATE sectors
+			SET (host_id, contract_id) = (
+				SELECT hosts.id, contracts.id
```
peterjan: we probably want to check for this inner select coming up empty

persist/postgres/sectors_test.go
```diff
@@ -262,6 +263,116 @@ func TestPinSlabs(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	one := int64(1)
+	assertPinned(1, &one)
+	assertPinned(2, nil)
+	assertPinned(3, &one)
+	assertPinned(4, nil)
+
+	// pin sectors 2 and 4 to contract 2
+	err = store.PinSectors(context.Background(), contractID2, []types.Hash256{{2}, {4}})
```
peterjan, line 366: extend with pinning sectors to non-existent contract id

persist/postgres/sectors.go
ChrisSchinnerl: Good catch

n8mgr (dismissed):

peterjan (dismissed):

peterjan (approved):

n8mgr (approved):

PR #118, opened 2025-04-16
Add SectorsForIntegrityCheck to database

ChrisSchinnerl:
Somehow I accidentally closed https://github.com/SiaFoundation/indexd/pull/110 with a force push that attempted to revert a wrong merge commit.

This PR attempts to recover that


slabs/slabs_test.go
ChrisSchinnerl, line 1: Not sure why this is part of the diff again but here we are unfortunately 😅

doc/rfc/001_db_schema.md
```diff
@@ -255,7 +255,7 @@ CREATE INDEX slabs_digest_idx ON slabs(digest);
 CREATE TABLE account_slabs (
```
peterjan, line 255: not sure if it's worth keeping a copy of `init.sql` in this doc, I'm getting annoyed at times having to copy over my changes to this file

peterjan (approved):
lgtm

doc/rfc/001_db_schema.md
ChrisSchinnerl: haha same. Maybe we should open a PR to shift all the numbers by 1 and get rid of the schema RFC

n8mgr (approved):

PR #123, opened 2025-04-17
Add Account Key Rotation

### LINKED ISSUE #86: RFC: Key Store
Write an RFC for a key store that allows for backing up the private keys for the registered public keys.

The overall idea is that when registering a public key, the user can optionally register an encrypted private key alongside it. `Indexd` has no means to use that private key other than storing it and returning it back to the user.

The user can then, assuming they still know how they encrypted it, decrypt it to get access to the private key.

This serves as a recovery mechanism for private keys in case a user loses theirs.

peterjan:
closes https://github.com/SiaFoundation/indexd/issues/86


ChrisSchinnerl (dismissed):

ChrisSchinnerl (dismissed):

persist/postgres/accounts_test.go
```diff
@@ -106,6 +106,76 @@ func TestDeleteAccount(t *testing.T) {
...
+	// add another account
+	pk2 := types.GeneratePrivateKey().PublicKey()
+	err = store.AddAccount(context.Background(), pk2)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// assert updating to an existing account fails
+	err = store.UpdateAccount(context.Background(), pk1, pk2)
+	if !errors.Is(accounts.ErrExists, err) {
```
n8mgr: arg ordering: `errors.Is(src, target)`

persist/postgres/accounts_test.go
```diff
@@ -106,6 +106,76 @@ func TestDeleteAccount(t *testing.T) {
...
+		t.Fatal("expected [accounts.ErrExists], got", err)
+	}
+
+	// assert updating a non existing account fails
+	pk3 := types.GeneratePrivateKey().PublicKey()
+	err = store.UpdateAccount(context.Background(), types.GeneratePrivateKey().PublicKey(), pk3)
+	if !errors.Is(accounts.ErrNotFound, err) {
```
n8mgr: arg ordering: errors.Is(src, target)

api/endpoints.go
```diff
@@ -73,6 +73,39 @@ func (a *api) handlePOSTAccount(jc jape.Context) {
...
+		jc.Error(err, http.StatusNotFound)
+		return
+	} else if errors.Is(err, accounts.ErrExists) {
+		jc.Error(err, http.StatusConflict)
+		return
+	} else if jc.Check("failed to rotate account key", err) != nil {
+		return
+	}
+
+	a.contracts.TriggerAccountFunding()
```
n8mgr, line 108: Use `jc.Encode(nil)` for `204 No Content`

persist/postgres/accounts.go
```diff
@@ -83,6 +83,27 @@ func (s *Store) DeleteAccount(ctx context.Context, ak types.PublicKey) error {
...
+		} else if exists {
+			return accounts.ErrExists
+		}
+
+		res, err := tx.Exec(ctx, `UPDATE accounts SET public_key = $1 WHERE public_key = $2`, sqlPublicKey(newAK), sqlPublicKey(oldAK))
+		if err != nil {
+			return fmt.Errorf("failed to update account: %w", err)
+		} else if res.RowsAffected() != 1 {
+			return accounts.ErrNotFound
+		}
```
n8mgr:
```suggestion
        res, err := tx.Exec(ctx, `UPDATE accounts SET public_key = $1 WHERE public_key = $2`, sqlPublicKey(newAK), sqlPublicKey(oldAK))  
        if err != nil {  
            var pgErr *pgconn.PgError  
            if errors.As(err, &pgErr) && pgErr.Code == pgerrcode.UniqueViolation {  
                return accounts.ErrExists  
            }  
            return fmt.Errorf("failed to update account: %w", err)  
        } else if res.RowsAffected() != 1 {  
            return accounts.ErrNotFound  
        } 
```

n8mgr (changes requested):

persist/postgres/accounts_test.go
peterjan: lol

persist/postgres/accounts.go
peterjan: `pgerrcode.UniqueViolation` man this is great, I thought about that but didn't want to do the `string.Contains` which we usually did in renterd with sqlite

ChrisSchinnerl (dismissed):

ChrisSchinnerl (approved):

n8mgr (approved):

PR #125, opened 2025-04-22
Fetch contracts using an offset and limit

peterjan:
This PR ensures we fetch contracts using an actual `offset` and `limit`. 

I wanted to bring attention to the fact that paging through a set of contracts, which might potentially be updated while we're paging through them, obviously has as a side effect that some contracts get skipped. I didn't think much of it, because of the nature of our loops, if a contract isn't handled in one iteration, it'll be handled the next time around...

The alternative is to use a (id) marker instead of an offset, but you'd still have the same problem, but less frequently. We could also get rid of filters, also not great. We could also get rid of the `offset` and `limit` but that's not great either imo. 

I think we have paging mostly because the API applies a default limit. **Perhaps we should make it an option** (`WithOffsetLimit` or something) and use it in the API, but in our backend we can use the filters to fetch all active (or whatever) contracts at once.


contracts/manager.go
```diff
@@ -155,6 +154,15 @@ func WithLogger(l *zap.Logger) ContractManagerOpt {
 	}
 }
 
+// WithNumThreads allows overriding the number of threads used by the contract
+// manager when performing various actions like refreshing or renewing
+// contracts, or syncing contract revisions.
+func WithNumThreads(n int) ContractManagerOpt {
```
ChrisSchinnerl: Refreshing and renewing probably isn't something we should do in parallel since we need to fund contracts when we do that. It's going to cause fewer issues to do it one after another.

I'd also not expose this and just default to 50.

contracts/chain_test.go
```diff
@@ -141,7 +141,7 @@ func (s *storeMock) Contracts(ctx context.Context, offset, limit int, queryOpts
 	filtered := make([]Contract, 0, len(s.contracts))
 	for _, c := range s.contracts {
 		if opts.Revisable != nil {
-			isRevisable := (c.State == ContractStatePending || c.State == ContractStateActive) && c.RenewedTo == (types.FileContractID{})
```
ChrisSchinnerl: why remove this again?

contracts/renew_test.go
```diff
@@ -168,13 +168,6 @@ func TestPerformContractRenewals(t *testing.T) {
 			}
 		}
 	}
-
-	// assert consecutive calls don't keep renewing the same contract
-	if err := contracts.performContractRenewals(context.Background(), period, renewWindow, zap.NewNop()); err != nil {
```
ChrisSchinnerl: why was this check removed? Shouldn't this still be the case?

contracts/chain_test.go
peterjan: not sure how that ended up being in the diff

ChrisSchinnerl (dismissed):

n8mgr (dismissed):

ChrisSchinnerl (approved):

PR #126, opened 2025-04-22
Take into account contract period when renewing contracts

peterjan:
I found out we weren't taking into account the `period` when renewing contracts in the contractor.

### DIFF
--- a/contracts/chain_test.go
+++ b/contracts/chain_test.go
@@ -132,8 +132,31 @@ func (s *storeMock) Contract(_ context.Context, contractID types.FileContractID)
 	return Contract{}, ErrNotFound
 }
 
-func (s *storeMock) Contracts(ctx context.Context, opts ...ContractQueryOpt) ([]Contract, error) {
-	return slices.Clone(s.contracts), nil
+func (s *storeMock) Contracts(ctx context.Context, queryOpts ...ContractQueryOpt) ([]Contract, error) {
+	cloned := slices.Clone(s.contracts)
+
+	var opts ContractQueryOpts
+	for _, opt := range queryOpts {
+		opt(&opts)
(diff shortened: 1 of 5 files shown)


contracts/chain_test.go
```diff
@@ -132,8 +132,31 @@ func (s *storeMock) Contract(_ context.Context, contractID types.FileContractID)
 	return Contract{}, ErrNotFound
 }
 
-func (s *storeMock) Contracts(ctx context.Context, opts ...ContractQueryOpt) ([]Contract, error) {
-	return slices.Clone(s.contracts), nil
+func (s *storeMock) Contracts(ctx context.Context, queryOpts ...ContractQueryOpt) ([]Contract, error) {
+	cloned := slices.Clone(s.contracts)
```
n8mgr, line 136: No benefit to cloning the slice here since it's going to be filtered below.


```suggestion
    var opts ContractQueryOpts
	for _, opt := range queryOpts {
		opt(&opts)
	}

	filtered := make([]Contract, 0, len(s.contracts))
	for _, c := range s.contracts {
		if opts.Revisable != nil {
			isRevisable := (c.State == ContractStatePending || c.State == ContractStateActive) && c.RenewedTo == (types.FileContractID{})
			if isRevisable != *opts.Revisable {
				continue
			}
		}
		if opts.Good != nil {
			if c.Good != *opts.Good {
				continue
			}
		}
		filtered = append(filtered, c)
	}
```

n8mgr (approved):

ChrisSchinnerl (approved):

contracts/chain_test.go
peterjan: removing in another PR 👍

PR #128, opened 2025-04-23
Add RecordIntegrityCheck and FailingSectors

ChrisSchinnerl:
For persisting the results of a data integrity check on a sector and fetching sectors that have failed the check a certain number of times.


n8mgr (approved):

persist/postgres/sectors.go
```diff
@@ -49,6 +49,34 @@ func (s *Store) MarkSectorsLost(ctx context.Context, hostKey types.PublicKey, ro
 	return err
 }
 
+// RecordIntegrityCheck records the result of integrity checks for the given
+// sectors stored on the given host.
+func (s *Store) RecordIntegrityCheck(ctx context.Context, success bool, nextCheck time.Time, hostKey types.PublicKey, roots []types.Hash256) error {
```
peterjan, line 54: `nit` order of the arguments is messed up

persist/postgres/sectors.go
```diff
@@ -82,8 +110,37 @@ func (s *Store) SectorsForIntegrityCheck(ctx context.Context, hostKey types.Publ
 	return sectors, err
 }
 
-// PinSlab adds a slab to the database for pinning. The slab is associated
-// with the provided account.
+// FailingSectors returns up to `limit` sectors that have failed integrity
+// checks at least 'minChecks' times.
+func (s *Store) FailingSectors(ctx context.Context, hostKey types.PublicKey, minChecks, limit int) ([]types.Hash256, error) {
```
peterjan, line 115: missing unit test, it's very simple but since we currently have a 1-1 thing going on we might as well keep going and have a unit test for every method on the store (I know it's covered by `TestRecordIntegrityCheck` but still)

peterjan (approved):

persist/postgres/sectors.go
ChrisSchinnerl: I considered it, but I honestly didn't see the value of effectively just copy pasting another unit test and renaming it unless there is some edge case the test doesn't cover.

PR #131, opened 2025-04-23
Broadcast revisions periodically

peterjan:


contracts/broadcast.go
```diff
@@ -0,0 +1,111 @@
...
+	"go.sia.tech/core/types"
+	"go.uber.org/zap"
+)
+
+func (cm *ContractManager) performBroadcastContractRevisions(ctx context.Context, log *zap.Logger) error {
+	broadcastLog := log.Named("broadcast")
+
+	const batchSize = 50
+	for offset := 0; ; offset += batchSize {
+		contracts, err := cm.store.Contracts(ctx, offset, batchSize, WithGood(true), WithRevisable(true))
```
n8mgr: Add a query for this. There is no reason to grab every contract from the database just to filter it locally using columns that are stored in the database. Just grab the ones that need broadcast. Using the generic query for everything in these hot loops is inefficient.

n8mgr (changes requested):

contracts/contract.go
```diff
@@ -148,6 +151,14 @@ func (c Contract) GoodForUpload(prices proto.HostPrices, maxCollateral types.Cur
 		c.UsedCollateral.Add(sectorAppendCollateral).Cmp(c.TotalCollateral) < 0
 }
 
+// NeedsBroadcast indicates that a contract should be broadcasted.
+func (c Contract) NeedsBroadcast(interval time.Duration) bool {
+	renewed := c.RenewedTo != (types.FileContractID{})
+	seenRecently := time.Since(c.LastChainUpdate) < interval
```
ChrisSchinnerl: I'm not sure what purpose `seenRecently` serves? If we always wait for a full interval to pass since the last broadcast anyway I'm not sure it adds any value considering the host won't broadcast it prematurely.

contracts/broadcast.go
```diff
@@ -0,0 +1,111 @@
...
+
+		if len(contracts) < batchSize {
+			break
+		}
+	}
+
+	return nil
+}
+
+func (cm *ContractManager) broadcastContractRevision(ctx context.Context, contract Contract, log *zap.Logger) error {
```
ChrisSchinnerl: If you use a specialized query as Nate suggested, you can only fetch the elements directly and don't need full contracts here.

peterjan:
(not quite ready yet, I'll re-request review - will remove the last_chain_update part)

contracts/chain_test.go
```diff
@@ -119,6 +119,24 @@ func (s *storeMock) BlockHosts(_ context.Context, hostKeys []types.PublicKey, re
...
+					LeafIndex:   1,
+					MerkleProof: []types.Hash256{{1}},
+				},
+				V2FileContract: types.V2FileContract{
+					HostPublicKey: c.HostKey,
+				},
+			}, nil
+		}
+	}
+	return types.V2FileContractElement{}, nil
```
ChrisSchinnerl: I'd return `ErrNotFound` to match the actual implementation

contracts/contract.go
```diff
@@ -27,24 +27,31 @@ var (
...
 	return func(opts *ContractQueryOpts) {
 		opts.Good = &good
 	}
 }
 
-// WithRevisable filters contracts by whether they can still be revised. This
-// defaults to 'true'.
+// WithMinBroadcast filters contracts by whether they have been broadcasted (or
+// seen on chain) recently.
+func WithMinBroadcast(threshold time.Time) ContractQueryOpt {
```
ChrisSchinnerl: Unused? I'd remove it until we have a need for it.

contracts/fund_test.go
```diff
@@ -32,12 +32,34 @@ func (am *accountsManagerMock) FundAccounts(ctx context.Context, host hosts.Host
 	return nil
 }
 
-func (s *storeMock) ContractsForFunding(_ context.Context, hk types.PublicKey, limit int) ([]types.FileContractID, error) {
-	cloned := slices.Clone(s.contracts)
+func (s *storeMock) ContractsForBroadcasting(_ context.Context, minBroadcast time.Time, limit int) ([]types.FileContractID, error) {
+	var contracts []Contract
+	for _, c := range s.contracts {
+		if c.Good &&
```
ChrisSchinnerl: Contracts should be broadcasted even when bad.

ChrisSchinnerl (changes requested):

contracts/broadcast.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	// fetch the host
+	host, err := cm.store.Host(ctx, fce.V2FileContract.HostPublicKey)
+	if err != nil {
+		return fmt.Errorf("failed to fetch host: %w", err)
+	}
+
+	// fetch the latest revision
+	resp, err := cm.contractor.LatestRevision(ctx, host.PublicKey, host.SiamuxAddr(), contractID)
```
n8mgr, line 59: This is less than ideal. What if the host is offline? What if the host lies about the revision number? We should store the current revision every time we revise the contract successfully and broadcast that instead.

peterjan: I guess we could extend `SyncContract` and save the revision separately as a blob in the contracts table. I'd probably lean towards storing them separately though. Storing a contract's revision is something we were happy to get rid of, I'm a little surprised we'd reintroduce it now for broadcasting purposes. The host being offline is indeed less than ideal, but at the same time broadcasting the revision is hardly critical. (cc @ChrisSchinnerl  - ok with keeping revisions in separate table?)

ChrisSchinnerl: I think both approaches are fine tbh. Hosts don't know when it's safe to lie about this so it's a gamble for them to try any funny business. 

If we decide to go down the perfectionist route I'd record the revision every time we revise the contract rather than upon syncing since that completely eliminates the host's ability to lie. Considering the size of this PR I'd move that new persistence code to a separate PR.

n8mgr: I'm not really sure why we don't store the revision tbh. Every RPC returns the revised contract. It'd be significantly more efficient than first pulling the revision every time we want to revise a contract or grabbing every revision from every host every 30 minutes to update the state of the store.

1. Try to do an RPC with our stored revision
2. If the host returns "invalid signature," sync the host's version
3. if revision number > existing revision number and the signatures are valid, update the revision state in the database

The only time revisions can realistically get out of sync is when an RPC fails or the host is malicious.

ChrisSchinnerl: It was just the simplest and most foolproof approach. We can do what you are suggesting but I think we need to update `coreutils` for that to work. Because right now, it seems only funding an account will return `invalid signature` and appending a sector will return `invalid challenge signature`.

So at the very least I'd add a test to coreutils that guarantees all RPCs that we send contracts to will respond with the right error and if not fix it.

peterjan: Considering this works as-is and now involves updating two repositories, do you mind if I create an issue titled "store revisions locally" and refactor this in a F/U? I'll assign the issue to myself of course.

persist/postgres/contracts_test.go
```diff
@@ -828,6 +951,46 @@ func TestMarkUnrenewableContractsBad(t *testing.T) {
 	assertContractGood(false)
 }
 
+func TestMarkBroadcastAttempted(t *testing.T) {
```
ChrisSchinnerl: (nit) doesn't match name of function it's testing

contracts/broadcast.go
ChrisSchinnerl: I think that's fine yeah. Switching from syncing all contracts at the beginning of the maintenance to syncing on demand + persisting revisions will probably take at least 2 PRs in indexd and 1 in coreutils anyway.

ChrisSchinnerl (dismissed):

n8mgr (dismissed):

ChrisSchinnerl (approved):

PR #133, opened 2025-04-24
Add `last_prune` column to contracts table

peterjan:
Getting ready for contract pruning.

### DIFF
--- a/contracts/contract.go
+++ b/contracts/contract.go
@@ -120,6 +120,8 @@ type (
 		InitialAllowance types.Currency `json:"initialAllowance"` // initial renter allowance locked in contract
 		MinerFee         types.Currency `json:"minerFee"`         // miner fee spent on formation txn
 
+		LastPrune time.Time `json:"lastPrune"` // last time the contract was pruned
+
 		// Good determines whether a contract is good or bad. A contract that
 		// is not good, will have its data migrated to a new contract.
 		//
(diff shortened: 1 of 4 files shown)


contracts/contract.go
```diff
@@ -120,6 +120,8 @@ type (
 		InitialAllowance types.Currency `json:"initialAllowance"` // initial renter allowance locked in contract
 		MinerFee         types.Currency `json:"minerFee"`         // miner fee spent on formation txn
 
+		LastPrune time.Time `json:"lastPrune"` // last time the contract was pruned
```
n8mgr, line 123: Do we need this? Can't it be part of the contract maintenance loop that runs every iteration?

peterjan: From the spec:

```
To avoid overwhelming the host, pruning sectors is limited to 1TiB worth of data
per batch. To avoid overwhelming the indexer, we limit pruning a contract to
once per day using a `last_prune` field on the contract within the database.
```

ChrisSchinnerl: Renterd had it in-memory I think but the annoying part is that it's such a heavy operation that you might not want it to start every time you reboot indexd.

peterjan: It's also kind of nice to have in the contract's metadata imo.

ChrisSchinnerl (approved):

n8mgr (approved):

PR #134, opened 2025-04-24
Add UnpinSlab to database

peterjan:


n8mgr (dismissed):

persist/postgres/sectors_test.go
```diff
@@ -262,6 +262,74 @@ func TestPinSlabs(t *testing.T) {
...
+	hk := types.PublicKey{1}
+	ha := chain.NetAddress{Protocol: quic.Protocol, Address: "[::]:4848"}
+	if err := store.UpdateChainState(context.Background(), func(tx subscriber.UpdateTx) error {
+		return tx.AddHostAnnouncement(hk, chain.V2HostAnnouncement{ha}, time.Now())
+	}); err != nil {
+		t.Fatal(err)
+	}
+
+	// add slab
+	slabID, err := store.PinSlab(context.Background(), account, time.Time{}, slabs.SlabPinParams{
```
ChrisSchinnerl: Already left a comment on the other PR but it would be nice to have more than 1 slab here to make sure unpinning a slab doesn't nuke the whole table.

ChrisSchinnerl (changes requested):

n8mgr (approved):

persist/postgres/sectors.go
```diff
@@ -108,6 +108,27 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
 	})
 }
 
+// UnpinSlab removes the association between the given account and the slab. If
+// a slab becomes unreferenced it will eventually be pruned by a background
+// process.
+func (s *Store) UnpinSlab(ctx context.Context, accountID proto.Account, slabID slabs.SlabID) error {
```
n8mgr, line 229: Is it that bad to try and prune the dangling slab references immediately? It should only be 31 references to check. Should be very performant with a batch. Seems like that would simplify pruning laster since you can focus on sectors rather than slabs + sectors.

peterjan: I think the reasoning is that we'd rather have the database be as consistent as possible and avoid having sectors that don't belong to slabs. Especially since we rely on that when we pin slabs now that we deduped sectors. We can play around with this though... I'll add an issue to explore the idea in a F/U?

n8mgr: I don’t want to get in the habit of merging code we’re immediately planning on refactoring.

n8mgr: There’s no difference from a db perspective to have dangling slabs or dangling sectors. If it simplifies the logic for #142, great. If it does nothing, at least we know.

n8mgr (changes requested):
Clean up the dangling slabs to simplify the pruning logic

persist/postgres/sectors.go
peterjan: Sounds good!

peterjan: It's a little annoying because #142 deduplicates the sectors. Since every slab is associated with its own set of sectors here, I have unpin slab remove the sectors if there's no other account referencing the slab. I'll update that again in #142

persist/postgres/sectors.go
```diff
@@ -108,6 +108,62 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
...
+		// fetch account id
+		var aID int64
+		err := tx.QueryRow(ctx, "SELECT id FROM accounts WHERE public_key = $1", sqlPublicKey(accountID)).Scan(&aID)
+		if errors.Is(err, sql.ErrNoRows) {
+			return accounts.ErrNotFound
+		} else if err != nil {
+			return fmt.Errorf("failed to get account id: %w", err)
+		}
+
+		// fetch slab id through the account_slabs table, this ensures the slab
```
peterjan: we could fetch the account id here as well but returning the appropriate error becomes messy

persist/postgres/sectors_test.go
ChrisSchinnerl, line 1: Might still be missing a benchmark for `UnpinSlab`

persist/postgres/sectors.go
ChrisSchinnerl: > There’s no difference from a db perspective to have dangling slabs or dangling sectors. If it simplifies the logic for #142, great. If it does nothing, at least we know.

While both work there is one thing we should consider (but we need a benchmark for it).
`UnpinSlab` is triggered by users and directly affects their experience when deleting a lot of files in their apps. So if we can speed `UnpinSlab` up significantly by pushing more logic into the background job, that might be a worthwhile UX tradeoff. If the speed is the same and we can clean up the background logic by keeping it here, we should do that instead.

ChrisSchinnerl (changes requested):

persist/postgres/sectors_test.go
peterjan: Forgot to push my commit 🤦

persist/postgres/sectors.go
peterjan: delete account_slab + slab:
`BenchmarkUnpinSlab-10                547           2228191 ns/op`

delete account_slab + slab + sectors: 
`BenchmarkUnpinSlab-10                448           2626225 ns/op`

it seems that we don't need to have a background process at all and when we unpin we can check whether that slab was pinned by another account, and remove it if not, and also remove dangling sectors providing they're not linked to any other slab

persist/postgres/sectors.go
```diff
@@ -222,6 +222,71 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
...
+		// providing they're not referenced by any other slab
+		batch := &pgx.Batch{}
+		batch.Queue(`DELETE FROM account_slabs WHERE account_id = $1 AND slab_id = $2`, aID, sID)
+		batch.Queue(`
+		WITH candidate_sectors AS (
+			SELECT ss.sector_id
+			FROM slab_sectors ss
+			WHERE ss.slab_id = $1 AND NOT EXISTS (SELECT 1 FROM slab_sectors ss2 WHERE ss2.sector_id = ss.sector_id AND ss2.slab_id <> $1)
+			)
+			DELETE FROM sectors
```
n8mgr: Will it be easier to implement pruning on the host if we keep these dangling refs around so we know what sectors we can prune?

peterjan: Yeah I thought about that, if we were to keep track of prunable data (total amount) on the contract we would need to either update the contract here, or stop deleting the sectors indeed. In [this comment](https://github.com/SiaFoundation/indexd/pull/142#discussion_r2068179264) though Chris proposed to:

> Short term it might be good enough to just always prune contracts instead of checking a field in our database to see whether it's necessary. Especially if contracts are capped at 10TB in size and will only be pruned every couple of weeks or so.

So that's why I think we can safely delete sectors here and then implement pruning in the most robust way possible, which is diffing every sector. I'm implementing that today, I think it'll be fast enough but if it's not perhaps we should indeed consider your proposal. I'm also fine with leaving the sectors but it's nice to avoid the background process. The contract pruning loop could also be responsible for removing the orphaned sectors from the database but that's not ideal imo.

ChrisSchinnerl: In theory that's very nice but in practice I think we will be missing any sectors that have been migrated.

e.g.
upload sector S1 to contract C1
move sector S1 to contract C2 on different host
information about sector being in C1 is now lost

persist/postgres/sectors.go
```diff
@@ -222,6 +222,71 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
...
+			 WHERE account_id = $1 AND slab_id = (SELECT id FROM slabs WHERE digest = $2)`, aID, sqlHash256(slabID)).Scan(&sID, &shared)
+		if errors.Is(err, sql.ErrNoRows) {
+			return slabs.ErrSlabNotFound
+		} else if err != nil {
+			return fmt.Errorf("failed to get slab id: %w", err)
+		}
+
+		// if slab is shared, only remove the association from the account_slabs
+		if shared {
+			_, err := tx.Exec(ctx, `DELETE FROM account_slabs WHERE account_id = $1 AND slab_id = $2`, aID, sID)
```
ChrisSchinnerl: Would this be a bit simpler if you switch this query and the one above? If you try to delete the slab from `account_slabs` first, you can still perform the not found check.

e.g.
```sql
DELETE FROM account_slabs WHERE account_id = $1 AND slab_id = (SELECT id FROM slabs WHERE digest = $2)
RETURNING slab_id
```

Then you check `RowsAffected` to see if the slab existed for that account and return an error if not.

Afterwards you could do

```sql
SELECT EXISTS (SELECT 1 FROM accounts_slabs WHERE slab_id = $1)
```
To figure out if there are more slabs in there and if not return early.

Seems like simpler queries to me slightly less prone to mistakes. If you don't care about the account error which might be alright considering that you need to authenticate before calling this as a user, you could return the account id from the delete query as well to maybe speed things up a bit more even.

n8mgr (approved):

persist/postgres/sectors.go
peterjan: Ok updated. I think the queries are simpler but it comes at the cost of returning proper errors.

ChrisSchinnerl (approved):

PR #135, opened 2025-04-25
Add SlabSectors

### LINKED ISSUE #129: Add `slab_sectors`
Right now the `sector_root` column of the `sectors` table is not `UNIQUE`. We can fix that and start deduplicating sectors by adding a `slab_sectors` join table between `slabs` and `sectors`.
Without it, things get weird in our data integrity checks and other spots where we pass sector roots as if they uniquely identified sectors. 

The reason why the root wasn't unique in the first place is the edge case where people either
- update a part of a file and reupload for some reason (e.g. versioning)
- people accidentally or on purpose pin invalid slabs

peterjan:
Closes #129

### DIFF
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -189,11 +189,37 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
 
 		// insert slab's sectors in a single batch
 		batch := &pgx.Batch{}
-		for i, sector := range slab.Sectors {
-			batch.Queue(`INSERT INTO sectors (sector_root, host_id, slab_id, slab_index, next_integrity_check) VALUES ($1, (SELECT id FROM hosts WHERE public_key = $2), $3, $4, $5)`, sqlHash256(sector.Root), sqlPublicKey(sector.HostKey), slabID, i, nextIntegrityCheck)
+		for _, sector := range slab.Sectors {
+			batch.Queue(`
+				INSERT INTO sectors (sector_root, host_id, next_integrity_check) 
+				VALUES ($1, (SELECT id FROM hosts WHERE public_key = $2), $3) 
+				ON CONFLICT (sector_root) DO UPDATE SET sector_root=EXCLUDED.sector_root 
+				RETURNING id
+			`, sqlHash256(sector.Root), sqlPublicKey(sector.HostKey), nextIntegrityCheck)
+		}
+
+		// fetch sector IDs
+		br := tx.SendBatch(ctx, batch)
+		sectorIDs := make([]int64, len(slab.Sectors))
+		for i := range slab.Sectors {
+			if err := br.QueryRow().Scan(&sectorIDs[i]); err != nil {
+				br.Close()
+				return fmt.Errorf("failed to get sector id for index %d: %w", i, err)
+			}
+		}
+		br.Close()
+
+		// insert slab sectors into join table
+		batch = &pgx.Batch{}
+		for i, sectorID := range sectorIDs {
+			batch.Queue(`
+				INSERT INTO slab_sectors (slab_id, slab_index, sector_id)
+				VALUES ($1, $2, $3)
+				ON CONFLICT (slab_id, slab_index) DO NOTHING
+			`, slabID, i, sectorID)
 		}
 		if err = tx.Tx.SendBatch(ctx, batch).Close(); err != nil {
-			return fmt.Errorf("failed to insert sectors: %w", err)
+			return fmt.Errorf("failed to insert slab sectors: %w", err)
 		}
 		return nil
 	})
@@ -237,11 +263,12 @@ func (s *Store) Slabs(ctx context.Context, accountID proto.Account, slabIDs []sl
 		for _, slabID := range dbIDs {
 			sectorsBatch.Queue(`SELECT s.sector_root, h.public_key, c.contract_id
 FROM sectors s
+INNER JOIN slab_sectors ss ON s.id = ss.sector_id
 LEFT JOIN hosts h ON h.id = s.host_id
 LEFT JOIN contract_sectors_map csm ON s.contract_sectors_map_id = csm.id
 LEFT JOIN contracts c ON c.contract_id = csm.contract_id
-WHERE s.slab_id = $1
-ORDER BY s.slab_index ASC`, slabID).Query(func(rows pgx.Rows) error {
+WHERE ss.slab_id = $1
+ORDER BY ss.slab_index ASC`, slabID).Query(func(rows pgx.Rows) error {
 				defer rows.Close()
 				for rows.Next() {
 					var sector slabs.Sector
@@ -364,7 +391,8 @@ func (s *Store) UnhealthySlab(ctx context.Context, maxRepairAttempt time.Time) (
 			WHERE id = (
 				SELECT slabs.id
 				FROM slabs
-				INNER JOIN sectors ON slabs.id = sectors.slab_id
+				INNER JOIN slab_sectors ON slabs.id = slab_sectors.slab_id
+				INNER JOIN sectors ON sectors.id = slab_sectors.sector_id 
 				LEFT JOIN contract_sectors_map csm ON sectors.contract_sectors_map_id = csm.id
 				LEFT JOIN contracts ON csm.contract_id = contracts.contract_id
 				WHERE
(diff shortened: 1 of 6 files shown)


persist/postgres/sectors.go
```diff
@@ -189,11 +189,37 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
...
+		// fetch sector IDs
+		br := tx.SendBatch(ctx, batch)
+		sectorIDs := make([]int64, len(slab.Sectors))
+		for i := range slab.Sectors {
+			if err := br.QueryRow().Scan(&sectorIDs[i]); err != nil {
+				br.Close()
+				return fmt.Errorf("failed to get sector id for index %d: %w", i, err)
+			}
+		}
+		br.Close()
```
n8mgr: You can simplify a bit and get rid of the `Close` calls in the loop.

```suggestion
		sectorIDs := make([]int64, len(slab.Sectors))
		for i, sector := range slab.Sectors {
			batch.Queue(`
				INSERT INTO sectors (sector_root, host_id, next_integrity_check) 
				VALUES ($1, (SELECT id FROM hosts WHERE public_key = $2), $3) 
				ON CONFLICT (sector_root) DO UPDATE SET sector_root=EXCLUDED.sector_root 
				RETURNING id
			`, sqlHash256(sector.Root), sqlPublicKey(sector.HostKey), nextIntegrityCheck).QueryRow(func(row pgx.Row) error {
				return row.Scan(&sectorIDs[i])
			})
		}

		// fetch sector IDs
		if err := tx.SendBatch(ctx, batch).Close(); err != nil {
			return fmt.Errorf("failed to insert sectors: %w", err)
		}
```

persist/postgres/sectors.go
```diff
@@ -237,11 +263,12 @@ func (s *Store) Slabs(ctx context.Context, accountID proto.Account, slabIDs []sl
 		for _, slabID := range dbIDs {
 			sectorsBatch.Queue(`SELECT s.sector_root, h.public_key, c.contract_id
 FROM sectors s
+INNER JOIN slab_sectors ss ON s.id = ss.sector_id
```
n8mgr, line 314: Might be nice to benchmark a few of these scenarios with natural keys and no joins in a f/u.

peterjan: Could definitely F/U with a benchmark using natural keys, but the join can not be avoided because we deduplicate the sector. Perhaps I'm not fully understanding what you are suggesting here.

persist/postgres/sectors_test.go
ChrisSchinnerl, line 1: Could you run all the benchmarks again to see if there are any regressions?

peterjan: Definitely, ran into a bit of a hickup with benchmarks but I'll get this sorted.

peterjan: Ok so the hickup is/was that some benchmarks run forever on my machine. I got this sorted but only through making the benchmarks exhaustive. Let's discuss that separately though. There is indeed a regression, quite a serious on in `BenchmarkUnhealthySlab`.. it's more than twice as slow now. I'll open a comment on the line itself to discuss this further.

persist/postgres/sectors.go
```diff
@@ -356,28 +382,45 @@ func (s *Store) UnpinnedSectors(ctx context.Context, hostKey types.PublicKey, li
...
+		INNER JOIN contract_sectors_map csm ON sectors.contract_sectors_map_id = csm.id
+		INNER JOIN contracts ON csm.contract_id = contracts.contract_id
+		WHERE slab_sectors.slab_id = slabs.id
+		AND contracts.good = FALSE`
+
+	if frand.Intn(2) == 0 {
+		conditionA, conditionB = conditionB, conditionA
+	}
+
+	existsExpr := strings.Join([]string{conditionA, conditionB}, " UNION ALL ")
```
peterjan: @ChrisSchinnerl @n8maninger I'm in a bit of a mess here with introducing `slab_sectors`, the problem is that there is definitely performance regression. 

Currently, `BenchmarkUnhealthySlab` claims it runs in 1.3ms, this is either incorrect or we experienced a regression over time and failed to update the benchmark results. On my machine, on `master`, it runs in 29.5ms. This branch however is close to 50ms... so quite the hit. I must say though that the original benchmark probably also short-circuited a bunch of slabs because of the original `WHERE ... OR ...` expression 

I (temporarily) switch between the two conditions 50-50. The main issue is that there's a huge difference between the performance of the first and second condition. If we were to only focus on bad contracts, the benchmark takes 1.5ms, while if we only focus on sectors not being stored on any hosts it takes 70ms. 

I split the `WHERE` into `EXISTS + UNION ALL` to avoid the `LEFT JOIN` and I checked for indices to add, which I did: 

```
-- speeds up lookup of unhealthy slabs
CREATE INDEX slabs_id_last_repair_attempt_idx ON slabs(last_repair_attempt ASC);
CREATE INDEX slab_sectors_sector_id_idx ON slab_sectors(sector_id);
CREATE INDEX sectors_id_idx ON sectors(id) INCLUDE (host_id, contract_sectors_map_id);
CREATE INDEX sectors_id_host_null_idx ON sectors(id) WHERE host_id IS NULL;
CREATE INDEX sectors_contract_sectors_map_id_host_id_idx ON sectors(contract_sectors_map_id, host_id);
CREATE INDEX contracts_contract_id_good_idx ON contracts(contract_id, good);
```

but I can't get it back to 30ms, or even below that obviously. Not sure how to proceed here. I have a couple of ideas of course, likely involving a materialized view but wanted to get your 4c before attempting to work my way around this. I'm not really opposed to doing this 50-50 in production by the way, I guess we would normally want to move away from bad contracts first, but we also don't want to starve the other condition.

n8mgr: Alternative to the view or `UNION`, you might try running each query separately in a batch then picking the first valid result. If no results are returned, then return the not found error. Might be quicker than union all.

persist/postgres/sectors.go
```diff
@@ -386,3 +396,16 @@ func (s *Store) UnhealthySlab(ctx context.Context, maxRepairAttempt time.Time) (
...
 	return slabID, err
 }
+
+// RefreshUnhealthySlabs refreshes a materialized view that contains slabs with
+// at least one sector that needs to be migrated to a new host. The condition
+// for such a sector is that it's either not stored on a host, or stored in a
+// bad contract. This function is meant to be called periodically to keep the view
+// up to date. The view is refreshed concurrently to avoid blocking other
+// transactions.
+func (s *Store) RefreshUnhealthySlabs(ctx context.Context) error {
```
ChrisSchinnerl: My preference is for this to be a background loop within the store because

- It avoids exposing implementation details of the store and makes the API simpler
- It removes the burden of potentially blocking for a minute on a large database from the caller

persist/postgres/sectors.go
peterjan: Tried it, was still slow. Added the view.

persist/postgres/sectors.go
peterjan: I don't really agree because it's an avoidable long-running background process. We don't really mind that the loop in the slab manager blocks for a minute (it has to be a very large database to get up to a minute), if we only update the materialized view right before doing migrations. I also kind of like that the slab manager decides when to refresh the view rather than have the store define an interval. I updated the code though, I really don't mind it, just prefer calling it explicitly over the background loop.

ChrisSchinnerl (dismissed):

ChrisSchinnerl (dismissed):

persist/postgres/sectors.go
```diff
@@ -344,17 +419,12 @@ func (s *Store) UnpinnedSectors(ctx context.Context, hostKey types.PublicKey, li
...
+// UnhealthySlab returns an unhealthy slab that hasn't had a repair attempted
+// since 'maxRepairAttempt'. A slab is considered unhealthy if it has at least
+// one sector that is not stored on a host or stored in a bad contract. When no
+// slab is found, ErrSlabNotFound is returned. If a slab is found, it will have
+// its last_repair_attempt updated to the time of the call. To prevent
 // subsequent or parallel calls from returning the same slab.
-//
-// NOTE: For the sake of scalability, we don't prioritize any slabs and instead
-// simply fetch the first one that we can get.
 func (s *Store) UnhealthySlab(ctx context.Context, maxRepairAttempt time.Time) (slabs.SlabID, error) {
```
n8mgr: Let's revert the change to `UnhealthySlabs` and deal with the performance degradation in a f/u. This PR has been open for a while and I'm not 100% happy with introducing a materialized view for data that really should be live. If we're all amenable to that, I'm fine with merging the rest of the PR during your day.

ChrisSchinnerl (approved):

peterjan:
@ChrisSchinnerl let's merge it? I think Nate was fine to merge providing I reverted the changes to `UnhealthySlabs`

PR #139, opened 2025-04-28
Perform integrity checks

ChrisSchinnerl:
This PR implements `performIntegrityChecks`

### DIFF
--- a/slabs/dataintegrity.go
+++ b/slabs/dataintegrity.go
@@ -52,6 +52,10 @@ func (v *sectorVerifier) VerifySector(ctx context.Context, prices proto.HostPric
 	return rhp.RPCVerifySector(ctx, v.tc, prices, token, root)
 }
 
+// verifySectors verifies a list of sectors on a host. If verifySectors returns
+// either errInsufficientServiceAccountBalance or context.Canceled, the caller
+// should stop handle any remaining results and then interrupt the integrity
+// checks for the host.
 func (c *SlabManager) verifySectors(ctx context.Context, hc SectorVerifier, host hosts.Host, roots []types.Hash256) ([]CheckSectorsResult, error) {
 	// check the account balance
 	cost := host.Settings.Prices.RPCVerifySectorCost().RenterCost()
--- a/slabs/dataintegrity.go
+++ b/slabs/dataintegrity.go
@@ -56,10 +56,10 @@ func (v *sectorVerifier) VerifySector(ctx context.Context, prices proto.HostPric
 // either errInsufficientServiceAccountBalance or context.Canceled, the caller
 // should stop handle any remaining results and then interrupt the integrity
 // checks for the host.
-func (c *SlabManager) verifySectors(ctx context.Context, hc SectorVerifier, host hosts.Host, roots []types.Hash256) ([]CheckSectorsResult, error) {
+func (m *SlabManager) verifySectors(ctx context.Context, hc SectorVerifier, host hosts.Host, roots []types.Hash256) ([]CheckSectorsResult, error) {
 	// check the account balance
 	cost := host.Settings.Prices.RPCVerifySectorCost().RenterCost()
-	balance, err := c.am.ServiceAccountBalance(ctx, host.PublicKey, c.serviceAccount)
+	balance, err := m.am.ServiceAccountBalance(ctx, host.PublicKey, m.serviceAccount)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get service account balance: %w", err)
 	} else if balance.Cmp(cost) < 0 {
@@ -82,14 +82,14 @@ func (c *SlabManager) verifySectors(ctx context.Context, hc SectorVerifier, host
 		}
 
 		// verify the sector
-		_, err := hc.VerifySector(ctx, host.Settings.Prices, c.serviceAccount.Token(c.serviceAccountKey, host.PublicKey), root)
+		_, err := hc.VerifySector(ctx, host.Settings.Prices, m.serviceAccount.Token(m.serviceAccountKey, host.PublicKey), root)
 		if errors.Is(err, context.Canceled) {
 			return results, err // interrupted
 		}
 
 		// adjust balance
 		balance = balance.Sub(cost)
-		if err := c.am.DebitServiceAccount(ctx, host.PublicKey, c.serviceAccount, cost); err != nil {
+		if err := m.am.DebitServiceAccount(ctx, host.PublicKey, m.serviceAccount, cost); err != nil {
 			return nil, fmt.Errorf("failed to debit service account: %w", err)
 		}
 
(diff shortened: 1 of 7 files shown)


slabs/manager.go
```diff
@@ -86,27 +116,93 @@ func NewManager(store Store, opts ...Option) (*SlabManager, error) {
...
 }
 
-func (m *SlabManager) performIntegrityChecks() error {
+func (m *SlabManager) performIntegrityChecks(ctx context.Context) error {
 	start := time.Now()
 	logger := m.log.Named("integrity")
 	logger.Debug("starting integrity checks", zap.Time("start", start))
 
-	// TODO: fetch sectors for integrity checks
+	usedHosts, err := m.store.Hosts(ctx, 0, math.MaxInt64,
```
n8mgr: Why not grab the sectors to check instead of starting with getting all active hosts? There's no guarantee a host has sectors so we're pulling data from the database for no reason. 

If you really wanted to keep it, you could introduce a query to ensure the host actually has sectors to check.

slabs/dataintegrity.go
```diff
@@ -0,0 +1,224 @@
...
+		for i, result := range results {
+			switch result {
+			case SectorLost:
+				lost = append(lost, toCheck[i])
+			case SectorFailed:
+				failed = append(failed, toCheck[i])
+			case SectorSuccess:
+				success = append(success, toCheck[i])
+			default:
+				hostLogger.Fatal("unknown result", zap.Int("result", int(result)))
```
peterjan, line 122: `panic` on developer error?

slabs/dataintegrity.go
```diff
@@ -0,0 +1,224 @@
...
+				return fmt.Errorf("failed to record integrity check for failed sectors: %w", err)
+			}
+			if err := m.store.RecordIntegrityCheck(ctx, true, time.Now().Add(m.integrityCheckInterval), host.PublicKey, success); err != nil {
+				hostLogger.Error("failed to record integrity check for successful sectors", zap.Error(err))
+				return fmt.Errorf("failed to record integrity check for successful sectors: %w", err)
+			}
+
+			// fetch sector roots for sectors that have now failed the check 5+
+			// times and mark them lost as well
+			for {
```
peterjan: Fetching sector roots to then mark them as lost in a loop is a bit wasteful. I would add `MarkFailingSectorsLost` where we pass `maxFailedIntegrityChecks` as threshold.

slabs/dataintegrity.go
```diff
@@ -0,0 +1,224 @@
...
+		if err != nil {
+			hostLogger.Error("failed to persist integrity check results", zap.Error(err))
+			return
+		}
+	}
+}
+
+// verifySectors verifies a list of sectors on a host. If verifySectors returns
+// either errInsufficientServiceAccountBalance or context.Canceled, the caller
+// should stop handle any remaining results and then interrupt the integrity
```
peterjan: weird sentence

slabs/dataintegrity.go
```diff
@@ -0,0 +1,224 @@
...
+
+		// verify the sector
+		_, err := hc.VerifySector(ctx, host.Settings.Prices, m.serviceAccount.Token(m.serviceAccountKey, host.PublicKey), root)
+		if errors.Is(err, context.Canceled) {
+			return results, err // interrupted
+		}
+
+		// adjust balance
+		balance = balance.Sub(cost)
+		if err := m.am.DebitServiceAccount(ctx, host.PublicKey, m.serviceAccount, cost); err != nil {
```
peterjan: I wonder if it's necessary to debit the service account for every root individually. That seems excessive. We could fetch the balance, multiply the cost b the number of sectors, potentially limit the amount of roots we check, debit the service account once and continue with verifying sectors.

slabs/manager.go
```diff
@@ -86,27 +116,93 @@ func NewManager(store Store, opts ...Option) (*SlabManager, error) {
...
+			// fetch good price table
+			host, err = m.hm.ScanHost(ctx, hostKey)
+			if err != nil {
+				logger.With(zap.Stringer("hostKey", hostKey)).Error("failed to scan host", zap.Error(err))
+				return
+			}
 
-	// TODO: mark successes/failures in database
+			// create verifier
+			verifier, err := newSectorVerifier(ctx, host.SiamuxAddr(), host.PublicKey)
```
peterjan: I'd consider passing the host prices to the sector verifier. You'd have to extend the interface to expose the VerifySectorCost but it would simplify the method signature a bit and we don't update the prices anyway.

slabs/manager.go
```diff
@@ -86,27 +116,93 @@ func NewManager(store Store, opts ...Option) (*SlabManager, error) {
...
+				// Instead we leave it up to the scan code to determine whether the host
+				// is offline.
+				logger.With(zap.Stringer("hostKey", host.PublicKey)).Warn("failed to create sector verifier", zap.Error(err))
+				return
+			}
+			defer verifier.Close()
 
-	// TODO: fetch sector roots for sectors that have now failed the check 3+
-	// times and mark them lost as well
+			m.performIntegrityChecksForHost(ctx, verifier, host, logger)
```
peterjan: If we move the prices to the verifier, we only need to pass the host key here. I would also probably require a logger in the goroutine and already pass the host logger, that saves us `logger.With(zap.Stringer("hostKey", host.PublicKey)` twice here and `performIntegrityChecksForHost` can use the logger that's passed in

slabs/manager.go
ChrisSchinnerl: > Why not grab the sectors to check instead of starting with getting all active hosts? There's no guarantee a host has sectors so we're pulling data from the database for no reason.
I think we had something like that in the initial RFC draft. But the required sorting of the sectors by host to then perform the checks on multiple hosts in parallel didn't spark joy.

I'll add a `HostsForIntegrityChecks` interface method for which I'll add the database implementation in a f/u. That should be super fast since we'll only need the public keys to be returned rather than the full hosts.

slabs/dataintegrity.go
ChrisSchinnerl: Won't `Fatal` also crash?

peterjan: missed it 👍

slabs/dataintegrity.go
ChrisSchinnerl: Added the interface method and will implement it in a f/u since this PR is massive already

slabs/dataintegrity.go
ChrisSchinnerl: I'd like to see how this goes and adjust if necessary. The more we batch, the higher the potential drift when we fund the account in the meantime.

slabs/manager.go
n8mgr: I do not want to get in the habit of merging code we're planning on replacing immediately. You can always add the database methods in another PR and merge that into master first if you really want to keep the scope of this PR clean.

peterjan (approved):

slabs/dataintegrity.go
peterjan: Verifying 1TB worth is like 250k database writes. That's a lot of writes 😁 but sure.

slabs/manager.go
ChrisSchinnerl: Me neither. I meant adding the interface method and mock in this PR (which I did [here](https://github.com/SiaFoundation/indexd/pull/139/commits/2e71099d89da0eb06206dcc3918682a3827f7d45)) since we need it in this PR and then implementing the production code separately. So nothing gets changed.

n8mgr (approved):

PR #142, opened 2025-04-29
Add slab pruning

peterjan:
This PR extends the slab manager to prune slabs


persist/postgres/sectors_test.go
```diff
@@ -380,6 +380,74 @@ func TestPinSlabs(t *testing.T) {
...
+	// add host
+	hk := types.PublicKey{1}
+	ha := chain.NetAddress{Protocol: quic.Protocol, Address: "[::]:4848"}
+	if err := store.UpdateChainState(context.Background(), func(tx subscriber.UpdateTx) error {
+		return tx.AddHostAnnouncement(hk, chain.V2HostAnnouncement{ha}, time.Now())
+	}); err != nil {
+		t.Fatal(err)
+	}
+
+	// add slab
```
ChrisSchinnerl: I'd add multiple slabs and make sure the number of slabs is correct as well as the right slab getting unpinned.. Right now this would also pass if `UnpinSlab` did `DELETE FROM account_slabs`. 

e.g. 4 slabs total over 2 accounts

persist/postgres/slabs.go
```diff
@@ -0,0 +1,141 @@
+package postgres
+
+import (
+	"context"
+	"errors"
+	"fmt"
+)
+
+const slabPruneBatchSize = 35
```
ChrisSchinnerl, line 10: How come this is so oddly specific? Maybe leave a comment to explain it?

persist/postgres/slabs.go
```diff
@@ -0,0 +1,141 @@
...
+// redundancy) and will continue until either the slabs to prune are exhausted
+// or the context is cancelled. It will return the total number of slabs pruned,
+// or an error if one occurred.
+func (s *Store) PruneSlabs(ctx context.Context) (int64, error) {
+
+	var pruned int64
+	for {
+		n, err := s.pruneSlabs(ctx, slabPruneBatchSize)
+		if err != nil {
+			if pruned > 0 && errors.Is(err, context.Canceled) {
```
ChrisSchinnerl: I think it's fine to do

```go
if err != nil {
  return pruned, err
}
```

If a method is expected to be cancellable, we can expect `context.Canceled` to be returned when it gets cancelled and not just when some edge case is met like `pruned > 0`.

persist/postgres/slabs.go
```diff
@@ -0,0 +1,141 @@
...
+			return fmt.Errorf("failed to fetch slabs to prune: %w", err)
+		} else if len(slabIDs) == 0 {
+			return nil
+		}
+
+		counts, err := prunableSectorCounts(ctx, tx, slabIDs)
+		if err != nil {
+			return fmt.Errorf("failed to fetch prunable sector counts: %w", err)
+		}
+		_ = counts // TODO: increment prunable data on the contract
```
ChrisSchinnerl, line 47: Not relevant to this PR but it might be easier to go the other way store the amount of used storage on a contract rather than the prunable storage. Then we don't run into a race when pruning a host while at the same time deleting sectors from our database.

Short term it might be good enough to just always prune contracts instead of checking a field in our database to see whether it's necessary. Especially if contracts are capped at 10TB in size and will only be pruned every couple of weeks or so.

persist/postgres/slabs.go
```diff
@@ -0,0 +1,139 @@
...
+			SELECT 1
+			FROM slab_sectors ss_in
+			WHERE ss_in.sector_id = s.id
+			AND ss_in.slab_id = ANY ($1)
+		)
+		AND NOT EXISTS (
+			SELECT 1
+			FROM slab_sectors ss_other
+			WHERE ss_other.sector_id = s.id
+			AND ss_other.slab_id <> ALL ($1)
```
ChrisSchinnerl, line 61: It may be faster to do a `RETURNING s.contract_id` here and drop the `prunableSectorCounts` query altogether. 

Then you can read the rows you deleted, group them by contract id in a map and figure out how much data you deleted from each contract.

slabs/manager.go
```diff
@@ -80,6 +81,10 @@ func NewManager(store Store, opts ...Option) (*SlabManager, error) {
 			if err := m.performSlabMigrations(); err != nil {
 				m.log.Error("failed to perform slab migrations", zap.Error(err))
 			}
+
+			if err := m.performSlabPruning(ctx); err != nil {
```
ChrisSchinnerl: Might be good to have that first to avoid migrating or integrity checking anything that we are going to delete a second later.

ChrisSchinnerl (changes requested):

persist/postgres/slabs.go
peterjan: Just roughly 1k sectors. At greater batch sizes I ran into slow query statements on delete calls that can't be further optimized.

persist/postgres/sectors.go
```diff
@@ -222,6 +222,27 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
 	})
 }
 
+// UnpinSlab removes the association between the given account and the slab. If
+// a slab becomes unreferenced it will eventually be pruned by a background
```
n8mgr: Is it that bad to try and prune the dangling slab reference immediately? It should only be 31 references to check. Should be very performant with a batch. Seems like that would simplify pruning laster since you can focus on sectors rather than slabs + sectors.

persist/postgres/slabs.go
peterjan: Resolving in favour of https://github.com/SiaFoundation/indexd/pull/142#discussion_r2068179264

persist/postgres/slabs.go
peterjan: Ok! I'll tackle these in the F/U that adds prunable data to the contract or whatever we end up deciding on. Fine to leave it as a TODO for this PR I think.

persist/postgres/sectors.go
peterjan: resolving in favour of https://github.com/SiaFoundation/indexd/pull/134#discussion_r2069749838

peterjan:
This ends up being redundant if we delete everything when we unpin a slab.

peterjan:
I'm closing this out for now. Pruning slabs when we unpin is fast enough, I don't think we need a background process.

Issue #137, opened 2025-04-28
Run benchmarks on CI


peterjan:
We should run our benchmarks on CI and catch performance regressions early. Not only that, but the benchmarks have a lot of database code which, if not run frequently, becomes outdated quite quickly, leaving us with broken benchmarks. Apparently there's numerous solutions though, the two main solutions are `benchstat` and https://github.com/rhysd/github-action-benchmark.

Example job

```
- name: Install benchstat
  run: go install golang.org/x/perf/cmd/benchstat@latest

- name: Fetch baseline benchmarks
  run: |
    git fetch origin main
    git checkout origin/main -- benchmark.txt
    mv benchmark.txt baseline.txt

- name: Compare benchmark results
  run: |
    benchstat baseline.txt benchmark.txt > comparison.txt
    cat comparison.txt
```

More intricate example:

```
name: Benchmark Check

on:
  push:
    branches: [ main ]
  pull_request:
    branches: [ main ]

jobs:
  benchmark:
    runs-on: ubuntu-latest

    steps:
      - name: Checkout code
        uses: actions/checkout@v4

      - name: Set up Go
        uses: actions/setup-go@v5
        with:
          go-version: '1.22'

      - name: Install benchstat
        run: go install golang.org/x/perf/cmd/benchstat@latest

      - name: Run benchmarks
        run: go test -bench=. -benchmem ./... > new_bench.txt

      - name: Fetch baseline benchmark
        run: |
          git fetch origin main
          git checkout origin/main -- baseline_bench.txt || echo "No baseline found, skipping"

      - name: Compare benchmarks
        run: |
          if [ -f baseline_bench.txt ]; then
            benchstat baseline_bench.txt new_bench.txt > benchdiff.txt
            cat benchdiff.txt
          else
            echo "No baseline to compare. Skipping comparison."
          fi

      - name: Fail on regression
        run: |
          if [ -f benchdiff.txt ]; then
            if grep -E '([5-9][0-9]%|[1-9][0-9][0-9]+%)' benchdiff.txt; then
              echo "Performance regression detected! 🔥"
              exit 1
            else
              echo "No significant regression detected. ✅"
            fi
          else
            echo "No benchmark comparison done."
          fi

      - name: Save latest benchmark as artifact
        uses: actions/upload-artifact@v4
        with:
          name: latest-benchmark
          path: new_bench.txt

```

PR #143, opened 2025-04-30
Run benchmarks on CI

peterjan:
This PR runs the benchmarks on our CI and saves the results as a GH artifact which we keep around for a month. The goal was to do a baseline comparison and catch regressions early. This proved to be pretty tricky. I think for that to be possible we would need to restructure some of our benchmarks but more importantly probably get a dedicated runner to get consistent results. This at the very least ensures the code in our benchmarks doesn't drift from the codebase, which happened quite a few times already. 

 https://github.com/SiaFoundation/indexd/issues/137


persist/postgres/sectors_test.go
```diff
@@ -952,8 +952,12 @@ func TestUnpinnedSectors(t *testing.T) {
...
 // Hardware |     Benchmark   |  ms/op  |  Throughput   |
+// M1 Max   |     PinSlab     |  3.8ms  | 10760.68 MB/s |
 // M2 Pro   |     PinSlab     |  1.4ms  | 26590.43 MB/s |
 //
+// M1 Max   |   Slabs-40MiB   |   1.8ms | 24220.77 MB/s |
+// M1 Max   |   Slabs-400MiB  |   3.1ms | 13591.64 MB/s |
+// M1 Max   |   Slabs-4GiB    |  12.5ms |  3344.74 MB/s |
 // M2 Pro   |   Slabs-40MiB   |   0.5ms | 64021.45 MB/s |
 // M2 Pro   |   Slabs-400MiB  |   0.8ms | 40447.43 MB/s |
 // M2 Pro   |   Slabs-4GiB    |   3.3ms |  9930.88 MB/s |
```
n8mgr: If we're going to add benchmarks, lets get rid of these and use the runner as the source of truth.

peterjan: That makes sense. It's a little unfortunate the results on the CI are somewhat inconsistent between runs, but overall they give a good indication of the performance. It's also quite a pain to update these and they rarely have proven useful, at least in my experience.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.23'
+
+      - name: Run benchmarks
+        run: |
+          set -euo pipefail
+      
+          if ! go test -run='^$' -benchmem -benchtime=10x -bench=. -timeout=1h ./... 2>&1 | tee indexd_benchmarks.txt; then
```
ChrisSchinnerl: Ngl it doesn't spark a ton of joy that our sweet and fast CI which usually passes in 1 minute right now takes 18 minutes to run the benchmarks 😅

peterjan: I considered running them when we merge into master instead of on every commit. I didn't do that because it kind of defeats the purpose a bit of not merging anything into master that breaks the code in our benchmarks. The setup in our benchmarks is simply quite heavy. I also considered doing `benchtime=1x` but I felt uneasy doing that too since we use the CI as SSOT now. I don't know, this PR was quite the pain in the ass and it didn't turn out as I had hoped. I still think we need to run our benchmarks on the CI though, numerous benchmarks broke over time and running them on our own machines makes it harder to spot regressions.

peterjan: I think it's probably fine to only run the benchmarks on `master` and then fix the issue if we end up merging something that breaks them. If we do so, we can also increase the benchtime to `200x`. I made it `10x` to not be completely worthless and also not take more than 20 minutes. Another thing I considered was running them in parallel somehow but that probably doesn't help things either.

ChrisSchinnerl (dismissed):

n8mgr (dismissed):

ChrisSchinnerl (approved):

PR #145, opened 2025-04-30
Add service accounts to account manager

ChrisSchinnerl:
This PR implements service accounts. Service accounts are regular accounts for internal use which have their balances tracked.

There is a bit of a race between funding an account and withdrawing from it but we accept that race since we should at most be off by 1 withdrawal. Meaning that an account might have the balance 1SC when it actually has slightly less. 
The idea is that when we run into an insufficient balance error on an RPC, we reset our local balance to 0 which pauses integrity checks until we manage to fund the account again.

That means a host will only fail the integrity check once for an insufficient balance per time we fund the account.Given that they have 5 tries before a sector is considered lost, this should work nicely. It also allows us to get rid of the locking I initially had in this PR.

### DIFF
--- a/accounts/manager.go
+++ b/accounts/manager.go
@@ -26,8 +26,8 @@ type (
 		Host(ctx context.Context, hostKey types.PublicKey) (hosts.Host, error)
 		HostAccountsForFunding(ctx context.Context, hk types.PublicKey, limit int) ([]HostAccount, error)
 		UpdateHostAccounts(ctx context.Context, accounts []HostAccount) error
-		UpdateServiceAccountBalance(ctx context.Context, account LockedServiceAccount, balance types.Currency) error
-		ServiceAccountBalance(ctx context.Context, account LockedServiceAccount) (types.Currency, error)
+		UpdateServiceAccountBalance(ctx context.Context, hostKey types.PublicKey, account proto.Account, balance types.Currency) error
+		ServiceAccountBalance(ctx context.Context, hostKey types.PublicKey, account proto.Account) (types.Currency, error)
 	}
 
 	// AccountFunder defines an interface to fund accounts.
@@ -43,7 +43,7 @@ type (
 		log        *zap.Logger
 
 		serviceAccountsMu sync.Mutex
-		serviceAccounts   map[proto.Account]map[types.PublicKey]*serviceAccount
+		serviceAccounts   map[proto.Account]map[types.PublicKey]struct{}
 	}
 )
 
@@ -60,7 +60,7 @@ func WithLogger(l *zap.Logger) Option {
 // NewManager creates a new AccountManager.
 func NewManager(store Store, funder AccountFunder, opts ...Option) *AccountManager {
 	m := &AccountManager{
-		serviceAccounts: make(map[proto.Account]map[types.PublicKey]*serviceAccount),
+		serviceAccounts: make(map[proto.Account]map[types.PublicKey]struct{}),
 		store:           store,
 		funder:          funder,
 		fundTarget:      types.Siacoins(1),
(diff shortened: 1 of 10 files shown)


accounts/manager.go
```diff
@@ -95,15 +104,45 @@ func (m *AccountManager) FundAccounts(ctx context.Context, host hosts.Host, cont
...
+			}
+
+			// update service accounts
+			fundedAccs := make(map[proto.Account]struct{})
+			for _, acc := range accounts[:funded] {
+				fundedAccs[acc.AccountKey] = struct{}{}
+			}
+			for _, serviceAccount := range serviceAccounts {
+				if _, ok := fundedAccs[serviceAccount.Account]; ok {
+					if err := m.UpdateServiceAccountBalance(ctx, serviceAccount, m.fundTarget); err != nil {
```
peterjan: we're setting the target as the balance here, but the actual host balance might be higher, probably fine but still a little bit awkward, I might fund it with 10SC and then with 1SC and the eventual service account balance will be 1 SC even though the ephemeral account has 10

accounts/manager.go
```diff
@@ -95,15 +104,45 @@ func (m *AccountManager) FundAccounts(ctx context.Context, host hosts.Host, cont
...
+			// lock service accounts
+			serviceAccounts := m.lockServiceAccounts(accounts)
+			defer func() {
+				for _, acc := range serviceAccounts {
+					acc.Unlock()
+				}
+			}()
+
+			// fund accounts
+			funded, drained, err = m.funder.FundAccounts(ctx, host, accounts, contractIDs, m.fundTarget, log)
```
peterjan: There's very tight coupling now between a service account and an account. E.g. for a service account to get funded, it has to be in the `accounts` table, which also means it'll get returned by the `/accounts` endpoint. In my head these were completely separate.

accounts/serviceaccounts.go
```diff
@@ -0,0 +1,106 @@
...
+		serviceAccount: acc,
+		Account:        account,
+		HostKey:        hostKey,
+	}, nil
+}
+
+// RegisterServiceAccount signals to the account manager that a specific
+// account is for internal use. The account first needs to be added. A service
+// account will have its balance tracked.
+func (m *AccountManager) RegisterServiceAccount(account proto.Account) {
```
peterjan, line 13: we'll probably need to be able to unregister them as well at some point

peterjan, line 13: Won't we get into trouble with the asynchronicity between registering a service account and having it funded?

ChrisSchinnerl: I don't see a reason to ever unregister a service account.

accounts/manager.go
ChrisSchinnerl: Reusing logic seems to make sense here. It also allows our own accounts to effectively hide amongst user accounts for the host to not be able to tell which ones are ours.

accounts/manager.go
ChrisSchinnerl: Why would it be 10SC if we only ever use the replenish RPC? I don't think this is something we need to handle. Especially since it doesn't cause us to waste any funds.

accounts/serviceaccounts.go
ChrisSchinnerl: I don't think so. Every manager that needs a service account should add that and register it within its constructor.
The contract manager is created last so the maintenance won't happen before all the service accounts are registered.

peterjan (dismissed):
LGTM 👍

n8mgr (approved):

accounts/serviceaccounts.go
```diff
@@ -0,0 +1,76 @@
...
+// DebitServiceAccount withdraws from a service account. This should be used
+// after successfully withdrawing from an account.
+func (m *AccountManager) DebitServiceAccount(ctx context.Context, hostKey types.PublicKey, account proto.Account, amount types.Currency) error {
+	if !m.serviceAccountExists(account) {
+		return ErrNotFound
+	}
+	return m.store.DebitServiceAccount(ctx, hostKey, account, amount)
+}
+
+func (m *AccountManager) serviceAccountExists(account proto.Account) bool {
```
peterjan, line 49: `nit` `isServiceAccount`

accounts/serviceaccounts_test.go
```diff
@@ -0,0 +1,144 @@
...
+	return accs[hostKey], nil
+}
+
+// DebitServiceAccount withdraws from a service account.
+func (s *mockStore) DebitServiceAccount(ctx context.Context, hostKey types.PublicKey, account proto.Account, amount types.Currency) error {
+	balance, err := s.serviceAccount(hostKey, account)
+	if err != nil {
+		return err
+	}
+	if balance.Cmp(amount) < 0 {
```
peterjan, line 30: we can just SubWithUnderflow and ignore the bool

peterjan (approved):

accounts/serviceaccounts_test.go
ChrisSchinnerl: No, the result of `SubWithUnderflow` is actually the underflown value and not 0.

PR #146, opened 2025-05-05
Get indexd ready to be deployed

ChrisSchinnerl:
I'm deploying indexd on my personal server right now and am adding any findings to this PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,4 +35,4 @@ EXPOSE 9980/tcp
 # Syncer port
 EXPOSE 9981/tcp
 
-ENTRYPOINT [ "indexd" ]
+ENTRYPOINT [ "indexd", "-api.admin", ":9980" ]
--- a/pins/manager.go
+++ b/pins/manager.go
@@ -39,10 +39,9 @@ type (
 		updatePriceFrequency time.Duration
 		rateWindow           time.Duration
 
-		mu       sync.Mutex
-		rates    []decimal.Decimal
-		average  decimal.Decimal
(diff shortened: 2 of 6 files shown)


Dockerfile
ChrisSchinnerl, line 1: One thing that is kind of awkward is setting up indexd on UI driven OSs and tools. Because you can't just configure the docker container and then run it to get a working node. Instead, you always have to use the CLI to generate a valid config which might not be very convenient depending on the platform you are using.

Maybe we should at least have env vars for the seed, network and database connection to allow people to get started with the default settings for everything else. Once the container is running successfully at least once, it's a lot easier to do a `docker exec` to generate a new config and restart.

n8mgr: After you create the container you can do `docker run -it <container name> config` regardless of whether the initial startup completed.

pins/manager.go
```diff
@@ -272,16 +274,10 @@ func (pm *PinManager) updatePrices(ctx context.Context, force bool, log *zap.Log
 // addRate adds the rate in the given currency to the list and returns a boolean
 // whether the prices should be updated, this happens when the average rate
 // exceeds a certain threshold.
-func (pm *PinManager) addRate(currency string, rate float64) bool {
```
peterjan: update docstring to reflect the change

peterjan: why is it safe to remove this behaviour?

peterjan (approved):

n8mgr (approved):

PR #149, opened 2025-05-06
Add MigrateSector method to database

ChrisSchinnerl:
Used for marking a sector as migrated after uploading it to a new host.

### DIFF
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -386,3 +386,20 @@ func (s *Store) UnhealthySlab(ctx context.Context, maxRepairAttempt time.Time) (
 	})
 	return slabID, err
 }
+
+// MigrateSector updates a sector that was just migrated in the database to be
+// linked to the new host identified by 'hostKey'. This will reset the contract
+// ID since a freshly migrated sector isn't pinned yet. To pin a sector
+// 'PinSectors' is used. If the host is not found, e.g. due to being deleted in
+// the meantime, this operation is a no-op.
+func (s *Store) MigrateSector(ctx context.Context, root types.Hash256, hostKey types.PublicKey) error {
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `
+			UPDATE sectors
+			SET host_id = host.id, contract_sectors_map_id = NULL
+			FROM ( SELECT id, public_key FROM hosts ) AS host
+			WHERE sector_root = $1 AND host.public_key = $2 AND host.id IS NOT NULL
+		`, sqlHash256(root), sqlPublicKey(hostKey))
+		return err
+	})
+}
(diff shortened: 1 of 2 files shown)


persist/postgres/sectors.go
```diff
@@ -386,3 +386,20 @@ func (s *Store) UnhealthySlab(ctx context.Context, maxRepairAttempt time.Time) (
...
+// linked to the new host identified by 'hostKey'. This will reset the contract
+// ID since a freshly migrated sector isn't pinned yet. To pin a sector
+// 'PinSectors' is used. If the host is not found, e.g. due to being deleted in
+// the meantime, this operation is a no-op.
+func (s *Store) MigrateSector(ctx context.Context, root types.Hash256, hostKey types.PublicKey) error {
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `
+			UPDATE sectors
+			SET host_id = host.id, contract_sectors_map_id = NULL
+			FROM ( SELECT id, public_key FROM hosts ) AS host
```
n8mgr: Pretty sure directly using FROM hosts will work instead of the subquery.

persist/postgres/sectors.go
```diff
@@ -386,3 +386,20 @@ func (s *Store) UnhealthySlab(ctx context.Context, maxRepairAttempt time.Time) (
...
 	return slabID, err
 }
+
+// MigrateSector updates a sector that was just migrated in the database to be
+// linked to the new host identified by 'hostKey'. This will reset the contract
+// ID since a freshly migrated sector isn't pinned yet. To pin a sector
+// 'PinSectors' is used. If the host is not found, e.g. due to being deleted in
+// the meantime, this operation is a no-op.
+func (s *Store) MigrateSector(ctx context.Context, root types.Hash256, hostKey types.PublicKey) error {
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
```
peterjan: I'd fetch host id separately and return `ErrHostNotFound` so we can debug log this event. We should probably also check `res.RowsAffected` and return an error when the sector isn't found

persist/postgres/sectors_test.go
```diff
@@ -1605,3 +1708,93 @@ func BenchmarkMarkSectorsLost(b *testing.B) {
 		})
 	}
 }
+
+// BenchmarkMarkSectorsLost benchmarks MarkSectorsLost in various batch sizes.
```
peterjan: outdated docstring

persist/postgres/sectors_test.go
```diff
@@ -1605,3 +1708,93 @@ func BenchmarkMarkSectorsLost(b *testing.B) {
...
+
+	// pin all sectors to contracts
+	for contractID, roots := range rootsByContract {
+		err := store.PinSectors(context.Background(), contractID, roots)
+		if err != nil {
+			b.Fatal(err)
+		}
+	}
+
+	b.Run("1", func(b *testing.B) {
```
peterjan: what's the point of `b.Run("1")`?

ChrisSchinnerl: Nothing. Before `b.Loop` it was a way to only run the setup code once but with `b.Loop` that's not necessary anymore apparently.

persist/postgres/sectors.go
ChrisSchinnerl: I added a boolean to indicate an update. We can then log the host and root and check the database if we need to debug.
Don't want to add an extra roundtrip on every call for something that is a potential race we know about.

n8mgr (approved):

peterjan (approved):

PR #150, opened 2025-05-06
Migrations: Contracts for repair

ChrisSchinnerl:
This PR adds a helper to get all sectors from a slab that require migration as well as the contracts that can be used for it.

### DIFF
--- a/slabs/migrations.go
+++ b/slabs/migrations.go
@@ -1,0 +1,52 @@
+package slabs
+
+import (
+	"go.sia.tech/core/types"
+	"go.sia.tech/indexd/contracts"
+	"go.sia.tech/indexd/hosts"
+)
+
+// contractsForRepair filters the sectors of a slab and returns the sectors that
+// require migration together with the contracts to use for them.
+func contractsForRepair(slab Slab, goodHosts []hosts.Host, goodContracts []contracts.Contract, period uint64) ([]Sector, []contracts.Contract) {
+	goodHostsMap := make(map[types.PublicKey]hosts.Host)
+	for _, host := range goodHosts {
+		if host.Usability.Usable() && !host.Blocked {
+			goodHostsMap[host.PublicKey] = host
+		}
+	}
+
+	// prepare a map of good-for-upload goodContractMap
+	goodContractMap := make(map[types.FileContractID]contracts.Contract)
+	for _, contract := range goodContracts {
+		host, ok := goodHostsMap[contract.HostKey]
+		if !ok {
+			continue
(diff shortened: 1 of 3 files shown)


n8mgr (dismissed):

slabs/migrations.go
```diff
@@ -0,0 +1,71 @@
...
+	"go.sia.tech/indexd/hosts"
+)
+
+// contractsForRepair filters the sectors of a slab and returns the indices of the sectors that
+// require migration together with the contracts to use for them.
+func contractsForRepair(slab Slab, availableHosts []hosts.Host, availableContracts []contracts.Contract, period uint64) ([]int, []contracts.Contract) {
+	// prepare a map of good hosts
+	hostsMap := make(map[types.PublicKey]hosts.Host)
+	for _, host := range availableHosts {
+		if host.Usability.Usable() && !host.Blocked && len(host.Networks) > 0 {
```
peterjan: could make this a method on the host type

slabs/migrations.go
```diff
@@ -0,0 +1,71 @@
...
+			hostsMap[host.PublicKey] = host
+		}
+	}
+
+	// prepare a map of good contracts
+	goodContractMap := make(map[types.FileContractID]contracts.Contract)
+	for _, contract := range availableContracts {
+		host, ok := hostsMap[contract.HostKey]
```
peterjan, line 23: `nit` could inverse it and avoid all the not-checks

```
if host, ok := hostMap[contract.HostKey]; ok && contract.GoodForUpload(host.Settings.Prices, host.Settings.MaxCollateral, period) { 
    goodContractMap[contract.ID] = contract 
}
```

slabs/migrations.go
```diff
@@ -0,0 +1,71 @@
...
+	for i, sector := range slab.Sectors {
+		isLost := sector.ContractID == nil && sector.HostKey == nil
+		goodContract := sector.ContractID != nil && goodContractMap[*sector.ContractID] != contracts.Contract{}
+		if isLost || !goodContract {
+			toMigrate = append(toMigrate, i)
+			continue
+		}
+
+		// remove contract from the map since we don't want to use it again
+		delete(goodContractMap, *sector.ContractID)
```
peterjan, line 48: I would add a sanity check here because we might be dereferencing `nil` here and end up with a production panic

peterjan (dismissed):

slabs/migrations.go
ChrisSchinnerl: That can't happen because `goodContract` is `true` at this point which means it is set. The test should also cover those combinations.

peterjan (approved):

PR #153, opened 2025-05-08
Add sector pinning

peterjan:
This PR pins sectors by appending them to a contract. Sectors that were missing are being marked as lost.


persist/postgres/init.sql
```diff
@@ -257,6 +257,12 @@ CREATE INDEX sectors_host_id_next_integrity_check_idx ON sectors(host_id, next_i
 -- speed up querying sectors of a host by root
 CREATE INDEX sectors_host_id_sector_root_idx ON sectors(host_id, sector_root);
 
+-- speed up lookup of sectors by root
+CREATE INDEX sectors_sector_root_idx ON sectors(sector_root);
+
+-- speed up removing sectors by host
+CREATE INDEX sectors_host_id_sector_root_contract_sectors_map_id_idx ON sectors(host_id, sector_root) WHERE contract_sectors_map_id IS NULL;
```
ChrisSchinnerl: Can't add indices without updating the corresponding benchmarks with results or adding new benchmarks.

contracts/pinning.go
```diff
@@ -0,0 +1,204 @@
...
+		// fetch hosts
+		batch, err := cm.store.Hosts(ctx, offset, batchSize, opts...)
+		if err != nil {
+			return fmt.Errorf("failed to fetch hosts for pinning: %w", err)
+		}
+
+		// pin sectors on each host in parallel
+		var wg sync.WaitGroup
+		for _, h := range batch {
+			wg.Add(1)
```
ChrisSchinnerl: You can make this more efficient by using a semaphore here. That way, as soon as the first batch was scheduled, you can fetch the next batch and start scheduling it as individual hosts finish. 
To make sure we are never blocking scheduling a full batch just because 1 host is slow.

contracts/pinning.go
```diff
@@ -0,0 +1,204 @@
...
+			go func(ctx context.Context, host hosts.Host, log *zap.Logger) {
+				defer wg.Done()
```
ChrisSchinnerl: I'd extract this into a `performSectorPinningOnHost`

contracts/pinning.go
```diff
@@ -0,0 +1,204 @@
...
+	dialCtx, cancel := context.WithTimeout(ctx, dialTimeout)
+	t, err := siamux.Dial(dialCtx, host.SiamuxAddr(), host.PublicKey)
+	cancel()
+	if err != nil {
+		return types.FileContractID{}, nil, fmt.Errorf("failed to dial host: %w", err)
+	}
+	defer t.Close()
+
+	// fetch settings
+	settings, err := rhp.RPCSettings(ctx, t)
```
ChrisSchinnerl: The used settings should be the ones who went through gouging checks on the `host`.

contracts/pinning.go
```diff
@@ -0,0 +1,204 @@
...
+		res, err := rhp.RPCAppendSectors(ctx, t, c.cm.TipState(), settings.Prices, c.sk, revision, sectors)
+		if err != nil {
+			contractLog.Debug("failed to pin sectors", zap.Error(err))
+			continue
+		} else if len(res.Sectors) == 0 {
+			contractLog.Debug("no sectors were pinned")
+			continue
+		}
+
+		// TODO: handle usage
```
ChrisSchinnerl: The `contractor` should only be a thin wrapper on top of the individual RPCs. So `PinSectors` should be `AppendSectors` and any other logic should happen outside the contractor interface.

contracts/pinning.go
```diff
@@ -0,0 +1,204 @@
...
+// returned, alongside with a list of missing sectors if any.
+func (c *contractor) PinSectors(ctx context.Context, host hosts.Host, contractIDs []types.FileContractID, sectors []types.Hash256, log *zap.Logger) (types.FileContractID, []types.Hash256, error) {
+	// sanity check
+	if len(sectors) > proto.MaxSectorBatchSize {
+		return types.FileContractID{}, nil, errors.New("too many sectors") // developer error
+	}
+
+	// dial host
+	dialCtx, cancel := context.WithTimeout(ctx, dialTimeout)
+	t, err := siamux.Dial(dialCtx, host.SiamuxAddr(), host.PublicKey)
```
ChrisSchinnerl: If you move this out of `PinSectors` and instead pass an interface in, you can reuse the connection between batches and easily write a unit test for `PinSectors`.

ChrisSchinnerl (changes requested):

persist/postgres/init.sql
peterjan: Leftover, thanks!

contracts/pinning.go
peterjan: I considered that but did not want to require a transport client in one of the methods on the  `Contractor`. I do think `AppendSectors` should be defined there, and we do want to reuse the connection (although I didn't care all that much originally seeing we append 1TB worth of sectors at a time), so I did it anyway.

contracts/manager.go
```diff
@@ -46,8 +46,10 @@ type (
 	}
 
 	// Contractor defines the dependencies required to form, renew and refresh
-	// contracts.
+	// contracts. It also provides a way to pin sectors to a contract.
 	Contractor interface {
+		AppendSectors(ctx context.Context, tc rhp.TransportClient, hostPrices proto.HostPrices, contractID types.FileContractID, sectors []types.Hash256) (rhp.RPCAppendSectorsResult, error)
```
ChrisSchinnerl: I agree that the transport client in this signature is pretty ugly. I had an idea for a refactor which I wasn't sure would work out that I started here: https://github.com/SiaFoundation/indexd/pull/159

This should help clean things up for you in this PR I think. It should lead to a nicer signature while also allowing us to reuse the transport client without reducing our ability to mock/test things.

contracts/pinning.go
```diff
@@ -0,0 +1,235 @@
...
+	sectorPinner struct {
+		host   HostClient
+		prices proto.HostPrices
+	}
+)
+
+// PinSectors pins a set of sectors using the given set of contracts The
+// contracts are tried in order, the contract ID that ends up being used is
+// returned, alongside with a list of missing sectors if any.
+func (p *sectorPinner) PinSectors(ctx context.Context, contractIDs []types.FileContractID, sectors []types.Hash256, log *zap.Logger) (usedContractID types.FileContractID, missing []types.Hash256, _ error) {
```
ChrisSchinnerl: At this point everything we need to mock is in the `hostClient` so this can be a method on `hostClient`. `sectorPinner` adds a lot of code for an abstraction that just serves avoiding passing in `prices`.

contracts/pinning.go
```diff
@@ -0,0 +1,235 @@
...
+				}()
+
+				client, err := cm.dialer.Dial(ctx, host.PublicKey, host.SiamuxAddr())
+				if err != nil {
+					hostLog.Debug("failed to create contractor", zap.Error(err))
+					return
+				}
+				defer client.Close()
+
+				err = cm.performSectorPinningOnHost(ctx, &sectorPinner{host: client, prices: host.Settings.Prices}, host, hostLog)
```
ChrisSchinnerl: `host.Settings.Prices` might be outdated at this point

contracts/pinning.go
```diff
@@ -0,0 +1,235 @@
...
+			for _, root := range roots {
+				if _, missing := lookup[root]; !missing {
+					filtered = append(filtered, root)
+				}
+			}
+			roots = slices.Clone(filtered)
+
+			for i := 0; i < len(missing); i += dbBatchSize {
+				end := min(i+dbBatchSize, len(missing))
+				if err := cm.store.MarkSectorsLost(ctx, host.PublicKey, missing[i:end]); err != nil {
```
ChrisSchinnerl: This is some super overcomplicated logic imo that you can avoid by just making `MarkSectorsLost` the first thing you call.

e.g.

```
if len(missing) > 0 {
  if err := cm.store.MarkSectorsLost(ctx, host.PublicKey, missing); err != nil {
    // error handling
  }

  // filter roots afterwards
  isMissing := make(map[types.Hash256]struct{}, len(missing))
  for _, sector := range missing {
    isMissing[sector] = struct{}{}
  }
  
  filtered := roots[:0]
  for _, root := range roots {
    if _, missing := isMissing[root]; !missing {
      filtered = append(filtered, root)
    }
  }
  roots = filtered
}

File: contracts/pinning.go [hunk 11]
```diff
@@ -0,0 +1,235 @@
...
+					return fmt.Errorf("failed to mark sectors as lost: %w", err)
+				}
+			}
+			nMissing += uint64(len(missing))
+		}
+
+		if len(roots) > 0 {
+			err = cm.store.PinSectors(ctx, contractID, roots)
```
ChrisSchinnerl: We could clean this up a bit if we checked for 0 roots within the store methods and made them no-ops in that case.

I'd also move 
```
nMissing += uint64(len(missing))
nPinned += uint64(len(roots))
```
to the end of the loop since they don't need to be within the if statements.
```

File: contracts/form_test.go [hunk 12]
```diff
@@ -63,15 +63,19 @@ func (d *dialerMock) Dial(ctx context.Context, hostKey types.PublicKey, addr str
 }
 
 type hostClientMock struct {
-	formCalls       []formContractCall
-	refreshCalls    []refreshContractCall
-	renewCalls      []renewContractCall
+	appendSectorCalls []appendSectorCall
```
ChrisSchinnerl, line 66: I don't think we absolutely need `appendSectorCalls`. For formation/refreshing/renewals storing the calls is convenient since contract formation is all about setting the right parameters but appending sectors is probably simple enough to trust it's working when the sectors end up in the mocked store.

We can keep it though since we already have it.

contracts/pinning.go
```diff
@@ -0,0 +1,235 @@
...
+		sectorsBatchSize = (1 << 40) / proto.SectorSize // 1TB of sectors
+	)
+
+	var wg sync.WaitGroup
+	sema := make(chan struct{}, 50)
+	defer close(sema)
+
+	for offset := 0; ctx.Err() == nil; offset += batchSize {
+		// fetch hosts
+		batch, err := cm.store.Hosts(ctx, offset, batchSize, opts...)
```
ChrisSchinnerl: Since we have already done it for integrity checks it probably makes sense to also have a specialized method here that only returns hosts that have at least 1 sector to pin and only their public key.

contracts/form.go
```diff
@@ -91,6 +92,7 @@ func (d *siamuxDialer) Dial(ctx context.Context, hostKey types.PublicKey, addr s
 		client:  client,
 		cm:      d.cm,
 		hostKey: hostKey,
+		ownKey:  d.ownKey,
 		signer: &formContractSigner{
 			renterKey: d.ownKey,
 			w:         d.w,
```
ChrisSchinnerl: This is getting kinda wild. We pass the same private key as fields to `hostClient` and `formContractSigner`.
`formContractSigner` holds a `rhp.Wallet` which also contains the same private key already. `formContractSigner` also exists only because `SingleAddressWallet` doesn't implement `rhp.Wallet`.

In theory we can clean this up if we changed our `rhp.RPC*` methods to never accept a private key directly and instead always a `rhp.Wallet` and by making `SingleAddressWallet` implement `rhp.Wallet`.

Then we can get rid of `formContractSigner` and also the `ownKey` field here.

Thoughts @n8maninger ?

ChrisSchinnerl (changes requested):

contracts/form_test.go
peterjan: Yeah that's true. I'd rather keep it though because it's an extra check and it comes in handy for when we want to test batching and order of events. E.g. using the `Xcalls` fields we can check whether we called sectors for contract x before contract y (just an example).

contracts/pinning.go
peterjan: Added an interface method 👍

contracts/pinning.go
peterjan: I don't really agree that `sectorPinner` doesn't serve a purpose, even though it only wraps the `client` with a set of prices. The `HostClient` currently defines a set of methods that very thinly wrap the RPCs, which is great. Extending it with things like `PinSectors` would needlessly bloat it. More importantly though, if we were to add it on the client, we would have to mock it. So we'd have two versions of `PinSectors`. In my tests I don't need to mock the `sectorPinner` and there's only one implementation of `PinSectors`. 

Definitely not a hill I'll die on, so say the word and I'll merge the two, I just don't see why we would want to do that to our client who now cleanly and thinly wraps some of our RPCs.

contracts/form.go
n8mgr: `SingleAddressWallet` should already implement `rhp.Wallet`, right? If you mean, `ContractSigner`, I don't particularly mind. I thought we didn't bother changing it because we derive a unique key per host though.

~~I don't believe any of the coreutils `rhp.RPC*` methods take a private key directly. It's all `rhp.ContractSigner` and `rhp.Wallet`~~

Looks like RPCAppend and RPCFreeSectors take a private key. Not really sure why they don't take a `ContractSigner`, but that's simple enough to change.

n8mgr: https://github.com/SiaFoundation/coreutils/pull/235

ChrisSchinnerl: You are right. It's `TransactionFunder` used by `FormContractSigner` that requires 
`FundV2Transaction(txn *types.V2Transaction, amount types.Currency) (types.ChainIndex, []int, error)` while `rhp.Wallet` uses 
`FundV2Transaction(txn *types.V2Transaction, amount types.Currency, useUnconfirmed bool) (types.ChainIndex, []int, error)
`

contracts/pinning.go
ChrisSchinnerl: Oh yeah I didn't mean to say `hostClient`. That should be lightweight. But it can be on the manager itself. 

Then
```go
	pinner := newSectorPinner(client, host.Settings.Prices)
	err = cm.performSectorPinningOnHost(ctx, pinner, host, hostLog)
```

becomes 
```
	err = cm.performSectorPinningOnHost(ctx, client, host, hostLog)
```

peterjan: Oh ok, yeah that makes sense. It makes even more sense to only pass the `host` to `performSectorPinningOnHost` and dial the host there before going through the batches. That also moves refreshing the prices inside of `performSectorPinningOnHost` which is also a good thing.

ChrisSchinnerl (dismissed):

n8mgr (dismissed):

ChrisSchinnerl (approved):

PR #154, opened 2025-05-08
Update BenchmarkMarkSectorsLost

peterjan:
Ok so I overlooked `MarkSectorsLost` 🤦  - so we don't need `RemoveSector`. 

However, `BenchmarkMarkSectorsLost` never finishes if I run it locally,  a couple of benchmarks actually. Every time we manipulate the state inside of the loop (even with timer stopped), `go bench` keeps updating `b.N` and it goes on forever because it can't find a reliable value. I understand this is a problem on my machine, but it's also a problem on the CI. In general it's better practice to try and avoid state updates within the `b.Loop` so I think we should start  writing exhaustive benchmarks with a sanity check it ran at least 25 times for instance.

I'm fine with just closing out the PR, but using `UNNEST` should be slightly better in our case since we're dealing with quite a lot of sectors. I'm also wondering whether we should move the batching inside of the function. At the very least we should probably add a sanity check we don't call this with batches over 1000 in size. When pinning sectors for instance, if all would be missing, we're passing 1TB worth of sectors, which would be bad.


n8mgr (dismissed):

persist/postgres/sectors.go
```diff
@@ -311,6 +311,30 @@ func (s *Store) PinSectors(ctx context.Context, contractID types.FileContractID,
 	})
 }
 
+// RemoveSectors removes the sectors for the given host from the database by
+// unsetting their host_id. This should be called when we failed to pin the
+// sectors on the host because the host could not find the sector.
+func (s *Store) RemoveSectors(ctx context.Context, hostKey types.PublicKey, roots []types.Hash256) error {
```
ChrisSchinnerl: This seems like it might be redundant with `MarkSectorsLost`? Which also mentions being used for pinning in its docstring.

The way this is right now, a host gets to walk away with losing uploaded data. Because we don't track it.

peterjan: ermahgerd

peterjan: I'll change this PR to update the method and the benchmark though. I'm using `UNNEST` which should be slightly better for calculating the query plan and the benchmark is a bit different too because the current one uses a single host where as it's probably better to use more hosts and select them randomly.

peterjan: Well I'm not sure if we should care I guess. I'll see if it makes a meaningful difference and close this out otherwise.

persist/postgres/sectors_test.go
```diff
@@ -1481,126 +1481,89 @@ func TestMarkSectorsLost(t *testing.T) {
 
 // BenchmarkMarkSectorsLost benchmarks MarkSectorsLost in various batch sizes.
 //
-// CPU    | BatchSize |	 Count  |     Time/op    |    Throughput
-// M2 Pro |     100   |    638  |   1.903132 ms  |   220389.59 MB/s
-// M2 Pro |    1000   |     94  |  12.430024 ms  |   337433.29 MB/s
-// M2 Pro |   10000   |     10  | 112.704779 ms  |   372149.61 MB/s
+// Hardware |  BatchSize |   ms/op    |  Throughput   |
+// M1 Max   |  	 100	 |   4.68ms   |   89.49 GB/s  |
```
peterjan, line 1485: @ChrisSchinnerl would you mind running this benchmark on your machine to see if we get similar throughput?

ChrisSchinnerl:
```
cpu: Apple M2 Pro
BenchmarkMarkSectorsLost
BenchmarkMarkSectorsLost/100
BenchmarkMarkSectorsLost/100-12         	     526	   2546219 ns/op	164726.74 MB/s
BenchmarkMarkSectorsLost/500
BenchmarkMarkSectorsLost/500-12         	     136	   8735221 ns/op	240080.02 MB/s
BenchmarkMarkSectorsLost/1000
BenchmarkMarkSectorsLost/1000-12        	      73	  16543916 ns/op	253525.47 MB/s
PASS
ok  	go.sia.tech/indexd/persist/postgres	33.326s
```

persist/postgres/sectors.go
```diff
@@ -26,10 +26,13 @@ func (s *Store) MarkSectorsLost(ctx context.Context, hostKey types.PublicKey, ro
 			sqlRoots[i] = sqlHash256(root)
 		}
 		resp, err := tx.Exec(ctx, `
+			WITH
+				hid AS (SELECT id FROM hosts WHERE public_key = $1),
+				removed(root) AS (SELECT UNNEST($2::bytea[]))
```
ChrisSchinnerl, line 31: I don't think there is any reason to use `UNNEST` unless you really need to turn your input into a table.
This performs the same for me as the other query against the master branch benchmark.

persist/postgres/sectors_test.go
```diff
@@ -1481,126 +1481,89 @@ func TestMarkSectorsLost(t *testing.T) {
...
-				toMark := sectorsToMark(batchSize)
-				if len(toMark) != batchSize {
-					b.Fatalf("expected %d sectors to mark, got %d", batchSize, len(toMark))
+				// break if we're exhausted
+				if len(hks) == 0 {
+					if iter < 25 {
+						b.Fatalf("expected to run at least 25 iterations, got %d", iter)
+					}
+					b.StopTimer()
+					break
```
ChrisSchinnerl, line 1556: This looks very illegal. We should never accept a result if we exhaust the database.

ChrisSchinnerl (changes requested):

peterjan:
Thanks for the review Chris, I'll just close it out as I just wanted to split off `RemoveSectors` for which I should've just used `MarkSectorsLost`.

PR #155, opened 2025-05-08
Add ContractsForPinning

peterjan:


n8mgr (dismissed):

persist/postgres/contracts.go
```diff
@@ -201,6 +201,36 @@ LIMIT $3
...
+// good for pinning sectors with. The contracts are sorted by size, capacity in
+// descending fashion.
+func (s *Store) ContractsForPinning(ctx context.Context, hk types.PublicKey) ([]types.FileContractID, error) {
+	var fcids []types.FileContractID
+	err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		rows, err := tx.Query(ctx, `
+SELECT c.contract_id
+FROM contracts c
+INNER JOIN hosts h ON c.host_id = h.id
+WHERE h.public_key = $1 AND c.good = TRUE AND c.state <= $2  AND c.remaining_allowance > 0 AND c.size < 1E13::NUMERIC
```
ChrisSchinnerl:
```suggestion
WHERE h.public_key = $1 AND c.good = TRUE AND c.state <= $2 AND c.remaining_allowance > 0 AND c.size < 1E13::NUMERIC 
```

persist/postgres/contracts.go
```diff
@@ -201,6 +201,36 @@ LIMIT $3
 	return fcids, nil
 }
 
+// ContractsForPinning returns all contracts for the given host key that are
+// good for pinning sectors with. The contracts are sorted by size, capacity in
+// descending fashion.
+func (s *Store) ContractsForPinning(ctx context.Context, hk types.PublicKey) ([]types.FileContractID, error) {
```
ChrisSchinnerl: My preference would be for the max contract size to be passed in by the contract manager. To make sure we only have that constant in one place and all our checks using it return consistent results.

persist/postgres/contracts_test.go
```diff
@@ -1303,7 +1367,67 @@ func BenchmarkContractsForFunding(b *testing.B) {
 	}
 
 	for b.Loop() {
-		if _, err := store.ContractsForFunding(context.Background(), hosts[b.N%numHosts], 50); err != nil {
+		rIdx := frand.Intn(len(hosts))
+		if _, err := store.ContractsForFunding(context.Background(), hosts[rIdx], 50); err != nil {
+			b.Fatal(err)
+		}
```
ChrisSchinnerl, line 1377: This is so insanely fast, can you add a check to make sure the returned contracts are not always 0?

persist/postgres/contracts_test.go
```diff
@@ -1303,7 +1367,67 @@ func BenchmarkContractsForFunding(b *testing.B) {
...
+			hosts = append(hosts, hk)
+		}
+		return nil
+	}); err != nil {
+		b.Fatal(err)
+	}
+
+	for b.Loop() {
+		rIdx := frand.Intn(len(hosts))
+		if _, err := store.ContractsForPinning(context.Background(), hosts[rIdx]); err != nil {
```
ChrisSchinnerl: Same thing here. Just want to make sure the number returned is reasonable.

persist/postgres/contracts.go
```diff
@@ -201,6 +201,36 @@ LIMIT $3
...
+	var fcids []types.FileContractID
+	err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		rows, err := tx.Query(ctx, `
+SELECT c.contract_id
+FROM contracts c
+INNER JOIN hosts h ON c.host_id = h.id
+WHERE h.public_key = $1 AND c.good = TRUE AND c.state <= $2  AND c.remaining_allowance > 0 AND c.size < 1E13::NUMERIC 
+ORDER BY c.size DESC, c.capacity DESC`, sqlPublicKey(hk), sqlContractState(contracts.ContractStateActive))
```
ChrisSchinnerl: I think if you want to save money, it's best to sort by `c.capacity DESC, c.size DESC`. 
That way, we reuse the free capacity of large contracts that we pruned first. e.g. if we have a 10TB contract that we pruned down to 8TB. Better ego use that one than a 9TB contract since the storage is effectively free.

ChrisSchinnerl (changes requested):

persist/postgres/contracts.go
peterjan: Wasn't really trying to save money. I was trying to neatly fill contracts up to 10TB as quickly as possible, then if they were identical try and save money 🤷‍♂️

ChrisSchinnerl (approved):

PR #157, opened 2025-05-12
Add store.ContractsForPruning

peterjan:
Adds `ContractsForPruning` which returns contracts depending on their `last_prune` time. Instead of adding a separate benchmark for it, I decided to tweak `BenchmarkContracts` a bit and just merge them all into a single benchmark to save on setup. I found we had some slowness in `ContractsForBroadcasting` and tweaked the index to exclude the `state`. For all indices that were added to support methods that return a list of contract ids, I added the `contract_id` to the index so those should all be `Index Only Scan`s

### DIFF
--- a/persist/postgres/contracts_test.go
+++ b/persist/postgres/contracts_test.go
@@ -446,6 +446,71 @@ func TestContractsForPinning(t *testing.T) {
 	}
 }
 
+func TestContractsForPruning(t *testing.T) {
+	store := initPostgres(t, zaptest.NewLogger(t).Named("postgres"))
+
+	addContract := func(hk types.PublicKey, fcid types.FileContractID, allowance types.Currency, size uint64, state contracts.ContractState, good bool, lastPrune time.Time) {
+		t.Helper()
+		if err := store.AddFormedContract(context.Background(), fcid, hk, 100, 200, types.ZeroCurrency, allowance, types.ZeroCurrency, types.ZeroCurrency); err != nil {
+			t.Fatal(err)
+		}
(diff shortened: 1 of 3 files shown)


persist/postgres/contracts_test.go
```diff
@@ -1255,187 +1337,40 @@ func BenchmarkContracts(b *testing.B) {
...
-	store := initPostgres(b, zap.NewNop())
-	hosts := make([]types.PublicKey, 0, numHosts)
-	if err := store.transaction(context.Background(), func(ctx context.Context, tx *txn) error {
-		for range numHosts {
-			var hostID int64
-			hk := types.GeneratePrivateKey().PublicKey()
-			err := tx.QueryRow(context.Background(), `INSERT INTO hosts (public_key, last_announcement) VALUES ($1, NOW()) RETURNING id;`, sqlPublicKey(hk)).Scan(&hostID)
+	b.Run("contracts_for_pruning", func(b *testing.B) {
+		for b.Loop() {
+			_, err := store.ContractsForPruning(context.Background(), hosts[frand.Intn(len(hosts))], randomTime())
```
ChrisSchinnerl, line 1370: Sanity check that contracts are returned

ChrisSchinnerl (approved):

n8mgr (approved):

persist/postgres/contracts_test.go
peterjan: `randomTime` complicates this a bit, I did verify it manually, this whole benchmark generates contracts with random parameters - I considered adding a count or something to check a contract was returned at least once but that looks so silly I don't care commit it

PR #159, opened 2025-05-13
Refactor `Contractor` into `Dialer` and `Contractor` for easier testing and nice interfaces

ChrisSchinnerl:
This is a small refactor which was triggered by https://github.com/SiaFoundation/indexd/pull/153/files#diff-52dce9968f7373dadf41cf819e65996c2a5cd28408843f8779cd04bf704d8b93R51

Pulling out dialing of methods like `AppendSectors` definitely makes sense to get better testability but passing a transport client limits it again since we can't easily mock that in testing.

So this PR isolates only the dialing of the connection into a `Dialer` while simultaneously making `Contractor` a very thin wrapper around the RHP4 client. Reducing the amount of code we need to mock to only RPC code.

### DIFF
--- a/contracts/form.go
+++ b/contracts/form.go
@@ -7,7 +7,6 @@ import (
 
 	proto "go.sia.tech/core/rhp/v4"
 	"go.sia.tech/core/types"
-	"go.sia.tech/coreutils/chain"
 	"go.sia.tech/coreutils/rhp/v4"
 	"go.sia.tech/coreutils/rhp/v4/siamux"
 	"go.sia.tech/indexd/hosts"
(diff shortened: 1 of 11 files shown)


contracts/form.go
```diff
@@ -51,50 +50,69 @@ func (s *formContractSigner) SignV2Inputs(txn *types.V2Transaction, toSign []int
 	s.w.SignV2Inputs(txn, toSign)
 }
 
+// dialer is an interface for dialing a host.
+type dialer interface {
+	NewContractor(ctx context.Context, hostKey types.PublicKey, addr string) (Contractor, error)
```
n8mgr: Why contractor instead of `Dial`?

n8mgr (dismissed):
>

contracts/form.go
peterjan: We could also get rid of `Contractor` and call it a `HostClient`, then the rename to `Dial` feels even more natural. Especially since we keep extending it with thin RPC wrappers it makes sense to rename it to client.

peterjan (approved):

PR #161, opened 2025-05-15
Add contract pruning

### LINKED ISSUE #45: Contract Maintenance
This serves as a collection of tasks that need to be finished in regards to contract maintenance 

- [x] Persistence
- [x] State updates
- [x] Formations
- [x] Renewals
- [x] Refreshes
- [x] Revision rebroadcasting
- [x] Pruning

peterjan:
closes https://github.com/SiaFoundation/indexd/issues/45


contracts/pruning.go
```diff
@@ -0,0 +1,226 @@
...
+	// free sectors
+	revision := rhp.ContractRevision{ID: contractID, Revision: rev.Contract}
+	return rhp.RPCFreeSectors(ctx, c.client, c.signer, c.cm.TipState(), hostPrices, revision, indices)
+}
+
+func (cm *ContractManager) performContractPruning(ctx context.Context, log *zap.Logger) error {
+	start := time.Now()
+	log = log.Named("contractpruning")
+
+	// prune sectors on usable hosts with active contracts
```
peterjan: figured we don't really need `HostsForX` here, we could in theory fetch fetch non-blocked hosts with a contract that's up for pruning

contracts/pruning.go
```diff
@@ -0,0 +1,226 @@
...
+	wg.Wait()
+
+	log.Debug("pruning finished", zap.Duration("duration", time.Since(start)))
+	return ctx.Err()
+}
+
+func (cm *ContractManager) performContractPruningOnHost(ctx context.Context, host hosts.Host, hostLog *zap.Logger) error {
+	// refresh prices if necessary
+	ts := host.Settings.Prices.ValidUntil
+	if !host.Usability.Usable() || time.Until(ts) < 30*time.Minute {
```
ChrisSchinnerl: Should use `IsGood` here to be consistent with the inner check. Otherwise we might not enter this `if` if a host is usable but not good.

contracts/pruning.go
```diff
@@ -0,0 +1,226 @@
...
+		host, err := cm.scanner.ScanHost(ctx, host.PublicKey)
+		if err != nil {
+			return fmt.Errorf("failed to scan host: %w", err)
+		} else if !host.IsGood() {
+			hostLog.Debug("host is not good for pinning", zap.Bool("blocked", host.Blocked), zap.Bool("usable", host.Usability.Usable()), zap.Bool("networks", len(host.Networks) > 0))
```
ChrisSchinnerl: `pinning`

contracts/pruning.go
ChrisSchinnerl: I'd keep it consistent. Only fetching public keys is a lot more lightweight than fetching full hosts. We also already have `MarkPruned` and `last_prune`. So it doesn't even require updating the schema with another column. 

We might just need to move from a `last_prune` to a `next_prune` and then set `next_prune` depending on whether the pruning of a contract failed to succeeded. That way we can avoid spinning on the same host. If marking a contract fails, which should not be the case, we'd interrupt the pruning loop altogether.

That said, you could just add the corresponding method to the store interface and add the actual implementation in a f/u.

ChrisSchinnerl (changes requested):

contracts/pruning.go
peterjan: Yeah `last_prune` is not usable if we have to get `HostsForX` but I'll update the database in a F/U which 'll also implement and benchmark the store method.

contracts/pinning.go
```diff
@@ -97,15 +94,18 @@ func (cm *ContractManager) performSectorPinning(ctx context.Context, log *zap.Lo
 }
 
 func (cm *ContractManager) performSectorPinningOnHost(ctx context.Context, host hosts.Host, hostLog *zap.Logger) error {
+	// check host is good
+	if !host.IsGood() {
```
peterjan, line 93: I moved it for pruning and just wanted it to be consistent with pinning. Apologies this is in my diff.

contracts/pinning_test.go
```diff
@@ -85,6 +85,22 @@ func (s *storeMock) HostsForPinning(ctx context.Context) ([]types.PublicKey, err
 	return hosts, nil
 }
 
+func (s *storeMock) HostsForPruning(ctx context.Context) ([]types.PublicKey, error) {
```
ChrisSchinnerl: (nit) I'd either move this to `pruning_test.go` to keep the relevant mock methods in the files using them or f/u with a PR that moves all the mock code into a single file. Right now we don't seem to be super consistent.

contracts/pruning.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to fetch hosts for pruning: %w", err)
+	} else if len(hosts) == 0 {
+		log.Warn("no hosts for pruning")
+		return nil
+	}
+
+	const (
+		nThreads         = 50
+		sectorsBatchSize = (1 << 40) / proto.SectorSize
```
ChrisSchinnerl: `core` exposes this as `MaxSectorBatchSize`

contracts/pruning.go
```diff
@@ -0,0 +1,225 @@
...
+
+		n, err := cm.pruneContract(ctx, client, host.Settings.Prices, contract)
+		if err != nil {
+			hostLog.Debug("failed to prune contract", zap.Error(err))
+			continue
+		} else if n > 0 {
+			hostLog.Debug("pruned contract", zap.Stringer("contractID", contract), zap.Int("sectors", n), zap.Int("bytes", n*proto.SectorSize))
+		}
+
+		err = cm.store.MarkPruned(ctx, contract)
```
ChrisSchinnerl: This also needs to change to work together with `ContractsForPruning`. You'll need to pass in a timestamp here for the next prune time and check if the pruning was successful or not.

Considering that this affects this method, the testing and mock I'd update the interface signature in this PR already to avoid adding a bunch of code that is broken.

contracts/pruning_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (c *hostClientMock) FreeSectors(ctx context.Context, hostPrices proto.HostPrices, contractID types.FileContractID, indices []uint64) (rhp.RPCFreeSectorsResult, error) {
+	c.freeSectorsCalls = append(c.freeSectorsCalls, freeSectorsCall{
+		hostPrices: hostPrices,
+		contractID: contractID,
+		indices:    indices,
+	})
+	return rhp.RPCFreeSectorsResult{}, nil
+}
+
```
ChrisSchinnerl, line 130: Would be good to also have a `TestPerformContractPruning` test with 1 good host, 1 bad host and 1 failing host to make sure only the good one gets pruned and to make sure we don't get stuck in a loop here if a host fails.

ChrisSchinnerl (changes requested):

contracts/pruning_test.go
peterjan: I'll extend the tests to cover that. The reason we don't have `TestPerform[Action]` but always do `TestPerform[Action]OnHost` is because the mocked store isn't thread safe. We do properly cover all components though, the only untested part is the thread part where we use a sema to do the action concurrently.

contracts/pruning_test.go
```diff
@@ -0,0 +1,371 @@
...
+	success := time.Now().Add(pruneIntervalSuccess)
+	failure := time.Now().Add(pruneIntervalFailure)
+	for _, c := range contracts {
+		switch c.ID {
+		case fcid1, fcid2, fcid3, fcid4:
+			if !(c.NextPrune.After(success.Add(-time.Minute)) && c.NextPrune.Before(success.Add(time.Minute))) {
+				t.Fatal("expected next prune to be scheduled 24h from now", c.ID, success, c.NextPrune)
+			}
+		case fcid4:
```
ChrisSchinnerl, line 368: Lint is failing because `fcid4` exists twice.

persist/postgres/hosts_test.go
ChrisSchinnerl, line 1: Did you target the wrong branch? The `postgres` changes were meant to be in a f/u right?

peterjan: Wasn't ready with the PR yet. Yeah I meant to base it off of `pj/hosts-for-pruning`.

ChrisSchinnerl (approved):

n8mgr (approved):

PR #162, opened 2025-05-16
Add HostsForIntegrityCheck implementation

ChrisSchinnerl:
```
goos: darwin
goarch: arm64
cpu: Apple M2 Pro
BenchmarkHostsForIntegrityCheck
BenchmarkHostsForIntegrityCheck/100
BenchmarkHostsForIntegrityCheck/100-12         	    2062	    491420 ns/op
BenchmarkHostsForIntegrityCheck/1000
BenchmarkHostsForIntegrityCheck/1000-12        	     915	   1244286 ns/op
BenchmarkHostsForIntegrityCheck/10000
BenchmarkHostsForIntegrityCheck/10000-12       	     136	   8736631 ns/op
```

### DIFF
--- a/persist/postgres/hosts.go
+++ b/persist/postgres/hosts.go
@@ -599,7 +599,7 @@ func scanHost(s scanner) (dbHost, error) {
 
 // HostsForIntegrityChecks returns a list of hosts that have sectors
 // requiring integrity checks.
-func (s *Store) HostsForIntegrityChecks(ctx context.Context) ([]types.PublicKey, error) {
+func (s *Store) HostsForIntegrityChecks(ctx context.Context, limit int) ([]types.PublicKey, error) {
 	var hosts []types.PublicKey
 	if err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
 		rows, err := tx.Query(ctx, `
@@ -611,7 +611,8 @@ func (s *Store) HostsForIntegrityChecks(ctx context.Context) ([]types.PublicKey,
 				WHERE sectors.host_id = hosts.id
 					AND sectors.next_integrity_check <= NOW()
 			)
-		`)
+			LIMIT $1
+		`, limit)
 		if err != nil {
 			return fmt.Errorf("failed to query hosts for scanning: %w", err)
 		}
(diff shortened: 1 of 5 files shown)


slabs/manager.go
```diff
@@ -159,7 +160,7 @@ func (m *SlabManager) performIntegrityChecks(ctx context.Context) error {
 	logger := m.log.Named("integrity")
 	logger.Debug("starting integrity checks", zap.Time("start", start))
 
-	usedHosts, err := m.store.HostsForIntegrityChecks(ctx)
+	usedHosts, err := m.store.HostsForIntegrityChecks(ctx, math.MaxInt)
```
ChrisSchinnerl: This is slightly awkward. We can't easily paginate here since only `SectorsForIntegrityCheck` will update the `next_integrity_check` column of the sectors. Meaning that if we fail to dial a host with missing integrity checks, it will just reappear forever. 

I don't expect fetching all hosts to be a problem before we start storing sectors on 10k+ hosts so not batching might be fine for now.

persist/postgres/hosts.go
```diff
@@ -599,6 +599,35 @@ func scanHost(s scanner) (dbHost, error) {
...
+			WHERE EXISTS (
+				SELECT 1
+				FROM sectors
+				WHERE sectors.host_id = hosts.id
+					AND sectors.next_integrity_check <= NOW()
+			)
+			LIMIT $1
+		`, limit)
+		if err != nil {
+			return fmt.Errorf("failed to query hosts for scanning: %w", err)
```
peterjan: update error message

slabs/manager.go
peterjan: If all call sites pass `math.MaxInt` we might as well drop the `limit` all together.

peterjan: We could add a field to the `host` that we immediately set upon retrieval. When you were fetching hosts before, were you filtering them with any options? I tried to go back to your PR and check but a force-push erased the history I think. I'm asking right now we return blocked hosts too.

n8mgr (dismissed):

slabs/manager.go
ChrisSchinnerl: Updated the query to ignore blocked hosts and update a new field on hosts.

New benchmark:
```
goos: darwin
goarch: arm64
cpu: Apple M2 Pro
BenchmarkHostsForIntegrityCheck
BenchmarkHostsForIntegrityCheck/50
BenchmarkHostsForIntegrityCheck/50-12 	    1034	    977290 ns/op
BenchmarkHostsForIntegrityCheck/100
BenchmarkHostsForIntegrityCheck/100-12         	     884	   1398005 ns/op
BenchmarkHostsForIntegrityCheck/200
BenchmarkHostsForIntegrityCheck/200-12         	     489	   2577229 ns/op
PASS
```

peterjan (approved):

PR #182, opened 2025-06-03
Change type of settings_max_contract_duration

### LINKED ISSUE #176: Wrong data type for `MaxContractDuration` on the `hosts` table schema
`MaxContractDuration` is currently defined as follows:
```SQL
settings_max_contract_duration NUMERIC(50,0) NOT NULL DEFAULT 0,
```

I think it should be consistent with the other `uint64` types, so the correct definition should be:
```SQL
settings_max_contract_duration BIGINT NOT NULL DEFAULT 0,
```

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/indexd/issues/176

### DIFF
--- a/persist/postgres/init.sql
+++ b/persist/postgres/init.sql
@@ -20,7 +20,7 @@ CREATE TABLE hosts (
     settings_wallet_address BYTEA NOT NULL DEFAULT '\x0000000000000000000000000000000000000000000000000000000000000000'::bytea CHECK (LENGTH(settings_wallet_address) = 32),
     settings_accepting_contracts BOOLEAN NOT NULL DEFAULT FALSE,
     settings_max_collateral NUMERIC(50,0) NOT NULL DEFAULT 0,
-    settings_max_contract_duration NUMERIC(50,0) NOT NULL DEFAULT 0,
+    settings_max_contract_duration BIGINT NOT NULL DEFAULT 0,
     settings_remaining_storage BIGINT NOT NULL DEFAULT 0,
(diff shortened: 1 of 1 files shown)


persist/postgres/init.sql
```diff
@@ -20,7 +20,7 @@ CREATE TABLE hosts (
     settings_wallet_address BYTEA NOT NULL DEFAULT '\x0000000000000000000000000000000000000000000000000000000000000000'::bytea CHECK (LENGTH(settings_wallet_address) = 32),
     settings_accepting_contracts BOOLEAN NOT NULL DEFAULT FALSE,
     settings_max_collateral NUMERIC(50,0) NOT NULL DEFAULT 0,
-    settings_max_contract_duration NUMERIC(50,0) NOT NULL DEFAULT 0,
+    settings_max_contract_duration BIGINT NOT NULL DEFAULT 0,
```
n8mgr, line 23: U64 will overflow this. Not sure if Postgres will error or let it overflow, but we should likely add a check that it is sane before adding to the database somewhere if we don’t have one already

ChrisSchinnerl: It will error with `failed to encode args[3]: unable to encode 0x8000000000000000 into binary format for int8 (OID 20): 18446744073709551605 is greater than maximum value for int64`

We can either rely on this or add checks but then we probably need checks for every BIGINT and INTEGER field since we do like our unsigned types.

n8mgr (approved):

peterjan (approved):

PR #183, opened 2025-06-05
Use local revision

### LINKED ISSUE #144: Store revisions in database
Instead of fetching the latest revision from the host, we should store the revision locally any time we interact with the host and it sends us the updated revision. We should sync revisions on demand when the host responds with an `invalid signature`. This will also involve updating `coreutils` to ensure all RPCs that we send contracts to will respond with the right error and if not fix it.

peterjan:
This PR gets rid of `RPCLatestRevision` calls and instead uses a new helper called `withRevision` which 'll fetch the revision from the database, try and revise the contract, sync the revision if necessary and update the database with the updated revision. 

I added a little benchmark that fetches and updates revisions just to be sure. On my machine it runs in under 2ms/op (that's fetch+update).

Closes https://github.com/SiaFoundation/indexd/issues/144


hosts/client.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+func (c *hostClient) AppendSectors(ctx context.Context, hostPrices proto.HostPrices, contractID types.FileContractID, sectors []types.Hash256) (rhp.RPCAppendSectorsResult, error) {
+	// sanity check
+	if len(sectors) > proto.MaxSectorBatchSize {
+		return rhp.RPCAppendSectorsResult{}, fmt.Errorf("too many sectors, %d > %d", len(sectors), proto.MaxSectorBatchSize) // developer error
+	}
+
+	// fetch revision and check if it meets the requirements
+	rev, err := rhp.RPCLatestRevision(ctx, c.client, contractID)
```
n8mgr: Should this not also be `withRevision`?

peterjan: 🤦‍♂️  forgot one, thank you

contracts/manager.go
```diff
@@ -46,19 +43,6 @@ type (
 		V2TransactionSet(basis types.ChainIndex, txn types.V2Transaction) (types.ChainIndex, []types.V2Transaction, error)
 	}
 
-	// HostClient defines the dependencies required to form, renew and refresh
-	// contracts.
-	HostClient interface {
```
ChrisSchinnerl: Tbh I'm not a huge fan of this since it ruins our nice separation of interfaces where every package only defines a `Client` interface with the method it actually needs.  It also moves the whole implementation into `hosts` where none of the methods that are exclusively used by `contracts` really belong.

I think the best solution here would be to explore moving the client implementation into a separate package for the rhp client and have `contracts` and `hosts` reference that while keeping their own minimal interface of the client.

hosts/client.go
```diff
@@ -0,0 +1,339 @@
...
+		update, err = reviseFn(rev)
+	}
+	if err != nil {
+		return err
+	}
+
+	// update revision in the database
+	err = c.store.UpdateContractRevision(ctx, contractID, update)
+	if err != nil {
+		c.log.Debug("failed to update contract revision", zap.Error(err))
```
ChrisSchinnerl: Imo if updating the database fails that usually warrants a `Error` log. There is no reason this should be the case.

hosts/client.go
```diff
@@ -0,0 +1,339 @@
...
+		return fmt.Errorf("%d > %d (%d+%d), %w", rev.ProofHeight, maxProofHeight, bh, revisionSubmissionBuffer, ErrContractNotRevisable)
+	} else if rev.RenterOutput.Value.IsZero() {
+		return ErrContractOutOfFunds
+	}
+
+	// revise the contract
+	update, err := reviseFn(rev)
+
+	// try and sync the revision if we got an error that indicates the revision is invalid
+	if err != nil && strings.Contains(err.Error(), proto.ErrInvalidSignature.Error()) {
```
ChrisSchinnerl: I'd drop a debug log here indicating the sync attempt.

hosts/client.go
```diff
@@ -0,0 +1,339 @@
...
+	err = c.store.UpdateContractRevision(ctx, contractID, resp.Contract)
+	if err != nil {
+		c.log.Debug("failed to update contract revision", zap.Error(err))
+		return types.V2FileContract{}, false, fmt.Errorf("failed to update contract revision: %w", err)
+	}
+
+	return resp.Contract, resp.Renewed, nil
+}
+
+func (c *hostClient) withRevision(ctx context.Context, contractID types.FileContractID, reviseFn func(revision types.V2FileContract) (types.V2FileContract, error)) error {
```
ChrisSchinnerl: probably worthwhile to add a docstring mentioning that whatever happens within reviseFn needs to be idempotent.

hosts/client.go
```diff
@@ -0,0 +1,339 @@
...
+			return types.V2FileContract{}, err
+		}
+		return res.Revision, nil
+	}); err != nil {
+		return rhp.RPCReplenishAccountsResult{}, 0, fmt.Errorf("failed to replenish accounts: %w", err)
+	}
+	return res, funded, nil
+}
+
+func (c *hostClient) syncRevision(ctx context.Context, contractID types.FileContractID, revision types.V2FileContract) (types.V2FileContract, bool, error) {
```
ChrisSchinnerl: `syncRevision` and `withRevision` are pretty much the core of this PR. Reminder to add thorough testing.

ChrisSchinnerl (changes requested):

contracts/manager.go
peterjan: The main issue is the `Dialer` actually. Because that returns the `HostClient` interface and you can't easily define a subset of the interface but have a shared dialer implementation. I have an idea though, it'll be ready for review tomorrow.

client/client.go
```diff
@@ -82,21 +103,22 @@ func (c *HostClient) AppendSectors(ctx context.Context, hostPrices proto.HostPri
...
+		if revision.Filesize > maxContractSize {
+			return types.V2FileContract{}, fmt.Errorf("contract is too large, %d > %d", revision.Filesize, maxContractSize)
+		}
+		res, err = rhp.RPCAppendSectors(ctx, c.client, c.signer, c.cm.TipState(), hostPrices, rhp.ContractRevision{ID: contractID, Revision: revision}, sectors)
+		if err != nil {
+			return types.V2FileContract{}, fmt.Errorf("failed to append sectors: %w", err)
+		}
+		return res.Revision, nil
+	}); err != nil {
+		return rhp.RPCAppendSectorsResult{}, fmt.Errorf("failed to fetch append sectors: %w", err)
```
ChrisSchinnerl: strange error message

client/client.go
```diff
@@ -82,21 +103,22 @@ func (c *HostClient) AppendSectors(ctx context.Context, hostPrices proto.HostPri
...
-		return rhp.RPCAppendSectorsResult{}, fmt.Errorf("failed to fetch latest revision: %w", err)
-	} else if !rev.Revisable {
-		return rhp.RPCAppendSectorsResult{}, errors.New("contract is not revisable")
-	} else if rev.Contract.RenterOutput.Value.IsZero() {
-		return rhp.RPCAppendSectorsResult{}, errors.New("contract is out of funds")
-	} else if rev.Contract.Filesize > maxContractSize {
-		return rhp.RPCAppendSectorsResult{}, fmt.Errorf("contract is too large, %d > %d", rev.Contract.Filesize, maxContractSize)
+	// append sectors
+	var res rhp.RPCAppendSectorsResult
+	if err := c.withRevision(ctx, contractID, func(revision types.V2FileContract) (_ types.V2FileContract, err error) {
```
ChrisSchinnerl:
```suggestion
	if err := c.withRevision(ctx, contractID, func(revision types.V2FileContract) (types.V2FileContract, error) {
```

client/client.go
```diff
@@ -116,118 +138,181 @@ func (c *HostClient) FormContract(ctx context.Context, settings proto.HostSettin
...
+	// fetch revision from database
+	rev, renewed, err := c.store.ContractRevision(ctx, contractID)
+	if err != nil {
+		return fmt.Errorf("failed to fetch contract revision: %w", err)
+	} else if renewed {
+		return ErrContractRenewed
+	} else if rev.ProofHeight > maxProofHeight {
+		return fmt.Errorf("%d > %d (%d+%d), %w", rev.ProofHeight, maxProofHeight, bh, revisionSubmissionBuffer, ErrContractNotRevisable)
+	} else if rev.RenterOutput.Value.IsZero() {
+		return ErrContractOutOfFunds
```
ChrisSchinnerl: I'm not sure this makes sense. Having a contract with exactly 0SC in it will likely never happen in the first place so it might be obsolete anyway but also we don't know what the caller has planned with the revision. So we might not know whether not having funds in it is an actual issue. The caller will need to check the funds anyway for whatever use-case they need the revision for.

client/client.go
```diff
@@ -116,118 +138,181 @@ func (c *HostClient) FormContract(ctx context.Context, settings proto.HostSettin
...
+	}
+
+	// revise the contract
+	update, err := reviseFn(rev)
+
+	// try and sync the revision if we got an error that indicates the revision is invalid
+	if err != nil && strings.Contains(err.Error(), proto.ErrInvalidSignature.Error()) {
```
ChrisSchinnerl, line 288: I really wish we had been a bit more explicit about revision mismatches than just returning a `ErrInvalidSignature` 😅 Guess it's too late now.

ChrisSchinnerl (changes requested):

client/client.go
peterjan: What would you return extra? There's not much context to give to a hash not matching up right?

n8mgr (dismissed):

ChrisSchinnerl (dismissed):

ChrisSchinnerl (approved):

PR #184, opened 2025-06-05
Add UTXO persistence

### LINKED ISSUE #173: Implement persistence for UTXOs
Similar to `renterd` and `hostd`. The 3 methods are already there but are currently no-ops.

peterjan:
This PR adds UTXO persistence, decided to add a benchmark for good measure.

Closes https://github.com/SiaFoundation/indexd/issues/173

### DIFF
--- a/persist/postgres/init.sql
+++ b/persist/postgres/init.sql
@@ -112,6 +112,13 @@ CREATE TABLE wallet_siacoin_elements (
     maturity_height INTEGER NOT NULL
 );
 
+CREATE TABLE wallet_locked_utxos (
+    id SERIAL PRIMARY KEY,
+	output_id BYTEA UNIQUE NOT NULL CHECK (LENGTH(output_id) = 32),
+	unlock_at TIMESTAMP WITH TIME ZONE NOT NULL
+);
+CREATE INDEX wallet_locked_utxos_unlock_at_idx ON wallet_locked_utxos(unlock_at);
+
 CREATE TABLE global_settings (
     id INTEGER PRIMARY KEY NOT NULL DEFAULT 0 CHECK (id = 0), -- enforce a single row
     db_version INTEGER NOT NULL, -- used for migrations
--- a/persist/postgres/init.sql
+++ b/persist/postgres/init.sql
@@ -114,8 +114,8 @@ CREATE TABLE wallet_siacoin_elements (
 
 CREATE TABLE wallet_locked_utxos (
     id SERIAL PRIMARY KEY,
-	output_id BYTEA UNIQUE NOT NULL CHECK (LENGTH(output_id) = 32),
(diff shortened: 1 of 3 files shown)


persist/postgres/init.sql
```diff
@@ -112,6 +112,13 @@ CREATE TABLE wallet_siacoin_elements (
     maturity_height INTEGER NOT NULL
 );
 
+CREATE TABLE wallet_locked_utxos (
+    id SERIAL PRIMARY KEY,
+	output_id BYTEA UNIQUE NOT NULL CHECK (LENGTH(output_id) = 32),
+	unlock_at TIMESTAMP WITH TIME ZONE NOT NULL
```
ChrisSchinnerl: Indentation

persist/postgres/wallet_test.go
```diff
@@ -4,12 +4,123 @@ import (
...
+	var count int
+	err = store.pool.QueryRow(context.Background(), `SELECT COUNT(*) FROM wallet_locked_utxos`).Scan(&count)
+	if err != nil {
+		t.Fatal(err)
+	} else if count != 0 {
+		t.Fatalf("expected 0 locked outputs, got %d", count)
+	}
+}
+
+func BenchmarkWalletLockUnlock(b *testing.B) {
```
ChrisSchinnerl, line 72: The benchmark results seem to be quite slow? Considering that we just insert into a table and it takes 50ms for 1000 rows.

ChrisSchinnerl (changes requested):

persist/postgres/wallet_test.go
peterjan: I was going to comment on that myself but figured I'd leave it. There's really not much indices to add considering the schema is that straightforward and we have indices on all 3 columns. I chose 1M database size which was very generous and I don't see us locking 1000 outputs any time soon either. Wouldn't you agree? I would look into it more closely but don't really see what more there is to do.

peterjan: I'll check the benchmark, run it with more batch sizes and remove the shufle from the loop.

peterjan: Do you mind running it on your machine Chris. If I alter the database within the loop and use stop-start timer I can only finish the benchmark if I manually set benchtime to 100x.


```
cpu: Apple M1 Max
BenchmarkWalletLockUnlock/LockedUTXOs     100    3171235 ns/op
BenchmarkWalletLockUnlock/LockUTXOs       100    16452679 ns/op
BenchmarkWalletLockUnlock/ReleaseUTXOs    100    5677362 ns/op
```

n8mgr (approved):

ChrisSchinnerl (approved):

PR #187, opened 2025-06-10
Ensure database exists

peterjan:
This PR avoids an unnecessary error at startup when starting the daemon using a database that doesn't exist yet.

```
daemon startup failed: failed to connect to postgres database: failed to initialize store: failed to initialize database: failed to begin transaction: failed to connect to `user=postgres database=indexd`: [::1]:5432 (localhost): server error: FATAL: database "indexd" does not exist (SQLSTATE 3D000)
```

### DIFF
--- a/persist/postgres/store.go
+++ b/persist/postgres/store.go
@@ -72,7 +72,7 @@ func Connect(ctx context.Context, ci ConnectionInfo, log *zap.Logger) (*Store, e
 		pool.Close()
 		return nil, fmt.Errorf("failed to connect to database: %w", err)
 	}
-	log.Sugar().Infof("successfully connected to database %q on %s:%d", ci.Database, ci.Host, ci.Port)
+	log.Info("connected", zap.String("database", ci.Database), zap.String("host", ci.Host), zap.Int("port", ci.Port))
 
 	store := &Store{
 		pool: pool,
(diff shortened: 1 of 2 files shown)


persist/postgres/store.go
```diff
@@ -61,10 +61,19 @@ func (s *Store) Close() error {
...
+
 	pool, err := pgxpool.New(ctx, ci.String())
 	if err != nil {
-		return nil, fmt.Errorf("failed to connect: %w", err)
+		return nil, fmt.Errorf("failed to create pool: %w", err)
+	} else if err := pool.Ping(ctx); err != nil {
+		pool.Close()
+		return nil, fmt.Errorf("failed to connect to database: %w", err)
 	}
+	log.Sugar().Infof("successfully connected to database %q on %s:%d", ci.Database, ci.Host, ci.Port)
```
ChrisSchinnerl: (nit) I'd prefer `.With(` fields over a sugared logger. Does it matter? Probably not. But if anyone ever feels like parsing any information out of our logs it's easier if we stick to fields.

ChrisSchinnerl (dismissed):

persist/postgres/store.go
peterjan: I actually considered that and opted to go this route. I don't mind changing it to structured logging but I really don't want to make the same mistake we made with `renterd` (in my opinion) where the structured logging is fairly useless and even worse we were repeating the same information twice.

n8mgr (approved):

ChrisSchinnerl (approved):

Issue #189, opened 2025-06-10
FTUE


peterjan:
Couple of findings after running `indexd` locally for the first time:

- we don't mention the location of the config file
- we don't ensure the database exists if you configure it to use a database different from the default postgres database
- we don't have default settings (both contract and gouging settings)
- we don't consider hosts to be usable (uptime of 1 week, price table validity)
- we don't ignore 0 values in max price settings

After fixing the above, we can successfully form contracts and fund accounts. 

We could probably use:
- a debug endpoint like `/debug/trigger` to easily (and forcefully) trigger certain loops
- a debug endpoint to write sectors to host so we can test sector pinning

PR #191, opened 2025-06-11
Update host checks and settings.

peterjan:
This PR changes 3 things:

It updates the `settings_valid_until` check and requires a minimum of 15 minutes instead of an hour. Most hosts set the validity to 30 minutes, so an hour definitely doesn't make sense unless we update the host code. Because it's a minimum value, I opted to go with 15 minutes as an absolute minimum. The hosts should really commit to prices for at least an hour though, and even more, in my opinion.

It allows disabling price checks on hosts by ignoring the check if the value in the database is 0. This allows the user to set the setting at 0 and essentially disable the check all together. An alternative would've been default settings but I think we should support disabling them all together anyway. This way we can keep the defaults at 0 for now.

I enabled contract maintenance by default. I think it's just annoying to disable it by default and have everyone manually update it before the indexer starts forming contracts. I understand why we didn't do this though and I'm fine with disabling it by default. Just a suggestion I had after running `indexd` for the first time locally.

Related https://github.com/SiaFoundation/indexd/issues/189

### DIFF
--- a/persist/postgres/hosts.go
+++ b/persist/postgres/hosts.go
@@ -89,13 +89,13 @@ WITH globals AS (
 	has_settings AND settings_max_contract_duration >= globals.contracts_period,
 	has_settings AND settings_max_collateral >= settings_collateral * globals.one_tb * globals.contracts_period,
 	has_settings AND settings_version >= globals.host_min_version,
-	has_settings AND settings_valid_until >= (NOW() + INTERVAL '1 hour'),
+	has_settings AND settings_valid_until >= (NOW() + INTERVAL '15 minutes'),
 	has_settings AND settings_accepting_contracts,
 	has_settings AND settings_contract_price <= globals.one_sc,
-	has_settings AND settings_collateral >= globals.hosts_min_collateral AND settings_collateral >= 2 * settings_storage_price,
-	has_settings AND settings_storage_price <= globals.hosts_max_storage_price,
-	has_settings AND settings_ingress_price <= globals.hosts_max_ingress_price,
-	has_settings AND settings_egress_price <= globals.hosts_max_egress_price,
+	has_settings AND (globals.hosts_min_collateral = 0 OR (settings_collateral >= globals.hosts_min_collateral AND settings_collateral >= 2 * settings_storage_price)),
+	has_settings AND (globals.hosts_max_storage_price = 0 OR settings_storage_price <= globals.hosts_max_storage_price),
+	has_settings AND (globals.hosts_max_ingress_price = 0 OR settings_ingress_price <= globals.hosts_max_ingress_price),
+	has_settings AND (globals.hosts_max_egress_price = 0 OR settings_egress_price <= globals.hosts_max_egress_price),
 	has_settings AND settings_free_sector_price <= globals.one_sc / globals.one_tb
 FROM hosts CROSS JOIN globals;`, sqlPublicKey(hk)))
 		if errors.Is(err, sql.ErrNoRows) {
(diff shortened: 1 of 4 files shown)


persist/postgres/hosts.go
```diff
@@ -89,13 +89,13 @@ WITH globals AS (
...
 	has_settings AND settings_accepting_contracts,
 	has_settings AND settings_contract_price <= globals.one_sc,
-	has_settings AND settings_collateral >= globals.hosts_min_collateral AND settings_collateral >= 2 * settings_storage_price,
-	has_settings AND settings_storage_price <= globals.hosts_max_storage_price,
-	has_settings AND settings_ingress_price <= globals.hosts_max_ingress_price,
-	has_settings AND settings_egress_price <= globals.hosts_max_egress_price,
+	has_settings AND (globals.hosts_min_collateral = 0 OR (settings_collateral >= globals.hosts_min_collateral AND settings_collateral >= 2 * settings_storage_price)),
+	has_settings AND (globals.hosts_max_storage_price = 0 OR settings_storage_price <= globals.hosts_max_storage_price),
+	has_settings AND (globals.hosts_max_ingress_price = 0 OR settings_ingress_price <= globals.hosts_max_ingress_price),
+	has_settings AND (globals.hosts_max_egress_price = 0 OR settings_egress_price <= globals.hosts_max_egress_price),
```
ChrisSchinnerl, line 98: I don't think using 0 to disable the check is a good idea. If you want to explicitly do so set it to a high value but 0 is a bit dangerous.

persist/postgres/hosts.go
```diff
@@ -163,13 +163,13 @@ WITH globals AS (
 	has_settings AND settings_max_contract_duration >= globals.contracts_period,
 	has_settings AND settings_max_collateral >= settings_collateral * globals.one_tb * globals.contracts_period,
 	has_settings AND settings_version >= globals.host_min_version,
-	has_settings AND settings_valid_until >= (NOW() + INTERVAL '1 hour'),
+	has_settings AND settings_valid_until >= (NOW() + INTERVAL '15 minutes'),
```
ChrisSchinnerl, line 166: Don't we have some place where we check if the settings are still valid for 30 minutes or so? Might need to update that now.

persist/postgres/init.sql
```diff
@@ -129,7 +129,7 @@ CREATE TABLE global_settings (
     scanned_block_id BYTEA NOT NULL DEFAULT '\x0000000000000000000000000000000000000000000000000000000000000000'::bytea CHECK (LENGTH(scanned_block_id) = 32),
 
     -- contract manager settings
-    contracts_maintenance_enabled BOOLEAN NOT NULL DEFAULT FALSE,
+    contracts_maintenance_enabled BOOLEAN NOT NULL DEFAULT TRUE,
```
ChrisSchinnerl, line 132: Didn't we decide to keep this `FALSE` to give people a chance to review the defaults before starting maintenance?

ChrisSchinnerl (changes requested):

persist/postgres/init.sql
peterjan: On first usage you generally need to sync the blockchain anyway, so in practice I feel it's just annoying to disable it by default. I don't mind reverting it.

persist/postgres/hosts.go
peterjan: Are you sure? I think disabling by setting 0 is fine and setting a high value is a bit odd because it's currencies.

persist/postgres/hosts.go
peterjan: Yeah we do that for long running operations like pinning and pruning. This asserts a minimum of 15 minutes, so I think the checks we have there can stay? Looks like this is going to change anyway with the PR you have in `DRAFT`

peterjan:
@ChrisSchinnerl I simply responded to your questions for the time being because it looks like we might even want to completely close it out if we don't want to do 0 for disable, disable by default and after your PR gets merged.

persist/postgres/hosts.go
ChrisSchinnerl: Yes, there is no reason you'd ever want to disable them in production. If you really don't care set them to $100/TB or whatever. 
Alternatively we can also choose not to accept 0 as a value at all. But a user who forgets to set a price limit shouldn't be able to accidentally disable it just because the default value of the currency type happens to be 0.

persist/postgres/init.sql
ChrisSchinnerl: It might be annoying right now without a UI but it's going to be the same as it is on renterd once we have it.

peterjan:
more or less redundant after https://github.com/SiaFoundation/indexd/pull/194 - especially if we don't want 0 to disable checks

PR #192, opened 2025-06-11
SDK

n8mgr:
Adds an initial SDK with support for uploading and downloading erasure-coded slabs. 

The existing API client is missing the signed methods for pinning. Is the idea that they would go on their own client rather than in the "admin" API client?

```
goos: darwin
goarch: arm64
pkg: go.sia.tech/indexd
cpu: Apple M4 Max
BenchmarkUpload-16      	       2	 649487958 ns/op	 394.16 MB/s	2055585020 B/op	   34318 allocs/op
BenchmarkDownload-16    	       3	 335777181 ns/op	 762.41 MB/s	1225120946 B/op	    4922 allocs/op
```


client.go
ChrisSchinnerl, line 1: How about moving this into an `sdk` package and then call `sdk.New`? 
Gives us a bit more flexibility with splitting things up into files, adding a separate `README` just for the sdk etc.

client.go
```diff
@@ -0,0 +1,474 @@
...
+		return Slab{}, fmt.Errorf("not enough hosts available: %d, required: %d", len(hosts), dataShards+parityShards)
+	}
+
+	errCh := make(chan error, len(shards))
+	nonce := make([]byte, 24)
+	for i := range shards {
+		// encrypt the shard before upload
+		nonce[0] = byte(i)
+		c, _ := chacha20.NewUnauthenticatedCipher(encryptionKey[:], nonce)
+		c.XORKeyStream(shards[i], shards[i])
```
ChrisSchinnerl: Did you check if moving this into the goroutine makes any difference in the benchmark?

client.go
```diff
@@ -0,0 +1,474 @@
...
+	// ErrNotEnoughShards is returned when not enough shards were
+	// uploaded or downloaded to satisfy the minimum required shards.
+	ErrNotEnoughShards = errors.New("not enough shards")
+)
+
+func (s *SDK) uploadSlab(ctx context.Context, encryptionKey [32]byte, shards [][]byte, dataShards, parityShards uint8, timeout time.Duration) (Slab, error) {
+	if len(shards) == 0 {
+		return Slab{}, errors.New("no shards to upload")
+	}
+
```
ChrisSchinnerl: I'd derive a `ctx, cancel := context.WithCancel(ctx)` up here and `defer` the `cancel`. So when you return at any point in time, you clean up any leftover routines immediately.

client.go
```diff
@@ -0,0 +1,474 @@
...
+	downloadOption struct {
+		hostTimeout time.Duration
+	}
+
+	// A HostDialer is an interface for writing and reading sectors to/from hosts.
+	HostDialer interface {
+		// Hosts returns the public keys of all hosts that are available for
+		// upload or download.
+		Hosts() []types.PublicKey
```
ChrisSchinnerl: Do you expect `HostDialer` to be responsible for geographically distributing the shards and making sure there is some randomness to the hosts picked? Trying to avoid all users of the SDK defaulting to the same 30 hosts and those 30 hosts being in the same data center.

Another thing that would be interesting to incorporate in the SDK eventually is backoffs for hosts that time out or just fail.

client.go
```diff
@@ -0,0 +1,474 @@
...
+			return Slab{}, ctx.Err()
+		case err := <-errCh:
+			if err == nil {
+				completed++
+			}
+		}
+	}
+
+	// redundancy is guaranteed to be at least 2x; 80% of shards
+	// being present is more than enough to repair.
```
ChrisSchinnerl: As of right now this won't work. I think the indexer expects sector information for every sector of the slab. 

I'm afraid if we loosen this rule we make it very easy to overwhelm the indexer. For every 40MiB chunk that is missing one 4MiB sector, the indexer needs to download 40MB and upload 4MB. 
So a single client uploading at 3Gbps to the network would require ~1Gbps of indexer bandwidth to keep up. 
Even at only 10s of users uploading at the same time on average with 1 failing shard this will quickly become unfeasible.

client.go
```diff
@@ -0,0 +1,474 @@
...
+
+			data, err := downloadShard(ctx, sector.Root, sector.HostKey, s.dialer, timeout)
+			if err == nil {
+				shards[i] = data[:]
+				if v := successful.Add(1); v >= uint32(slab.MinSectors) {
+					// got enough pieces to recover
+					cancel()
+				}
+			}
+		}(ctx, sector, i)
```
ChrisSchinnerl: I think we can follow up with an improvement here as well. This is quite wasteful in terms of bandwidth and chances are that all hosts start sending data at the same time. Worst case you end up with 1/3 the download speed of the best case using 3x redundancy. The worst case will also become increasingly likely on slower internet connections and parallel downloads. 

An easy and straightforward improvement could be to just use 12 shards out of 30 and launch more once we have 2+ failure.

client.go
```diff
@@ -0,0 +1,474 @@
...
+				return pinned, nil
+			} else if work.err != nil {
+				return nil, work.err
+			}
+			slab, err := s.uploadSlab(ctx, shardKey, shards, uo.dataShards, uo.parityShards, uo.hostTimeout)
+			if err != nil {
+				return nil, fmt.Errorf("failed to upload slab %d: %w", i, err)
+			}
+			slab.Length = uint32(work.length)
+			// TODO: pin slab
```
ChrisSchinnerl: Another todo might be cleanup of pinned slabs when the upload fails or giving the caller an easy way to do it themselves.

client_test.go
```diff
@@ -0,0 +1,214 @@
...
+	data := frand.Bytes(benchmarkSize)
+
+	b.SetBytes(benchmarkSize)
+	b.ResetTimer()
+	for b.Loop() {
+		if _, err := sdk.Upload(context.Background(), bytes.NewReader(data)); err != nil {
+			b.Fatalf("failed to upload: %v", err)
+		}
+	}
+}
```
ChrisSchinnerl: One more variation of the benchmarks that would be great to have is one with hosts timing out. Ideally for various numbers such as 1, 3 and maybe 5 since those seem realistic in production. To rule out that we are loosing a lot of performance on a minority of hosts acting up.

client.go
```diff
@@ -0,0 +1,474 @@
...
+				if len(hosts) == 0 {
+					errCh <- errors.New("no hosts available")
+					hostsMu.Unlock()
+					return
+				}
+				hostKey := hosts[0]
+				hosts = hosts[1:]
+				hostsMu.Unlock()
+
+				root, err := uploadShard(ctx, sector, hostKey, s.dialer, timeout) // error can be ignored, hosts will be retried until none are left and the upload fails.
```
peterjan: I was going to comment that it should be debug logged at least but I see there's no logger. Should we add one? Did you leave it out on purpose?

client.go
```diff
@@ -0,0 +1,474 @@
...
+	// A SlabSector represents a sector in a slab
+	SlabSector struct {
+		Root    types.Hash256
+		HostKey types.PublicKey
+	}
+
+	// A Slab represents a collection of erasure-coded sectors
+	Slab struct {
+		SectorKey  [32]byte
+		MinSectors uint8
```
peterjan: `nit` I would call this `MinShards` because all error messages mention shards, number of required shards etc

client.go
```diff
@@ -0,0 +1,474 @@
...
+)
+
+var (
+	// ErrNotEnoughShards is returned when not enough shards were
+	// uploaded or downloaded to satisfy the minimum required shards.
+	ErrNotEnoughShards = errors.New("not enough shards")
+)
+
+func (s *SDK) uploadSlab(ctx context.Context, encryptionKey [32]byte, shards [][]byte, dataShards, parityShards uint8, timeout time.Duration) (Slab, error) {
+	if len(shards) == 0 {
```
peterjan: We can return early here if we don't have enough shards to meet min. redundancy

client.go
```diff
@@ -0,0 +1,474 @@
...
+					}
+					errCh <- nil
+					break
+				}
+			}
+		}(ctx, shards[i], i)
+	}
+
+	var completed uint8
+	for range len(shards) {
```
peterjan: It's not that I don't like the pattern but I feel the combination of an `errCh` and a `completed` count is relatively useless because we're not evaluating the error at all and we require sending a `nil` value to make the count. Would it not be slightly cleaner if we add use a waitgroup, then wait for it and loop the slab's sectors and compare the root against an empty root. It would also be consistent with the pattern on the download.

client.go
```diff
@@ -0,0 +1,474 @@
...
+}
+
+// Upload uploads the data to hosts and pins it to the indexer.
+//
+// Returns the metadata of the slabs that were pinned
+func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) ([]Slab, error) {
+	uo := uploadOption{
+		dataShards:   10,
+		parityShards: 20,
+		hostTimeout:  4 * time.Second, // ~10 Mbps
```
peterjan: This is very aggressive I think, wouldn't a more reasonable value be something like 10 seconds.

client.go
```diff
@@ -0,0 +1,474 @@
...
+// TODO: support seeks
+func (s *SDK) Download(ctx context.Context, w io.Writer, metadata []Slab, opts ...DownloadOption) error {
+	do := downloadOption{
+		hostTimeout: 4 * time.Second, // ~10 Mbps
+	}
+	for _, opt := range opts {
+		opt(&do)
+	}
+
+	if len(metadata) == 0 {
```
peterjan: `nit` would move it up

client.go
```diff
@@ -0,0 +1,474 @@
...
+				return err
+			}
+			writeLen -= n
+		}
+	}
+	return nil
+}
+
+// downloadShard reads a sector from a host
+func downloadShard(ctx context.Context, root types.Hash256, hostKey types.PublicKey, dialer HostDialer, timeout time.Duration) (*[proto4.SectorSize]byte, error) {
```
peterjan: do we want to check if timeout is greater than zero?

client.go
```diff
@@ -0,0 +1,474 @@
...
+	}
+
+	type work struct {
+		shards [][]byte
+		err    error
+	}
+	workCh := make(chan work, 1)
+	go func(ctx context.Context, slabs []Slab) {
+		for i, slab := range metadata {
+			shards, err := s.downloadSlab(ctx, slab, do.hostTimeout)
```
peterjan: Another todo in a f/u would be to keep track of what hosts timed out so we can retry those hosts if we don't manage to collect enough pieces to recover from the other hosts.

client.go
```diff
@@ -0,0 +1,474 @@
...
+func WithRedundancy(dataShards, parityShards uint8) UploadOption {
+	return func(uo *uploadOption) {
+		uo.dataShards = dataShards
+		uo.parityShards = parityShards
+	}
+}
+
+// WithUploadHostTimeout sets the timeout for writing sectors to individual
+// hosts. This avoids long hangs when a host is unresponsive or slow.
+// The default timeout is 4 seconds, worst case around 300Mbps.
```
peterjan: Shouldn't this be 10Mbps?  (x2)

client.go
n8mgr: I don’t mind, but generic names for Sdk/api has been getting on my nerves lately. For example, importing the hostd, renterd, and walletd API package into a single app. `index` maybe?

client.go
n8mgr: Yeah. The host dialer will be primarily responsible for grabbing hosts from an indexd service and picking which hosts to try.

client.go
n8mgr: There's nothing an SDK user can do to react to these errors. No need to add a dependency if we don't have to.

client.go
n8mgr: I would agree that we can switch to a WaitGroup + `atomic.Int32` instead of an error channel. With Chris' comment that we shouldn't consider files with `> 80% uploaded` successful, we would end up returning the error from the channel and cancelling early though.

client.go
n8mgr: Significant slowdown
```
goos: darwin
goarch: arm64
pkg: go.sia.tech/indexd
cpu: Apple M4 Max
BenchmarkUpload-16    	       2	 746132354 ns/op	 343.10 MB/s	2055771804 B/op	   34745 allocs/op
PASS
ok  	go.sia.tech/indexd	1.903s
```

client.go
n8mgr: Is minimum of 10Mbps transfer speed really that aggressive? I don't mind for upload so much as I'd rather wait a little longer than fail the upload, but it does bring up a question of what are our minimum speed expectations? 

Increasing to 10s would be a transfer speed of 3Mbps per sector. At that point, wouldn't you rather time out and try for a host with better peering?

client.go
n8mgr: Yeah, comment is a little misleading as it's taking the transfer speed and multiplying by the default shards length to get the theoretical maximum. i.e. if you're concurrently uploading to 30 locations each at 10Mbps, your aggregate transfer speed is 300Mbps.

client.go
n8mgr: I don't think it matters since it will instantly timeout

client_test.go
n8mgr: That's expected for uploads, no? We have to wait on all shards to complete so upload speed is hard capped by the slowest host.

client.go
peterjan: Hm yeah, definitely tricky. I think minimum speed expectations should probably be measured across a certain time frame so the host can be excluded are marked as bad based on historical performance rather than setting a very strict timeout for every sector write.

The thing that worries me most with host timeouts is how we handle retries, or failure in the event we run out of hosts. It would be a shame if your batch upload fails because of a failed slab upload only because we timed out a host after 4 seconds. That doesn't mean though that I don't think 10 Mbps is not a good minimum value...

Generally we use an overdrive mechanism to tackle this situation because you get quite a decent tradeoff, in the best case you get lucky and you have fast hosts and the upload/download completes before the mechanism kicks in. On average though some overdrive requests launch but the slow ones aren't cancelled, and might even beat the overdrive requests to the punch. That obviously adds a lot of complexity but it is quite nice and avoids awkward retry mechanisms on failure.

client_test.go
ChrisSchinnerl: In the basic case (1 goroutine for each shard and spinning up new goroutines only for failing shards) that is true. Worst case might actually be the slowest host timing out and then following up with another slow host. 

A benchmark could provide us with baseline performance stats to later optimize this.

e.g. iirc what renterd does is add 2 new config values. One is a more ambitious timeout and the other one a number of shards. When this more ambitious timeout has passed and the remaining shards are <= the configured threshold of 3 or so, we launch one more worker for each of them. To effectively optimize around the edge case of 1-3 hosts being too slow.

As I'm typing this out I'm realizing that RHP4 is making this pretty cheap now. Because uploading a sector no longer involves paying for 6 weeks or so of storage. Only pinning does.

sdk/client.go
```diff
@@ -0,0 +1,515 @@
...
+	for i := range shards {
+		// encrypt the shard before upload
+		nonce[0] = byte(i)
+		c, _ := chacha20.NewUnauthenticatedCipher(encryptionKey[:], nonce)
+		c.XORKeyStream(shards[i], shards[i])
+
+		select {
+		case <-ctx.Done():
+			return Slab{}, ctx.Err()
+		case sema <- struct{}{}:
```
ChrisSchinnerl, line 127: I like the semaphore approach as a simpler alternative for renterd's overdrive. But I think the upload code won't benefit as it is right now since failed shards won't get relaunched if the semaphore allows for it.

client.go
ChrisSchinnerl: This number also caught my attention but I initially didn't want to comment on it. I like the fact that this PR doesn't use the exact same overdrive mechanism and instead relies on a semaphore which serves a similar purpose really. With the only exception being that you overdrive downloads from the start.

However, what I'm worried about isn't really slow hosts but slow clients. Not because their ISP is bad but because they might have parallel stuff happening on their machine.

So what if we had 2 timeouts?
1. soft timeout -> 4s; we move on from a host and try another one but don't abort the upload/download
2. hard timeout -> 30-60s; the connection closes

I have to admit the code might end up being a bit more messy but that way uploads/downloads at least won't fail if your connection is temporarily slow.

sdk/client.go
n8mgr: I'm not sure I understand. Currently failed shards will be retried until there are no more hosts available.

client.go
n8mgr: Currently both upload and download implement "soft timeout" until there are no more hosts to try. I'm not sure it's beneficial to add a hard timeout where we abort the download.

n8mgr: Uploads and downloads from centralized services fail all the time due to timeouts/bad peering. I'm not sure it's beneficial to add significantly more complexity to try and tackle every edge-case. Particularly when the target transfer speed is low by consumer internet standards and when we're talking about an MVP SDK.

sdk/client.go
ChrisSchinnerl: heh I completely missed the loop.

client.go
ChrisSchinnerl: Yeah we can give this a try. Complexity is pretty much the reason I didn't comment on it on the first review.

ChrisSchinnerl (dismissed):

peterjan (dismissed):

peterjan (approved):

ChrisSchinnerl (approved):

PR #194, opened 2025-06-12
Update ScanHost to WithScannedHost

ChrisSchinnerl:
After a discussion we had yesterday we decided that rather than relying on the price table validity at all, we'd just fetch a new price table on demand. It's a free and fast operation that should only be necessary once per 30 minutes for most hosts.

To make this more convenient in code, this PR changes `ScanHost` to `WithScannedHost`. This will also implicitly check that the host is good which is a nice DX improvement as well and helps deduplicate some code.


contracts/form.go
```diff
@@ -215,60 +216,54 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
...
-		host, err := cm.scanner.ScanHost(scanCtx, candidates[i].PublicKey)
-		cancel()
-		if err != nil {
-			hostLog.Warn("failed to scan host", zap.Error(err))
-			continue
-		}
+		err := cm.scanner.WithScannedHost(ctx, hostKey, func(host hosts.Host) error {
+			// make sure host is still good
+			if !isGood(host, hostLog) {
+				return fmt.Errorf("host is not good: %s", host.PublicKey)
```
peterjan, line 163: Shouldn't we wrap `hosts.ErrBadHost` here? We could also move this check inside `WithScannedHost` to ensure we never forget to check whether the host is still good after scanning it.

edit: seems you're already doing that

contracts/form_test.go
```diff
@@ -155,14 +155,20 @@ func (s *storeMock) Scanner() *scannerMock {
 
 // ScanHost returns the preconfigured settings for the host or no settings to
 // simulate a failing scan. Upon success, the underlying store is updated.
-func (s *scannerMock) ScanHost(ctx context.Context, hk types.PublicKey) (hosts.Host, error) {
+func (s *scannerMock) WithScannedHost(ctx context.Context, hk types.PublicKey, fn func(h hosts.Host) error) error {
```
peterjan: update docstring

hosts/manager.go
```diff
@@ -278,6 +270,55 @@ func (m *HostManager) pruneHosts(ctx context.Context) {
...
+// settings and checks.
+func (m *HostManager) scanHost(ctx context.Context, hk types.PublicKey) (Host, error) {
+	logger := m.log.With(zap.Stringer("hk", hk))
+
+	host, err := m.store.Host(ctx, hk)
+	if err != nil {
+		return Host{}, fmt.Errorf("failed to get host, %w", err)
+	}
+
+	scanCtx, cancel := context.WithTimeout(ctx, scanTimeout)
```
peterjan: `nit` might as well move to the top and overwrite `ctx`

hosts/manager.go
```diff
@@ -178,53 +183,40 @@ func (m *HostManager) UpdateUsabilitySettings(ctx context.Context, us UsabilityS
...
+	logger.Debug("host has outdated prices, rescan")
 
-	settings, err := fetchSettings(scanCtx, m.scanner, hk, addrs, logger)
+	// scan the host if the prices were outdated
+	host, err = m.scanHost(ctx, hk)
 	if err != nil {
-		return Host{}, fmt.Errorf("failed to fetch settings, %w", err)
+		return fmt.Errorf("failed to scan host, %w", err)
+	} else if !host.IsGood() {
+		return ErrBadHost
```
peterjan: `return fmt.Errorf("%w: blocked=%t, usable=%t, networks=%d", ErrBadHost, host.Blocked, host.Usability.Usable(), len(host.Networks))` ?

peterjan (commented):
It feels weird to me to call `WithScannedHost` on a scanner. A scanner should expose `ScanHost`. The name doesn't really sit well because it also implies we scan the host first, and then execute the function. Perhaps we should get rid of the scanner interface, just call it `HostManager` and then rename it to `WithValidPrices` or `WithPriceRetry` or something.

contracts/form.go
ChrisSchinnerl: `WithScannedHost` performs the `h.IsGood` check indeed but `isGood` is the inline function that performs additional checks.

contracts/form_test.go
```diff
@@ -142,8 +142,8 @@ type scannerMock struct {
 // Scanner is a convenience method to create a scanner from a store mock. The
 // scanner contains all the settings of the hosts from the mocked store and will
 // be updating the store upon scanning.
-func (s *storeMock) Scanner() *scannerMock {
-	scannerMock := &scannerMock{
+func (s *storeMock) Scanner() *hostManagerMock {
```
peterjan: I would get rid of this convenience method. I'm not 100% sure but I touched this code recently and we very rarely make use of the fact that this convenience method transfers over the host settings. In most tests it's cleaner to do init a host manager mock and then do `hm.settings[pk] = settings` and not mention scanning at all.

peterjan (dismissed):

contracts/form.go
```diff
@@ -215,60 +216,54 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
...
+			}
 
-		// make sure the host is still good
-		if !isGood(host, hostLog) {
-			continue // ignore bad host
-		}
+			allowance, collateral := initialContractFunding(host.Settings.Prices, host.Settings.MaxCollateral, period)
+			formationCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
n8mgr, line 167: You can defer this cancel now since it's wrapped in a function

peterjan (dismissed):

n8mgr (dismissed):

PR #197, opened 2025-06-13
Move host client to its own package

peterjan:
This PR introduces a new package called `client` which houses the implementation of the host client. This allows us to easily mock a portion of the client by defining an interface that contains a subset of RPC methods required for that particular case, e.g. in the `accounts` we only need to replenish the accounts. 

This further cleans up the diff in https://github.com/SiaFoundation/indexd/pull/183 - which 'll only contain switching out the `RPCLatestRevision` for `withRevision`.

### DIFF
--- a/accounts/funder.go
+++ b/accounts/funder.go
@@ -23,15 +23,34 @@ type (
 		ReplenishAccounts(context.Context, types.FileContractID, []proto.Account, types.Currency) (rhp.RPCReplenishAccountsResult, int, error)
 	}
 
+	// Dialer defines an interface for dialing the host and returning a host client. This client can be used to
+	// interact with the host using the RHP methods. The client is expected to be closed when no longer needed.
+	Dialer interface {
+		DialHost(ctx context.Context, hostKey types.PublicKey, addr string) (HostClient, error)
+	}
+
 	// Funder dials a host and replenish a set of ephemeral accounts.
 	Funder struct {
-		dialer client.Dialer[HostClient]
+		dialer Dialer
 	}
 )
 
+type wrapper struct {
+	d *client.SiamuxDialer
+}
+
+// DialHost dials the host and returns a HostClient.
+func (w *wrapper) DialHost(ctx context.Context, hostKey types.PublicKey, addr string) (HostClient, error) {
+	client, err := w.d.DialHost(ctx, hostKey, addr)
+	if err != nil {
+		return nil, err
+	}
+	return client, nil
+}
+
 // NewFunder creates a new Funder.
-func NewFunder(dialer client.Dialer[HostClient]) *Funder {
-	return &Funder{dialer: dialer}
+func NewFunder(dialer *client.SiamuxDialer) *Funder {
+	return &Funder{dialer: &wrapper{d: dialer}}
 }
 
 // FundAccounts tops up the provided accounts to the target balance using the
--- a/accounts/funder.go
+++ b/accounts/funder.go
@@ -35,12 +35,12 @@ type (
 	}
 )
 
-type wrapper struct {
-	d rhp.Dialer
+type dialer struct {
+	d *rhp.SiamuxDialer
 }
 
-func (w *wrapper) DialHost(ctx context.Context, hostKey types.PublicKey, addr string) (HostClient, error) {
-	client, err := w.d.DialHost(ctx, hostKey, addr)
+func (d *dialer) DialHost(ctx context.Context, hostKey types.PublicKey, addr string) (HostClient, error) {
+	client, err := d.d.DialHost(ctx, hostKey, addr)
 	if err != nil {
 		return nil, err
 	}
@@ -48,8 +48,8 @@ func (w *wrapper) DialHost(ctx context.Context, hostKey types.PublicKey, addr st
 }
 
 // NewFunder creates a new Funder.
-func NewFunder(dialer rhp.Dialer) *Funder {
-	return &Funder{dialer: &wrapper{d: dialer}}
+func NewFunder(d *rhp.SiamuxDialer) *Funder {
+	return &Funder{dialer: &dialer{d: d}}
 }
 
 // FundAccounts tops up the provided accounts to the target balance using the
--- a/accounts/funder.go
+++ b/accounts/funder.go
@@ -4,98 +4,52 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"time"
+	"io"
 
-	"go.sia.tech/core/consensus"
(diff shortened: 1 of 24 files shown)


peterjan:
not sure if I like the package name myself, it annoyingly conflicts with coreutils obviously but at the same time it's most fitting

accounts/funder.go
```diff
@@ -4,98 +4,52 @@ import (
...
-	signer rhp.ContractSigner
-}
-
-func (c *hostClient) Dial(ctx context.Context, addr string, peerKey types.PublicKey) (rhp.TransportClient, error) {
-	return siamux.Dial(ctx, addr, peerKey)
-}
-
-func (c *hostClient) RPCLatestRevision(ctx context.Context, t rhp.TransportClient, fcid types.FileContractID) (proto.RPCLatestRevisionResponse, error) {
-	return rhp.RPCLatestRevision(ctx, t, fcid)
+type wrapper struct {
```
ChrisSchinnerl, line 38: I'd call these `dialer`.

rhp/client.go
```diff
@@ -0,0 +1,238 @@
...
+type (
+	// ChainManager defines an interface to access the chain state as well as
+	// interact with the transaction pool.
+	ChainManager interface {
+		TipState() consensus.State
+		rhp.TxPool
+	}
+
+	// Dialer is an interface for dialing a host.
+	Dialer interface {
```
ChrisSchinnerl: This interface is redundant. You can remove it and instead export `SiamuxDialer` and use that directly.

contracts/manager.go
```diff
@@ -165,6 +170,19 @@ type (
 	}
 )
 
+type wrapper struct {
+	d rhp.Dialer
+}
```
ChrisSchinnerl, line 183: I'm absolutely not used to Go generics and can't get it to work but something like this would be too sweet if it worked.

```go

type wrapper[OUT any] struct {
	d *siamuxDialer
}

// DialHost dials the host and returns a HostClient.
func (w *wrapper[OUT]) DialHost(ctx context.Context, hostKey types.PublicKey, addr string) (OUT, error) {
	client, err := w.d.DialHost(ctx, hostKey, addr)
	if err != nil {
		return nil, err
	}
	return client, nil
}

func Adapt[OUT any](d *siamuxDialer) OUT {
	// wrap the dialer to return the correct type
	return &wrapper[OUT]{d: d}
}
```

ChrisSchinnerl (changes requested):
Pretty decent refactor overall. Regarding the name of the package, I'd also not be opposed to just call it the `client` or `dialer` package.

rhp/client.go
```diff
@@ -0,0 +1,238 @@
...
+	if err != nil {
+		return rhp.RPCReplenishAccountsResult{}, 0, fmt.Errorf("failed to fetch latest revision: %w", err)
+	} else if rev.Contract.RenterOutput.Value.Cmp(target) < 0 {
+		return rhp.RPCReplenishAccountsResult{}, 0, ErrContractInsufficientFunds
+	}
+
+	// prepare batch
+	batchSize := int(min(rev.Contract.RenterOutput.Value.Div(target).Big().Uint64(), proto.MaxAccountBatchSize))
+	funded = min(batchSize, len(accounts))
+	batch := accounts[:funded]
```
n8mgr: Why not return an error if funded > batch size? This adds an unnecessary extra return value to check.

rhp/client.go
peterjan: I know but I liked that we obfuscated it was a siamux dialer. I removed it though

contracts/manager.go
peterjan: I tried generics but failed initially. I took another stab at it though and succeeded. There's some trade-offs but I like it.

n8mgr (changes requested):
This seems to be getting out of hand lol.. Why are we using generics?

I am probably missing a lot of context, so forgive my ignorance. Is there a reason a `Dialer interface`  that returns a `TransportClient` in each package that needs it and calling `rhp.RPC*` with it is untenable? Are we trying to avoid passing in fund and sign? We're effectively doing the same thing, just in a more roundabout manner.

```go
transport, err := siamux.Dial(...)
resp, err := rhp4.RPCLastestRevision(ctx,transport)
```

vs

```go
hc, err := dialer.Dial(...)
resp, err := hc.LatestRevision(ctx)
```

rhp/client.go
peterjan: If you look at `Funder.FundAccounts`, my intention was to allow passing in a large number of contracts and accounts to fund. Potentially larger than `proto.MaxAccountBatchSize` since it's a batch size after all. We loop the contracts and pass at most `proto.MaxAccountBatchSize` for every contract. The contract might not have enough funds to fund all accounts, but we do want to drain them as close to zero as possible, so that's why I return the number of funded accounts and limit the batch size in the client to the max accounts that can be funded with that contract. I think this is better than skipping over the contract because we can't fill all accounts. 

Preparing this answer I did have to alter the implementation of `Funder.FundAccounts` though. Refactoring this left an unwanted check that was limiting the number of accounts you can pass to the max batch size.

peterjan:
> This seems to be getting out of hand lol.. Why are we using generics?

Lol yeah I can see that. It's not that bad though. The snippet Chris left looks like overkill but what I ended up with is not that big of a deal. It allows getting a `Dialer` and essentially passing what interface it should return, as long as it's a valid subset of the methods the client supports.

 > I am probably missing a lot of context, so forgive my ignorance. Is there a reason a Dialer interface that returns a TransportClient in each package that needs it and calling rhp.RPC* with it is untenable? Are we trying to avoid passing in fund and sign? We're effectively doing the same thing, just in a more roundabout manner.

There's no reason we can't do it your way, and no, we are not trying to avoid passing in fund and sign. Moving the client implementation into its own package has nothing but benefits though. We used to have a `hostClient` in both the `accounts` and `contracts` package, that's duplicated code in a way, even though the client was only implementing the methods that  it needed.

For example, the change we want to make is switch from fetching the latest revision, to optimistically using the local revision from the database. Having a centralised client already pays off because you only have to update the code in one location. The most important benefit though is that we maintain the ability to only mock the interface you care about in the package you're in. E.g. in the `accounts` package you only have to mock the `HostClient` defined there. But you can get the actual implementation from the `client` package and pass your own interface... I mean that's kind of nice, don't see how we could be against that. It perfectly follows our interface rules, centralises the implementation and allows mocking a subset of the client that you care about in the package you're in.

ChrisSchinnerl (approved):

n8mgr (approved):

PR #119, opened 2025-04-16
Add `UnhealthySlab` method to database

ChrisSchinnerl:
That way we can fetch slabs that require migration

PR #198, opened 2025-06-13
Migrations

### LINKED ISSUE #50: Data integrity
This issue serves as a collection of data integrity related tasks

- [ ] Slab repair loop
- [x] Sector integrity checks

ChrisSchinnerl:
closes https://github.com/SiaFoundation/indexd/issues/50


slabs/migrations_test.go
```diff
@@ -162,5 +313,86 @@ func TestContractsForRepair(t *testing.T) {
 	// should use them
 	allHosts = append(allHosts, goodHost2, goodHost3)
 	allContracts = append(allContracts, cGoodHost2, cGoodHost3)
-	assertResult(allHosts, allContracts, []int{1, 2}, []int{8, 9})
+	assertResult(allHosts, allContracts, []int{1, 2}, []int{7, 8})
+}
+
+func TestRSReconstructSome(t *testing.T) {
```
peterjan, line 321: @n8maninger I had to swap out `ReconstructSome` for `Reconstruct` because it was throwing a segfault in `TestMigrateSlab`  👀 . I isolated what's happening in a test, it looks like a bug but that would surprise me of course... Initially I thought it was because the shard wasn't allocated and RS was trying to write into it, but even if I preallocate the required shard it still segfaults. I can make the segfaults go away if I init the byte slice with a capacity, but not length. Problem then is that the required data isn't being written into the shard, it's just all 0s.

peterjan: It might be a bug actually. Does it segfault on your machine as well?

n8mgr: Screenshot 2025-07-14 at 4 42 17 PM

Test passes with no issue

n8mgr: Oh. I have to uncomment lines. I see.

n8mgr: If I allocate the slice, I no longer get a panic, but it just fills it with junk data instead not zeroes.

```go
downloaded := make([][]byte, len(shards))
downloaded[0] = slices.Clone(shards[0])
downloaded[1] = make([]byte, 0, proto.SectorSize)
downloaded[2] = slices.Clone(shards[2])
downloaded[3] = make([]byte, 0, proto.SectorSize)
err = enc.ReconstructSome(downloaded, required)
if err != nil {
	t.Fatal(err)
}
for i := range downloaded {
	if required[i] && !bytes.Equal(downloaded[i], shards[i]) {
		t.Fatalf("expected shard %d to be reconstructed", i)
	}
}
```

n8mgr: Looks like a bug. I can't find anything wrong with our code.

peterjan: I'll open an [issue](https://github.com/klauspost/reedsolomon/issues/302). In the mean time @n8maninger  I propose we merge this PR as is and use `Reconstruct`. It would be nice to get this merged before Chris returns from holiday.

slabs/manager.go
```diff
@@ -277,12 +284,42 @@ func (m *SlabManager) performIntegrityChecks(ctx context.Context) error {
...
+func (m *SlabManager) performSlabMigrations(ctx context.Context) error {
 	start := time.Now()
 	logger := m.log.Named("migrations")
 	logger.Debug("starting slab migrations", zap.Time("start", start))
 
-	// TODO: implement
+	const slabsPerBatch = 10
+	nextBatch := func(ctx context.Context) (batch []Slab, _ error) {
+		for len(batch) < slabsPerBatch {
+			slab, err := m.store.UnhealthySlab(ctx, start)
```
n8mgr, line 295: Why are we doing it this way instead of adding a limit arg to the store?

peterjan: I'm not sure to be honest. Chris added it back in april in [this PR](https://github.com/SiaFoundation/indexd/pull/119). It looks like there's no reason not to go for a `limit` approach. I'm assuming he preferred this pattern and I apparently never questioned it. 

edit: I don't see a good reason not to add a limit though, if you want to I can add that in? Chris would usually have had a reason though, especially since he wrote a little `NOTE` in the docstring that mentions scalability. If anything though the `NOTE` would be an argument in favour of using a `LIMIT` approach if anything... Perhaps we can change it in a F/U so Chris can review it on Monday.

PR #200, opened 2025-06-17
Add `--debug` flag that exposes a debug endpoint

peterjan:
This PR add a `--debug` flag that when enabled exposes a debug endpoint `POST /api/debug/trigger/:action` that allows triggering the following 3 actions: 'funding', 'maintenance' or 'scanning'

Behaviour:
```
curl -X POST localhost:9980/api/debug/trigger/scanning -> 401
curl -X POST -u :test  localhost:9980/api/debug/trigger/foo -> 400
curl -X POST -u :test  localhost:9980/api/debug/trigger/scanning -> 204 (OK)
```

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -28,6 +28,12 @@ type (
 	// to ensure the account is funded as soon as possible.
 	ContractManager interface {
 		TriggerAccountFunding()
+		TriggerMaintenance()
+	}
+
+	// HostManager defines an interface that allows trigger a host scan.
+	HostManager interface {
+		TriggerHostScanning()
 	}
 
 	// Explorer retrieves data about the Sia network from an external source.
@@ -77,8 +83,10 @@ type (
 type (
 	// An api provides an HTTP API for the indexer
 	api struct {
+		debug     bool
 		chain     ChainManager
 		contracts ContractManager
+		hosts     HostManager
 		explorer  Explorer
 		store     Store
 		syncer    Syncer
@@ -88,10 +96,11 @@ type (
 )
 
 // NewServer initializes the API
-func NewServer(chain ChainManager, contracts ContractManager, syncer Syncer, wallet Wallet, store Store, opts ...ServerOption) http.Handler {
+func NewServer(chain ChainManager, contracts ContractManager, hosts HostManager, syncer Syncer, wallet Wallet, store Store, opts ...ServerOption) http.Handler {
 	a := &api{
 		chain:     chain,
 		contracts: contracts,
+		hosts:     hosts,
 		store:     store,
 		syncer:    syncer,
 		wallet:    wallet,
@@ -101,7 +110,7 @@ func NewServer(chain ChainManager, contracts ContractManager, syncer Syncer, wal
 		opt(a)
 	}
 
-	return jape.Mux(map[string]jape.Handler{
+	routes := map[string]jape.Handler{
 		"GET /state": a.handleGETState,
 
 		// accounts endpoints
@@ -135,5 +144,12 @@ func NewServer(chain ChainManager, contracts ContractManager, syncer Syncer, wal
 		"GET /wallet/events":  a.handleGETWalletEvents,
 		"GET /wallet/pending": a.handleGETWalletPending,
 		"POST /wallet/send":   a.handlePOSTWalletSend,
-	})
+	}
+
+	// debug endpoints
+	if a.debug {
+		routes["GET /debug/trigger/:action"] = a.handleGETTrigger
+	}
+
+	return jape.Mux(routes)
 }
--- a/api/api.go
+++ b/api/api.go
@@ -148,6 +148,7 @@ func NewServer(chain ChainManager, contracts ContractManager, hosts HostManager,
 
 	// debug endpoints
 	if a.debug {
+		routes["GET /debug/pprof/:handler"] = a.handleGETPProf
 		routes["POST /debug/trigger/:action"] = a.handlePOSTTrigger
 	}
 
--- a/api/api.go
+++ b/api/api.go
@@ -31,7 +31,7 @@ type (
 		TriggerMaintenance()
 	}
 
-	// HostManager defines an interface that allows trigger a host scan.
+	// HostManager defines an interface that allows triggering a host scan.
 	HostManager interface {
 		TriggerHostScanning()
 	}
@@ -148,7 +148,7 @@ func NewServer(chain ChainManager, contracts ContractManager, hosts HostManager,
 
 	// debug endpoints
 	if a.debug {
-		routes["GET /debug/trigger/:action"] = a.handleGETTrigger
+		routes["POST /debug/trigger/:action"] = a.handleGETTrigger
 	}
 
 	return jape.Mux(routes)
--- a/api/api.go
+++ b/api/api.go
@@ -148,7 +148,7 @@ func NewServer(chain ChainManager, contracts ContractManager, hosts HostManager,
 
 	// debug endpoints
 	if a.debug {
-		routes["POST /debug/trigger"] = a.handlePOSTTrigger
+		routes["POST /debug/trigger/:action"] = a.handlePOSTTrigger
 	}
 
 	return jape.Mux(routes)
--- a/api/api.go
+++ b/api/api.go
@@ -148,7 +148,7 @@ func NewServer(chain ChainManager, contracts ContractManager, hosts HostManager,
 
 	// debug endpoints
 	if a.debug {
-		routes["POST /debug/trigger/:action"] = a.handleGETTrigger
+		routes["POST /debug/trigger"] = a.handlePOSTTrigger
 	}
 
 	return jape.Mux(routes)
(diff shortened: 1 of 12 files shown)


api/api.go
```diff
@@ -28,6 +28,12 @@ type (
 	// to ensure the account is funded as soon as possible.
 	ContractManager interface {
 		TriggerAccountFunding()
+		TriggerMaintenance()
+	}
+
+	// HostManager defines an interface that allows trigger a host scan.
```
ChrisSchinnerl:
```suggestion
	// HostManager defines an interface that allows triggering a host scan.
```

api/api.go
```diff
@@ -135,5 +144,12 @@ func NewServer(chain ChainManager, contracts ContractManager, syncer Syncer, wal
 		"GET /wallet/events":  a.handleGETWalletEvents,
 		"GET /wallet/pending": a.handleGETWalletPending,
 		"POST /wallet/send":   a.handlePOSTWalletSend,
-	})
+	}
+
+	// debug endpoints
+	if a.debug {
+		routes["GET /debug/trigger/:action"] = a.handleGETTrigger
```
ChrisSchinnerl: POST would be better

contracts/manager.go
```diff
@@ -260,6 +270,12 @@ func (cm *ContractManager) maintenanceLoop(ctx context.Context) {
 				log.Error("account funding failed", zap.Error(err))
 			}
 			continue
+		case <-cm.triggerMaintenanceChan:
+			// reset ticker
+			ticker.Stop()
```
ChrisSchinnerl, line 275: Do we no longer need to `select` on the channel to drain it after stopping it? There was an update related to tickers but I'm not sure.

hosts/manager.go
```diff
@@ -150,8 +157,14 @@ func NewManager(syncer Syncer, store Store, opts ...Option) (*HostManager, error
 			select {
 			case <-pruneTicker.C:
 				m.pruneHosts(ctx)
+			case <-m.triggerHostScanningChan:
+				// reset ticker
+				scanTicker.Stop()
```
ChrisSchinnerl, line 162: Same question regarding the channel draining. Although honestly it might also be fine to not reset the ticker at all? This isn't going to do a lot if no hosts require scanning.

hosts/manager.go
```diff
@@ -259,6 +280,35 @@ func (m *HostManager) UpdateChainState(tx UpdateTx, applied []chain.ApplyUpdate)
...
+		hosts, err := m.store.HostsForScanning(ctx)
+		if err != nil {
+			m.log.Error("failed to get hosts for scanning", zap.Error(err))
+			return nil
+		}
+		return hosts
+	}
+
+	// forcing a rescan of all hosts is only exposed with the debug flag
+	// enabled, therefor it's fine to pay the price here and fetch all hosts
```
ChrisSchinnerl:
```suggestion
	// enabled, therefore it's fine to pay the price here and fetch all hosts
```

api/api.go
```diff
@@ -135,5 +144,12 @@ func NewServer(chain ChainManager, contracts ContractManager, syncer Syncer, wal
...
+	}
+
```
ChrisSchinnerl, line 154: Might as well add the pprof routes too

ChrisSchinnerl (commented):
Might as well add `pp

contracts/manager.go
peterjan: That's no longer necessary I believe

```
The reset problem is fixed in Go 1.23. Quoting the docs again:

The timer channel associated with a Timer or Ticker is now unbuffered, with capacity 0. The main effect of this change is that Go now guarantees that for any call to a Reset or Stop method, no stale values prepared before that call will be sent or received after the call.
```

src https://go.dev/doc/go1.23?utm_source=chatgpt.com#timer-changes

hosts/manager.go
peterjan: Trigger a host scan will force a rescan for all hosts. So I think we do want to reset the ticker.

api/api.go
peterjan: `pprof` routes have already been added, see https://github.com/SiaFoundation/indexd/blob/master/cmd/indexd/web.go#L5 . I thought about that actually because I didn't like that we have `/api/debug` and `/debug/[pprof_routes]` but at the same time this is a debug endpoint for the indexd API and should be a handler on the server to easily access the managers. I considered dropping the `debug` from the route, but in the end decided against it... Perhaps we should reconsider?

api/api.go
peterjan: I considered that but it's an extra `-X POST` for no good reason other than the conventions but.. meh. I made it a `POST ` though, I personally also care about that a lot but I know Luke has a different opinion on that for instance.

api/api.go
ChrisSchinnerl: Ha I forgot about that. We are not going to register multiple sub routes under `/api` anymore right? So it might be fine to merge them all under `/api/debug`.

peterjan: I'm not a big fan of that. Not sure what Nate thinks about that but don't you feel the pprof routes should always be just `/debug/pprof`? I prefer the way it is now over merging them all under `/api/debug`.

edit: what if we merge them under `/debug`, I like that a lot better because pprof gets its familiar path and also we don't have to house a debug endpoint under `/api`, which usually means it's a static, well documented route. Doing it this way leaves our api space clean.

api/types.go
```diff
@@ -12,6 +12,12 @@ type (
 		NewAccountKey types.PublicKey `json:"newAccountKey"`
 	}
 
+	// DebugTriggerRequest is the request body for the [POST] /debug/trigger
+	// endpoint.
+	DebugTriggerRequest struct {
```
peterjan: figured this was a bit more flexible towards the future and also warrants it being a POST request

api/api.go
ChrisSchinnerl: I don't care as long as it's a single debug route.

api/types.go
ChrisSchinnerl: I'd do `/debug/trigger/:action` and encode it in the path. As soon as you need to type out json by hand on the CLI it gets pretty annoying. The `POST` is already warranted by the fact that it updates state.

ChrisSchinnerl (changes requested):

api/types.go
peterjan: Hm ok. Had that initially. I'll revert back.

cmd/indexd/web.go
```diff
@@ -13,6 +13,9 @@ type webRouter struct {
 
 func (wr webRouter) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	switch {
+	case strings.HasPrefix(req.URL.Path, "/api/debug/trigger"):
```
ChrisSchinnerl: This doesn't feel like a very clean solution. We need a seemingly empty case here just to prevent an issue that we get from reusing the treemux for both `/api` and `/debug` routes.
Which we do because otherwise we'd need to return another handler from the constructor which is also weird.
All of which are worse in my opinion than just housing the debug endpoints under `/api/debug`.

peterjan: I personally don't mind the empty case and prefer not housing the `debug` route under `/api`, but I reverted it and made it available under `/api/debug`.

api/api.go
```diff
@@ -135,5 +144,12 @@ func NewServer(chain ChainManager, contracts ContractManager, syncer Syncer, wal
 		"GET /wallet/events":  a.handleGETWalletEvents,
 		"GET /wallet/pending": a.handleGETWalletPending,
 		"POST /wallet/send":   a.handlePOSTWalletSend,
-	})
+	}
+
+	// debug endpoints
+	if a.debug {
+		routes["POST /debug/trigger/:action"] = a.handlePOSTTrigger
```
n8mgr, line 152: Should add `pprof` here too.

```go
func (a *api) handleGETPProf(jc jape.Context) {
	var handler string
	if err := jc.DecodeParam("handler", &handler); err != nil {
		return
	}

	switch handler {
	case "cmdline":
		pprof.Cmdline(jc.ResponseWriter, jc.Request)
	case "profile":
		pprof.Profile(jc.ResponseWriter, jc.Request)
	case "symbol":
		pprof.Symbol(jc.ResponseWriter, jc.Request)
	case "trace":
		pprof.Trace(jc.ResponseWriter, jc.Request)
	default:
		pprof.Index(jc.ResponseWriter, jc.Request)
	}
}

routes["GET /debug/pprof/:handler"] = a.handleGETPProf
```

peterjan: Ok. I thought it was bad practice to house pprof routes under anything other than `/debug/pprof` but if you both feel that is fine I'm also fine. Tested the following `go tool pprof http://:[password]@localhost:9880/api/debug/pprof/profile\?seconds\=30` and it works.

n8mgr (approved):

ChrisSchinnerl (approved):

PR #202, opened 2025-06-18
Reset chain state if consensus db is not found

peterjan:
This allows forcing a resync by removing consensus.db, similar to how we do it for other daemons.

### DIFF
--- a/cmd/indexd/run.go
+++ b/cmd/indexd/run.go
@@ -40,15 +40,7 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
 	}
 	defer store.Close()
 
-	// ensure deleting the database forces a resync
-	chainPath := filepath.Join(cfg.Directory, "consensus.db")
-	if _, err := os.Stat(chainPath); os.IsNotExist(err) {
-		if err := store.ResetChainState(context.Background()); err != nil {
-			return fmt.Errorf("failed to reset chain state: %w", err)
-		}
-	}
-
-	bdb, err := coreutils.OpenBoltChainDB(chainPath)
+	bdb, err := coreutils.OpenBoltChainDB(filepath.Join(cfg.Directory, "consensus.db"))
 	if err != nil {
 		return fmt.Errorf("failed to open consensus database: %w", err)
 	}
--- a/cmd/indexd/run.go
+++ b/cmd/indexd/run.go
@@ -40,7 +40,15 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
 	}
 	defer store.Close()
 
-	bdb, err := coreutils.OpenBoltChainDB(filepath.Join(cfg.Directory, "consensus.db"))
+	// ensure deleting the database forces a resync
+	chainPath := filepath.Join(cfg.Directory, "consensus.db")
+	if _, err := os.Stat(chainPath); os.IsNotExist(err) {
+		if err := store.ResetChainState(context.Background()); err != nil {
+			return fmt.Errorf("failed to reset chain state: %w", err)
+		}
+	}
+
+	bdb, err := coreutils.OpenBoltChainDB(chainPath)
 	if err != nil {
 		return fmt.Errorf("failed to open consensus database: %w", err)
 	}
--- a/persist/postgres/chain.go
+++ b/persist/postgres/chain.go
@@ -17,11 +17,11 @@ type updateTx struct {
 // global settings. This is typically used to force a resync of consensus.
 func (s *Store) ResetChainState(ctx context.Context) error {
 	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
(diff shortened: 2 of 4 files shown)


cmd/indexd/run.go
```diff
@@ -40,7 +40,15 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
 	}
 	defer store.Close()
 
-	bdb, err := coreutils.OpenBoltChainDB(filepath.Join(cfg.Directory, "consensus.db"))
+	// ensure deleting the database forces a resync
+	chainPath := filepath.Join(cfg.Directory, "consensus.db")
+	if _, err := os.Stat(chainPath); os.IsNotExist(err) {
+		if err := store.ResetChainState(context.Background()); err != nil {
+			return fmt.Errorf("failed to reset chain state: %w", err)
```
ChrisSchinnerl: Problem with triggering it like this is that it won't cover the common use-case of people replacing the database with a bootstrap one. So this won't trigger. I'd move this to where we call `UpdatesSince` and check for `ErrMissingBlock` and then call `ResetChainState`. Similar to how `hostd` and `walletd` do it.

We might also want to upgrade `renterd` to do it like that since Nate had issues resetting a renter the other day.

persist/postgres/chain.go
```diff
@@ -11,6 +12,27 @@ type updateTx struct {
 	tx  *txn
 }
 
+// ResetChainState resets the chain state in the store, clearing all
+// wallet-related data and resetting the scanned height and block ID in the
+// global settings. This is typically used to force a resync of consensus.
+func (s *Store) ResetChainState(ctx context.Context) error {
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
```
ChrisSchinnerl, line 19: I'd consider using [Truncate](https://www.postgresql.org/docs/current/sql-truncate.html) as a faster alternative to `DELETE` for when we want to delete all contents of a table.

ChrisSchinnerl (changes requested):

subscriber/subscriber.go
```diff
@@ -163,7 +165,13 @@ func (s *Subscriber) Sync(ctx context.Context) error {
 		}
 
 		rus, aus, err := s.cm.UpdatesSince(index, s.updateBatchSize)
-		if err != nil {
+		if err != nil && strings.Contains(err.Error(), "missing block at index") {
```
ChrisSchinnerl: The `chain` package in `coreutils` exposes that error now as `ErrMissingBlock` so you can use `errors.Is`

subscriber/subscriber.go
```diff
@@ -163,7 +165,13 @@ func (s *Subscriber) Sync(ctx context.Context) error {
...
+			s.log.Warn("missing block at index, resetting chain state")
+			if err := s.store.ResetChainState(ctx); err != nil {
+				return fmt.Errorf("failed to reset consensus state: %w", err)
+			}
+			return nil
```
ChrisSchinnerl: Should we `index, err = s.store.LastScannedIndex(ctx)` after the reset here and then `continue` instead? That way we don't need to wait for `reorgChan` to signal again before the actual syncing begins.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (dismissed):

subscriber/subscriber.go
```diff
@@ -163,7 +164,15 @@ func (s *Subscriber) Sync(ctx context.Context) error {
 		}
 
 		rus, aus, err := s.cm.UpdatesSince(index, s.updateBatchSize)
-		if err != nil {
+		if errors.Is(err, chain.ErrMissingBlock) {
```
n8mgr: We ran into this in `hostd` and `renterd`. You should attempt to reset on any error here. There are a few different errors we're not catching that cause the indexer to lock up.

subscriber/subscriber.go
```diff
@@ -164,7 +165,13 @@ func (s *Subscriber) Sync(ctx context.Context) error {
 
 		rus, aus, err := s.cm.UpdatesSince(index, s.updateBatchSize)
 		if err != nil {
-			return fmt.Errorf("failed to fetch updates since %v: %w", index, err)
+			s.log.Warn("failed to fetch updates, resetting chain state", zap.Uint64("height", index.Height), zap.Stringer("id", index.ID), zap.Error(err))
+			if err := s.store.ResetChainState(ctx); err != nil {
+				return fmt.Errorf("failed to reset consensus state: %w", err)
+			} else if index, err = s.store.LastScannedIndex(ctx); err != nil {
+				return fmt.Errorf("failed to get last scanned index after reset: %w", err)
+			}
```
ChrisSchinnerl, line 178: We should also update this and `renterd` with the max retry protection that `hostd` has. Otherwise we might be stuck in a loop here.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

n8mgr (approved):

PR #208, opened 2025-06-24
Add contracts API

peterjan:
The diff still includes https://github.com/SiaFoundation/indexd/pull/223 for some reason. Ideally we merge that first and F/U with this one.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -41,6 +41,8 @@ type (
 		AddAccount(ctx context.Context, ak types.PublicKey) error
 		DeleteAccount(ctx context.Context, ak types.PublicKey) error
 		UpdateAccount(ctx context.Context, oldAK, newAK types.PublicKey) error
+		Contract(ctx context.Context, contractID types.FileContractID) (contracts.Contract, error)
+		Contracts(ctx context.Context, offset, limit int, queryOpts ...contracts.ContractQueryOpt) ([]contracts.Contract, error)
 		BlockHosts(ctx context.Context, hks []types.PublicKey, reason string) error
 		BlockedHosts(ctx context.Context, offset, limit int) ([]types.PublicKey, error)
 		Host(ctx context.Context, hk types.PublicKey) (hosts.Host, error)
@@ -109,6 +111,12 @@ func NewServer(chain ChainManager, contracts ContractManager, syncer Syncer, wal
 		"POST   /account/:accountkey": a.handlePOSTAccount,
 		"PUT    /account/:accountkey": a.handlePUTAccount,
 		"DELETE /account/:accountkey": a.handleDELETEAccount,
+
+		// contract endpoints
+		"GET    /contract/:contractid": a.handleGETContract,
+
+		// contracts endpoints
+		"GET    /contracts": a.handleGETContracts,
 
 		// explorer endpoints
 		"GET /explorer/exchange-rate/siacoin/:currency": a.handleGETExplorerSiacoinExchangeRate,
--- a/api/api.go
+++ b/api/api.go
@@ -28,6 +28,12 @@ type (
 	// to ensure the account is funded as soon as possible.
 	ContractManager interface {
 		TriggerAccountFunding()
+		TriggerMaintenance()
+	}
+
+	// HostManager defines an interface that allows triggering a host scan.
+	HostManager interface {
+		TriggerHostScanning()
 	}
 
 	// Explorer retrieves data about the Sia network from an external source.
@@ -79,8 +85,10 @@ type (
 type (
 	// An api provides an HTTP API for the indexer
 	api struct {
+		debug     bool
 		chain     ChainManager
 		contracts ContractManager
+		hosts     HostManager
 		explorer  Explorer
 		store     Store
 		syncer    Syncer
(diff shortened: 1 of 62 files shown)


api/api.go
```diff
@@ -110,6 +112,12 @@ func NewServer(chain ChainManager, contracts ContractManager, syncer Syncer, wal
 		"PUT    /account/:accountkey": a.handlePUTAccount,
 		"DELETE /account/:accountkey": a.handleDELETEAccount,
 
+		// contract endpoints
+		"GET    /contract/:contractid": a.handleGETContract,
+
+		// contracts endpoints
+		"GET    /contracts": a.handleGETContracts,
```
ChrisSchinnerl:
```suggestion
		"GET /contracts": a.handleGETContracts,
```

api/api.go
```diff
@@ -110,6 +112,12 @@ func NewServer(chain ChainManager, contracts ContractManager, syncer Syncer, wal
 		"PUT    /account/:accountkey": a.handlePUTAccount,
 		"DELETE /account/:accountkey": a.handleDELETEAccount,
 
+		// contract endpoints
+		"GET    /contract/:contractid": a.handleGETContract,
```
ChrisSchinnerl:
```suggestion
		"GET /contract/:contractid": a.handleGETContract,
```

api/options.go
```diff
@@ -42,6 +42,24 @@ func WithLimit(limit int) URLQueryParameterOption {
...
+type ContractQueryParameterOption URLQueryParameterOption
+
+// WithRevisable sets the 'revisable' parameter.
+func WithRevisable(revisable bool) ContractQueryParameterOption {
+	return func(q url.Values) {
+		q.Set("revisable", fmt.Sprint(revisable))
+	}
+}
+
+// WithGood sets the 'good' parameter.
```
ChrisSchinnerl: These docstrings are not super useful from a DX because you'd need to search the codebase for what those parameters actually mean. Better to explain it here as well.

api/endpoints.go
```diff
@@ -125,6 +125,52 @@ func (a *api) handleDELETEAccount(jc jape.Context) {
...
+
+func (a *api) handleGETContracts(jc jape.Context) {
+	offset, limit, ok := parseOffsetLimit(jc)
+	if !ok {
+		return
+	}
+
+	var opts []contracts.ContractQueryOpt
+	if jc.Request.FormValue("revisable") != "" {
+		var revisable bool
```
ChrisSchinnerl, line 195: Not sure what @n8maninger 's thoughts are but I'm leaning towards defaulting to revisable contracts in the API. Seems like the response you'd want most often.

ChrisSchinnerl (changes requested):
I think it makes sense to pause this, merge the fixes needed for contract formation in the meantime and then add an integration test here.

If `recent_uptime` is the only issue I'd default it to 95% for now when we insert a host. We can then try and be fancy again later.

api/endpoints.go
peterjan: Yeah I agree, I thought about that but as soon as you default these options it becomes slightly awkward. I guess instead of passing `revisable=true|false` you almost want to pass an `all` parameter or something.

peterjan:
> I think it makes sense to pause this, merge the fixes needed for contract formation in the meantime and then add an integration test here.

Ok, I already had most of that so I pushed it and will prepare one or two tiny PRs that get us to contracts being formed.

api/endpoints.go
n8mgr: Yeah, I'd agree to default to the contracts that are actually in use.

persist/postgres/contracts.go
```diff
@@ -145,7 +145,7 @@ WHERE
 	-- active filter
 	(
 		$2::boolean IS NULL OR 
-		($2::boolean = TRUE AND c.state <= 1) OR 
+		($2::boolean = TRUE AND c.state <= 1 AND c.renewed_to IS NULL) OR
```
ChrisSchinnerl, line 148: Did you check if that slows down any benchmarks? We might not have an index on renewed_to. We might want a composite index now over state and renewed_to.

This also slightly changes functionality and might warrant extending the appropriate test.

contracts/renew.go
```diff
@@ -67,6 +67,7 @@ func (cm *ContractManager) renewContract(ctx context.Context, contract Contract,
 		if err := cm.store.AddRenewedContract(ctx, contract.ID, renewed.ID, renewed.Revision, host.Settings.Prices.ContractPrice, minerFee, types.ZeroCurrency); err != nil {
 			return fmt.Errorf("failed to store renewed contract: %w", err)
 		}
+		contractLog.Debug("renewed")
```
ChrisSchinnerl:
```suggestion
		contractLog.Debug("successfully renewed contract")
```

persist/postgres/contracts.go
peterjan: I did actually but there's no noticeable difference, `renewed_to` is `UNIQUE` so it's backed by an index. I didn't consider the composite index though, I'll test that. Edit: I added a composite index but the benchmarks did not improve. I added an extra assertion in `TestContracts`

api/api_test.go
```diff
@@ -404,3 +526,21 @@ func TestWalletAPI(t *testing.T) {
...
+	if !enable {
+		return zap.NewNop()
+	}
+	config := zap.NewProductionEncoderConfig()
+	config.EncodeTime = zapcore.RFC3339TimeEncoder
+	config.EncodeLevel = zapcore.CapitalColorLevelEncoder
+	config.StacktraceKey = ""
+	consoleEncoder := zapcore.NewConsoleEncoder(config)
+
+	return zap.New(
```
ChrisSchinnerl, line 541: I'd try `zaptest.NewLogger` since that logs using `testing.T` rather than `stdout`.

ChrisSchinnerl (approved):

api/api_test.go
peterjan: I'll follow up with this one since I got Chris's approval.

n8mgr (approved):

PR #209, opened 2025-06-26
Fix host usability

### LINKED ISSUE #189: FTUE

peterjan:
This PR essentially sets us up to start forming contracts with hosts as soon as we enable maintenance. Currently, a freshly announced host fails three checks: uptime, gouging and price table validity. I added default settings that should match what we want to use in production. I also updated the uptime requirement and set price table validity to 24 hours, those are not ideal but probably good enough for now, we can always revisit these later.

Closes #189

### DIFF
--- a/internal/testutils/host.go
+++ b/internal/testutils/host.go
@@ -17,6 +17,11 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	blocksPerMonth = 144 * 30
+	oneTB          = 1e12
+)
+
 type (
 	// A Host is an ephemeral host that can be used for testing.
 	Host struct {
@@ -108,16 +113,16 @@ func (c *ConsensusNode) NewHost(t testing.TB, pk types.PrivateKey, log *zap.Logg
 		Release:             "test",
 		AcceptingContracts:  true,
 		WalletAddress:       w.Address(),
-		MaxCollateral:       types.Siacoins(10000),
-		MaxContractDuration: 1000,
+		MaxCollateral:       types.Siacoins(1000), // 1 KS
+		MaxContractDuration: blocksPerMonth * 3,   // 3 months
 		RemainingStorage:    100 * proto4.SectorSize,
 		TotalStorage:        100 * proto4.SectorSize,
 		Prices: proto4.HostPrices{
-			ContractPrice: types.Siacoins(1).Div64(5), // 0.2 SC
-			StoragePrice:  types.NewCurrency64(100),   // 100 H / byte / block
-			IngressPrice:  types.NewCurrency64(100),   // 100 H / byte
-			EgressPrice:   types.NewCurrency64(100),   // 100 H / byte
-			Collateral:    types.NewCurrency64(200),
+			ContractPrice: types.Siacoins(1).Div64(5),                                      // 0.2 SC
+			StoragePrice:  types.Siacoins(150).Div64(oneTB).Div64(blocksPerMonth),          // 150 SC / TB / month
+			EgressPrice:   types.Siacoins(500).Div64(oneTB),                                // 500 SC / TB
+			IngressPrice:  types.Siacoins(10).Div64(oneTB),                                 // 10 SC / TB
+			Collateral:    types.Siacoins(150).Div64(oneTB).Div64(blocksPerMonth).Mul64(2), // 2x storage
 		},
 	})
 	ss := testutil.NewEphemeralSectorStore()
@@ -143,7 +148,7 @@ func (c *ConsensusNode) NewHost(t testing.TB, pk types.PrivateKey, log *zap.Logg
 	}
 	c.addSyncFn(syncFn)
 
-	rs := rhp4.NewServer(pk, c.cm, s, contractor, w, sr, ss, rhp4.WithPriceTableValidity(2*time.Minute))
+	rs := rhp4.NewServer(pk, c.cm, s, contractor, w, sr, ss, rhp4.WithPriceTableValidity(24*time.Hour))
 	rhp4Listener, err := net.Listen("tcp", ":0")
 	if err != nil {
 		t.Fatal(err)
--- a/internal/testutils/host.go
+++ b/internal/testutils/host.go
@@ -148,7 +148,7 @@ func (c *ConsensusNode) NewHost(t testing.TB, pk types.PrivateKey, log *zap.Logg
 	}
 	c.addSyncFn(syncFn)
 
-	rs := rhp4.NewServer(pk, c.cm, s, contractor, w, sr, ss, rhp4.WithPriceTableValidity(24*time.Hour))
+	rs := rhp4.NewServer(pk, c.cm, s, contractor, w, sr, ss, rhp4.WithPriceTableValidity(30*time.Minute))
 	rhp4Listener, err := net.Listen("tcp", ":0")
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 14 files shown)


internal/testutils/host.go
```diff
@@ -143,7 +148,7 @@ func (c *ConsensusNode) NewHost(t testing.TB, pk types.PrivateKey, log *zap.Logg
 	}
 	c.addSyncFn(syncFn)
 
-	rs := rhp4.NewServer(pk, c.cm, s, contractor, w, sr, ss, rhp4.WithPriceTableValidity(2*time.Minute))
+	rs := rhp4.NewServer(pk, c.cm, s, contractor, w, sr, ss, rhp4.WithPriceTableValidity(24*time.Hour))
```
ChrisSchinnerl: I'd make this match the `hostd` default in production

persist/postgres/hosts.go
```diff
@@ -85,7 +85,7 @@ WITH globals AS (
 	WHERE hosts.public_key = $1
 ) SELECT
 	hosts.*,
-	recent_uptime > 0.9,
+	recent_uptime >= 0.894,
```
ChrisSchinnerl: Why change from 90% uptime to 89.4%? That's an oddly specific number. Rather than lower the uptime requirement I'd increase the `recent_uptime` default for hosts.

persist/postgres/init.sql
```diff
@@ -136,10 +136,10 @@ CREATE TABLE global_settings (
...
     -- host manager settings
     hosts_min_protocol_version BYTEA NOT NULL DEFAULT '\x010000', -- used for host checks
-    hosts_min_collateral NUMERIC(50,0) NOT NULL DEFAULT 0, -- hastings / byte / block
-    hosts_max_storage_price NUMERIC(50,0) NOT NULL DEFAULT 0, -- hastings / byte / block
-    hosts_max_ingress_price NUMERIC(50,0) NOT NULL DEFAULT 0, -- hastings / byte
-    hosts_max_egress_price NUMERIC(50,0) NOT NULL DEFAULT 0, -- hastings / byte
+    hosts_min_collateral NUMERIC(50,0) NOT NULL DEFAULT ((100 * POWER(10,24)) / POWER(10,12) / 4320), -- 100 SC / TB month
+    hosts_max_storage_price NUMERIC(50,0) NOT NULL DEFAULT ((3000 * POWER(10,24)) / POWER(10,12) / 4320),-- 3000 SC / TB / month
+    hosts_max_ingress_price NUMERIC(50,0) NOT NULL DEFAULT ((3000 * POWER(10,24)) / POWER(10,12)),-- 3000 SC / TB
+    hosts_max_egress_price NUMERIC(50,0) NOT NULL DEFAULT ((3000 * POWER(10,24)) / POWER(10,12)), -- 3000 SC / TB
```
ChrisSchinnerl: Might be better to remove the default value from these columns and explicitly set the values in `initSettings`. The value of SC can change drastically and we don't want to write a database migration every time we change the defaults.

ChrisSchinnerl (changes requested):

persist/postgres/hosts.go
peterjan: I did it to keep the assertions in `TestHostsRecentUptime` but added a note instead.

persist/postgres/store.go
```diff
@@ -73,15 +75,33 @@ func Connect(ctx context.Context, ci ConnectionInfo, log *zap.Logger) (*Store, e
 		return nil, fmt.Errorf("failed to connect to database: %w", err)
 	}
 	log.Info("connected", zap.String("database", ci.Database), zap.String("host", ci.Host), zap.Int("port", ci.Port))
+	return pool, nil
+}
 
-	store := &Store{
+// NewStore creates a new Store instance, initializing the database if necessary.
+func NewStore(ctx context.Context, pool *pgxpool.Pool, defaultMaintenanceSettings contracts.MaintenanceSettings, defaultUsabilitySettings hosts.UsabilitySettings, log *zap.Logger) (*Store, error) {
```
peterjan: this sets us up to override the defaults with specific values for testing

persist/postgres/settings.go
```diff
@@ -79,3 +71,15 @@ func (u *updateTx) UpdateLastScannedIndex(ctx context.Context, ci types.ChainInd
...
+func setMaintenanceSettings(ctx context.Context, tx *txn, settings contracts.MaintenanceSettings) error {
+	_, err := tx.Exec(ctx, `UPDATE global_settings SET contracts_maintenance_enabled = $1, contracts_wanted = $2, contracts_renew_window = $3, contracts_period = $4`,
+		settings.Enabled, settings.WantedContracts, settings.RenewWindow, settings.Period)
+	return err
+}
+
+func setUsabilitySettings(ctx context.Context, tx *txn, settings hosts.UsabilitySettings) error {
+	query := `UPDATE global_settings SET hosts_max_egress_price = $1, hosts_max_ingress_price = $2, hosts_max_storage_price = $3, hosts_min_collateral = $4, hosts_min_protocol_version = $5`
+	_, err := tx.Exec(ctx, query, sqlCurrency(settings.MaxEgressPrice), sqlCurrency(settings.MaxIngressPrice), sqlCurrency(settings.MaxStoragePrice), sqlCurrency(settings.MinCollateral), sqlProtocolVersion(settings.MinProtocolVersion))
+	return err
```
ChrisSchinnerl, line 84: I'd combine these two helper with `initSettings` and remove the `DEFAULT 0` for those columns in `init.sql` now. No point in defaulting to a value that is not the actual default we want to set.

persist/postgres/hosts_test.go
```diff
@@ -614,7 +610,16 @@ func TestHostsRecentUptime(t *testing.T) {
 	// add a host
 	hk = db.addTestHost(t)
 
-	// assert default uptime
+	// manually override the default uptime to .894, this very specific value
+	// gives us the uptime properties that are laid out in the spec, for the
+	// time being though the recent uptime defaults to 0.9 to ensure host pass
```
ChrisSchinnerl: You heard it here first but I think "for the time being" is going to end up unexpectedly permanent.

ChrisSchinnerl (changes requested):

persist/postgres/settings.go
peterjan: That's what I had initially, I split that up because of the `CHECK`s we do in our maintenance settings. I wanted to either have defaults on both, or on neither. I moved it though, removed the defaults on our maintenance settings but removed them for our usability settings.

persist/postgres/hosts_test.go
peterjan: I rewrote the comment. Were you suggesting to alter the test? If we're going in to this thinking it's a permanent change we should probably also update the spec and come up with a new uptime specification, which I'd probably want to F/U with.

n8mgr (approved):

ChrisSchinnerl (approved):

persist/postgres/store.go
ChrisSchinnerl: Do we need that? For testing the defaults we could just use the standard defaults and if we want different settings we can update them after creating the store?

PR #212, opened 2025-06-26
Add application API

peterjan:


internal/testutils/indexer.go
```diff
@@ -122,7 +154,9 @@ func NewIndexer(t testing.TB, c *ConsensusNode, log *zap.Logger) *Indexer {
 		}
 	})
 	return &Indexer{
-		Client: api.NewClient(fmt.Sprintf("http://%s", httpListener.Addr().String()), password),
+		Client: api.NewClient(fmt.Sprintf("http://%s", adminListener.Addr().String()), password),
+
+		applicationAPIAddress: fmt.Sprintf("http://%s", applicationListener.Addr().String()),
```
peterjan: this is temporary, when we add our first actual route I plan to create a separate client to mirror what we do with the admin API, and get rid of the `/foo` route

api/options.go
```diff
@@ -7,29 +7,47 @@ import (
...
+		applyOption(Option)
+	}
 
-// WithDebug sets the debug mode for the API server. In debug mode, the server
-// exposes additional debug endpoints that allow triggering certain actions.
-func WithDebug() ServerOption {
-	return func(a *api) {
-		a.debug = true
+	// Option is an interface for options that can be applied to the API.
+	Option interface {
```
n8mgr: Should separate these into `AppOption` and `AdminOption`

api/application.go
```diff
@@ -0,0 +1,43 @@
...
+// key that corresponds to a previously registered public key.
+func NewApplicationAPI(hostname string, store AccountStore, opts ...Option) http.Handler {
+	a := &applicationAPI{
+		log: zap.NewNop(),
+	}
+	for _, opt := range opts {
+		a.applyOption(opt)
+	}
+
+	return wrapSignedAPI(hostname, store, map[string]authedHandler{
```
n8mgr: nit: probably worth inlining `wrapSignedAPI`

api/options.go
peterjan: `WithLogger` triggered me, I had to make it `WithAppLogger` and `WithAdminLogger`

n8mgr (approved):

n8mgr:
Wondering how you would feel about splitting this into two packages: `api/admin` and `api/app`. They don't need to interact and we get to drop the `App` and `Admin` prefix from everything. `admin.NewClient` vs `api.NewAdminClient`

peterjan:
> Wondering how you would feel about splitting this into two packages: `api/admin` and `api/app`. They don't need to interact and we get to drop the `App` and `Admin` prefix from everything. `admin.NewClient` vs `api.NewAdminClient`

I considered that actually, I don't have anything against it but in general I try and avoid subpackages as much as possible. I still have `proto` nightmares from `siad. Here it probably works though you are right. I'll F/U with that immediately after this. 

edit: it's actually a good idea because the clients make this problem even worse

PR #213, opened 2025-06-27
Update port when trying to add peer with same address.

peterjan:


persist/postgres/peers.go
```diff
@@ -20,7 +20,7 @@ func (s *Store) AddPeer(addr string) error {
 		return fmt.Errorf("invalid peer address: %w", err)
 	}
 	return s.transaction(context.Background(), func(ctx context.Context, tx *txn) error {
-		const query = `INSERT INTO syncer_peers (ip_address, port) VALUES ($1, $2) ON CONFLICT (ip_address) DO NOTHING`
+		const query = `INSERT INTO syncer_peers (ip_address, port) VALUES ($1, $2) ON CONFLICT (ip_address) DO UPDATE SET port = EXCLUDED.port`
```
ChrisSchinnerl: wouldn't it be better to change the primary key to be a composite key of ip_address and port? Because it's not illegal to have multiple syncers running on different ports.

peterjan: Hm I kind of remember there being a reason for it but can't recall. I don't see a reason now though so I updated it 👍 . Thanks for catching that.

n8mgr: I think we should keep the primary on ip_address. Don't want to waste a slot syncing with what is effectively the same peer.

ChrisSchinnerl: Won't that cause issues for anyone (like some of our own servers) running multiple instances of renterd for whatever reason? Or in case you rely on more exotic setups such as tunneling via some service to bypass your a CGNAT?

n8mgr: True 🤔

peterjan: Seeing as this is a bit of an edge case, is it that bad to waste a slot?

ChrisSchinnerl (approved):

n8mgr (approved):

PR #214, opened 2025-06-27
Download only required data when recovering slabs

### LINKED ISSUE #206: SDK: Only download parts of shards we need rather than all of them and then slicing the data using the input `length`

n8mgr:
Fixes #206

### DIFF
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -177,6 +177,8 @@ func (s *SDK) downloadSlab(ctx context.Context, slab Slab, maxInflight int, time
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	offset, length := sectorRegion(slab.MinShards, slab.Offset, slab.Length)
+
 	var successful atomic.Uint32
 	var wg sync.WaitGroup
 	shards := make([][]byte, len(slab.Shards))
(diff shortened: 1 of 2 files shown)


sdk/client.go
```diff
@@ -404,11 +408,21 @@ func stripedJoin(dst io.Writer, dataShards [][]byte, writeLen int) error {
 	return nil
 }
 
+func sectorRegion(minShards uint8, so, sl uint32) (offset, length uint64) {
```
ChrisSchinnerl, line 411: (nit) considering that both `sector` as well as `slab` start with an `s`, `so` and `sl` aren't extra descriptive. Maybe call it `slabOffset` and `sectorOffset` respectively? Sure it's a bit longer but also very clear.

sdk/client.go
```diff
@@ -404,11 +408,21 @@ func stripedJoin(dst io.Writer, dataShards [][]byte, writeLen int) error {
...
+	minChunkSize := proto4.LeafSize * uint32(minShards)
+	start := (so / minChunkSize) * proto4.LeafSize
+	end := ((so + sl) / minChunkSize) * proto4.LeafSize
+	if (so+sl)%minChunkSize != 0 {
+		end += proto4.LeafSize
+	}
+	return uint64(start), uint64(end - start)
+}
```
ChrisSchinnerl, line 419: Looks good but might be nice to have a unit test for this logic regardless.

sdk/client_test.go
```diff
@@ -132,23 +132,39 @@ func TestRoundtrip(t *testing.T) {
...
 	}
 
-	if !bytes.Equal(buf.Bytes(), data) {
-		t.Fatal("data mismatch")
+	sizes := []int{
+		100,
+		1024,  // 1 KiB
+		4096,  // 4 KiB
+		1e6,   // 1 MB
+		100e6, // 100 MB
```
ChrisSchinnerl, line 164: This currently only tests downloads of various lengths but not offsets. The offset is always 0. 
Might be good to have a separate test with a single slab from which we download arbitrary ranges.

ChrisSchinnerl (changes requested):

peterjan (approved):
I have nothing to add to Chris's review.

PR #215, opened 2025-06-28
Upload refactor

n8mgr:
Small cleanup of the upload code. Doesn't affect benchmarks, but makes it slightly easier to follow. Requires #214

### DIFF
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -1,7 +1,6 @@
 package sdk
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -29,6 +28,13 @@ type (
 	downloadOption struct {
 		hostTimeout time.Duration
 		maxInflight int
+	}
+
+	slabUploadWork struct {
+		shards        [][]byte
+		length        uint32
+		encryptionKey [32]byte
+		err           error
 	}
 
 	// A HostDialer is an interface for writing and reading sectors to/from hosts.
@@ -220,6 +226,9 @@ top:
 //
 // Returns the metadata of the slabs that were pinned
 func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) ([]Slab, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
 	uo := uploadOption{
 		dataShards:   10,
 		parityShards: 20,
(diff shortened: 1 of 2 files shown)


sdk/client.go
```diff
@@ -445,6 +429,44 @@ func uploadShard(ctx context.Context, sector *[proto4.SectorSize]byte, hostKey t
...
+			ch <- slabUploadWork{err: io.EOF} // signal done with EOF
+			return
+		}
+
+		if err := enc.Encode(shards); err != nil {
+			ch <- slabUploadWork{err: fmt.Errorf("failed to encode slab %d shards: %w", i, err)}
+			return
+		}
+		encryptionKey := h.Sum()
+		ch <- slabUploadWork{length: n, encryptionKey: encryptionKey, shards: shards}
```
peterjan, line 466: instead of these naked sends (error cases), we might as well take the context into account and use a select?

sdk/client.go
```diff
@@ -445,6 +429,44 @@ func uploadShard(ctx context.Context, sector *[proto4.SectorSize]byte, hostKey t
...
+		for i := range shards {
+			shards[i] = make([]byte, proto4.SectorSize)
+		}
+		tr := io.TeeReader(r, h.E)
+		n, err := stripedSplit(tr, shards[:uo.dataShards])
+		if err != nil {
+			ch <- slabUploadWork{err: fmt.Errorf("failed to read slab %d: %w", i, err)}
+			return
+		} else if n == 0 {
+			ch <- slabUploadWork{err: io.EOF} // signal done with EOF
```
peterjan, line 457: slightly more idiomatic to signal it's done by closing the channel, fine to keep as-is though

sdk/client.go
```diff
@@ -408,6 +372,26 @@ func stripedJoin(dst io.Writer, dataShards [][]byte, writeLen int) error {
...
+// `proto4.SectorSize` bytes. The function returns the total number of bytes
+// read, which is the sum of the lengths of all shards.
+func stripedSplit(r io.Reader, shards [][]byte) (length uint32, err error) {
+	for i := 0; i < proto4.SectorSize; i += proto4.LeafSize {
+		for j := range shards {
+			n, err := io.ReadFull(r, shards[j][i:i+proto4.LeafSize])
+			if errors.Is(err, io.EOF) {
+				return length, nil // no more data to read
+			} else if err != nil && !errors.Is(err, io.ErrUnexpectedEOF) {
+				return 0, fmt.Errorf("failed to read shard %d: %w", i, err)
```
peterjan, line 387: you are passing the offset as shard index, should be `j`

sdk/client.go
```diff
@@ -445,6 +429,44 @@ func uploadShard(ctx context.Context, sector *[proto4.SectorSize]byte, hostKey t
...
 
+// processSlabs reads slabs from the provided reader, encodes them using the
+// provided Reed-Solomon encoder, and sends the encoded shards to the provided
+// channel. Slabs are processed in a loop until EOF is reached.
+
+// The encryption key is derived for each slab by hashing the app key slab's raw data.
+func processSlabs(ctx context.Context, enc reedsolomon.Encoder, uo uploadOption, r io.Reader, appKey types.PrivateKey, ch chan<- slabUploadWork) {
+	h := types.NewHasher()
+	for i := 0; ; i++ {
+		select {
```
peterjan, line 440: `nit` slightly prefer `if ctx.Err() != nil { return }`

sdk/client.go
```diff
@@ -234,55 +243,21 @@ func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) ([]
...
-	}()
+
+	slabCh := make(chan slabUploadWork, 1)
+	go processSlabs(ctx, enc, uo, r, s.appKey, slabCh)
 
 	// TODO: cleanup on failure
 	var pinned []Slab
 	for i := 0; ; i++ {
 		select {
 		case <-ctx.Done():
```
peterjan, line 258: I'm not sure how to properly do the error handling, but it is kind of sad to return `nil` here if the context gets cancelled. We might've uploaded a bunch of data already. It's not great but perhaps we should consider return an upload result:

```go
type UploadResult struct {
    Slabs []Slab
    Err   error
}
if ctx.Err() != nil {
    return &UploadResult{Slabs: slabsSoFar, Err: ctx.Err()}, nil
}
```

Or just return the pinned slabs alongside the `ctx.Err()`... it's a little unexpected perhaps for the caller but if properly mentioned in the docstring I think it's ok. If the caller can check `err` against `context.Cancelled`. If we use the `UploadResult` we could also return the total number of bytes read.

ChrisSchinnerl: Are we ever going to care about the errors if the upload was cancelled on purpose? Seems like something that we would just end up ignoring to keep the logs clean anyway.

sdk/client.go
ChrisSchinnerl: I definitely prefer `select` for anything related to goroutine synchronization. It's a lot more obvious imo.

sdk/client.go
ChrisSchinnerl: probably have to since otherwise the calling loop will exit when the context is closed and this will never unblock

sdk/client.go
```diff
@@ -234,55 +243,21 @@ func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) ([]
...
-			if err := enc.Encode(shards); err != nil {
-				workCh <- work{err: fmt.Errorf("failed to encode slab %d shards: %w", i, err)}
-				return
-			}
-			encryptionKey := types.HashBytes(append(s.appKey[:], slabBuf[:n]...))
-			workCh <- work{length: n, encryptionKey: encryptionKey, shards: shards}
-		}
-	}()
+
+	slabCh := make(chan slabUploadWork, 1)
```
ChrisSchinnerl, line 251: Maybe leave a todo here for a potential future refactor that allows us to upload multiple slabs for a single object in parallel. 
e.g. if the last sector of a slab takes 1 second to finish it would be good if the next slab already started uploading to reduce the typical zig-zag pattern on single file uploads.

sdk/client.go
peterjan: It's an SDK though

ChrisSchinnerl: It is, but I think both successful and cancelled uploads are cases where returning errors for failed shards isn't particularly useful.

That said I noticed we are hiding individual errors even on failed uploads so maybe we need to f/u with better error handling anyway. We could have an `UploadError` type for `Upload` that implements `error` that you can cast the returned error to if you want more information. Similar to the host error set in `renterd`.

PR #223, opened 2025-07-02
Update `withRevision`

peterjan:
The `reviseFn` in `withRevision` doesn't always revise the contract, this is the case when renewing a contract. In that case it's better to return the original revision and have the caller only update the revision in the store if the revision number is greater.

I extended the unit test and it'll be followed by https://github.com/SiaFoundation/indexd/pull/208 in which the integration tests highlighted the issue in the first place.


client/client.go
```diff
@@ -298,16 +306,18 @@ func (c *HostClient) withRevision(ctx context.Context, contractID types.FileCont
...
 
 		// try and revise the contract again
-		update, err = reviseFn(rev)
+		revised, err = reviseFn(rev)
 	}
 	if err != nil {
 		return err
-	} else if update.RevisionNumber <= rev.RevisionNumber {
-		return fmt.Errorf("new revision number %d is not greater than current revision number %d", update.RevisionNumber, rev.RevisionNumber)
+	} else if revised.ID != contractID {
```
ChrisSchinnerl, line 312: Imo this is incorrect because we are adding a wrong sanity check to engineer our way around another wrong sanity check.

If `withRevision` isn't always guaranteed to revise a contract we should just allow the return value of the `reviseFn` callback to return `nil` to indicate that. If the returned revision isn't `nil` the ID should always match. Either that or instead of a point we can just return the same revision that was passed to `reviseFn` and then we avoid persisting it if the revision number hasn't changed. 

Definitely `withRevision` shouldn't implicitly be used to persist a renewed contract. That's what `AddRenewedContract` is for.

ChrisSchinnerl (changes requested):

client/client.go
peterjan: > Definitely withRevision shouldn't implicitly be used to persist a renewed contract. That's what AddRenewedContract is for.

That is not what was happening. I did not consider to return the same revision in `reviseFn` though, that's a good idea.

client/client.go
```diff
@@ -105,11 +105,11 @@ func (c *HostClient) AppendSectors(ctx context.Context, hostPrices proto.HostPri
 
 	// append sectors
 	var res rhp.RPCAppendSectorsResult
-	if err := c.withRevision(ctx, contractID, func(revision types.V2FileContract) (types.V2FileContract, error) {
+	if err := c.withRevision(ctx, contractID, func(revision types.V2FileContract) (_ types.V2FileContract, err error) {
```
ChrisSchinnerl: leftovers? Also in a few other places

client/client.go
```diff
@@ -272,44 +272,43 @@ func (c *HostClient) withRevision(ctx context.Context, contractID types.FileCont
...
 
 		// try and revise the contract again
-		update, err = reviseFn(rev)
+		revised, err = reviseFn(local)
 	}
 	if err != nil {
 		return err
-	} else if update.RevisionNumber <= rev.RevisionNumber {
-		return fmt.Errorf("new revision number %d is not greater than current revision number %d", update.RevisionNumber, rev.RevisionNumber)
 	}
```
ChrisSchinnerl, line 315: I'd keep the check you had before but invert it. To sanity check that the contract id didn't change before updating the revision.

ChrisSchinnerl (changes requested):

client/client.go
peterjan: These are not leftovers, we weren't overwriting `res` in the client methods. Figured it was probably fine to take it along in this PR since all of these would've become immediately apparent as soon as we add integration tests that cover these RPCs.

The reason there are no unit tests by the way is because we call `rhp.RPCX` directly. Adding unit tests would mean we'd have to add `rpcFnX` to the client to be able to mock it in testing. Very similar to the `latestRevisionFn: rhp.RPCLatestRevision, // allows mocking in tests` we already have.

client/client.go
peterjan: I guess that's safer indeed but it requires returning `rhp.ContractRevision` in `withRevision`. I kind of wanted to avoid that but I can re-add it. Currently `withRevision` always returns the updated revision, except for renewals, so the contract ID is always the same. It's safer for sure, I just felt it wasn't really necessary.

client/client.go
```diff
@@ -266,50 +277,49 @@ func (c *HostClient) syncRevision(ctx context.Context, contractID types.FileCont
 // reports an invalid signature, suggesting the local revision is out of sync,
 // it will synchronize with the host and retry the function using the updated
 // revision. Therefore, the revise function must be idempotent.
-func (c *HostClient) withRevision(ctx context.Context, contractID types.FileContractID, reviseFn func(revision types.V2FileContract) (types.V2FileContract, error)) error {
+func (c *HostClient) withRevision(ctx context.Context, contractID types.FileContractID, reviseFn func(revision types.V2FileContract) (rhp.ContractRevision, error)) error {
```
ChrisSchinnerl:
```suggestion
func (c *HostClient) withRevision(ctx context.Context, contractID types.FileContractID, reviseFn func(revision rhp.ContractRevision) (rhp.ContractRevision, error)) error {
```

This is better. That way the input and output of the passed in function are the same so the caller doesn't need to construct it. Especially since RPCs expect a `ContractRevision` anyway.

ChrisSchinnerl (changes requested):

peterjan:
@n8maninger , the contracts API was merged into this, sorry about that, I figured `withRevision` would get merged in quicker and then forgot to remind you

n8mgr (approved):

PR #225, opened 2025-07-02
Register alert for hosts that lost sectors

chris124567:
https://github.com/SiaFoundation/indexd/issues/222

Re-used alert message/id from renterd


slabs/dataintegrity.go
```diff
@@ -86,6 +91,20 @@ func (v *sectorVerifier) VerifySector(ctx context.Context, root types.Hash256) (
...
 
+func newLostSectorsAlert(hk types.PublicKey, lostSectors int) alerts.Alert {
+	return alerts.Alert{
+		ID:       alerts.IDForHost(alertLostSectorsID, hk),
+		Severity: alerts.SeverityWarning,
+		Message:  "Host has lost sectors",
+		Data: map[string]interface{}{
+			"lostSectors": lostSectors,
+			"hostKey":     hk.String(),
+			"hint":        "The host has reported that it can't serve at least one sector. Consider blocking this host through the blocklist feature. If you think this was a mistake and you want to ignore this warning for now you can reset the lost sector count",
```
peterjan: I wouldn't mention the reset for the time being since indexd does not have that endpoint yet.

slabs/manager.go
```diff
@@ -85,6 +87,11 @@ type (
 		SectorsForIntegrityCheck(ctx context.Context, hostKey types.PublicKey, limit int) ([]types.Hash256, error)
 		Slabs(ctx context.Context, accountID proto.Account, slabIDs []SlabID) ([]Slab, error)
 	}
+
+	// AlertsManager defines an interface to register and dismiss alerts.
```
peterjan:
```suggestion
	// AlertsManager defines an interface to register alerts.
```

slabs/dataintegrity.go
```diff
@@ -144,6 +163,9 @@ func (m *SlabManager) performIntegrityChecksForHost(ctx context.Context, verifie
 				hostLogger.Error("failed to record integrity check for successful sectors", zap.Error(err))
 				return fmt.Errorf("failed to record integrity check for successful sectors: %w", err)
 			}
+			if err := m.alerter.RegisterAlert(newLostSectorsAlert(verifier.HostKey(), len(lost))); err != nil {
```
ChrisSchinnerl: Registering the alert here is pretty straightforward but will only show us an alert when a host gets integrity checked and only display the number of lost sectors from that one check. 

Instead, register the alert in `performIntegrityChecks`. At the end of that method, fetch all hosts from the database that have >0 lost sectors and then register an alert for each one. You'll need to extend the `Store` interface and add a method on the mock to satisfy it. Then you can extend the existing test to check if the right alerts were registered in the mock store.

Usually we do this in 2 steps with a separate PR for each one.

- Add the database code for the interface method you want to add with testing and potentially benchmarking
- Add the method to the interface and make use of it (that would be this PR)

ChrisSchinnerl (changes requested):

slabs/manager.go
```diff
@@ -181,6 +190,19 @@ func (m *SlabManager) Close() error {
 	return nil
 }
 
+func newLostSectorsAlert(hk types.PublicKey) alerts.Alert {
```
peterjan: It's a little odd we don't report the number of sectors that are missing... It'll be hard to gauge for the user how big of a problem it really is and whether it's getting worse or not. It might make sense to return the lost sector counts alongside the public keys. It hurts the eyes but at the same time we're currently nudging users towards blocking a host as soon as the count is non zero.

slabs/manager_test.go
```diff
@@ -47,8 +49,25 @@ func (s *mockStore) Hosts(ctx context.Context, offset, limit int, queryOpts ...h
...
 
-func (s *mockStore) HostsForIntegrityChecks(ctx context.Context, limit int) ([]types.PublicKey, error) {
-	panic("not implemented")
+func (s *mockStore) HostsForIntegrityChecks(ctx context.Context, limit int) (result []types.PublicKey, err error) {
+	for hk := range s.hosts {
+		if len(result) >= limit {
+			break
+		}
+
+		delete(s.hosts, hk)
```
peterjan: Why do you delete the host? The store returns hosts that haven't been checked for a certain amount of time

slabs/dataintegrity_test.go
```diff
@@ -280,3 +285,75 @@ func TestPerformIntegrityChecksForHost(t *testing.T) {
 		t.Fatalf("expected lost sector %v, got %v", rootBad, store.lostSectors)
 	}
 }
+
+func TestIntegrityChecksAlert(t *testing.T) {
+	store := newMockStore()
+	am := newMockAccountManager(store)
+	hm := newMockHostManager()
+	account := types.GeneratePrivateKey()
```
peterjan: `nit` inline this, if you make it a variable you give the impression it'll be used down the  line

slabs/dataintegrity_test.go
```diff
@@ -280,3 +285,75 @@ func TestPerformIntegrityChecksForHost(t *testing.T) {
 		t.Fatalf("expected lost sector %v, got %v", rootBad, store.lostSectors)
 	}
 }
+
+func TestIntegrityChecksAlert(t *testing.T) {
```
peterjan, line 287: This test is doing a lot of stuff that's not really related to the alert. You can get rid of most of it:

```go
func TestIntegrityChecksAlert(t *testing.T) {
	store := newMockStore()
	alerter := alerts.NewManager()
	sm, err := newSlabManager(newMockAccountManager(store), nil, store, nil, alerter, types.GeneratePrivateKey(), types.GeneratePrivateKey())
	if err != nil {
		t.Fatal(err)
	}

	// assert there are no alerts
	if alerts, err := alerter.Alerts(0, math.MaxInt64); err != nil {
		t.Fatal(err)
	} else if len(alerts) != 0 {
		t.Fatalf("expected 0 alerts, got %d", len(alerts))
	}

	// mock a lost sector
	hk := types.PublicKey{1}
	store.lostSectors[hk] = make(map[types.Hash256]struct{})
	store.lostSectors[hk][types.Hash256{1}] = struct{}{}

	// perform integrity checks
	sm.performIntegrityChecks(context.Background())
	
	// assert alert was generated
	var got alerts.Alert
	if alerts, err := alerter.Alerts(0, math.MaxInt64); err != nil {
		t.Fatal(err)
	} else if len(alerts) != 1 {
		t.Fatalf("expected 1 alert, got %d", len(alerts))
	} else {
		got = alerts[0]
	}

	expected := newLostSectorsAlert(hk)
	got.Timestamp = expected.Timestamp // ignore timestamp
	if !reflect.DeepEqual(expected, got) {
		t.Fatal("unexpected alert", expected, got)
	}
}
```

peterjan (commented):
When registering alerts it's important to think about what mechanism will unregister them. In the case of lost sectors this is only possible if the lost sector count gets reset, but still.. in the renter we do this by grabbing all used hosts and then diffing them with the ones that have lost sectors, for every host in that diff we unregister the alert ID regardless of whether it exists or not. That way we're sure any potential alerts are unregistered if the host has no lost sectors.

slabs/manager.go
```diff
@@ -182,6 +192,20 @@ func (m *SlabManager) Close() error {
 	return nil
 }
 
+func newLostSectorsAlert(hk types.PublicKey, lostSectors uint64) alerts.Alert {
```
n8mgr: I'm gonna go against the grain here and say we don't want an alert per host. The most "useful" UX, imo, would be:

1. If any host has lost sectors, register an alert: "hosts have lost sectors. check hosts page for details"
2. Surface the number of lost sectors and action items in the table on the hosts page. Could even make a filter for it.

I'm happy to be overruled here, but returning an unbounded list of hosts from the db and surfacing an alert for each of them is not a great UX. Gives `renterd` 4MM "upload failed" alert vibes.

n8mgr: To expand, the DB query would become:
```go
// HasLostSectors returns (true, nil) if there are any hosts with lost sectors.
func (s *Store) HasLostSectors(ctx context.Context) (lost bool, err error) {
	err = s.transaction(ctx, func(ctx context.Context, tx *txn) error {
		return tx.QueryRow(ctx, `SELECT
			EXISTS (
				SELECT 1
				FROM hosts
				WHERE lost_sectors > 0
			)`).Scan(&lost)
	})
	return
}
```

peterjan: Historically the lost sectors alert didn't give us much issues though seeing as it's usually a handful if any.  That said, though I think I do indeed prefer your idea. It's kind of similar to contract updates and the churn alert, that also aggregates a bunch of stuff into a single alert. I would probably send the host keys as alert context so the UI can link to specific hosts from the alert itself if need be. Either way is fine.

chris124567: OK now if there are hosts with lost sectors, we generate one alert which has the list of relevant hosts in the data.  If there are no hosts with lost sectors, that alert is dismissed.

slabs/manager.go
```diff
@@ -234,6 +261,36 @@ func (m *SlabManager) performIntegrityChecks(ctx context.Context) error {
 		wg.Wait()
 	}
 
+	offset := 0
+	const limit = 100
+	activeHosts := make(map[types.PublicKey]struct{})
```
peterjan: not used?

slabs/manager_test.go
```diff
@@ -47,11 +51,55 @@ func (s *mockStore) FailingSectors(ctx context.Context, hostKey types.PublicKey,
 }
 
 func (s *mockStore) Hosts(ctx context.Context, offset, limit int, queryOpts ...hosts.HostQueryOpt) ([]hosts.Host, error) {
-	panic("not implemented")
+	var allHosts []hosts.Host
```
peterjan: If no longer needed, please revert to `panic("not implemented")` because it'll cause a merge conflict in my PR. If you need the `Hosts` method I'll fix the conflict obviously. I think you only need `HostsWithLostSectors` though.

peterjan (approved):

n8mgr (approved):

PR #233, opened 2025-07-09
Update `downloadShards`

peterjan:
This PR is a refactor of `downloadShards`, I think it's a bit easier to read now and it mirrors the upload code. Functionality remains exactly the same, to showcase that I didn't touch `TestDownloadSlab`, I'll move that into `download_test.go` though in a follow up.


slabs/migrations_test.go
```diff
@@ -337,7 +337,7 @@ func TestDownloadSlab(t *testing.T) {
 		unavailableSlab.Sectors = slices.Clone(unavailableSlab.Sectors)
 		unavailableSlab.Sectors[0].HostKey = nil
 		unavailableSlab.Sectors[1].HostKey = nil
-		_, err := sm.downloadShards(context.Background(), slab, nil, zap.NewNop())
```
n8mgr, line 340: > didn't touch TestDownloadSlab

😛

Issue #217, opened 2025-07-01
API docs


ChrisSchinnerl:
- [x] OpenAPI spec
- [x] Hosted docs at `api.sia.tech/indexd`

PR #237, opened 2025-07-11
Add OpenAPI spec

peterjan:
It's obviously incomplete but it's definitely a decent start. Every time we add routes now we should extend it. I think Alex and John will be a valuable resource in telling us where the docs are lacking more explanation and descriptions. If the UI reaches MVP stage it's probably a good idea to go over the entire document again and fill in the blanks, add more descriptions and so on.

Related to https://github.com/SiaFoundation/indexd/issues/217


n8mgr (changes requested):
Would probably separate this into `admin` and `app` since users will likely be looking for one or the other.

peterjan:
> Would probably separate this into `admin` and `app` since users will likely be looking for one or the other.

I know that's what we want eventually but I think we should keep it in one file for as long as possible. It's kind of a pain to bundle/unbundle these yaml files because the openapi spec keeps changing how you can use `$ref`. I suggested to split the spec up into multiple files for `renterd` too and we decided against it at the time. This case is very different though, seeing we have a clear separation between two APIs. I pushed a commit that splits it up, you can bundle it by running something like  `redocly bundle openapi/admin.yml -o openapi/admin-bundled.yml`. 

[admin-bundled.yml.zip](https://github.com/user-attachments/files/21230986/admin-bundled.yml.zip)

n8mgr (approved):

PR #238, opened 2025-07-11
Add `/hosts` endpoint to application API.

peterjan:
This PR adds a `GET /hosts` endpoint to the application API that returns a paginated list of usable, non-blocked hosts that have an active contract.

### DIFF
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -8,6 +8,8 @@ import (
 
 	proto "go.sia.tech/core/rhp/v4"
 	"go.sia.tech/core/types"
+	"go.sia.tech/indexd/api"
+	"go.sia.tech/indexd/hosts"
 	"go.sia.tech/indexd/slabs"
 	"go.sia.tech/jape"
 	"go.uber.org/zap"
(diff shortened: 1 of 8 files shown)


api/app/app.go
```diff
@@ -19,6 +21,7 @@ type (
 
 	// Store defines the store interface for the application API.
 	Store interface {
+		Hosts(ctx context.Context, offset, limit int, queryOpts ...hosts.HostQueryOpt) ([]hosts.Host, error)
```
n8mgr: No need to return the full host object. Things like blocked, scan times, and usability are internal and shouldn't be exposed. Should ideally only ever return hosts that are usable by the app with as little unnecessary state as possible.

I would say it could probably be pared down to:

```
Host struct {
	PublicKey              types.PublicKey     `json:"publicKey"`
	Networks               []net.IPNet         `json:"networks"`
}
```

Maybe even drop the `Networks` field.

peterjan: I added a method on the store called `UsableHosts` that returns a `HostInfo` with only that info.

n8mgr (approved):

Issue #227, opened 2025-07-03
Correct handling of `renewed_to` during reorg


ChrisSchinnerl:
I think we right now have an edge case where reverting a renewed contract will change the state of its parent back from "complete" to "active", but `renewed_to` will remain set.

We should add some testing that for a contract C_parent and its renewed contract C_child does the following.

1. renew `C_parent` to `C_child`
2. revert `C_child`
3. checks `C_parent`'s state changed from `complete` to `active`, `renewed_to` to `NULL` and `C_child`'s changed to `pending`
4. reapply `C_child`
5. makes sure `C_parent`'s state is `complete` again, `renewed_to` is set  to `C_child` and `C_child`'s state is `active`

and another test case that does:

1. renew `C_parent` to `C_child`
2. revert `C_child`
3. checks `C_parent`'s state changed from `complete` to `active`, `renewed_to` to `NULL` and `C_child`'s changed to `pending`
4. renew `C_parent` to another contract `C_child2`
5. mine enough blocks for `C_child` to get pruned
6. make sure `C_parent`'s `renewed_to` is set to `C_child2` and not cleared when `C_child` gets pruned

and finally implement any fixes necessary to get those tests to pass

PR #239, opened 2025-07-12
Correct handling of renewed_to

chris124567:
This implements the first part of https://github.com/SiaFoundation/indexd/issues/227... I wanted to make sure I was on the right track before continuing with the pruning test since I'm still somewhat inexperienced with indexd.


contracts/chain_test.go
```diff
@@ -340,7 +343,7 @@ func (tx *mockUpdateTx) ContractElements() ([]types.V2FileContractElement, error
 	return stateElements, nil
 }
 
-func (tx *mockUpdateTx) Contract(contractID types.FileContractID) (types.V2FileContractElement, ContractState) {
+func (tx *mockUpdateTx) Contract(contractID types.FileContractID) (types.V2FileContractElement, ContractState, *types.FileContractID) {
```
peterjan: is updating the signature here a leftover? 

edit: I missed the test, it's a bit of a `nit` but I would remove this and add an extra method called `Renewal` or `RenewedContract`, just because the signature is a little weird and it's not obvious what you're returning.

persist/postgres/contracts.go
```diff
@@ -497,6 +497,19 @@ func (tx *updateTx) UpdateContractState(contractID types.FileContractID, state c
 	return nil
 }
 
+// UpdateContractRenewedTo updates the renewed to ID of a contract to the provided one.
+func (tx *updateTx) UpdateContractRenewedTo(contractID types.FileContractID, renewedTo *types.FileContractID) error {
+	var value any
+	if renewedTo != nil {
+		value = sqlHash256(*renewedTo)
+	}
+	_, err := tx.tx.Exec(tx.ctx, `UPDATE contracts SET renewed_to = $1 WHERE contract_id = $2`, value, sqlHash256(contractID))
```
peterjan: I think it makes sense to make this query a bit more strict, we really want this to error out if this is not doing exactly what we expect it's doing. E.g. we want to allow going from `renewed_to` -> `NULL` and from `NULL` to `renewed_to`, we don't want to allow `NULL` -> `NULL` (fine of course, but not expected to happen) or `renwed_to_a` -> `renewed_to_b` (not good, we should've nulled it before the update). 

I would split it up depending on whether `renewedTo` is nil, then do `WHERE contract_id = $2 AND renwed_to IS NULL` and use `res.RowsAffected` to ensure we've updated a row.

peterjan:
On the right track 👍

chris124567:
There are a few ways to proceed here but I'm not confident in any of them.  One easy way to do the pruning part would be to adapt the existing `TestPruneExpiredContractElements` in the postgres store,  Unfortunately, the `applyContractDiff`/`revertContractDiff` methods on the contract manager are private and it would seem inconsistent with the other managers to make those public.  There's no way to create a consensus.ApplyUpdate with arbitrary diffs as those fields are private too.  Alternatively, we could mine/revert blocks with renewals and create a sort of local test node like we have in the explored tests.  Or maybe we could mock a bunch of things in `contracts/chain_test.go`, but that wouldn't test the postgres store.

persist/postgres/contracts.go
```diff
@@ -497,6 +497,21 @@ func (tx *updateTx) UpdateContractState(contractID types.FileContractID, state c
 	return nil
 }
 
+// UpdateContractRenewedTo updates the renewed to ID of a contract to the provided one.
+func (tx *updateTx) UpdateContractRenewedTo(contractID types.FileContractID, renewedTo *types.FileContractID) error {
```
ChrisSchinnerl: This is missing a test

ChrisSchinnerl:
> There are a few ways to proceed here but I'm not confident in any of them. One easy way to do the pruning part would be to adapt the existing `TestPruneExpiredContractElements` in the postgres store, Unfortunately, the `applyContractDiff`/`revertContractDiff` methods on the contract manager are private and it would seem inconsistent with the other managers to make those public. There's no way to create a consensus.ApplyUpdate with arbitrary diffs as those fields are private too. Alternatively, we could mine/revert blocks with renewals and create a sort of local test node like we have in the explored tests. Or maybe we could mock a bunch of things in `contracts/chain_test.go`, but that wouldn't test the postgres store.

I think the second test will need to be in `persist/postgres/contracts_test.go` like `TestPruneExpiredContractElements` but instead of using `applyContractDiff`/`revertContractDiff` you fake the revert by manually updating the contracts.
Meaning you set the `state` of each contract and the `renewed_to` field of the parent manually. Then you add the new renewed contract followed by rejecting and pruning expired contracts.

ChrisSchinnerl (changes requested):

persist/postgres/contracts.go
```diff
@@ -498,6 +498,21 @@ func (s *Store) RejectPendingContracts(ctx context.Context, maxFormation time.Ti
 	})
 }
 
+// UpdateContractRenewedTo updates the renewed to ID of a contract to the provided one.
+func (tx *updateTx) UpdateContractRenewedTo(contractID types.FileContractID, renewedTo *types.FileContractID) error {
+	var value any
+	if renewedTo != nil {
+		value = sqlHash256(*renewedTo)
+	}
+	res, err := tx.tx.Exec(tx.ctx, `UPDATE contracts SET renewed_to = $1 WHERE contract_id = $2`, value, sqlHash256(contractID))
```
ChrisSchinnerl: You should be able to drop the if condition if you do the following

```suggestion
	res, err := tx.tx.Exec(tx.ctx, `UPDATE contracts SET renewed_to = $1 WHERE contract_id = $2`, (*sqlHash256)(renewedTo), sqlHash256(contractID))
```

contracts/chain.go
```diff
@@ -123,6 +124,12 @@ func (m *ContractManager) applyContractDiff(tx *updateTx, diff consensus.V2FileC
 		switch {
 		case diff.Resolution != nil:
 			state = ContractStateResolved
+			if _, ok := diff.Resolution.(*types.V2FileContractRenewal); ok {
+				renewedTo := diff.V2FileContractElement.ID.V2RenewalID()
+				if err := tx.UpdateContractRenewedTo(diff.V2FileContractElement.ID, &renewedTo); err != nil {
+					return fmt.Errorf("failed to update renewed to for %v: %w", diff.V2FileContractElement.ID, err)
+				}
+			}
```
ChrisSchinnerl, line 132: I think we might be running into a potential panic here when somehow the contract that we renewed to gets deleted. Because `renewed_to` has a foreign key constraint on `contracts.contract_id`. So we need to change `UpdateContractRenewedTo` to not update `renewed_to` if the contract doesn't exist and explain why that is in the docstring. And add that case to `TestUpdateContractRenewedTo`.

persist/postgres/contracts_test.go
```diff
@@ -1270,6 +1270,199 @@ func TestUpdateContractRevision(t *testing.T) {
...
+	}
+
+	if contract, err := store.Contract(context.Background(), contractID); err != nil {
+		t.Fatal(err)
+	} else if contract.RenewedTo != (types.FileContractID{}) {
+		t.Fatal("expected contract to not be renewed")
+	}
+}
+
+func TestPruneRenewedTo(t *testing.T) {
```
ChrisSchinnerl: While looking at this I realized that what we do in this test can't happen in RHP4 making this test obsolete. Because a contract can't renew to 2 different IDs in RHP4 anymore. Meaning that the other tests in this PR should already cover everything we need.

However, this reminded me of another edge case that I opened an issue for: https://github.com/SiaFoundation/indexd/issues/347

ChrisSchinnerl (changes requested):

contracts/chain.go
chris124567: Do we want to return nil in that case?  The query as it is now would error if executing it would cause a constraint to be violated.

ChrisSchinnerl (approved):

n8mgr (approved):

Issue #220, opened 2025-07-01
Add integration tests


ChrisSchinnerl:
We need integration test coverage for all the common features

- [x] pinning slabs
- [x] migrations
- [x] contract pruning
- [x] integrity checks
- [x] funding accounts

PR #240, opened 2025-07-14
Add TestAccountFunding (e2e)

peterjan:
Related to https://github.com/SiaFoundation/indexd/issues/220

### DIFF
--- a/accounts/manager_test.go
+++ b/accounts/manager_test.go
@@ -98,6 +98,13 @@ func (s *mockStore) HostAccountsForFunding(ctx context.Context, hk types.PublicK
 	return
 }
 
+func (s *mockStore) ScheduleAccountsForFunding(ctx context.Context, hostKey types.PublicKey) error {
+	for _, acc := range s.eas[hostKey] {
+		acc.NextFund = time.Now()
+	}
+	return nil
+}
+
 func (s *mockStore) UpdateHostAccounts(ctx context.Context, accounts []HostAccount) error {
 	for _, acc := range accounts {
 		_, ok := s.eas[acc.HostKey]
@@ -159,7 +166,7 @@ func TestAccountManager(t *testing.T) {
 	}
 
 	contractIDs := []types.FileContractID{{1}}
-	err := am.FundAccounts(context.Background(), host, contractIDs, zap.NewNop())
+	err := am.FundAccounts(context.Background(), host, contractIDs, false, zap.NewNop())
 	if !errors.Is(err, hosts.ErrNotFound) {
 		t.Fatal("expected host not found error")
 	}
@@ -170,7 +177,7 @@ func TestAccountManager(t *testing.T) {
 	s.accounts[types.GeneratePrivateKey().PublicKey()] = struct{}{}
 
 	// fund accounts
-	err = am.FundAccounts(context.Background(), host, contractIDs, zap.NewNop())
+	err = am.FundAccounts(context.Background(), host, contractIDs, false, zap.NewNop())
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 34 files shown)


persist/postgres/store.go
```diff
@@ -32,6 +32,11 @@ type (
 	}
 )
 
+// Database returns the underlying database connection pool.
+func (s *Store) Database() *pgxpool.Pool {
```
n8mgr: I'm very against exposing this just for some tests. Usually this suggests the need for a more testable interface. It looks like the primary issue is the account funding interval. That could pretty easily be an option on the account manager instead of hard coded to an hour.

internal/test/e2e/accounts_test.go
```diff
@@ -0,0 +1,102 @@
+package e2e
```
n8mgr: Why do we need a dedicated package for this? Can it not go in `accounts_test`? Ideally tests would be close to the code they're wanting to test.

peterjan: We had an `e2e` package in `renterd` and it allowed us to run integration tests separately, which I liked. I didn't think too much about that though, I guess we can probably accomplish the same thing if we annotate end to end tests with `E2E` or something and run tests using a regex pattern. I moved it 👍

persist/postgres/store.go
peterjan: Yeah that was a pretty bold move indeed. This is definitely the tricky part of integration testing a highly concurrent and asynchronous system. The main problem I see with options is that you have to inject config everywhere and have to define options that are only used for testing purposes. That's kind of what we ended up with in `renterd` and I wanted to avoid that at all cost. That's not the only thing though.. if you do expose an option and you define a small interval, it is still _very_ much prone to NDFs because you don't know when it'll run/trigger. So you can't do the assertion steps you want to which are 1) check not OK 2) trigger 3) check OK.

In this case though I think I found a really good middle ground. We already have the `Trigger` endpoint in the debug API, but it wasn't exposed on the client yet (just an oversight). I exposed that and I allowed to pass a `force` boolean. When an account funding is forcefully triggered, it will reset the `next_fund` for all accounts before paging though the accounts. I feel this is useful outside of testing too. I could not pass `math.MaxInt` as a limit (which is what I wanted to do initially) because that would imply updating a large batch of accounts at once. This is where the way we exhaust accounts (updating a timestamp) bites us in the ass a bit. I added `ScheduleAccountsForFunding` which I execute on a per host basis before funding accounts for that particular host.

I think in this case it worked out really well, in general though the older I get, the more I don't care about exposing the database like this if all developers "simply agree" to never ever misuse it. Sometimes it doesn't make sense to overcomplicate tests, constructors and method signatures and work your way around something that's not exposed, if the upside is a valuable integration test. But that's just my 2c, I used to feel very strongly against this as well.

n8mgr (approved):

accounts/manager_test.go
```diff
@@ -248,20 +255,28 @@ func TestAccountManager(t *testing.T) {
 	}
 
 	// assert all accounts next fund was updated and consecutive failed funds was reset
-	expected = time.Now().Add(accountFundInterval)
+	expected = time.Now().Add(time.Hour)
```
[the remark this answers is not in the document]
peterjan: going to take it along in another PR

api/admin/admin.go
```diff
@@ -268,7 +271,10 @@ func (a *admin) handlePOSTAccount(jc jape.Context) {
 		return
 	}
 
-	a.contracts.TriggerAccountFunding()
+	err = a.contracts.TriggerAccountFunding(false)
+	if err != nil {
+		a.log.Warn("failed to trigger account funding after account ", zap.Error(err))
```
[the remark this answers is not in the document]
peterjan: going to take it along in another PR

PR #242, opened 2025-07-14
Add `/slab/:id` to App API

n8mgr:
Adds a new endpoint `[GET] /slab/:id` to the App SDK to fetch slab metadata from the indexer. 

Also adjusts the SDK to use data from the AppAPI

```
BenchmarkSlabs/Slab-16                       531           2234295 ns/op        18772.38 MB/s      15079 B/op        296 allocs/op
```

Prefer this over `[GET] /slabs/:id` in #235 since the pointers to `HostKey` and `ContractID` are only useful for the Admin API and not actionable from an App.

### DIFF
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"time"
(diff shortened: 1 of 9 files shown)


peterjan (approved):
Yeah that's better 👍 . Could've been two PRs though, it was a little hard to follow. Do you still want to expose an endpoint that returns a paginated list of slab digests for a certain account? I was not too sure whether that endpoint was necessary.

n8mgr:
> Yeah that's better 👍 . Could've been two PRs though, it was a little hard to follow. Do you still want to expose an endpoint that returns a paginated list of slab digests for a certain account? I was not too sure whether that endpoint was necessary.

I think it’s necessary. Particularly if users are trying to recover. Maybe the object endpoints make it obsolete though.. 🤔

PR #250, opened 2025-07-16
Add TestDataIntegrity (e2e)

peterjan:
Related to #220


client/client.go
```diff
@@ -183,8 +183,14 @@ func (c *HostClient) ReadSector(ctx context.Context, hostPrices proto.HostPrices
 func (c *HostClient) RefreshContract(ctx context.Context, settings proto.HostSettings, params proto.RPCRefreshContractParams) (rhp.RPCRefreshContractResult, error) {
 	var res rhp.RPCRefreshContractResult
 	if err := c.withRevision(ctx, params.ContractID, func(contract rhp.ContractRevision) (_ rhp.ContractRevision, err error) {
-		res, err = rhp.RPCRefreshContract(ctx, c.client, c.cm, c.signer, c.cm.TipState(), settings.Prices, contract.Revision, params)
-		return res.Contract, err
+		res, err = rhp.RPCRefreshContractPartialRollover(ctx, c.client, c.cm, c.signer, c.cm.TipState(), settings.Prices, contract.Revision, params)
```
ChrisSchinnerl: We can't default to a partial rollover. Instead we need to check the host version and then decide whether to do a full or a partial one.

api/admin/admin_test.go
ChrisSchinnerl, line 1: NDF in `TestContractsAPI` it seems

slabs/e2e_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	proto "go.sia.tech/core/rhp/v4"
+	"go.sia.tech/core/types"
+	"go.sia.tech/indexd/api/admin"
+	"go.sia.tech/indexd/internal/testutils"
+	"go.sia.tech/indexd/slabs"
+	"lukechampine.com/frand"
+)
+
+func TestDataIntegrityE2E(t *testing.T) {
```
ChrisSchinnerl: Probably fine to not add the `E2E` suffix?

slabs/e2e_test.go
```diff
@@ -0,0 +1,91 @@
...
+	// create cluster
+	logger := testutils.NewLogger(false)
+	cluster := testutils.NewCluster(t, testutils.WithLogger(logger), testutils.WithHosts(3), testutils.WithIndexer(testutils.WithSlabOptions(slabs.WithHealthCheckInterval(100*time.Millisecond), slabs.WithIntegrityCheckIntervals(50*time.Millisecond, 50*time.Millisecond))))
+	indexer := cluster.Indexer
+	time.Sleep(time.Second)
```
ChrisSchinnerl: Why do we need to sleep directly after creating the cluster?

ChrisSchinnerl (changes requested):
Looking pretty good. Probably even better after merging the slab manager PR and rebasing

peterjan:
> Probably even better after merging the slab manager PR and rebasing

Oh man, I wish.

slabs/e2e_test.go
peterjan: Sure. Not all integration tests are in an `e2e` file. Perhaps I should move the ones that aren't. I personally like the suffix for quickly jumping or running only the e2e tests using a regex.

slabs/e2e_test.go
peterjan: It takes about two seconds to reliably scan and form contracts with 3 hosts. The sleeps are temporary anyway, I just don't want to add the `.Retry` because I kind of hope we can come up with something better than simply polling at intervals.

api/admin/admin_test.go
peterjan: yeah `UsedCollateral` issue, had it fixed elsewhere but fixed it here too

client/client.go
```diff
@@ -183,8 +189,18 @@ func (c *HostClient) ReadSector(ctx context.Context, hostPrices proto.HostPrices
 func (c *HostClient) RefreshContract(ctx context.Context, settings proto.HostSettings, params proto.RPCRefreshContractParams) (rhp.RPCRefreshContractResult, error) {
 	var res rhp.RPCRefreshContractResult
 	if err := c.withRevision(ctx, params.ContractID, func(contract rhp.ContractRevision) (_ rhp.ContractRevision, err error) {
-		res, err = rhp.RPCRefreshContractFullRollover(ctx, c.client, c.cm, c.signer, c.cm.TipState(), settings.Prices, contract.Revision, params)
-		return res.Contract, err
+		if settings.ProtocolVersion.Cmp(minVersionPartialRefresh) >= 0 {
```
ChrisSchinnerl: I'd stick to the tagged `core` and `coreutils` and change this to a simple `if settings.ProtocolVersion[0] >= 5`.
Once the version helper is merged in `core` I'd like to add a constant for version 5.0.0 to `coreutils` and then we can check for that here.

client/client.go
```diff
@@ -183,8 +183,18 @@ func (c *HostClient) ReadSector(ctx context.Context, hostPrices proto.HostPrices
...
+			res, err = rhp.RPCRefreshContractFullRollover(ctx, c.client, c.cm, c.signer, c.cm.TipState(), settings.Prices, contract.Revision, params)
+		}
+		if err != nil {
+			return rhp.ContractRevision{}, err
+		}
+		// renewals return the old (or 'renewed') revision, the revision of the
+		// renewal will be persisted in the database when the renewed contract
+		// is added
+		return contract, nil
 	}); err != nil {
```
n8mgr, line 198: Does it make sense to separate this logic instead of sending the same values for both?

Either way, feel like this is out of scope for this PR.

ChrisSchinnerl (approved):

Issue #246, opened 2025-07-15
Add missing UI endpoints


peterjan:
Add some endpoints that would be useful for the UI
- [x] /syncer/connect
- [x] /host/:hostkey/scan
- [x] /wallet/event/:id
- [x] /txpool/recommendedfee

alexfreska:
@peterjan I see /state has a scanHeight is that the consensus or the wallet? Could we expose both:
- the consensus state with sync height and either a synced flag or the prev timestamps so I can decide if synced
- the wallet scan height

alexfreska:
- Do we need the ability to reset lost sectors in indexd?

peterjan:
> Do we need the ability to reset lost sectors in indexd?

Probably but I would hold off on that for a bit perhaps.

> scanHeight is that the consensus or the wallet? Could we expose both: sync height and scan height

The scan height is indeed the height up until which we've synced up with the chain manager. It's a little confusing terminology though because it's not only the wallet but the entire chain state and we refer to it as "syncing" so `syncHeight` could be confusing in that way. I'll add `scanHeight` | `syncHeight` | `synced` 

cc @alexfreska

chris124567:
@alexfreska Would the endpoint for individual events be useful or is just the existing endpoint for the list of events with limit/offset OK?

alexfreska:
@chris124567 I added it because I noticed we have a txn/event detail view in all the apps that assumes the event is in the current paginated list data. This breaks if we ever wanted to be able to pull up the view for an event that is not in the current data. But its also not super important if its annoying to add, I can adjust the UX to make it clear its a detail view for something in the current page only.

peterjan:
@chris124567 OK let's add it in that case 🤷‍♂️ just trying to avoid copying over technical debt or legacy from `renterd`. The indexer is a chance to do better and it'll probably be the software we work on for the next year or two so anything we don't need we should probably not add at this point.

alexfreska:
Could you all add `network: 'mainnet' | 'zen'` to /state response?

chris124567:
Once #301 is merged in coreutils I can add the event endpoint.

chris124567:
@alexfreska Is there anything else that'd be helpful besides the (pending) event endpoint?

alexfreska:
@chris124567 I've designed some features for the app and the main two areas that would be useful are:
[...]

PR #251, opened 2025-07-16
Add UI API endpoints

chris124567:
Add:
- `/syncer/connect`
- `/host/:hostkey/scan`
- `/txpool/recommendedfee`

with tests.  It doesn't seem possible to get a single event from a `SingleAddressWallet` which is needed for `/wallet/event/:id` if we want it to be efficient.  Do we want to make a PR in coreutils?

#246

### DIFF
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -13,6 +13,7 @@ import (
 
 	"go.sia.tech/core/consensus"
 	"go.sia.tech/core/types"
+	"go.sia.tech/coreutils/syncer"
 	"go.sia.tech/coreutils/wallet"
 	"go.sia.tech/indexd/accounts"
 	"go.sia.tech/indexd/api"
@@ -82,6 +83,7 @@ type (
 
 	// A Syncer can connect to other peers and synchronize the blockchain.
 	Syncer interface {
+		Connect(ctx context.Context, addr string) (*syncer.Peer, error)
 		BroadcastV2TransactionSet(index types.ChainIndex, txns []types.V2Transaction) error
 	}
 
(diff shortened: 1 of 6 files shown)


api/admin/admin.go
```diff
@@ -135,6 +138,12 @@ func NewAPI(chain ChainManager, contracts ContractManager, hosts HostManager, sy
 	routes := map[string]jape.Handler{
 		"GET /state": a.handleGETState,
 
+		// syncer endpoints
```
peterjan: `nit` can you move them down so the routes are listed in a (somewhat) alphabetical order - we're pretty big on that, goes for everything really, e.g. extending the manager with a new method, properties in a struct and so on

hosts/manager.go
```diff
@@ -320,9 +320,9 @@ func (m *HostManager) pruneHosts(ctx context.Context) {
 	}
 }
 
-// scanHost scans the host with given host key and returns it with updated
+// ScanHost scans the host with given host key and returns it with updated
 // settings and checks.
-func (m *HostManager) scanHost(ctx context.Context, hk types.PublicKey) (Host, error) {
+func (m *HostManager) ScanHost(ctx context.Context, hk types.PublicKey) (Host, error) {
```
peterjan: `nit` methods that are exposed are usually above private methods, so if you expose it please move it up 

generally it's:
- const
- vars
- types
- constructors
- public methods
- private methods
- helpers

peterjan (commented):
Looks good! Even though the PR is still small, it would've been better as three separate PRs imo. I would probably comment on the issue on why we need `/wallet/events/:id` per se, we didn't have it in `renterd` either. It's easy to add though but the fewer routes we have to maintain the better.

chris124567:
> Looks good! Even though the PR is still small, it would've been better as three separate PRs imo. I would probably comment on the issue on why we need `/wallet/events/:id` per se, we didn't have it in `renterd` either. It's easy to add though but the fewer routes we have to maintain the better.

I'll comment on the issue.  If we decide to add that endpoint, I'll make a separate PR :+1:

peterjan (approved):

PR #255, opened 2025-07-17
Add `settings` to `HostInfo`

peterjan:


hosts/host.go
```diff
@@ -102,8 +102,9 @@ type (
 	// key and addresses. It is used for listing usable hosts in the
 	// application API.
 	HostInfo struct {
-		PublicKey types.PublicKey    `json:"publicKey"`
-		Addresses []chain.NetAddress `json:"addresses"`
+		PublicKey types.PublicKey     `json:"publicKey"`
+		Addresses []chain.NetAddress  `json:"addresses"`
+		Settings  proto4.HostSettings `json:"settings"`
```
n8mgr, line 107: Would rather have apps scan this themselves

peterjan: Hm ok. It's kind of an annoying extra step. Needed it for integration tests.

PR #257, opened 2025-07-18
Add `limit` to UnhealthySlab

### LINKED ISSUE #252: Consider adding `limit` to `UnhealthySlab`
see https://github.com/SiaFoundation/indexd/pull/198#discussion_r2208196841

chris124567:
Closes https://github.com/SiaFoundation/indexd/issues/252

### DIFF
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -458,7 +458,7 @@ func (s *Store) UnpinnedSectors(ctx context.Context, hostKey types.PublicKey, li
 	return roots, err
 }
 
-// UnhealthySlab returns a slab that has at least one sector that needs to be
+// UnhealthySlabs returns slabs which have at least one sector that needs to be
 // migrated to a new host and hasn't had a repair attempted since
 // 'maxRepairAttempt'. The condition for such a slab is that it either has:
 // a). a sector that is not stored on a host (host_id == null)
@@ -468,14 +468,14 @@ func (s *Store) UnpinnedSectors(ctx context.Context, hostKey types.PublicKey, li
 // subsequent or parallel calls from returning the same slab.
 //
 // NOTE: For the sake of scalability, we don't prioritize any slabs and instead
-// simply fetch the first one that we can get.
-func (s *Store) UnhealthySlab(ctx context.Context, maxRepairAttempt time.Time) (slabs.SlabID, error) {
-	var slabID slabs.SlabID
+// simply fetch the first ones that we can get.
+func (s *Store) UnhealthySlabs(ctx context.Context, maxRepairAttempt time.Time, limit int) ([]slabs.SlabID, error) {
+	var slabIDs []slabs.SlabID
 	err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
-		err := tx.QueryRow(ctx, `
+		rows, err := tx.Query(ctx, `
 			UPDATE slabs
 			SET last_repair_attempt = NOW()
-			WHERE id = (
+			WHERE id IN (
 				SELECT slabs.id
 				FROM slabs
 				INNER JOIN slab_sectors ON slabs.id = slab_sectors.slab_id
@@ -490,16 +490,32 @@ func (s *Store) UnhealthySlab(ctx context.Context, maxRepairAttempt time.Time) (
 						(sectors.host_id IS NULL)
 					)
 					AND (slabs.last_repair_attempt <= $1)
-				LIMIT 1
+				LIMIT $2
 			)
 			RETURNING digest
-		`, maxRepairAttempt).Scan((*sqlHash256)(&slabID))
-		if errors.Is(err, sql.ErrNoRows) {
+		`, maxRepairAttempt, limit)
+		if err != nil {
+			return fmt.Errorf("failed to query unhealthy slabs: %w", err)
+		}
+		defer rows.Close()
+
+		for rows.Next() {
+			var slabID slabs.SlabID
+			if err := rows.Scan((*sqlHash256)(&slabID)); err != nil {
+				return fmt.Errorf("failed to scan slab ID: %w", err)
+			}
+			slabIDs = append(slabIDs, slabID)
+		}
+		if err := rows.Err(); err != nil {
+			return err
+		}
+
+		if len(slabIDs) == 0 {
 			return slabs.ErrSlabNotFound
 		}
-		return err
+		return nil
 	})
-	return slabID, err
+	return slabIDs, err
 }
 
 // MigrateSector updates a sector that was just migrated in the database to be
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -471,7 +471,6 @@ func (s *Store) UnpinnedSectors(ctx context.Context, hostKey types.PublicKey, li
 func (s *Store) UnhealthySlabs(ctx context.Context, maxRepairAttempt time.Time, limit int) ([]slabs.Slab, error) {
 	var results []slabs.Slab
 	err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
-		dbIDMap := make(map[int64]int)
 		rows, err := tx.Query(ctx, `
 			UPDATE slabs
 			SET last_repair_attempt = NOW()
@@ -492,7 +491,7 @@ func (s *Store) UnhealthySlabs(ctx context.Context, maxRepairAttempt time.Time, 
 					AND (slabs.last_repair_attempt <= $1)
 				LIMIT $2
 			)
-			RETURNING id, digest, encryption_key, min_shards, pinned_at
+			RETURNING digest
 		`, maxRepairAttempt, limit)
 		if err != nil {
 			return fmt.Errorf("failed to query unhealthy slabs: %w", err)
@@ -500,54 +499,14 @@ func (s *Store) UnhealthySlabs(ctx context.Context, maxRepairAttempt time.Time, 
 		defer rows.Close()
 
 		for rows.Next() {
-			var dbID int64
 			var slab slabs.Slab
-			if err := rows.Scan(&dbID, (*sqlHash256)(&slab.ID), (*sqlHash256)(&slab.EncryptionKey), &slab.MinShards, &slab.PinnedAt); err != nil {
+			if err := rows.Scan((*sqlHash256)(&slab.ID)); err != nil {
 				return fmt.Errorf("failed to scan slab ID: %w", err)
 			}
 
-			dbIDMap[dbID] = len(results)
 			results = append(results, slab)
 		}
-		if err := rows.Err(); err != nil {
-			return err
-		}
-
-		sectorsBatch := &pgx.Batch{}
-		for slabID := range dbIDMap {
-			sectorsBatch.Queue(`SELECT s.sector_root, h.public_key, csm.contract_id
-FROM sectors s
-INNER JOIN slab_sectors ss ON s.id = ss.sector_id
-LEFT JOIN hosts h ON h.id = s.host_id
-LEFT JOIN contract_sectors_map csm ON s.contract_sectors_map_id = csm.id
-WHERE ss.slab_id = $1
-ORDER BY ss.slab_index ASC`, slabID).Query(func(rows pgx.Rows) error {
-				defer rows.Close()
-				for rows.Next() {
-					var sector slabs.Sector
-					var hostKey sql.Null[sqlPublicKey]
-					var contractID sql.Null[sqlHash256]
-
-					if err := rows.Scan((*sqlHash256)(&sector.Root), &hostKey, &contractID); err != nil {
-						return fmt.Errorf("failed to scan sector: %w", err)
-					}
-
-					if hostKey.Valid {
-						sector.HostKey = (*types.PublicKey)(&hostKey.V)
-					}
-					if contractID.Valid {
-						sector.ContractID = (*types.FileContractID)(&contractID.V)
-					}
-					results[dbIDMap[slabID]].Sectors = append(results[dbIDMap[slabID]].Sectors, sector)
-				}
-				return rows.Err()
-			})
-		}
-		if err := tx.Tx.SendBatch(ctx, sectorsBatch).Close(); err != nil {
-			return fmt.Errorf("failed to get slab sectors: %w", err)
-		}
-
-		return nil
+		return rows.Err()
 	})
 	return results, err
 }
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -463,9 +463,8 @@ func (s *Store) UnpinnedSectors(ctx context.Context, hostKey types.PublicKey, li
 // 'maxRepairAttempt'. The condition for such a slab is that it either has:
 // a). a sector that is not stored on a host (host_id == null)
 // b). a sector that is stored in a bad contract (contract_id != null && contract.good = false)
-// When no slab is found, ErrSlabNotFound is returned. If a slab is found, it
-// will have its last_repair_attempt updated to the time of the call. To prevent
-// subsequent or parallel calls from returning the same slab.
+// If a slab is found, it will have its last_repair_attempt updated to the time
+// of the call. To prevent subsequent or parallel calls from returning the same slab.
 //
 // NOTE: For the sake of scalability, we don't prioritize any slabs and instead
 // simply fetch the first ones that we can get.
@@ -506,14 +505,7 @@ func (s *Store) UnhealthySlabs(ctx context.Context, maxRepairAttempt time.Time, 
 			}
 			slabIDs = append(slabIDs, slabID)
 		}
-		if err := rows.Err(); err != nil {
-			return err
-		}
-
-		if len(slabIDs) == 0 {
-			return slabs.ErrSlabNotFound
-		}
-		return nil
+		return rows.Err()
 	})
 	return slabIDs, err
 }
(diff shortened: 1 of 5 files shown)


persist/postgres/sectors.go
```diff
@@ -490,16 +490,32 @@ func (s *Store) UnhealthySlab(ctx context.Context, maxRepairAttempt time.Time) (
...
+			if err := rows.Scan((*sqlHash256)(&slabID)); err != nil {
+				return fmt.Errorf("failed to scan slab ID: %w", err)
+			}
+			slabIDs = append(slabIDs, slabID)
+		}
+		if err := rows.Err(); err != nil {
+			return err
+		}
+
+		if len(slabIDs) == 0 {
```
peterjan: It no longer makes sense to return `slabs.ErrSlabNotFound` not found here. We should return an empty slice.

persist/postgres/sectors.go
```diff
@@ -468,14 +468,14 @@ func (s *Store) UnpinnedSectors(ctx context.Context, hostKey types.PublicKey, li
 // subsequent or parallel calls from returning the same slab.
 //
 // NOTE: For the sake of scalability, we don't prioritize any slabs and instead
-// simply fetch the first one that we can get.
-func (s *Store) UnhealthySlab(ctx context.Context, maxRepairAttempt time.Time) (slabs.SlabID, error) {
-	var slabID slabs.SlabID
+// simply fetch the first ones that we can get.
+func (s *Store) UnhealthySlabs(ctx context.Context, maxRepairAttempt time.Time, limit int) ([]slabs.SlabID, error) {
```
peterjan, line 471: It would be great if we would return the full slab here. I actually need this for migrations and it was on my TODO, but we might as well do it here.

persist/postgres/sectors_test.go
```diff
@@ -949,11 +951,13 @@ func TestUnhealthySlabs(t *testing.T) {
...
 	}
-	unhealthyID, err = store.UnhealthySlab(context.Background(), time.Now())
+	unhealthyIDs, err = store.UnhealthySlabs(context.Background(), time.Now(), 1)
 	if err != nil {
 		t.Fatal(err)
-	} else if slabID != unhealthyID {
-		t.Fatalf("expected slab ID %v, got %v", slabID, unhealthyID)
+	} else if len(unhealthyIDs) != 1 {
+		t.Fatalf("expected 1 slab, got %d", len(unhealthyIDs))
+	} else if slabID != unhealthyIDs[0] {
```
peterjan, line 970: If you add a `limit` it's usually best to add a couple of assertions that the limit is being applied. E.g. make sure you have two slabs and assert you only get 1 if you apply a limit of 1.

persist/postgres/sectors_test.go
```diff
@@ -1397,8 +1401,8 @@ func BenchmarkPinSectors(b *testing.B) {
 	}
 }
 
-// BenchmarkUnhealthySlab benchmarks UnhealthySlab
-func BenchmarkUnhealthySlab(b *testing.B) {
+// BenchmarkUnhealthySlabs benchmarks UnhealthySlab
```
peterjan:
```suggestion
// BenchmarkUnhealthySlabs benchmarks UnhealthySlabs
```

persist/postgres/sectors_test.go
```diff
@@ -1397,8 +1401,8 @@ func BenchmarkPinSectors(b *testing.B) {
...
+func BenchmarkUnhealthySlabs(b *testing.B) {
```
peterjan, line 1554: We changed the implementation so the benchmark should reflect that. If we allow passing a limit the benchmark should/could check for 10/100/1000. Especially if we'll start hydrating the slabs instead of returning the digests.

slabs/manager.go
```diff
@@ -290,19 +290,16 @@ func (m *SlabManager) performSlabMigrations(ctx context.Context) error {
...
-			slab, err := m.store.UnhealthySlab(ctx, start)
-			if errors.Is(err, ErrSlabNotFound) {
-				return batch, nil
-			} else if errors.Is(err, context.Canceled) {
-				return nil, nil
-			} else if err != nil {
-				return nil, err
-			}
-			batch = append(batch, slab)
+	nextBatch := func(ctx context.Context) ([]Slab, error) {
```
peterjan: I would get rid of the helper all together. Something like this or do a length check and `break` is fine.
```go
var exhausted bool
for !exhausted {
    batch, err := store.UnhealthySlabs(ctx, time, 10)
    if err != nil {
        return err
    } else if len(batch) < 10 {
        exhausted = true
    }
}
```

peterjan (changes requested):

persist/postgres/sectors_test.go
chris124567: I might be misunderstanding what you're saying but for all the calls to UnhealthySlabs in tests we check the length of the slice that is returned

peterjan: Yes but I think you are missing the case where there are 2 unhealthy slabs and you only return 1 if the limit is 1. Might be mistaken though. Didn't really check just noticed you updated the test to make it pass but didn't add an extra assertion to check we only return at most `limit` items.

chris124567:
```
BenchmarkUnhealthySlabs/10-4                  15      75067405 ns/op
BenchmarkUnhealthySlabs/100-4                 12      84047137 ns/op
BenchmarkUnhealthySlabs/1000-4                 7     153831419 ns/op
```

persist/postgres/sectors_test.go
chris124567: I see what you mean.  I added a test to test the limit.

peterjan (dismissed):

persist/postgres/sectors_test.go
```diff
@@ -1397,98 +1534,111 @@ func BenchmarkPinSectors(b *testing.B) {
...
+		_, err = store.pool.Exec(context.Background(), `UPDATE sectors SET host_id = NULL, contract_sectors_map_id = NULL WHERE id % 4 = 1`)
+		if err != nil {
+			b.Fatal(err)
+		}
+	}
+
+	for _, batchSize := range []int{10, 100, 1000} {
+		b.Run(fmt.Sprint(batchSize), func(b *testing.B) {
+			reset(b)
+			b.ReportMetric(float64(b.N), "slabs")
```
peterjan: I'd drop this metric, it's no longer correct

persist/postgres/sectors_test.go
```diff
@@ -1397,98 +1534,111 @@ func BenchmarkPinSectors(b *testing.B) {
...
-		hks = append(hks, store.addTestHost(b, types.PublicKey{i}))
-	}
-
-	// add 2 contracts
-	store.addTestContract(b, hks[0])
-	store.addTestContract(b, hks[1])
+// BenchmarkUnhealthySlabs benchmarks UnhealthySlabs
+func BenchmarkUnhealthySlabs(b *testing.B) {
+	var store *Store
+	reset := func(b *testing.B) {
```
peterjan: Couldn't we limit `reset` to the following instruction? There's no need to redo the whole database setup

```go
	reset := func(b *testing.B) {
		// make sure the slabs have a last_repair_attempt time between 1 and 7 days
		// in the past
		_, err = store.pool.Exec(context.Background(), "UPDATE slabs SET last_repair_attempt = NOW() - interval '1 day' - interval '1 week' * random()")
		if err != nil {
			b.Fatal(err)
		}
	}
```

peterjan (commented):
The benchmarks are a little disappointing :( I was hoping it would be faster and we could avoid fetching the entire slab again using the ID in the migration code. It looks like we might not get away with it though and we'll have to revert back to returning a set of slab IDs.

persist/postgres/sectors.go
```diff
@@ -490,16 +490,66 @@ func (s *Store) UnhealthySlab(ctx context.Context, maxRepairAttempt time.Time) (
 						(sectors.host_id IS NULL)
 					)
 					AND (slabs.last_repair_attempt <= $1)
-				LIMIT 1
+				LIMIT $2
```
peterjan: I think this might be missing a `GROUP BY slabs.id`?

persist/postgres/sectors.go
```diff
@@ -458,24 +458,24 @@ func (s *Store) UnpinnedSectors(ctx context.Context, hostKey types.PublicKey, li
...
-// subsequent or parallel calls from returning the same slab.
+// If a slab is found, it will have its last_repair_attempt updated to the time
+// of the call. To prevent subsequent or parallel calls from returning the same slab.
 //
 // NOTE: For the sake of scalability, we don't prioritize any slabs and instead
-// simply fetch the first one that we can get.
-func (s *Store) UnhealthySlab(ctx context.Context, maxRepairAttempt time.Time) (slabs.SlabID, error) {
-	var slabID slabs.SlabID
+// simply fetch the first ones that we can get.
+func (s *Store) UnhealthySlabs(ctx context.Context, maxRepairAttempt time.Time, limit int) ([]slabs.Slab, error) {
```
ChrisSchinnerl: I don't think there is any need to fetch full slabs. Similar to how we only return host keys when we fetch hosts for integrity checks it's going to be more efficient to fetch a batch of slab ids and then have the individual workers fetch slabs they want to repair from the database.

ChrisSchinnerl (changes requested):

persist/postgres/sectors.go
peterjan: Yeah I already DM'ed Christopher. That's on me, I figured it was probably fine to return the full slab (since the limit is so low) but I was mistaken. The performance of `UnhealthySlabs` is quite poor, even with only returning the IDs... I do feel we should try and return more than one unhealthy slab ID at a time and change the implementation to take a `limit`.

This PR is a result of a suggestion Nate made on a different PR, asking why `UnhealthySlabs` does not take a `limit` parameter. Should we all agree on whether that is indeed something we should add? I agree(d) with Nate, I figured it was probably an oversight but I underestimated how slow this query is.

This made me go back to find https://github.com/SiaFoundation/indexd/pull/135#discussion_r2078796206 - we tried a materialized view and then decided to revert and revisit in a F/U. Perhaps now is the time or do we want to leave it as is.

chris124567:
```
BenchmarkUnhealthySlabs/10
BenchmarkUnhealthySlabs/10-4                  13          82066843 ns/op
BenchmarkUnhealthySlabs/25
BenchmarkUnhealthySlabs/25-4                  15          77077162 ns/op
BenchmarkUnhealthySlabs/50
BenchmarkUnhealthySlabs/50-4                  16          74697334 ns/op
```

Here are the results for the benchmark on a Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz.

@peterjan @ChrisSchinnerl Do we think a materialized view is warranted here to improve performance?

persist/postgres/sectors.go
chris124567: This makes the query 100x slower.  I think it shouldn't matter because the slab would still only be updated once?

persist/postgres/sectors.go
ChrisSchinnerl: I mean if the single slab id query is that much faster we could just stop trying to be smart for now and slap a good old for-loop at the problem (Inside the transaction).

This gets me the following results
```
cpu: Apple M2 Pro
BenchmarkUnhealthySlab
BenchmarkUnhealthySlab/Limit-1
BenchmarkUnhealthySlab/Limit-1-12         	      54	  22047807 ns/op
BenchmarkUnhealthySlab/Limit-100
BenchmarkUnhealthySlab/Limit-100-12       	      55	  23070129 ns/op
BenchmarkUnhealthySlab/Limit-1000
BenchmarkUnhealthySlab/Limit-1000-12      	      49	  25018747 ns/op
```

Compared to changing the query to have a higher limit
```
cpu: Apple M2 Pro
BenchmarkUnhealthySlab
BenchmarkUnhealthySlab/Limit-1
BenchmarkUnhealthySlab/Limit-1-12         	      50	  23341210 ns/op
BenchmarkUnhealthySlab/Limit-100
BenchmarkUnhealthySlab/Limit-100-12       	      49	  30562014 ns/op
BenchmarkUnhealthySlab/Limit-1000
BenchmarkUnhealthySlab/Limit-1000-12      	      25	  79011920 ns/op
```

The latter might not scale as well due to duplicate results from the subquery.

That cleans up the call site for now, makes things a bit more efficient since we reuse a database transaction and we can then come back later to optimize the query.

peterjan: Works for me!

persist/postgres/sectors.go
peterjan: True but at the call site you would request a batch of 10 unhealthy slabs, get one, and figure that must be it and never try and fetch the next batch. Chris's for loop idea is probably good enough for now.

persist/postgres/sectors.go
chris124567: @ChrisSchinnerl Can you post your code for UnhealthySlabs?  I can't get that level of performance.  I tried with a batch:

```
func (s *Store) UnhealthySlabs(ctx context.Context, maxRepairAttempt time.Time, limit int) ([]slabs.Slab, error) {
	var results []slabs.Slab
		slabBatch := &pgx.Batch{}
		for range limit {
			slabBatch.Queue(`UPDATE slabs
				SET last_repair_attempt = NOW()
				WHERE id = (
					SELECT slabs.id
					FROM slabs
					INNER JOIN slab_sectors ON slabs.id = slab_sectors.slab_id
					INNER JOIN sectors ON slab_sectors.sector_id = sectors.id
					LEFT JOIN contract_sectors_map csm ON sectors.contract_sectors_map_id = csm.id
					LEFT JOIN contracts ON csm.contract_id = contracts.contract_id
					WHERE
						(
							-- stored on bad contract
							(sectors.contract_sectors_map_id IS NOT NULL AND contracts.good = FALSE) OR
							-- not stored on any host
							(sectors.host_id IS NULL)
						)
						AND (slabs.last_repair_attempt <= $1)
					LIMIT 1
				)
				RETURNING digest
		`, maxRepairAttempt).QueryRow(func(row pgx.Row) error {
				var slab slabs.Slab
				if err := row.Scan((*sqlHash256)(&slab.ID)); err != nil {
					return err
				}
				results = append(results, slab)
				return nil
			})
		}

		if err := tx.Tx.SendBatch(ctx, slabBatch).Close(); errors.Is(err, sql.ErrNoRows) {
			return nil
		} else if err != nil {
			return fmt.Errorf("failed to get unhealthy slabs: %w", err)
		}

		return nil
	})
	return results, err
}
```

and with a plain loop:
```
func (s *Store) UnhealthySlabs(ctx context.Context, maxRepairAttempt time.Time, limit int) ([]slabs.Slab, error) {
	var results []slabs.Slab
		for range limit {
			var slab slabs.Slab
			err := tx.QueryRow(ctx, `UPDATE slabs
				SET last_repair_attempt = NOW()
				WHERE id = (
					SELECT slabs.id
					FROM slabs
					INNER JOIN slab_sectors ON slabs.id = slab_sectors.slab_id
					INNER JOIN sectors ON slab_sectors.sector_id = sectors.id
					LEFT JOIN contract_sectors_map csm ON sectors.contract_sectors_map_id = csm.id
					LEFT JOIN contracts ON csm.contract_id = contracts.contract_id
					WHERE
						(
							-- stored on bad contract
							(sectors.contract_sectors_map_id IS NOT NULL AND contracts.good = FALSE) OR
							-- not stored on any host
							(sectors.host_id IS NULL)
						)
						AND (slabs.last_repair_attempt <= $1)
					LIMIT 1
				)
				RETURNING digest
		`, maxRepairAttempt).Scan((*sqlHash256)(&slab.ID))
				break
			} else if err != nil {
				return fmt.Errorf("failed to query unhealthy slabs: %w", err)
			}
			results = append(results, slab)
		}
		return nil
	})
	return results, err
}
```

and both had a similar result:
```
BenchmarkUnhealthySlabs/10-4                   2     718282860 ns/op
BenchmarkUnhealthySlabs/100-4                  1    7373388605 ns/op
BenchmarkUnhealthySlabs/1000
SIGQUIT: quit
```

n8mgr:
> Do we think a materialized view is warranted here to improve performance?

A materialized view comes with its own problems (now you need to refresh it). It's never the answer unless retrieving old data is fine. In this case, it is not.

The best solution to the performance issues is most likely modifying the schema to simplify the query and reduce LEFT joins.

persist/postgres/sectors.go
ChrisSchinnerl: It was the same as your plain loop but I don't have it anymore. Running it again I get similar results to yours which probably means I messed something up in the benchmark.

We probably need to revisit the query/schema to speed things up like Nate said but we could merge this with the plain loop for now to get the changes to the interface, the testing and benchmarks in.

chris124567: Agree :+1:

persist/postgres/sectors.go
```diff
@@ -458,48 +458,52 @@ func (s *Store) UnpinnedSectors(ctx context.Context, hostKey types.PublicKey, li
...
```
peterjan: We merged a PR that adds `store.Slab` so that should enable you to return `[]SlabID` here and fetch the slab right before passing it for migration, like here https://github.com/SiaFoundation/indexd/pull/267/files#diff-7716dfaf14f62cdc8291e3287d61ec72dcc04bb2b8d12ef819106924eadea6eaR60

ChrisSchinnerl (dismissed):

ChrisSchinnerl (commented):
Lgtm apart from PJ's remark

peterjan (approved):

ChrisSchinnerl (approved):

Issue #203, opened 2025-06-18
SDK: Wrap the `io.Writer` in `Download` function in a buffer


ChrisSchinnerl:
One thing that killed download performance for us in `renterd` was `Recover` writing directly to the `http.ResponseWriter`. This might not seem like a big deal at first, but `stripedJoin` actually writes 64 bytes at a time to the writer. 

A good way to verify the impact might be a custom writer in a test that counts the number of times `Write` has been called.

PR #260, opened 2025-07-21
SDK: Wrap the io.Writer in Download function in a buffer

chris124567:
#203 

Before:
```
    client_test.go:88: Downloaded: 1048576 bytes, Write calls: 16390
```

After:
```
    client_test.go:88: Downloaded: 1048576 bytes, Write calls: 16
```

Not sure if we actually want to keep the test since it isn't testing anything that `TestRoundtrip` isn't.

### DIFF
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -1,6 +1,7 @@
 package sdk
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"errors"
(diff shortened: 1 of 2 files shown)


peterjan (approved):
LGTM, the default buffer size of 4kb is probably fine also. I'd get rid of the test personally, or extend `TestRoundtrip`.

PR #261, opened 2025-07-22
Add `store.Slab`

peterjan:
```
BenchmarkSlabs/Slab
BenchmarkSlabs/Slab-10                       100           1894078 ns/op        22144.31 MB/s
BenchmarkSlabs/PinnedSlab
BenchmarkSlabs/PinnedSlab-10                 100            828057 ns/op        50652.38 MB/s
```

### DIFF
--- a/persist/postgres/sectors_test.go
+++ b/persist/postgres/sectors_test.go
@@ -1130,6 +1130,22 @@ func BenchmarkSlabs(b *testing.B) {
 		}
 	})
 
+	b.Run("PinnedSlab", func(b *testing.B) {
+		id, err := store.PinSlab(context.Background(), proto.Account{1}, time.Time{}, newSlab())
+		if err != nil {
+			b.Fatal(err)
+		}
+
+		b.SetBytes(slabSize)
+		b.ResetTimer()
+		for b.Loop() {
+			_, err := store.PinnedSlab(context.Background(), id)
+			if err != nil {
+				b.Fatal(err)
+			}
+		}
+	})
+
 	// fetch 40MiB of data at a time
 	b.Run("Slabs-40MiB", func(b *testing.B) {
 		runSlabsBenchmark(b, 1)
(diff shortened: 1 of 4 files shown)


persist/postgres/slabs_test.go
```diff
@@ -29,6 +29,70 @@ func TestSlab(t *testing.T) {
...
+	for _, host := range hosts {
+		root := frand.Entropy256()
+		params.Sectors = append(params.Sectors, slabs.SectorPinParams{
+			Root:    root,
+			HostKey: host,
+		})
+		expectedSectors = append(expectedSectors, slabs.Sector{
+			Root:       root,
+			HostKey:    &host,
+			ContractID: nil, // not pinned to a contract
```
ChrisSchinnerl, line 48: Considering that the only difference between a pinned slab and a slab is that the slab can have a contract id and a pinnedAt timestamp I'd make sure the test covers those edge cases as well 😅 For the latter by at least checking it's not zero.

persist/postgres/sectors_test.go
```diff
@@ -1130,6 +1130,22 @@ func BenchmarkSlabs(b *testing.B) {
 		}
 	})
 
+	b.Run("PinnedSlab", func(b *testing.B) {
```
ChrisSchinnerl, line 1133: Considering that a pinned slab is a subset of a slab right now, how about making a pinned slab a conversion method on the slab rather than a separate database method? Sure, a pinned slab requires one less join but the query is pretty fast.

cc @n8maninger

peterjan: I considered that actually but I don't like conversion methods in general.

edit: just remembered there's also `slabs.ErrUnrecoverable` which `PinnedSlab` returns if it's unrecoverable, so if we want to keep that the method would have to potentially return an error

edit2: avoiding the left joins is probably worth the trouble of adding a separate method on the store imo

ChrisSchinnerl (approved):

n8mgr (approved):

PR #262, opened 2025-07-23
Use `testutils.NewCluster` in integration tests

peterjan:
This PR updates our integration tests to make use of our `Cluster` type where possible.

### DIFF
--- a/api/admin/admin_test.go
+++ b/api/admin/admin_test.go
@@ -12,6 +12,7 @@ import (
 	"go.sia.tech/coreutils/chain"
 	"go.sia.tech/coreutils/testutil"
 	"go.sia.tech/coreutils/wallet"
+	"go.sia.tech/indexd/accounts"
 	"go.sia.tech/indexd/api"
 	"go.sia.tech/indexd/api/admin"
 	"go.sia.tech/indexd/contracts"
@@ -24,39 +25,45 @@ import (
 )
 
 func TestAccountsAPI(t *testing.T) {
-	cluster := testutils.NewCluster(t)
-	indexer := cluster.Indexer
+	c := testutils.NewConsensusNode(t, zap.NewNop())
+	indexer := testutils.NewIndexer(t, c, zap.NewNop())
+
+	var accs []types.PublicKey
+	for range 10 {
+		accs = append(accs, types.GeneratePrivateKey().PublicKey())
+		err := indexer.AccountsAdd(context.Background(), accs[len(accs)-1])
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	err := indexer.AccountsAdd(context.Background(), accs[len(accs)-1])
+	if err == nil || !strings.Contains(err.Error(), accounts.ErrExists.Error()) {
+		t.Fatal("expected ErrExists", err)
+	}
 
 	accounts, err := indexer.Accounts(context.Background())
 	if err != nil {
 		t.Fatal(err)
-	} else if len(accounts) != 2 {
-		t.Fatal("unexpected accounts", len(accounts))
-	}
-
-	acc := types.GeneratePrivateKey()
-	err = indexer.AccountsAdd(context.Background(), acc.PublicKey())
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	accounts, err = indexer.Accounts(context.Background(), api.WithOffset(2), api.WithLimit(1))
-	if err != nil {
-		t.Fatal(err)
-	} else if len(accounts) != 1 {
-		t.Fatal("unexpected accounts", len(accounts))
-	} else if accounts[0] != acc.PublicKey() {
-		t.Fatal("unexpected account", accounts[0])
-	}
-
-	accounts, err = indexer.Accounts(context.Background())
-	if err != nil {
-		t.Fatal(err)
-	} else if len(accounts) != 3 {
+	} else if !reflect.DeepEqual(accs, accounts) {
 		t.Fatal("unexpected accounts", accounts)
 	}
 
-	for _, acc := range accounts {
+	accounts, err = indexer.Accounts(context.Background(), api.WithOffset(7), api.WithLimit(2))
+	if err != nil {
+		t.Fatal(err)
+	} else if !reflect.DeepEqual(accs[7:9], accounts) {
+		t.Fatal("unexpected accounts", accounts)
+	}
+
+	accounts, err = indexer.Accounts(context.Background(), api.WithOffset(10), api.WithLimit(2))
+	if err != nil {
+		t.Fatal(err)
+	} else if len(accounts) != 0 {
+		t.Fatal("unexpected accounts", accounts)
+	}
+
+	for _, acc := range accs {
 		err = indexer.AccountsDelete(context.Background(), acc)
 		if err != nil {
 			t.Fatal(err)
--- a/api/admin/admin_test.go
+++ b/api/admin/admin_test.go
@@ -207,7 +207,10 @@ func TestContractsAPI(t *testing.T) {
 	}
 }
 func TestExplorerAPI(t *testing.T) {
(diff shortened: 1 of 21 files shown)


api/admin/admin_test.go
```diff
@@ -26,45 +24,39 @@ import (
 )
 
 func TestAccountsAPI(t *testing.T) {
-	c := testutils.NewConsensusNode(t, zap.NewNop())
-	indexer := testutils.NewIndexer(t, c, zap.NewNop())
+	cluster := testutils.NewCluster(t)
```
ChrisSchinnerl: The cluster can be nice for tests where you create hosts, connect to them and fund them but it's overkill for this test. It only gets rid of a for-loop here at the cost of becoming harder to follow since some accounts are created outside the test.

api/admin/admin_test.go
```diff
@@ -111,22 +103,10 @@ func TestAccountsAPI(t *testing.T) {
 func TestContractsAPI(t *testing.T) {
 	// create cluster with one host
 	logger := newTestLogger(false)
-	c := testutils.NewConsensusNode(t, logger)
```
ChrisSchinnerl, line 114: I think this is a nice example of a test to use it for. When we don't care about the details. Just the fact that we have a funded, announced host and everyone being funded.

api/admin/admin_test.go
```diff
@@ -221,17 +201,13 @@ func TestContractsAPI(t *testing.T) {
...
 		t.Fatal(err)
 	} else if len(contracts) != 1 {
-		t.Fatal("expected 1 contract", len(contracts))
+		t.Fatal("expected 1 contract, got", len(contracts))
 	} else if contracts[0].RenewedFrom != contract.ID {
 		t.Fatal("expected contract to be renewed", contracts[0].RenewedFrom, contract.ID)
 	}
 }
-
 func TestExplorerAPI(t *testing.T) {
```
ChrisSchinnerl, line 216: I don't think we need the cluster here.

api/admin/admin_test.go
```diff
@@ -240,9 +216,6 @@ func TestExplorerAPI(t *testing.T) {
 }
 
 func TestSyncerAPI(t *testing.T) {
-	c := testutils.NewConsensusNode(t, zap.NewNop())
-	indexer := testutils.NewIndexer(t, c, zap.NewNop())
```
ChrisSchinnerl: Same

api/admin/admin_test.go
```diff
@@ -253,16 +226,13 @@ func TestSyncerAPI(t *testing.T) {
...
 	defer s.Close()
 
-	if err := indexer.SyncerConnect(s.Addr()); err != nil {
+	if err := testutils.NewCluster(t).Indexer.SyncerConnect(s.Addr()); err != nil {
 		t.Fatal(err)
 	}
 }
 
 func TestTxpoolAPI(t *testing.T) {
-	c := testutils.NewConsensusNode(t, zap.NewNop())
```
ChrisSchinnerl: Same

api/admin/admin_test.go
```diff
@@ -505,22 +466,14 @@ func TestWalletAPI(t *testing.T) {
 		t.Fatalf("expected miner payout, %+v", events[0])
 	}
 
-	// assert wallet is empty
-	res, err := indexer.Wallet(context.Background())
-	if err != nil {
-		t.Fatal(err)
-	} else if !res.Confirmed.Add(res.Unconfirmed).IsZero() {
-		t.Fatal("expected wallet to be empty")
-	}
```
ChrisSchinnerl: Similar to the first test. We are just testing a wallet and want maximum control but adding the cluster makes this harder to follow.

api/admin/admin_test.go
```diff
@@ -417,8 +378,8 @@ func TestHostsAPI(t *testing.T) {
 }
 
 func TestSettingsAPI(t *testing.T) {
```
ChrisSchinnerl: Same

client/client.go
```diff
@@ -206,10 +206,17 @@ func (c *HostClient) RenewContract(ctx context.Context, settings proto.HostSetti
 		// 1. Contracts drain over time if they contain more funds than needed
 		// 2. Renewals are very "cheap" since no party needs to lock away
 		//    additional funds. Only the fees need to be paid.
+		collateral := contract.Revision.TotalCollateral
```
ChrisSchinnerl: unrelated change snuck in?

cmd/indexd/run.go
```diff
@@ -120,6 +122,12 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
 	}
 	defer contracts.Close()
 
+	slabs, err := slabs.NewManager(am, hm, store, dialer, alerts.NewManager(), types.GeneratePrivateKey(), types.GeneratePrivateKey())
```
ChrisSchinnerl: not making use of the accounts you pass in

contracts/form.go
```diff
@@ -78,14 +78,25 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
 	// helpers for CIDR check
 	usedCidrs := make(map[string]types.PublicKey)
 	addHost := func(host hosts.Host) {
-		for _, cidr := range host.Networks {
-			usedCidrs[cidr.IP.String()] = host.PublicKey
+		for _, network := range host.Networks {
+			// in testing the host's IP is often unspecified, to ensure we don't
```
ChrisSchinnerl: I'd make this an option on the manager that we set in testing. Such as `DisableCIDRChecks`. 
To avoid weird things happening in production if we ever come across a host with unspecified IP here even though it shouldn't happen.

ChrisSchinnerl (changes requested):

peterjan:
@ChrisSchinnerl I just wanted to get some misc changes in and gave the PR a name. I reverted the unrelated changes to focus this PR on the cluster. I'm not sure what's up with the diff though, guess merging `store.Slab` will clear it up for us.

n8mgr (dismissed):
Rebase this on master with the reverts squashed

client/client.go
```diff
@@ -183,8 +183,14 @@ func (c *HostClient) ReadSector(ctx context.Context, hostPrices proto.HostPrices
 func (c *HostClient) RefreshContract(ctx context.Context, settings proto.HostSettings, params proto.RPCRefreshContractParams) (rhp.RPCRefreshContractResult, error) {
 	var res rhp.RPCRefreshContractResult
 	if err := c.withRevision(ctx, params.ContractID, func(contract rhp.ContractRevision) (_ rhp.ContractRevision, err error) {
-		res, err = rhp.RPCRefreshContract(ctx, c.client, c.cm, c.signer, c.cm.TipState(), settings.Prices, contract.Revision, params)
-		return res.Contract, err
+		res, err = rhp.RPCRefreshContract(ctx, c.client, c.cm, c.signer, c.cm.TipState(), settings.Prices, contract.Revision, proto.RPCRefreshContractParams{
+			Allowance:  contract.Revision.RenterOutput.Value,
```
n8mgr: Be careful with merges. This is reintroducing a bug that was fixed in #259

contracts/form.go
```diff
@@ -78,12 +78,20 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
 	// helpers for CIDR check
 	usedCidrs := make(map[string]types.PublicKey)
 	addHost := func(host hosts.Host) {
-		for _, cidr := range host.Networks {
-			usedCidrs[cidr.IP.String()] = host.PublicKey
+		if cm.disableCIDRChecks {
```
ChrisSchinnerl, line 81: (nit) I'd leave a comment here about why the public key is added instead of a network.

ChrisSchinnerl (approved):

contracts/form.go
peterjan: I'll take that along in another PR so you don't need to reapprove

n8mgr (approved):

Issue #247, opened 2025-07-15
Extend `contracts.Contract`


peterjan:
Ensure `contracts.Contract` has the following properties
- contractHeightStart
- contractHeightEnd
- proofWindowHeightStart
- proofWindowHeightEnd
- proofHeight
- revisionHeight

peterjan:
| Original Field                | Status        | New / Existing Name    |
|------------------------------|---------------|-------------------------|
| `contractHeightStart`        | Missing       | `formationHeight`       |
| `contractHeightEnd`          | Exists        | `expirationHeight`      |
| `proofWindowHeightStart`     | Exists        | `proofHeight`           |
| `proofWindowHeightEnd`       | Same as above | `expirationHeight`      |
| `proofHeight`                | Exists        | `proofHeight`           |
| `revisionHeight`             | Missing       | `revisionHeight`        |

ChrisSchinnerl:
Let's not add things just for the sake of having the same fields or endpoints as `renterd`, `hostd` etc. We should carefully reevaluate everything we had in `renterd` and decide whether it makes sense to also have that in `indexd`. Especially things that were exclusively added just to be displayed in the UI.

PR #265, opened 2025-07-23
Add contract formation height to `contracts.Contract`

chris124567:
https://github.com/SiaFoundation/indexd/issues/247


ChrisSchinnerl (commented):
This looks good. My only nit would be a different name for the height since it is easy to confuse with the formation time since those are not the same thing. Maybe activation height would be more accurate since it's also the height at which the contract status changes to "active".

However, I'd like to hold off on merging this since it's a field that serves no purpose apart from being displayed in the UI and we should try to reevaluate any legacy UI features rather than just migrating them over to indexd.

ChrisSchinnerl:
Closing this for now since we'll probably not need it. We can reopen it if that changes.

PR #267, opened 2025-07-24
Add slab manager

peterjan:


cmd/indexd/run.go
```diff
@@ -222,6 +231,15 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
 	return nil
 }
 
+func deriveKey(key types.PrivateKey, purpose string) types.PrivateKey {
+	seed := blake2b.Sum256(append(key[:], []byte(purpose)...))
```
ChrisSchinnerl: We might as well start using a standardized key derivation function like `hkdf` instead of rolling our own.
I'd also add a unit test to prevent regressions.

e.g.

```go
func deriveKey(key types.PrivateKey, purpose string) types.PrivateKey {
	seed := make([]byte, 32)
        defer clear(seed)
	hkdf := hkdf.New(blake2b.New256, key[:], []byte(purpose), nil)
	if err := io.ReadFull(hkdf, seed); err != nil {
          panic(err) // never happens
        }
	return types.NewPrivateKeyFromSeed(seed)
}
```

internal/testutils/indexer.go
```diff
@@ -239,6 +250,15 @@ func closeWithTimeout(closeFn func() error) error {
 	return closeFn()
 }
 
+func deriveKey(key types.PrivateKey, purpose string) types.PrivateKey {
```
ChrisSchinnerl: we can probably get rid of this and instead just generate random keys in testing to avoid the duplicate function

ChrisSchinnerl (changes requested):

internal/testutils/indexer.go
peterjan: Moved it to `internal/utils`, figured I might as well be the first. Even in testing it's probably useful to derive the keys and avoid a bug we have to search for because the keys are random. Thinking about `cluster.Reboot` stuff etc.

internal/utils/keys_test.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"bytes"
+	"crypto/ed25519"
+	"encoding/hex"
+	"testing"
+)
+
+var (
+	skHex      = "b3383d90bf82c4c9bfa445a9d6090977dc9333d302309bd002ba9407c42674b4581010a63b0682cc447cd970fdb8b7d634e1e6fec0631a310e53069c7e1b61d9"
+	derivedHex = "5d83d7281a0f5741684c57f95689b2710e1f2a1b32d67a6c6c036c18b61380467a785e0fc02b806005ec2a56c0bd5648539e2ec3093cbd74a0af7ac52ce45912"
```
ChrisSchinnerl: (nit) can be inlined

ChrisSchinnerl (dismissed):

internal/utils/keys.go
```diff
@@ -0,0 +1,20 @@
+package utils
```
n8mgr: 1. This can't be in `internal` because it's used in `cmd`.
2. Generic `utils` packages make it difficult to remember what is in them. Throw it in a dedicated `keys` package, a `crypto` package, or even just the `slabs` package.

peterjan: I'll never learn, one day I'll write a static analyzer that prevents it

ChrisSchinnerl (approved):

n8mgr (approved):

Issue #151, opened 2025-05-07
Add protection against malicious/incorrect slabs during migration


ChrisSchinnerl:
Since users can pin any slab, we need to be careful when reconstructing sectors during migration. We need to make sure the reconstructed sector's root matches the root of the sector we actually want to repair. 

If not, the slab was constructed incorrectly and can never be repaired. We should think about a way to mitigate this. First thing that comes to mind is marking the slab as not repairable and skipping it.

PR #268, opened 2025-07-24
Make uploadShards fail if the root doesn't match the expected root

ChrisSchinnerl:
Related to https://github.com/SiaFoundation/indexd/issues/151

### DIFF
--- a/slabs/uploads.go
+++ b/slabs/uploads.go
@@ -15,7 +15,10 @@ import (
 	"lukechampine.com/frand"
 )
 
-var errNotEnoughHosts = errors.New("not enough hosts")
+var (
+	errNotEnoughHosts = errors.New("not enough hosts")
+	errRootMismatch   = errors.New("sector root of shard doesn't match expected root")
+)
 
 type (
 	// Shard represents a sector present on a host.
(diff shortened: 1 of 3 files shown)


slabs/uploads.go
```diff
@@ -96,6 +103,11 @@ func (m *SlabManager) uploadShards(ctx context.Context, shards [][]byte, goodHos
 			goto nextCandidate
 		}
 
+		if root != slab.Sectors[i].Root {
+			uploadErr = fmt.Errorf("uploaded %d out of %d shards: %w", len(uploaded), len(shards), errRootMismatch)
```
peterjan, line 107: I think it's fine to log this and `goto nextCandidate` - right? I'm not sure when this would happen but if it does it's reasonably safe to assume it's the host's fault.

ChrisSchinnerl: No, `RPCWriteSector` would have failed if the host returns a root that doesn't match the actual data. We also know that according to the Slab metadata we downloaded the right data. So at this point the slab metadata is just wrong and there isn't really anything we can do for that slab anymore.

peterjan (approved):

n8mgr:
~~This is already done in the RPC https://github.com/SiaFoundation/coreutils/blob/0ce26cea9f6e0f4ceea2f190b4b8048b9052969e/rhp/v4/rpc.go#L351~~

I see what you're trying to do now

n8mgr (approved):

PR #276, opened 2025-07-25
Form contracts with full hosts

### LINKED ISSUE #275: Form contracts with full hosts
Right now when a host has a full contract that contract becomes !GoodForUpload. That means is opens up a slot for another contract to be formed assuming we have fewer than `wanted` contracts. However, the full host isn't guaranteed to have a new contract formed with it.

To make sure we don't punish good hosts like that and instead stick with them, the code should be changed to always form a new contract with a full host and to prioritize full hosts over random candidates.

ChrisSchinnerl:
Fixes https://github.com/SiaFoundation/indexd/issues/275


contracts/form.go
```diff
@@ -156,14 +171,20 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
 	// randomize their order to avoid preferring any host
 	cm.shuffle(len(candidates), func(i, j int) { candidates[i], candidates[j] = candidates[j], candidates[i] })
 
+	// move full hosts to the front of the list
+	sort.SliceStable(candidates, func(i, j int) bool {
+		return isFull[candidates[i].PublicKey] && !isFull[candidates[j].PublicKey]
+	})
+
+	// we form contracts with all full hosts and until we reach the wanted
```
peterjan, line 178: `nit` comment is technically not correct right? we form contracts with all full hosts regardless of `wanted` 
`nit` considering you renamed `wanted` to `min` I would get rid of the wanted variable all together, it's two terms that more or less mean the same thing but with a slight nuance

peterjan (dismissed):

contracts/form.go
ChrisSchinnerl: I reverted it back to 1 variable but I think the comment is correct since it says `and` instead of `either... or`.

n8mgr (approved):

peterjan (approved):

Issue #271, opened 2025-07-24
Add host dialer impl to SDK


n8mgr:
The SDK currently does not support connecting to hosts. Implement the HostDialer interface and add it to the SDK. It doesn't need to be exposed. This should be fairly simple:

- Retrieve usable hosts from app API
- Lazy open connections when `Read` or `Write` is called
- Cache the open connections for future calls

chris124567:
```
package sdk

import (
    "bytes"
    "context"
    "errors"
    "fmt"

    proto "go.sia.tech/core/rhp/v4"
    "go.sia.tech/core/types"
    "go.sia.tech/coreutils/rhp/v4"
    "go.sia.tech/coreutils/rhp/v4/quic"
    "go.sia.tech/coreutils/rhp/v4/siamux"
    "go.sia.tech/indexd/hosts"
)

var _ HostDialer = (*Dialer)(nil)

type Dialer struct {
    c      AppClient
    appKey types.PrivateKey

    hosts map[types.PublicKey]hosts.Host
    conns map[types.PublicKey]rhp.TransportClient
}

func newDialer(c AppClient, appKey types.PrivateKey) *Dialer {
    return &Dialer{
        c:      c,
        appKey: appKey,

        hosts: make(map[types.PublicKey]hosts.Host),
    }
}

// Hosts returns the public keys of all hosts that are available for
// upload or download.
func (d *Dialer) Hosts(ctx context.Context) ([]types.PublicKey, error) {
    hosts, err := d.c.Hosts(ctx)
    if err != nil {
        return nil, err
    }

    pks := make([]types.PublicKey, 0, len(hosts))
    for _, host := range hosts {
        pks = append(pks, host.PublicKey)
        d.hosts[host.PublicKey] = host
    }
    return pks, nil
}

func (d *Dialer) dialHost(ctx context.Context, hostKey types.PublicKey, reuse bool) (rhp.TransportClient, error) {
    if tc, ok := d.conns[hostKey]; ok && reuse {
        // we have an existing connection
        return tc, nil
    }

    h, ok := d.hosts[hostKey]
    if !ok {
        return nil, fmt.Errorf("we haven't seen host")
    }

    for _, addr := range h.Addresses {
        if addr.Protocol == siamux.Protocol {
            tc, err := siamux.Dial(ctx, addr.Address, hostKey)
            if err != nil {
                return nil, fmt.Errorf("failed to dial host over siamux: %w", err)
            }
[...]

PR #280, opened 2025-07-25
Add host dialer

chris124567:
Implement the HostDialer interface in the SDK and add an associated test.

- Cache results of Hosts() to get net addresses of hosts
- Cache open connections with hosts
- Cache host settings until they expire
- Retry if we get transport related failures due to connections randomly dropping

https://github.com/SiaFoundation/indexd/issues/271

### DIFF
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -112,7 +112,7 @@ func (s *SDK) uploadSlab(ctx context.Context, encryptionKey [32]byte, shards [][
 	}
 
 	var hostsMu sync.Mutex
-	hosts := s.dialer.Hosts()
+	hosts := shuffle(s.dialer.Hosts())
 	if len(hosts) < len(shards) {
 		return slabs.SlabPinParams{}, fmt.Errorf("not enough hosts available: %d, required: %d", len(hosts), len(shards))
 	}
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -48,7 +48,7 @@ type (
 	HostDialer interface {
 		// Hosts returns the public keys of all hosts that are available for
 		// upload or download.
-		Hosts(context.Context) ([]types.PublicKey, error)
+		Hosts() []types.PublicKey
 
 		// WriteSector writes a sector to the host identified by the public key.
 		WriteSector(context.Context, types.PublicKey, *[proto4.SectorSize]byte) (types.Hash256, error)
@@ -112,10 +112,7 @@ func (s *SDK) uploadSlab(ctx context.Context, encryptionKey [32]byte, shards [][
 	}
 
 	var hostsMu sync.Mutex
-	hosts, err := s.dialer.Hosts(ctx)
-	if err != nil {
-		return slabs.SlabPinParams{}, fmt.Errorf("failed to get usable hosts: %w", err)
-	}
+	hosts := s.dialer.Hosts()
 	if len(hosts) < len(shards) {
 		return slabs.SlabPinParams{}, fmt.Errorf("not enough hosts available: %d, required: %d", len(hosts), len(shards))
 	}
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -14,7 +14,9 @@ import (
 	"github.com/klauspost/reedsolomon"
 	proto4 "go.sia.tech/core/rhp/v4"
 	"go.sia.tech/core/types"
+	"go.sia.tech/indexd/api"
 	"go.sia.tech/indexd/api/app"
+	"go.sia.tech/indexd/hosts"
 	"go.sia.tech/indexd/slabs"
 	"golang.org/x/crypto/chacha20"
 	"lukechampine.com/frand"
@@ -38,6 +40,7 @@ type (
 		PinSlab(context.Context, slabs.SlabPinParams) (slabs.SlabID, error)
 		UnpinSlab(context.Context, slabs.SlabID) error
 
+		Hosts(context.Context, ...api.URLQueryParameterOption) ([]hosts.Host, error)
 		Slab(context.Context, slabs.SlabID) (slabs.PinnedSlab, error)
 	}
 
@@ -45,7 +48,7 @@ type (
 	HostDialer interface {
 		// Hosts returns the public keys of all hosts that are available for
 		// upload or download.
-		Hosts() []types.PublicKey
+		Hosts(context.Context) ([]types.PublicKey, error)
 
 		// WriteSector writes a sector to the host identified by the public key.
 		WriteSector(context.Context, types.PublicKey, *[proto4.SectorSize]byte) (types.Hash256, error)
@@ -109,7 +112,10 @@ func (s *SDK) uploadSlab(ctx context.Context, encryptionKey [32]byte, shards [][
 	}
 
 	var hostsMu sync.Mutex
-	hosts := shuffle(s.dialer.Hosts())
+	hosts, err := s.dialer.Hosts(ctx)
+	if err != nil {
+		return slabs.SlabPinParams{}, fmt.Errorf("failed to get usable hosts: %w", err)
+	}
 	if len(hosts) < len(shards) {
 		return slabs.SlabPinParams{}, fmt.Errorf("not enough hosts available: %d, required: %d", len(hosts), len(shards))
 	}
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -40,7 +40,7 @@ type (
 		PinSlab(context.Context, slabs.SlabPinParams) (slabs.SlabID, error)
 		UnpinSlab(context.Context, slabs.SlabID) error
 
-		Hosts(context.Context, ...api.URLQueryParameterOption) ([]hosts.Host, error)
+		Hosts(context.Context, ...api.URLQueryParameterOption) ([]hosts.HostInfo, error)
 		Slab(context.Context, slabs.SlabID) (slabs.PinnedSlab, error)
 	}
 
--- a/sdk/dialer.go
+++ b/sdk/dialer.go
@@ -32,12 +32,12 @@ type connEntry struct {
 
 // Dialer implements the HostDialer interface.
 type Dialer struct {
-	mu  sync.Mutex
 	log *zap.Logger
 
 	c      AppClient
 	appKey types.PrivateKey
 
+	mu             sync.Mutex
 	tg             *threadgroup.ThreadGroup
 	addrs          map[types.PublicKey][]chain.NetAddress
 	conns          map[types.PublicKey]*connEntry
@@ -57,30 +57,11 @@ func NewDialer(c AppClient, appKey types.PrivateKey, log *zap.Logger) (*Dialer, 
 		conns:          make(map[types.PublicKey]*connEntry),
 		cachedSettings: make(map[types.PublicKey]proto.HostSettings),
 	}
-
-	// Run immediately
-	if err := d.updateHosts(context.Background()); err != nil {
+	if err := d.initHosts(); err != nil {
 		return nil, err
 	}
-	go d.refreshHosts()
 
 	return d, nil
-}
-
-func (d *Dialer) refreshHosts() {
-	ticker := time.NewTicker(10 * time.Minute)
-	defer ticker.Stop()
-
-	for {
-		select {
-		case <-ticker.C:
-			if err := d.updateHosts(context.Background()); err != nil {
-				d.log.Warn("Failed to refresh hosts list", zap.Error(err))
-			}
-		case <-d.tg.Done():
-			return
-		}
-	}
 }
 
 // Close closes all the open connections on the dialer.
@@ -101,6 +82,43 @@ func (d *Dialer) Close() {
 		entry.mu.Unlock()
 	}
 	clear(d.conns)
+}
+
+func (d *Dialer) initHosts() error {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	// init hosts
+	err := d.updateHosts(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to refresh hosts list: %w", err)
+	}
+
+	// refresh in bg thread
+	ctx, cancel, err = d.tg.AddContext(context.Background())
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		defer cancel()
+
+		ticker := time.NewTicker(10 * time.Minute)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ticker.C:
+				if err := d.updateHosts(context.Background()); err != nil {
+					d.log.Warn("Failed to refresh hosts list", zap.Error(err))
+				}
+			case <-d.tg.Done():
+				return
+			}
+		}
+	}()
+
+	return nil
 }
 
 func (d *Dialer) updateHosts(ctx context.Context) error {
--- a/sdk/dialer.go
+++ b/sdk/dialer.go
@@ -27,7 +27,9 @@ var _ HostDialer = (*Dialer)(nil)
 type connEntry struct {
 	mu   sync.Mutex
 	dial chan struct{} // signals dial completion
-	tc   rhp.TransportClient
+
+	hostKey types.PublicKey
+	tc      rhp.TransportClient
 }
 
 // Dialer implements the HostDialer interface.
@@ -69,19 +71,23 @@ func (d *Dialer) Close() {
 	d.tg.Stop()
 
 	d.mu.Lock()
-	defer d.mu.Unlock()
+	conns := slices.Collect(maps.Values(d.conns))
+	d.mu.Unlock()
 
-	for hostKey, entry := range d.conns {
+	for _, entry := range conns {
 		entry.mu.Lock()
 		if entry.tc != nil {
 			if err := entry.tc.Close(); err != nil {
-				d.log.Debug("Failed to close connection", zap.Stringer("pk", hostKey), zap.Error(err))
+				d.log.Debug("Failed to close connection", zap.Stringer("pk", entry.hostKey), zap.Error(err))
 			}
 		}
 		entry.tc = nil
 		entry.mu.Unlock()
 	}
+
+	d.mu.Lock()
 	clear(d.conns)
+	d.mu.Unlock()
 }
 
 func (d *Dialer) initHosts() error {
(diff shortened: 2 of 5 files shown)


sdk/dialer.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+	}
+	clear(d.conns)
+
+	return nil
+}
+
+// Hosts returns the public keys of all hosts that are available for
+// upload or download.
+func (d *Dialer) Hosts(ctx context.Context) ([]types.PublicKey, error) {
```
n8mgr: We don't want to refetch this list every time we call `Hosts()`. It should be cached, potentially in the constructor.

chris124567: The list will change if contracts are formed with new hosts, so should we periodically refresh it too?

n8mgr: The list of hosts shouldn't change that often. Probably cache it for 10 minutes or so.

sdk/dialer.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// Close closes all the open connections on the dialer.
+func (d *Dialer) Close() error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	for _, conn := range d.conns {
+		if err := conn.Close(); err != nil {
+			return err
```
peterjan: We probably want to collect errors here and continue closing the rest.

sdk/dialer.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	clear(d.conns)
+
+	return nil
+}
+
+// Hosts returns the public keys of all hosts that are available for
+// upload or download.
+func (d *Dialer) Hosts(ctx context.Context) ([]types.PublicKey, error) {
+	d.mu.Lock()
```
peterjan: You are acquiring the lock here for quite a long time potentially. Looking at what the mutex is used for that's not a good idea because the dialer would freeze during calls to `prices` and so on.

sdk/dialer.go
```diff
@@ -0,0 +1,216 @@
...
+	for _, addr := range h {
+		if addr.Protocol == siamux.Protocol {
+			tc, err := siamux.Dial(ctx, addr.Address, hostKey)
+			if err != nil {
+				return nil, fmt.Errorf("failed to dial host over siamux: %w", err)
+			}
+			d.conns[hostKey] = tc
+			return tc, nil
+		} else if addr.Protocol == quic.Protocol {
+			tc, err := quic.Dial(ctx, addr.Address, hostKey)
```
peterjan: I haven't tried this yet but should work. I'm wondering if we want to prefer one of the other? If we support both we might as well give preference to whatever's the fastest or most reliable. Not relevant now but would be good to have some metrics on that too.

sdk/dialer.go
```diff
@@ -0,0 +1,216 @@
...
+		return fmt.Errorf("failed to redial host: %w", err)
+	}
+	d.mu.Unlock()
+
+	return fn(tc)
+}
+
+func (d *Dialer) prices(ctx context.Context, hostKey types.PublicKey) (proto.HostPrices, error) {
+	d.mu.Lock()
+	if settings, ok := d.cachedSettings[hostKey]; ok && time.Now().Before(settings.Prices.ValidUntil) {
```
peterjan: `nit` build in a 5s leeway here

sdk/dialer.go
```diff
@@ -0,0 +1,216 @@
...
+
+func (d *Dialer) dialHost(ctx context.Context, hostKey types.PublicKey, reuse bool) (rhp.TransportClient, error) {
+	if tc, ok := d.conns[hostKey]; ok && reuse {
+		// we have an existing connection
+		return tc, nil
+	}
+
+	h, ok := d.addrs[hostKey]
+	if !ok {
+		return nil, fmt.Errorf("we haven't seen host")
```
peterjan: `nit` bit of an odd error message

sdk/dialer.go
```diff
@@ -0,0 +1,216 @@
...
+// WriteSector writes a sector to the host identified by the public key.
+func (d *Dialer) WriteSector(ctx context.Context, hostKey types.PublicKey, sector *[proto.SectorSize]byte) (types.Hash256, error) {
+	prices, err := d.prices(ctx, hostKey)
+	if err != nil {
+		return types.Hash256{}, fmt.Errorf("failed to get prices: %w", err)
+	}
+
+	var result rhp.RPCWriteSectorResult
+	err = d.retry(ctx, hostKey, func(tc rhp.TransportClient) (err error) {
+		token := (&proto.Account{}).Token(d.appKey, hostKey)
```
peterjan: Probably overkill but these tokens are valid for up to 5' so we could cache them on a per host basis and save calculating the signature hash

sdk/dialer.go
```diff
@@ -0,0 +1,216 @@
...
+	"go.sia.tech/coreutils/rhp/v4/quic"
+	"go.sia.tech/coreutils/rhp/v4/siamux"
+	"go.sia.tech/indexd/api"
+)
+
+var _ HostDialer = (*Dialer)(nil)
+
+// Dialer implements the HostDialer interface.
+type Dialer struct {
+	mu sync.Mutex
```
peterjan: We are going to have way too much lock contention on this mutex as it is now. `dialHost` is executed under the lock, that means a`siamux.Dial` (or quic.Dial) to a host can essentially block the whole SDK. It's not only `dialHost` though, but that's definitely going to be the biggest issue.

I always split it up in a `Dialer` and a `HostClient`, so the dialer hands out host clients and the clients then cache the settings and the connection. It's also the client that has a `Read` and `Write` sector. I can see benefits to this approach too though, I'm not really advocating for it but on the other hand..  a dialer handing out a client would be consistent with the rest of the codebase.

peterjan (changes requested):
Pretty sweet. Still some locking issues to figure out though.

sdk/dialer.go
ChrisSchinnerl: I think that's overkill indeed.

sdk/dialer.go
```diff
@@ -0,0 +1,216 @@
...
+// upload or download.
+func (d *Dialer) Hosts(ctx context.Context) ([]types.PublicKey, error) {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	clear(d.addrs)
+	offset, limit := 0, 100
+	var pks []types.PublicKey
+	for {
+		hosts, err := d.c.Hosts(ctx, api.WithOffset(offset), api.WithLimit(limit))
```
ChrisSchinnerl, line 163: As PJ already mentioned, this is an I/O bound operation and potentially very slow. 

Ideally, fetching hosts should not interrupt uploads or downloads. Instead, as Nate already said, fetch the hosts in the constructor and then update them every 10 minutes or so in the background to avoid interrupting uploads/downloads.

sdk/dialer.go
```diff
@@ -0,0 +1,216 @@
...
+
+		offset += len(hosts)
+		if len(hosts) < limit {
+			break
+		}
+	}
+	return pks, nil
+}
+
+func (d *Dialer) dialHost(ctx context.Context, hostKey types.PublicKey, reuse bool) (rhp.TransportClient, error) {
```
ChrisSchinnerl: Similar thing here. `dialHost` is surrounded by a single lock which will degrade performance quite a bit since we can only dial hosts sequentially. 

It would be better to rely on a pattern that allows for dialing any number of hosts in parallel. e.g. by updating the `conns` field to not contain a `rhp.TransportClient` directly but a wrapper type which contains a lock and a channel. The first goroutine that dials a host will lock that entry, create a channel, unlock it and then proceed to dial the host. When the host is dialed it can lock the entry again, send 'nil' or an error down the channel, close the channel and unlock the entry. Any goroutines attempting to call the same host while a dial is in progress will just fetch the channel under the lock and then wait on it until the host is dialed.

ChrisSchinnerl (changes requested):

sdk/dialer.go
n8mgr: It’s not a huge deal if we occasionally race with multiple dialers. I’d change it to a double lock and move on over adding the complexity of waiters and passing stuff between goroutines.

```
mu.Lock
If client, ok := clients {
  mu.Unlock()
  return client, nil
}
mu.Unlock()

// dial host

mu.Lock()
defer mu.Unlock()
if existing, ok := clients; ok {
  client.Close()
  return existing, nil
}
clients[peerKey] = client
```

sdk/dialer.go
n8mgr: Just ignore any errors. They are unlikely to matter.

sdk/dialer.go
n8mgr: SiaMux is faster in our benchmarks. We should prefer it, but support both. If we want to sort the list, we should do it on the server.

sdk/dialer.go
```diff
@@ -5,151 +5,260 @@ import (
...
 		addrs:          make(map[types.PublicKey][]chain.NetAddress),
-		conns:          make(map[types.PublicKey]rhp.TransportClient),
+		conns:          make(map[types.PublicKey]*connEntry),
 		cachedSettings: make(map[types.PublicKey]proto.HostSettings),
 	}
 }
 
+// Start starts the goroutine that periodically refreshes the map of host
+// addresses
+func (d *Dialer) Start(ctx context.Context) (func(), error) {
```
n8mgr: Having to remember to call `Start` is poor devex. Stick it in the constructor. For cleanup, consider adding a threadgroup or storing a ticker.

Benefit of a global threadgroup is you can also easily wait for inflight RPCs to complete.

sdk/dialer.go
```diff
@@ -0,0 +1,329 @@
...
+func (d *Dialer) Close() error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	var errs []error
+	for _, entry := range d.conns {
+		entry.mu.Lock()
+		if entry.tc != nil {
+			if err := entry.tc.Close(); err != nil {
+				errs = append(errs, err)
```
n8mgr: No reason to collect or return these. 9 times out of 10, this will be `defer Close()`. Worst case scenario, we've just given an external developer an error string with 100+ errors that aren't actionable.

Should `log.Debug` them and move on.

n8mgr (changes requested):

sdk/client.go
```diff
@@ -38,6 +40,7 @@ type (
 		PinSlab(context.Context, slabs.SlabPinParams) (slabs.SlabID, error)
 		UnpinSlab(context.Context, slabs.SlabID) error
 
+		Hosts(context.Context, ...api.URLQueryParameterOption) ([]hosts.Host, error)
```
n8mgr: This should return `hosts.HostInfo`

n8mgr: This should return `[]hosts.HostInfo`

sdk/dialer.go
```diff
@@ -0,0 +1,323 @@
...
+			}
+		}
+		entry.tc = nil
+		entry.mu.Unlock()
+	}
+	clear(d.conns)
+}
+
+func (d *Dialer) updateHosts(ctx context.Context) error {
+	offset, limit := 0, 100
```
n8mgr, line 160: Should `d.tg.AddContext(ctx)` here

sdk/dialer.go
```diff
@@ -0,0 +1,323 @@
...
+	d.mu.Lock()
+	d.cachedSettings[hostKey] = settings
+	d.mu.Unlock()
+
+	return settings.Prices, nil
+}
+
+// WriteSector writes a sector to the host identified by the public key.
+func (d *Dialer) WriteSector(ctx context.Context, hostKey types.PublicKey, sector *[proto.SectorSize]byte) (types.Hash256, error) {
+	prices, err := d.prices(ctx, hostKey)
```
n8mgr, line 334: Should `d.tg.AddContext(ctx)` here

sdk/dialer.go
```diff
@@ -0,0 +1,323 @@
...
+	if err != nil {
+		return types.Hash256{}, fmt.Errorf("failed to write sector: %w", err)
+	}
+
+	return result.Root, nil
+}
+
+// ReadSector reads a sector from the host identified by the public key.
+func (d *Dialer) ReadSector(ctx context.Context, hostKey types.PublicKey, sectorRoot types.Hash256) (*[proto.SectorSize]byte, error) {
+	prices, err := d.prices(ctx, hostKey)
```
n8mgr, line 360: Should `d.tg.AddContext(ctx)` here

n8mgr (changes requested):

sdk/dialer.go
```diff
@@ -0,0 +1,323 @@
...
+
+type connEntry struct {
+	mu   sync.Mutex
+	dial chan struct{} // signals dial completion
+	tc   rhp.TransportClient
+}
+
+// Dialer implements the HostDialer interface.
+type Dialer struct {
+	mu  sync.Mutex
```
peterjan: `nit` the mutex should be moved down to cover the fields it protects, usually it's static fields, dependencies, config variables and then state fields below a mutex. It really does help the reader to immediately understand how the dialer works.

So something like this
```
type Dialer struct {
    appKey types.PrivateKey

    c    AppClient
    tg   *threadgroup.ThreadGroup
    log  *zap.Logger
    
    mu                    sync.Mutex
    addrs                 map[types.PublicKey][]chain.NetAddress
    conns                 map[types.PublicKey]*connEntry
    cachedSettings map[types.PublicKey]proto.HostSettings
}

File: sdk/dialer.go [hunk 18]
```diff
@@ -0,0 +1,323 @@
...
+		appKey: appKey,
+
+		tg:             threadgroup.New(),
+		addrs:          make(map[types.PublicKey][]chain.NetAddress),
+		conns:          make(map[types.PublicKey]*connEntry),
+		cachedSettings: make(map[types.PublicKey]proto.HostSettings),
+	}
+
+	// Run immediately
+	if err := d.updateHosts(context.Background()); err != nil {
```
peterjan: Passing a background context is not ideal here. Maybe something like this, which also takes into account the threadgroup.

```
func (d *Dialer) initHosts() error {
    ctx, cancel := context.WithTimeout(context.Background(), 10 * time.Second)
    defer cancel()
    
    // init hosts
    err := d.updateHosts(ctx)
    if err != nil {
        return fmt.Errorf(...)
    }
    
    // refresh in bg thread
    ctx, cancel, err := cm.tg.AddContext(context.Background())
    if err != nil {
        return err
    }
	
    go func() {
        defer cancel()
        
        ticker := time.NewTicker(10 * time.Minute)
        defer ticker.Stop()
        
        for {...}
    }()
    
    return nil
}
```

sdk/dialer.go
```diff
@@ -0,0 +1,323 @@
...
+
+// Close closes all the open connections on the dialer.
+func (d *Dialer) Close() {
+	d.tg.Stop()
+
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	for hostKey, entry := range d.conns {
+		entry.mu.Lock()
```
peterjan: Holding two locks at the same time is always a risk. We should grab the connections under lock. And then close them outside of the main mutex on the dialer. To do that though we have to make sure `d.conns` is not appended to after `Close` is called. You can do that by doing `tg.AddContext` in all exposed methods and it'll return `ErrClosed` when it's closed, preventing new connections being added.

n8mgr (dismissed):

sdk/dialer.go
```diff
@@ -0,0 +1,365 @@
...
+	"go.sia.tech/coreutils/rhp/v4/quic"
+	"go.sia.tech/coreutils/rhp/v4/siamux"
+	"go.sia.tech/coreutils/threadgroup"
+	"go.sia.tech/indexd/api"
+	"go.uber.org/zap"
+)
+
+var _ HostDialer = (*Dialer)(nil)
+
+type connEntry struct {
```
peterjan, line 27: `nit` reorder properties

sdk/dialer.go
```diff
@@ -0,0 +1,365 @@
...
+
+// Dialer implements the HostDialer interface.
+type Dialer struct {
+	log *zap.Logger
+
+	c      AppClient
+	appKey types.PrivateKey
+
+	mu             sync.Mutex
+	tg             *threadgroup.ThreadGroup
```
peterjan: `nit` the threadgroup is never reassigned and is not covered by the mutex

sdk/dialer.go
```diff
@@ -0,0 +1,365 @@
...
+			if err := entry.tc.Close(); err != nil {
+				d.log.Debug("Failed to close connection", zap.Stringer("pk", entry.hostKey), zap.Error(err))
+			}
+		}
+		entry.tc = nil
+		entry.mu.Unlock()
+	}
+
+	d.mu.Lock()
+	clear(d.conns)
```
peterjan, line 113: the threadgroup should make it safe to do this immediately after you collect them

sdk/dialer.go
```diff
@@ -0,0 +1,365 @@
...
+	for _, addr := range h {
+		if addr.Protocol == quic.Protocol {
+			tc, err = quic.Dial(ctx, addr.Address, hostKey)
+			if err == nil {
+				return tc, nil
+			}
+		}
+	}
+
+	err = errors.New("host has no supported protocols")
```
peterjan, line 275: `nit` we're overwriting the error here, technically the host might have both siamux and quic addresses, we just failed to dial it

sdk/dialer_test.go
```diff
@@ -0,0 +1,138 @@
...
+			}
+
+			if !bytes.Equal(data[:], sector[:]) {
+				t.Error("retrieved sector does not match")
+			}
+		}()
+	}
+	wg.Wait()
+
+	dialer.Close()
```
peterjan: `nit` would defer this (also in TestHostDialer)

sdk/dialer_test.go
```diff
@@ -0,0 +1,138 @@
...
+	indexer := cluster.Indexer
+
+	// add an account
+	a1 := types.GeneratePrivateKey()
+	app := indexer.App(a1)
+	err := indexer.AccountsAdd(context.Background(), a1.PublicKey())
+	if err != nil {
+		t.Fatal(err)
+	}
+	time.Sleep(3 * time.Second)
```
peterjan: `nit` adding 2 hosts should only take at most 1 second I think (also in TestHostDialer)

peterjan (dismissed):
Left some nits but looks good over all. I would probably move some logic over to `connEntry`, essentially almost all places where you grab its lock are good candidates.

peterjan (approved):

ChrisSchinnerl (approved):

PR #281, opened 2025-07-28
Update UnhealthySlabs

peterjan:
This PR updates the signature of `UnhealthySlabs` which was still returning `[]Slab` instead of the IDs. I've also slightly updated the implementation of `UnhealthySlabs` to make it safer and to ensure we are returning unique slab IDs (this was not the case). I got rid of the extra test that covered the `limit` and merged the two, updated the test a bit to get rid of the sleeps.

### DIFF
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -112,7 +112,7 @@ func (s *Store) SectorsForIntegrityCheck(ctx context.Context, hostKey types.Publ
 
 // MarkFailingSectorsLost marks sectors as lost if they have failed the
 // integrity checks >= maxChecks times.
-func (s *Store) MarkFailingSectorsLost(ctx context.Context, hostKey types.PublicKey, maxChecks int) error {
+func (s *Store) MarkFailingSectorsLost(ctx context.Context, hostKey types.PublicKey, maxChecks uint) error {
 	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
 		resp, err := tx.Exec(ctx, `
 			UPDATE sectors
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -458,8 +458,8 @@ func (s *Store) UnpinnedSectors(ctx context.Context, hostKey types.PublicKey, li
 	return roots, err
 }
 
-// UnhealthySlabs returns slabs which have at least one sector that needs to be
-// migrated to a new host and hasn't had a repair attempted since
+// UnhealthySlabs returns the ID of slabs which have at least one sector that
+// needs to be migrated to a new host and hasn't had a repair attempted since
 // 'maxRepairAttempt'. The condition for such a slab is that it either has:
 // a). a sector that is not stored on a host (host_id == null)
 // b). a sector that is stored in a bad contract (contract_id != null && contract.good = false)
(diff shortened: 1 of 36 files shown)


persist/postgres/sectors_test.go
```diff
@@ -2150,3 +2026,23 @@ func BenchmarkMigrateSector(b *testing.B) {
 		}
 	}
 }
+
+func (s *Store) addTestSlab(t testing.TB, account proto.Account, dataShards, parityShards int, hks []types.PublicKey) slabs.SlabID {
```
ChrisSchinnerl: (nit) `pinTestSlab`

Maybe instead of providing the `parityShards` use `len(hks)` as the parity shards and add 1 sector for each host. That seems a bit more predictable than the `i%len(hks)`

persist/postgres/sectors.go
```diff
@@ -458,23 +458,28 @@ func (s *Store) UnpinnedSectors(ctx context.Context, hostKey types.PublicKey, li
...
 // If a slab is found, it will have its last_repair_attempt updated to the time
 // of the call. To prevent subsequent or parallel calls from returning the same slab.
 //
 // NOTE: For the sake of scalability, we don't prioritize any slabs and instead
-// simply fetch the first ones that we can get.
+// simply fetch the first batch we find.
 func (s *Store) UnhealthySlabs(ctx context.Context, maxRepairAttempt time.Time, limit int) ([]slabs.SlabID, error) {
+	now := time.Now()
+	if maxRepairAttempt.After(now) {
+		return nil, fmt.Errorf("maxRepairAttempt must be in the past: %v", maxRepairAttempt)
```
ChrisSchinnerl, line 474: Do we need this? Seems like it might end up being annoying if we ever expose this via the API and then don't exactly agree on `Now`.

ChrisSchinnerl (changes requested):

persist/postgres/sectors.go
peterjan: Hm that's fair but it's definitely safer and we do need it because `UnhealthySlabs` is returning the same slab 10x over in my integration test. It's an NDF so I chalked it up to `NOW()` and incorrectly rounding off microseconds.

peterjan: The fact that the pattern is potentially non-exhaustive and we wouldn't be able to tell is prone to errors to begin with. I think at the very least we should have input validation on `maxRepairAttempt` and ensure that our update to the LRA timestamp excludes the slab we are returning to (somewhat) guarantee it's exhaustive.

slabs/migrations.go
```diff
@@ -52,11 +52,17 @@ func (m *SlabManager) migrateSlabs(ctx context.Context, slabs []Slab, l *zap.Log
 	}
 
 	var wg sync.WaitGroup
-	for _, slab := range slabs {
+	for _, slabID := range slabIDs {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
 
+			slab, err := m.store.Slab(ctx, slabID)
```
n8mgr: nit: probably better to grab this in the goroutine so it's concurrent.

peterjan: I grabbed it outside to avoid needlessly spinning up a new goroutine 😂 guessing the concurrency argument is better idd.

ChrisSchinnerl (approved):

n8mgr (approved):

PR #282, opened 2025-07-28
Add `HostsWithUnpinnableSectors`

ChrisSchinnerl:
Related to https://github.com/SiaFoundation/indexd/issues/278

Will eventually be used to form contracts with these hosts.

### DIFF
--- a/persist/postgres/hosts.go
+++ b/persist/postgres/hosts.go
@@ -305,11 +305,11 @@ func (s *Store) HostsWithUnpinnedSectors(ctx context.Context) ([]types.PublicKey
 			WHERE EXISTS (
 				SELECT 1
 				FROM sectors
-				WHERE sectors.host_id = hosts.id
+				WHERE sectors.host_id = hosts.id AND sectors.contract_sectors_map_id IS NULL
 			) AND NOT EXISTS (
 				SELECT 1
 				FROM contracts
-				WHERE contracts.host_id = hosts.id AND contracts.state >= $1 AND contracts.state <= $2
+				WHERE contracts.host_id = hosts.id AND (contracts.state <> $1 OR contracts.state <> $2)
 			)
 		`, contracts.ContractStatePending, contracts.ContractStateActive)
 		if err != nil {
--- a/persist/postgres/hosts.go
+++ b/persist/postgres/hosts.go
@@ -302,8 +302,11 @@ func (s *Store) HostsWithUnpinnedSectors(ctx context.Context) ([]types.PublicKey
 		rows, err := tx.Query(ctx, `
 			SELECT public_key
 			FROM hosts
-			INNER JOIN sectors ON sectors.host_id = hosts.id
-			WHERE NOT EXISTS (
+			WHERE EXISTS (
+				SELECT 1
+				FROM sectors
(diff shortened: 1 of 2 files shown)


persist/postgres/hosts_test.go
```diff
@@ -1767,6 +1836,61 @@ func BenchmarkHostsWithLostSectors(b *testing.B) {
 	}
 }
 
+func BenchmarkHostsWithUnpinnedSectors(b *testing.B) {
```
ChrisSchinnerl:
```
goos: darwin
goarch: arm64
cpu: Apple M2 Pro
BenchmarkHostsWithUnpinnedSectors
BenchmarkHostsWithUnpinnedSectors-12    	     787	   1493412 ns/op
PASS
```

persist/postgres/hosts.go
```diff
@@ -294,6 +294,41 @@ func (s *Store) BlockHosts(ctx context.Context, hks []types.PublicKey, reason st
...
+			SELECT public_key
+			FROM hosts
+			WHERE EXISTS (
+				SELECT 1
+				FROM sectors
+				WHERE sectors.host_id = hosts.id
+			) AND NOT EXISTS (
+				SELECT 1
+				FROM contracts
+				WHERE contracts.host_id = hosts.id AND contracts.state >= $1 AND contracts.state <= $2
```
n8mgr: I'd feel better if this was `contracts.state <>` instead of `>=` and `<=`

peterjan: It should probably be `contract.state <= $1` and pass in `contracts.ContractStateActive`

persist/postgres/hosts.go
```diff
@@ -294,6 +294,41 @@ func (s *Store) BlockHosts(ctx context.Context, hks []types.PublicKey, reason st
 	})
 }
 
+// HostsWithUnpinnedSectors returns a list of host public keys for hosts that
+// don't have any contracts but unpinned sectors.
+func (s *Store) HostsWithUnpinnedSectors(ctx context.Context) ([]types.PublicKey, error) {
```
peterjan: I feel the name is a bit misleading. It's not necessarily `HostsWithUnpinnedSectors` but `HostsWithSectorsAndNoActiveContracts` 😛 . Maybe `-DanginlingSectors` or `-RiskedSectors` or `PrunableSectors`.

persist/postgres/hosts_test.go
ChrisSchinnerl: Updated:

```
goos: darwin
goarch: arm64
cpu: Apple M2 Pro
BenchmarkHostsWithUnpinnedSectors
BenchmarkHostsWithUnpinnedSectors-12    	    1483	    804151 ns/op
PASS
```

persist/postgres/hosts.go
ChrisSchinnerl: Updated it to be more explicit

persist/postgres/hosts.go
ChrisSchinnerl: Changed it to `Unpinnable`. Is that acceptable? ^^

n8mgr (approved):

PR #284, opened 2025-07-29
Update price validity check

peterjan:
Checking minimum price validity shouldn't take into account the current time but check whether the price table we received from the host was valid for at least 15' when we scanned it. If we don't, a failure to scan the host every 15 minutes would result in an usable host.


n8mgr:
I might go with removing the check entirely. This feels simultaneously correct and incorrect.

n8mgr (approved):

ChrisSchinnerl (approved):

PR #285, opened 2025-07-29
Treat `accounts.ErrExists` as no-op in accounts API

peterjan:


ChrisSchinnerl (dismissed):
Alternatively we could also have the database not return an error if the account exists in the first place.

peterjan:
> Alternatively we could also have the database not return an error if the account exists in the first place.

I think internally it might still prove useful to know we are re-adding an account, but definitely when we expose it in the API it's very annoying to handle the 409 conflict if you just want to ensure the account is added.

ChrisSchinnerl (approved):

n8mgr (approved):

PR #287, opened 2025-07-29
Add `junkd` upload command

peterjan:
Very simple upload command that allows pinning slabs on `indexd`


cmd/filld/main.go
```diff
@@ -0,0 +1,250 @@
...
+					log.Error("failed to load client, timing out for 5 minutes", zap.Error(err))
+					if ok := <-waitFor(ctx, 5*time.Minute); ok {
+						continue loop
+					}
+				}
+
+				// prepare random data
+				data := make([]byte, slabSize)
+				rand.Read(data)
+				r := bytes.NewReader(data)
```
ChrisSchinnerl: `r := io.LimitReader(frand.Reader, slabSize)`

Reduces RAM usage a bit since we don't need to keep a full slab in RAM twice.

cmd/filld/main.go
```diff
@@ -0,0 +1,250 @@
...
+
+				// upload slab
+				start := time.Now()
+				slabs, err := client.Upload(ctx, r, sdk.WithRedundancy(dataShards, parityShards))
+				if err != nil {
+					log.Error("failed to upload slab, timing out for 5 minutes", zap.Error(err), zap.Duration("duration", time.Since(start)))
+					if ok := <-waitFor(ctx, 5*time.Minute); ok {
+						continue loop
```
ChrisSchinnerl: Shouldn't you return if `!ok` rather than going on to print `upload completed`?

cmd/filld/main.go
```diff
@@ -0,0 +1,250 @@
...
+
+		loop:
+			for {
+				// prepare client (refresh the app key daily)
+				client, err := loadClient(ctx, sk)
+				if err != nil {
+					log.Error("failed to load client, timing out for 5 minutes", zap.Error(err))
+					if ok := <-waitFor(ctx, 5*time.Minute); ok {
+						continue loop
+					}
```
ChrisSchinnerl: Same as below. Probably need to abort here on `!ok`

cmd/filld/main.go
```diff
@@ -0,0 +1,250 @@
...
+
+	elapsedMu sync.Mutex
+	elapsed   []time.Duration
+)
+
+func init() {
+	flag.StringVar(&adminURL, "admin.url", "http://localhost:9980/api", "the URL of the admin API")
+	flag.StringVar(&adminPassword, "admin.password", "", "the password for the admin API")
+
+	flag.StringVar(&appSeed, "app.seed", "", "the seed phrase for the application key")
```
ChrisSchinnerl: Using a full wallet seed here seems a bit overkill. We can use something like pbkdf2 to turn an arbitrary string into an account.

Then we can add rotating keys further down the road.

cmd/filld/main.go
ChrisSchinnerl, line 1: Not sure I have a better name but `filld` isn't great. Maybe `junkd`?

peterjan: Yeah updated it. Figured `junkd` was just a bit boring.

sdk/client.go
```diff
@@ -536,11 +537,32 @@ func initSDK(client AppClient, dialer HostDialer, appKey types.PrivateKey) (*SDK
 	}, nil
 }
 
+type (
+	sdkOptions struct {
+		logger *zap.Logger
+	}
+
+	SDKOption func(*sdkOptions)
```
n8mgr: Fix lint + no need to call it `SDKOption`. It's an `Option` in the `sdk` package.

n8mgr (changes requested):

Issue #279, opened 2025-07-25
Set the `host_id` for  sectors that haven't been pinned within 72 hours to `NULL`


ChrisSchinnerl:
That will get them caught by the health code and migrated to a new host.

PR #288, opened 2025-07-30
Set the `host_id` for sectors that haven't been pinned within 72 hours to NULL

chris124567:
https://github.com/SiaFoundation/indexd/issues/279

### DIFF
--- a/persist/postgres/init.sql
+++ b/persist/postgres/init.sql
@@ -292,6 +292,9 @@ CREATE INDEX sectors_host_id_null_contract_map_idx ON sectors(host_id) WHERE con
 -- speed up querying sectors of a host by root
 CREATE INDEX sectors_host_id_sector_root_idx ON sectors(host_id, sector_root);
 
+-- for pruning unpinned sectors
+CREATE INDEX sectors_uploaded_at_unpinned_idx ON sectors(uploaded_at) WHERE host_id IS NOT NULL AND contract_sectors_map_id IS NULL;
+
 CREATE TABLE slab_sectors (
     slab_id BIGINT REFERENCES slabs(id) ON DELETE CASCADE,
     sector_id BIGINT REFERENCES sectors(id) ON DELETE CASCADE,
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -430,15 +430,14 @@ func (s *Store) PinSectors(ctx context.Context, contractID types.FileContractID,
 
 // PruneUnpinnableSectors sets the the host ID for sectors that haven't been
 // pinned within 3 days to NULL.
-func (s *Store) PruneUnpinnableSectors(ctx context.Context) error {
+func (s *Store) PruneUnpinnableSectors(ctx context.Context, threshold time.Time) error {
 	err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
 		_, err := tx.Exec(ctx, `
             UPDATE sectors
             SET host_id = NULL
             WHERE host_id IS NOT NULL
 	            AND contract_sectors_map_id IS NULL
-	            AND uploaded_at <= NOW() - INTERVAL '3 days'
-		`)
+	            AND uploaded_at <= $1`, threshold)
 		return err
 	})
 	return err
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -428,8 +428,8 @@ func (s *Store) PinSectors(ctx context.Context, contractID types.FileContractID,
 	})
 }
 
-// PruneUnpinnableSectors sets the the host ID for sectors that haven't been
-// pinned within 3 days to NULL.
+// PruneUnpinnableSectors sets the host ID for sectors that haven't been pinned
+// by the threshold time to NULL.
 func (s *Store) PruneUnpinnableSectors(ctx context.Context, threshold time.Time) error {
 	err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
 		_, err := tx.Exec(ctx, `
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -194,7 +194,7 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
 			batch.Queue(`
 				INSERT INTO sectors (sector_root, host_id, next_integrity_check)
 				VALUES ($1, (SELECT id FROM hosts WHERE public_key = $2), $3)
-				ON CONFLICT (sector_root) DO UPDATE SET sector_root=EXCLUDED.sector_root
+				ON CONFLICT (sector_root) DO UPDATE SET sector_root=EXCLUDED.sector_root, uploaded_at=NOW()
 				RETURNING id
 			`, sqlHash256(sector.Root), sqlPublicKey(sector.HostKey), nextIntegrityCheck).QueryRow(func(row pgx.Row) error {
 				return row.Scan(&sectorIDs[i])
@@ -539,7 +539,7 @@ func (s *Store) MigrateSector(ctx context.Context, root types.Hash256, hostKey t
 	err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
 		resp, err := tx.Exec(ctx, `
 			UPDATE sectors
-			SET host_id = hosts.id, contract_sectors_map_id = NULL, consecutive_failed_checks = 0
+			SET host_id = hosts.id, contract_sectors_map_id = NULL, consecutive_failed_checks = 0, uploaded_at=NOW()
 			FROM hosts
 			WHERE sector_root = $1 AND hosts.public_key = $2
 		`, sqlHash256(root), sqlPublicKey(hostKey))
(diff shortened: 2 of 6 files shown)


persist/postgres/sectors.go
```diff
@@ -425,6 +425,22 @@ func (s *Store) PinSectors(ctx context.Context, contractID types.FileContractID,
...
+// PruneUnpinnableSectors sets the the host ID for sectors that haven't been
+// pinned within 3 days to NULL.
+func (s *Store) PruneUnpinnableSectors(ctx context.Context) error {
+	err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `
+            UPDATE sectors
+            SET host_id = NULL
+            WHERE host_id IS NOT NULL
+	            AND contract_sectors_map_id IS NULL
+	            AND uploaded_at <= NOW() - INTERVAL '3 days'
```
peterjan: `upload_at` needs to change or we need another timestamp I think because what's going to happen now is that if a sector crosses this 3 day mark, it'll be migrated to new hosts over and over until it's finally appended to a contract before `PruneUnpinnableSectors` got to it again

persist/postgres/sectors_test.go
```diff
@@ -1855,6 +1937,74 @@ func BenchmarkMarkFailingSectorsLost(b *testing.B) {
 	}
 }
 
+// BenchmarkPruneUnpinnableSectors benchmarks PruneUnpinnableSectors
+func BenchmarkPruneUnpinnableSectors(b *testing.B) {
```
peterjan, line 1903: could you post the results

persist/postgres/sectors.go
ChrisSchinnerl: Yeah we need `MigrateSector` and `PinSlab` to update the `uploaded_at` column to `NOW()` and extend the corresponding tests to make sure that's happening.

I'd also play around with indices to see if we could add one to speed this up. A good one might be a conditional index like `CREATE INDEX sectors_uploaded_at_unpinned ON sectors(uploaded_at) WHERE host_id IS NOT NULL AND contract_sectors_map IS NULL; -- for pruning unpinnable sectors
`. But if it doesn't improve the benchmark significantly and it's plenty fast, we don't need it.

persist/postgres/sectors.go
```diff
@@ -425,6 +425,22 @@ func (s *Store) PinSectors(ctx context.Context, contractID types.FileContractID,
 	})
 }
 
+// PruneUnpinnableSectors sets the the host ID for sectors that haven't been
+// pinned within 3 days to NULL.
+func (s *Store) PruneUnpinnableSectors(ctx context.Context) error {
```
ChrisSchinnerl: I'd pass the threshold for `uploaded_at` as an argument to be consistent with other pruning store methods.

ChrisSchinnerl (changes requested):

chris124567:
```
BenchmarkPruneUnpinnableSectors-4              1    1395458381 ns/op
```

is the same with and without that index :thinking:  Will see if there's anything else I can do

chris124567:
OK, if I disable:
```
-- CREATE INDEX sectors_host_id_uploaded_at_idx ON sectors(host_id, uploaded_at ASC) WHERE contract_sectors_map_id IS NULL;
```

The Postgres planner will use the new index:
```
CREATE INDEX sectors_uploaded_at_unpinned_idx ON sectors(uploaded_at) WHERE host_id IS NOT NULL AND contract_sectors_map_id IS NULL;
```

but it still doesn't make much of a difference:

```
BenchmarkPruneUnpinnableSectors-4   	       1	1284628786 ns/op
```

That's with the most specific index you can get.  The query would be updating 10% of sectors in the benchmark (so 0.1*262144=26214 sectors updated), so  I don't think 1.2 or 1.3s for that is unreasonable.

persist/postgres/sectors_test.go
```diff
@@ -1731,6 +1862,76 @@ func BenchmarkMarkFailingSectorsLost(b *testing.B) {
...
+	hk := store.addTestHost(b)
+
+	// prepare base db
+	const (
+		dbBaseSize = 1 << 40 // 1TiB of sectors
+		nSectors   = dbBaseSize / proto.SectorSize
+	)
+
+	// insert sectors in batches
+	sectorRoots := make([]types.Hash256, 0, nSectors)
```
peterjan: unused

persist/postgres/sectors_test.go
```diff
@@ -1731,6 +1862,76 @@ func BenchmarkMarkFailingSectorsLost(b *testing.B) {
...
+			b.Fatal(err)
+		}
+		_, err = store.pool.Exec(context.Background(), `UPDATE sectors SET uploaded_at = $1 WHERE id % 10 = 0`, now.Add(-4*day))
+		if err != nil {
+			b.Fatal(err)
+		}
+	}
+	reset()
+
+	for b.Loop() {
```
peterjan: you could add `b.SetBytes` to record amount of data we can process

peterjan (dismissed):
Yeah there's not much left to optimize by restructuring the query or adding indices. Perhaps we could try updating smaller batches in a loop but that's not likely to give a huge speedup. Definitely on the slower side but should probably be fine.

persist/postgres/sectors_test.go
```diff
@@ -1731,6 +1862,76 @@ func BenchmarkMarkFailingSectorsLost(b *testing.B) {
...
+		if _, err := store.PinSlab(context.Background(), account, time.Now().Add(time.Hour), slabs.SlabPinParams{
+			MinShards:     1,
+			EncryptionKey: frand.Entropy256(),
+			Sectors:       sectors,
+		}); err != nil {
+			b.Fatal(err)
+		}
+	}
+
+	// 10% of the sectors are candidates to be marked
```
ChrisSchinnerl: I'd run a few sub-benchmarks with varying numbers of percentages as well as `b.SetBytes` set to tell us the actual data pruned. A result of `1` operation and `100ms` doesn't give us a ton of information without checking out the benchmark code.

10% means that we failed to pin 100GB of data which is already quite a lot. I'd run the benchmark for `0.1%`, `1%` and `10%` to get results for pruning 1GB, 10GB and 100GB.

ChrisSchinnerl (changes requested):

chris124567:
```
BenchmarkPruneUnpinnableSectors
BenchmarkPruneUnpinnableSectors/10.000%
BenchmarkPruneUnpinnableSectors/10.000%-4                  1    1885673395 ns/op    58308.70 MB/s
BenchmarkPruneUnpinnableSectors/1.000%
BenchmarkPruneUnpinnableSectors/1.000%-4                   6     218199937 ns/op    50390.10 MB/s
BenchmarkPruneUnpinnableSectors/0.100%
BenchmarkPruneUnpinnableSectors/0.100%-4                  56      19919721 ns/op    55197.14 MB/s
```

persist/postgres/sectors.go
```diff
@@ -428,6 +428,28 @@ func (s *Store) PinSectors(ctx context.Context, contractID types.FileContractID,
 	})
 }
 
+// PruneUnpinnableSectors sets the host ID for sectors that haven't been pinned
+// by the threshold time to NULL.
+func (s *Store) PruneUnpinnableSectors(ctx context.Context, threshold time.Time) error {
+	err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `SET LOCAL enable_seqscan = off;`)
```
ChrisSchinnerl: Not sure how @n8maninger feels about it but I'd drop this and let the query planner decide what to do unless the impact is significant. If it's not, my guess is the query planner expected it to not matter and went for the scan.

persist/postgres/init.sql
```diff
@@ -282,6 +282,9 @@ CREATE INDEX sectors_host_id_null_contract_map_idx ON sectors(host_id) WHERE con
 -- speed up querying sectors of a host by root
 CREATE INDEX sectors_host_id_sector_root_idx ON sectors(host_id, sector_root);
 
+-- for pruning unpinned sectors
+CREATE INDEX sectors_uploaded_at_unpinned_idx ON sectors(uploaded_at) WHERE host_id IS NOT NULL AND contract_sectors_map_id IS NULL;
```
ChrisSchinnerl, line 296: We need a migration for this since we auto-deploy indexd to our servers now.

ChrisSchinnerl (changes requested):
It's a shame that this isn't faster but it's probably alright. Ideally we never have to prune any sectors since we are able to pin them.

chris124567:
```
goos: linux
goarch: amd64
BenchmarkPruneUnpinnableSectors
BenchmarkPruneUnpinnableSectors/10.000%
BenchmarkPruneUnpinnableSectors/10.000%-4         	       1	1386670341 ns/op	79291.49 MB/s
BenchmarkPruneUnpinnableSectors/1.000%
BenchmarkPruneUnpinnableSectors/1.000%-4          	      10	 190795868 ns/op	57627.64 MB/s
BenchmarkPruneUnpinnableSectors/0.100%
BenchmarkPruneUnpinnableSectors/0.100%-4          	      79	  19986206 ns/op	55013.52 MB/s
```

n8mgr (approved):

ChrisSchinnerl (approved):

PR #292, opened 2025-07-30
Add `waitUntilSynced` to Contract manager

peterjan:
Instead of waiting until our chain manager's tip is synced once, we should instead make sure that our wallet is synced, and check this before doing any contract maintenance that requires the wallet. If we don't we run into reorg errors because the basis is too far in the past.

```
 ERROR   contracts.maintenance.formation failed to form contract {“hostKey”: “ed25519:3077686e977dbe016b6f03ddf6fb9828af8136c10ae6062ec7b2ef99ab3098c2", “full”: false, “error”: “failed to form contract: failed to form contract: failed to get transaction set: failed to update transaction set basis: reorg path from 435999::b78a6a52 to 534253::05e0519a is too long (-0 +98254)“}
 ```

### DIFF
--- a/contracts/manager.go
+++ b/contracts/manager.go
@@ -392,37 +392,40 @@ func (cm *ContractManager) maintenanceLoop(ctx context.Context) {
 }
 
 func (cm *ContractManager) waitUntilSynced(ctx context.Context, log *zap.Logger) bool {
-	var once sync.Once
-	for {
-		select {
-		case <-ctx.Done():
-			return false
-		default:
-		}
-
+	isSynced := func() (bool, error) {
 		ci, err := cm.store.LastScannedIndex(ctx)
 		if err != nil {
-			log.Debug("failed to get last scanned index", zap.Error(err))
-			continue
+			return false, fmt.Errorf("failed to get last scanned index: %w", err)
 		}
 
 		block, ok := cm.cm.Block(ci.ID)
 		if !ok {
-			log.Debug("failed to get block for last scanned index", zap.Stringer("id", ci.ID))
-			continue
-		}
-		if time.Since(block.Timestamp) < 3*time.Hour {
-			return true
+			return false, fmt.Errorf("failed to get block for last scanned index %v", ci.ID)
 		}
 
-		once.Do(func() {
-			log.Info("waiting for wallet to be synced before doing contract maintenance")
-		})
+		return time.Since(block.Timestamp) < 3*time.Hour, nil
+	}
 
+	// check if we are synced
+	synced, _ := isSynced()
+	if synced {
+		return true
+	}
+
+	// if not, check again every minute
+	log.Info("waiting for wallet to be scanned before doing contract maintenance")
+	for {
 		select {
 		case <-ctx.Done():
 			return false
 		case <-time.After(time.Minute):
+		}
+
+		synced, err := isSynced()
+		if synced {
+			return true
+		} else if err != nil {
+			log.Debug(err.Error())
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: subscriber/subscriber.go [hunk 1]
```diff
@@ -181,7 +181,7 @@ func (s *Subscriber) Sync(ctx context.Context) error {
 			break
 		}
 
-		updateCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+		updateCtx, cancel := context.WithTimeout(ctx, 30*time.Minute)
```
peterjan: Seeing as we panic when this times out, we can be pretty generous here. We just want it to fail eventually in case of a deadlock or something.

contracts/manager.go
```diff
@@ -389,20 +391,39 @@ func (cm *ContractManager) maintenanceLoop(ctx context.Context) {
...
-func (cm *ContractManager) blockUntilReady(log *zap.Logger) bool {
+func (cm *ContractManager) waitUntilSynced(ctx context.Context, log *zap.Logger) bool {
 	var once sync.Once
 	for {
 		select {
-		case <-cm.tg.Done():
+		case <-ctx.Done():
 			return false
-		case <-time.After(time.Second):
+		default:
```
ChrisSchinnerl: As already mentioned by copilot, this is a potentially hot loop now every time we call `continue`.

subscriber/subscriber.go
ChrisSchinnerl: 5 minutes should already be plenty generous. 30 minutes means that syncing a block takes longer than for the next block to arrive on average.

contracts/manager.go
```diff
@@ -389,20 +391,39 @@ func (cm *ContractManager) maintenanceLoop(ctx context.Context) {
...
+			log.Debug("failed to get block for last scanned index", zap.Stringer("id", ci.ID))
+			continue
+		}
+		if time.Since(block.Timestamp) < 3*time.Hour {
 			return true
 		}
+
 		once.Do(func() {
-			log.Info("waiting for consensus to be synced before starting maintenance")
+			log.Info("waiting for wallet to be synced before doing contract maintenance")
```
ChrisSchinnerl: I think the message was fine? Or maybe `scanned` would be the better word. Technically we are not waiting for the wallet to sync here but everything subscribed to consensus.

ChrisSchinnerl (changes requested):

contracts/manager.go
peterjan: Consensus is technically synced, since the chain manager has all blocks up to the current tip. However, contract maintenance is currently paused because the wallet is undergoing a rescan. To avoid confusion, it's better to communicate that contract maintenance is blocked due to the wallet scan rather than implying consensus is still catching up.

contracts/manager.go
```diff
@@ -389,20 +401,42 @@ func (cm *ContractManager) maintenanceLoop(ctx context.Context) {
...
-		case <-time.After(time.Second):
+		case <-time.After(cm.syncPollInterval):
 		}
-		if time.Since(cm.cm.TipState().PrevTimestamps[0]) < 3*time.Hour {
+
+		synced, err := isSynced()
+		if synced {
 			return true
+		} else if err != nil {
+			log.Debug(err.Error())
```
n8mgr: `log.Debug("failed to check synced", zap.Error(err))`

n8mgr (changes requested):

ChrisSchinnerl (approved):

n8mgr (approved):

PR #305, opened 2025-08-06
Extend TestUnhealthySlabs

### LINKED ISSUE #270: Prune `contract_sectors_map` of expired contracts
Right now we don't ever delete a contract. Only when a contract is renewed, we update the contract's row in the `contract_sectors_map` to make sure the sectors stored in that contract are linked to the new contract.

However, when a contract expires, we don't update that map. Meaning that the uploaded sector will remain healthy according to `UnhealthySlabs`.

To fix this we should do the following:

- [ ]  Add testing to reproduce this issue by asserting that slabs should become unhealthy when contracts expire
- [ ]  Add testing to make sure renewals don't cause a slab to become unhealthy
- [x]  [Fix the issue by pruning that map of contracts that have been expired for 12 blocks or so similar to `PruneExpiredContractElements`](https://github.com/SiaFoundation/indexd/pull/277)
- [x]  [Add a test for the actual pruning of the map](https://github.com/SiaFoundation/indexd/pull/277)

ChrisSchinnerl:
This PR adds 2 missing edge cases to `TestUnhealthySlabs` and fixes `PruneContractSectorsMap` which broke due to one of those edge cases.

Closes https://github.com/SiaFoundation/indexd/issues/270

### DIFF
--- a/persist/postgres/contracts.go
+++ b/persist/postgres/contracts.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/jackc/pgx/v5"
 	"go.sia.tech/core/types"
 	"go.sia.tech/coreutils/rhp/v4"
 	"go.sia.tech/indexd/contracts"
@@ -386,6 +387,9 @@ func (s *Store) PruneContractSectorsMap(ctx context.Context, maxBlocksSinceExpir
 			INNER JOIN contracts ON csm.contract_id = contracts.contract_id
 			WHERE current_height.scanned_height >= contracts.expiration_height + $1
 		`, maxBlocksSinceExpiry)
+		if err != nil {
+			return fmt.Errorf("failed to query contract_sectors_map for pruning: %w", err)
+		}
 		defer rows.Close()
 
 		for rows.Next() {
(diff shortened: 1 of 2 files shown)


n8mgr (approved):

persist/postgres/contracts.go
```diff
@@ -373,15 +374,56 @@ WHERE fces.contract_id = contracts.contract_id AND current_height.scanned_height
...
+			return fmt.Errorf("failed to iterate over contract_sectors_map rows: %w", rows.Err())
+		} else if len(toPrune) == 0 {
+			return nil
+		}
+
+		s.log.Debug("pruning contract sectors map", zap.Uint64("maxBlocksSinceExpiry", maxBlocksSinceExpiry), zap.Int("toPrune", len(toPrune)))
+
+		for _, id := range toPrune {
+			// update sectors table
+			_, err = tx.Exec(ctx, `
```
n8mgr: Probably best to batch these either with WHERE s.contract_sectors_map_id = ANY($1) or two pgx.Batch

n8mgr (changes requested):

n8mgr (approved):

PR #311, opened 2025-08-06
Add app auth flow

### LINKED ISSUE #313: Add Auth flow to App API
Add the Auth flow to the App API with an embedded UI

n8mgr:
Adds an auth flow for registering apps with an indexer, changes the app API to require TLS, and changes the junk uploader to use the new auth flow.

note: you will need to explicitly disable TLS to use the junk uploader locally.

Flow:
1. App requests connection
2. User visits response URL and approves or denies the connection
3. App can poll `/auth/check` to verify status

Good follow ups would be to style the page and redirect to `CallbackURL` if provided. Closes #313

### DIFF
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -17,6 +17,7 @@ import (
 	"go.sia.tech/coreutils/wallet"
 	"go.sia.tech/indexd/accounts"
 	"go.sia.tech/indexd/api"
+	"go.sia.tech/indexd/api/app"
 	"go.sia.tech/indexd/build"
 	"go.sia.tech/indexd/contracts"
 	"go.sia.tech/indexd/explorer"
@@ -81,6 +82,10 @@ type (
 		UpdateMaintenanceSettings(ctx context.Context, ms contracts.MaintenanceSettings) error
 		PinnedSettings(ctx context.Context) (pins.PinnedSettings, error)
 		UpdatePinnedSettings(ctx context.Context, ps pins.PinnedSettings) error
+
+		AddAppConnectKey(context.Context, app.AddConnectKey) error
+		DeleteAppConnectKey(context.Context, string) error
+		AppConnectKeys(ctx context.Context, offset, limit int) ([]app.ConnectKey, error)
 	}
 
 	// A Syncer can connect to other peers and synchronize the blockchain.
@@ -179,6 +184,11 @@ func NewAPI(chain ChainManager, contracts ContractManager, hosts HostManager, sy
 		// txpool endpoints
 		"GET /txpool/recommendedfee": a.handleGETTxpoolRecommendedFee,
 
+		// connect endpoints
+		"GET /apps/connect/keys":         a.handleGETAppConnectKeys,
+		"PUT /apps/connect/keys":         a.handlePUTAppConnectKeys,
+		"DELETE /apps/connect/keys/:key": a.handleDELETEAppConnectKeys,
+
 		// wallet endpoints
 		"GET /wallet":            a.handleGETWallet,
 		"GET /wallet/events":     a.handleGETWalletEvents,
@@ -247,6 +257,45 @@ func (a *admin) handlePOSTTrigger(jc jape.Context) {
 		return
 	}
 
+	jc.Encode(nil)
+}
+
+func (a *admin) handleGETAppConnectKeys(jc jape.Context) {
+	offset, limit, ok := api.ParseOffsetLimit(jc)
+	if !ok {
+		return
+	}
+
+	keys, err := a.store.AppConnectKeys(jc.Request.Context(), offset, limit)
+	if jc.Check("failed to get app connect keys", err) != nil {
+		return
+	}
+	jc.Encode(keys)
+}
+
+func (a *admin) handlePUTAppConnectKeys(jc jape.Context) {
+	var key app.AddConnectKey
+	if jc.Decode(&key) != nil {
+		return
+	}
+
+	err := a.store.AddAppConnectKey(jc.Request.Context(), key)
+	if jc.Check("failed to add app connect key", err) != nil {
+		return
+	}
+	jc.Encode(nil)
+}
+
+func (a *admin) handleDELETEAppConnectKeys(jc jape.Context) {
+	var key string
+	if jc.DecodeParam("key", &key) != nil {
+		return
+	}
+
+	err := a.store.DeleteAppConnectKey(jc.Request.Context(), key)
+	if jc.Check("failed to delete app connect key", err) != nil {
+		return
+	}
 	jc.Encode(nil)
 }
 
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -2,6 +2,7 @@ package admin
 
 import (
 	"context"
+	"encoding/hex"
 	"net/http"
 
 	"errors"
@@ -25,6 +26,7 @@ import (
 	"go.sia.tech/indexd/pins"
 	"go.sia.tech/jape"
 	"go.uber.org/zap"
+	"lukechampine.com/frand"
 )
 
 const (
@@ -83,7 +85,8 @@ type (
 		PinnedSettings(ctx context.Context) (pins.PinnedSettings, error)
 		UpdatePinnedSettings(ctx context.Context, ps pins.PinnedSettings) error
 
-		AddAppConnectKey(context.Context, app.AddConnectKey) error
+		AddAppConnectKey(context.Context, app.UpdateAppConnectKey) (app.ConnectKey, error)
+		UpdateAppConnectKey(context.Context, app.UpdateAppConnectKey) (app.ConnectKey, error)
 		DeleteAppConnectKey(context.Context, string) error
 		AppConnectKeys(ctx context.Context, offset, limit int) ([]app.ConnectKey, error)
 	}
@@ -186,6 +189,7 @@ func NewAPI(chain ChainManager, contracts ContractManager, hosts HostManager, sy
 
 		// connect endpoints
 		"GET /apps/connect/keys":         a.handleGETAppConnectKeys,
+		"POST /apps/connect/keys":        a.handlePOSTAppConnectKeys,
 		"PUT /apps/connect/keys":         a.handlePUTAppConnectKeys,
 		"DELETE /apps/connect/keys/:key": a.handleDELETEAppConnectKeys,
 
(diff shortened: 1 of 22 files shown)


cmd/indexd/run.go
```diff
@@ -191,15 +223,51 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
 		app.WithLogger(log.Named("api.application")),
 	}
 
+	hostname := cfg.ApplicationAPI.Hostname
+	if hostname == "" {
+		host, port, _ := net.SplitHostPort(appAPIListener.Addr().String())
+		if ip := net.ParseIP(host); ip == nil || ip.IsUnspecified() {
+			hostname = net.JoinHostPort("127.0.0.1", port)
```
ChrisSchinnerl: Could you leave a comment here explaining what edge case you ran into?

cmd/indexd/run.go
```diff
@@ -39,8 +45,35 @@ import (
...
+		return nil, fmt.Errorf("failed to create cert: %w", err)
+	}
+	keyPEM := pem.EncodeToMemory(&pem.Block{Type: "RSA PRIVATE KEY", Bytes: x509.MarshalPKCS1PrivateKey(key)})
+	certPEM := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: certDER})
+
+	cert, err := tls.X509KeyPair(certPEM, keyPEM)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create tls cert: %w", err)
+	}
+	return []tls.Certificate{cert}, nil
```
ChrisSchinnerl: I'd write the certificates to disk to reuse them between reboots. That way, if you wanted to, you could add the self signed certificate to your client to trust it and to avoid potential errors.

You might run into weird issues if the certificate changes every time you reboot.

cmd/indexd/main.go
```diff
@@ -40,7 +40,8 @@ var cfg = config.Config{
 		Password: os.Getenv(indexdAdminPasswordEnvVar),
 	},
 	ApplicationAPI: config.ApplicationAPI{
-		Address: ":9982",
+		Address:  ":9982",
+		Hostname: "127.0.0.1:9982",
```
ChrisSchinnerl: I think you can get rid of `Hostname` here since you implemented the `if hostname == ""` check where you split the listener address which is better UX anyway.

api/app/app.go
```diff
@@ -27,21 +33,78 @@ type (
...
+		Expiration  time.Time
+	}
+
+	// A RegisterAppRequest is the request body for registering a new application.
+	RegisterAppRequest struct {
+		Name        string `json:"name"`
+		Description string `json:"description"`
+		LogoURL     string `json:"logoURL"`
+		ServiceURL  string `json:"serviceURL"`
+		CallbackURL string `json:"callbackURL"`
```
ChrisSchinnerl, line 57: I assume those are meant to be persisted alongside the account so that the UI can then show some information about the app? I created an issue for that to f/u with.

config/config.go
```diff
@@ -18,11 +18,23 @@ type (
...
+		KeyFile  string `yaml:"keyFile"`
+	}
+
 	// ApplicationAPI contains the configuration for the HTTP server serving the
 	// application API
 	ApplicationAPI struct {
-		Address  string `yaml:"address"`
+		Address string `yaml:"address"`
+		// Password is used to register new app keys with the indexer.
+		Password string `yaml:"password"`
```
ChrisSchinnerl: How about we try and avoid adding another password to the config and instead perform the following flow:

1. The app calls `/auth/connect`
2. The endpoint returns the link to the UI and a random request id 

Here is where it diverges for Indexd and Siahub

Indexd
3. The login UI shows you the text "Please approve Request XYZ in the Admin UI"
4. You go to `Admin UI` -> `Pending Auth Requests` -> click approve on the request in the list
5. The `Admin UI` sends a request to the admin API's `/auth/approve` endpoint to approve the request

Siahub
3. The login UI asks you to log in with your email and password and confirms whether you want to connect the app
4. SiaHub sends a request to the admin APIO's `/auth/approve` request 

That way we only need the admin password and still get a similar flow for both indexd and siahub users. The nice thing is that the client app doesn't need to care whether you are connecting to indexd, siahub or another third party service. In any case it gets a link to a UI that gives you further instructions on how to finish adding your app.

ChrisSchinnerl (commented):
Left a few comments but overall this is great. Feels like a massive step towards having an MVP.

config/config.go
n8mgr: One concern with adding it to the admin UI is that anyone could theoretically spam the user's UI with requests if they know the app api url.

n8mgr:
@ChrisSchinnerl I removed the TLS stuff from this PR since it's quite a bit larger now.

- Adds the app_connect_keys table
- Adds CLUD endpoints to the admin API
- Verifies the connect key when adding a new app
- Replaces `Hostname` with `AdvertiseURL` in the App API config, effectively the same except it expects a protocol.

api/admin/admin.go
```diff
@@ -179,6 +184,11 @@ func NewAPI(chain ChainManager, contracts ContractManager, hosts HostManager, sy
 		// txpool endpoints
 		"GET /txpool/recommendedfee": a.handleGETTxpoolRecommendedFee,
 
+		// connect endpoints
+		"GET /apps/connect/keys":         a.handleGETAppConnectKeys,
+		"PUT /apps/connect/keys":         a.handlePUTAppConnectKeys,
```
ChrisSchinnerl, line 193: I probably would have gone for a `POST /apps/connect/key` route which generates a key and returns it. That way the backend controls how the keys are generated rather than the UI or third party.

api/app/app_test.go
```diff
@@ -35,8 +66,42 @@ func TestApplicationAPI(t *testing.T) {
...
+
+	// check the app is not authenticated yet
+	if ok, err := client.CheckAppAuth(ctx); err != nil {
+		t.Fatal(err)
+	} else if ok {
+		t.Fatal("expected app to not be authenticated yet")
+	}
+
+	// approve the app
+	respondToAppConnection(t, connectResp.ResponseURL, true)
```
ChrisSchinnerl: Could you pass the appkey as a parameter here? It's not quite obvious why this works since `foobar` is hardcoded in the helper but more importantly it would be good to verify that authentication works as well.

api/app/app_test.go
```diff
@@ -35,8 +66,42 @@ func TestApplicationAPI(t *testing.T) {
...
+
+	// check the app is not authenticated yet
```
ChrisSchinnerl: wrong comment

persist/postgres/apps.go
```diff
@@ -0,0 +1,95 @@
...
+	"database/sql"
+	"errors"
+	"fmt"
+
+	"go.sia.tech/core/types"
+	"go.sia.tech/indexd/api/app"
+)
+
+// AddAppConnectKey adds or updates an application connection key in the database.
+func (s *Store) AddAppConnectKey(ctx context.Context, key app.AddConnectKey) error {
```
ChrisSchinnerl: (nit) maybe check that remaining uses > 0 before adding it. No point in adding a key with no uses to it.

persist/postgres/apps.go
```diff
@@ -0,0 +1,95 @@
...
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `
+			INSERT INTO app_connect_keys (app_key, use_description, remaining_uses) VALUES ($1, $2, $3)
+			ON CONFLICT (app_key) DO UPDATE SET (use_description, remaining_uses) = (EXCLUDED.use_description, EXCLUDED.remaining_uses);
+		`, key.Key, key.Description, key.RemainingUses)
+		return err
```
ChrisSchinnerl, line 56: (nit) as the bot already said maybe updating the timestamp here makes sense as well. Then it might need a better name though like `modified_at` or something. `data_created` is already a bit inconsistent with the remaining timestamps in the database.

If someone says "hey devs please update foo's key to 4 uses" and all I see is the creation data and 0 remaining uses, I don't know if I'm the first to update that key or if someone else has done that already and the user has claimed the key 4 times. So maybe an additional `total_uses` would be nice as well.

persist/postgres/apps.go
```diff
@@ -0,0 +1,95 @@
...
+	})
+}
+
+// ValidAppConnectKey checks if an application connection key is valid.
+func (s *Store) ValidAppConnectKey(ctx context.Context, key string) (bool, error) {
+	var count int
+	err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		return tx.QueryRow(ctx, `
+			SELECT remaining_uses FROM app_connect_keys WHERE app_key = $1 AND remaining_uses > 0
```
ChrisSchinnerl: I'd drop the `remaining_uses` filter considering that you `return count > 0` and the `app_key` being unique.

Then you can nicely distinguish between 
`true`, nil -> success
`false`, nil -> key has no more uses
`false`, ErrNoRows -> unknown key
`false`, err -> something else

Which is nice because a person without admin access and only a connection key has no idea what's wrong otherwise. Since they can't check the remaining uses.

persist/postgres/migrations.go
```diff
@@ -6,4 +6,15 @@ import (
 	"go.uber.org/zap"
 )
 
-var migrations = []func(context.Context, *txn, *zap.Logger) error{}
+var migrations = []func(context.Context, *txn, *zap.Logger) error{
```
ChrisSchinnerl, line 9: So it begins

ChrisSchinnerl (commented):
Left a few comments but this is quite exquisite. Especially having multi-use connect keys too.

ChrisSchinnerl (approved):

sdk/connect.go
```diff
@@ -0,0 +1,63 @@
...
+	case "darwin":
+		return exec.Command("open", url).Start()
+	default:
+		return fmt.Errorf("unsupported platform %q", runtime.GOOS)
+	}
+}
+
+// Connect requests permission to connect an application to the indexer.
+// If the app is already connected, it returns (true, nil).
+// If the user rejects the connection,
```
ChrisSchinnerl: Unfinished comment

sdk/connect.go
```diff
@@ -0,0 +1,63 @@
...
+		return false, fmt.Errorf("failed to check app auth: %w", err)
+	} else if ok {
+		return true, nil
+	}
+
+	resp, err := client.RequestAppConnection(ctx, meta)
+	if err != nil {
+		return false, fmt.Errorf("failed to request app connection: %w", err)
+	}
+	openBrowser(resp.ResponseURL)
```
ChrisSchinnerl: Not handling the error here means that the app might get stuck until the request expires without the user knowing why. This is also going to fail when run on a server to connect an app without UI. 

We should probably pass a callback, logger or something so that a user can log the URL to stdout for manual pasting.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (dismissed):
Lgtm but linter is unhappy

PR #319, opened 2025-08-07
Add sector stats to store

ChrisSchinnerl:
This PR adds sector stats to the database. They can be used to get an overview of how many sectors the indexer currently manages, how many of them are unpinned, are stored on bad contracts or are not stored on a host at all.

I'm playing around with materialized views to see how long it takes to refresh them.

### DIFF
--- a/persist/postgres/stats_test.go
+++ b/persist/postgres/stats_test.go
@@ -1,0 +1,83 @@
+package postgres
+
+import (
+	"context"
+	"reflect"
+	"testing"
+	"time"
+
+	"go.sia.tech/core/rhp/v4"
+	"go.sia.tech/core/types"
+	"go.sia.tech/indexd/api/admin"
+	"go.uber.org/zap/zaptest"
+)
+
+func TestSectorStats(t *testing.T) {
+	store := initPostgres(t, zaptest.NewLogger(t))
+
+	// prepare a host with a contract
+	hk := store.addTestHost(t)
+	fcid := store.addTestContract(t, hk)
+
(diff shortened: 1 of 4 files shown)


persist/postgres/stats_test.go
```diff
@@ -0,0 +1,163 @@
...
+	// we now got 1 sector that is still unpinned, 1 sector pinned to a bad
+	// contract and 1 sector that is on no host at all
+	assertStats(3, 1, 0, 0)
+	if err := store.RefreshSectorStats(context.Background()); err != nil {
+		t.Fatal(err)
+	}
+	assertStats(3, 1, 1, 1)
+}
+
+func BenchmarkSectorStats(b *testing.B) {
```
ChrisSchinnerl, line 86:
```
goos: darwin
goarch: arm64
cpu: Apple M2 Pro
BenchmarkSectorStats
BenchmarkSectorStats-12    	      20	  55946179 ns/op
PASS
```

As expected this is not very fast. The alternative is to update the currently 6 queries which potentially modify these stats to keep the stats up-to-date. Which might slow down these queries since we need to fetch the rows before modifying them to know how to update the stats or use some subquery/CTE magic to have the updates return the pre-update values.

Depending on how we feel about these tradeoffs we can also start with a materialized view since the stats are already useful for debugging in the early stages of `indexd` and once development on the persistence code has slowed down move to the more complex approach, or go for it right away.

n8mgr:
Since we likely also want to tie storage stats to individual accounts as well, it may be worth doing this the hard way up front. A materialized view SUM GROUP BY will likely take a lot longer to refresh for a large number of accounts.

PR #327, opened 2025-08-11
Sign Request Body and Method

n8mgr:
Changes the signed URL to include the HTTP method and request body. I felt this change was necessary because without it the API cannot be 100% sure the rotated key is the new key the owner intended.

### DIFF
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -13,6 +13,7 @@ import (
 
 	proto "go.sia.tech/core/rhp/v4"
 	"go.sia.tech/core/types"
+	"go.sia.tech/indexd/accounts"
 	"go.sia.tech/indexd/api"
 	"go.sia.tech/indexd/hosts"
 	"go.sia.tech/indexd/slabs"
@@ -38,7 +39,12 @@ type (
 		ValidAppConnectKey(context.Context, string) (bool, error)
 		UseAppConnectKey(context.Context, string, types.PublicKey) error
 
-		HasAccount(ctx context.Context, ak types.PublicKey) (bool, error)
+		HasAccount(context.Context, types.PublicKey) (bool, error)
+	}
+
+	// Accounts defines the account management interface for the application API.
+	Accounts interface {
+		TriggerAccountFunding(force bool) error
 	}
 
 	authReq struct {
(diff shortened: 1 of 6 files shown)


api/app/app.go
```diff
@@ -421,6 +485,10 @@ func NewAPI(advertiseURL string, store Store, opts ...Option) (http.Handler, err
 	}
 
 	return jape.Mux(map[string]jape.Handler{
+		"GET /account":    wrapCORS(wrapSignedAuth(a.handleGETAccount)),
+		"PUT /account":    wrapCORS(wrapSignedAuth(a.handlePUTAccount)), // rotate account
```
ChrisSchinnerl: I already responded on Slack with a suggestion but I think the problem with this here is that it's useless garbage. 

I think a user is only going to need key rotation when they are either moving the app from one device to another or recovering from data loss. Neither of which can be achieved with this endpoint since it requires knowledge of the current keys.

A better approach might be adding the account's public key to `AddConnectKeyRequest` and then associate the account with that key. When that key is used, instead of creating a new account, we update the existing one. 
The flow would then be the same as before. You install the app on a new machine, connect using that key and all your slabs are available. 

Sure it requires admin interaction when only using `indexd` but SiaHub would just let users generate a new app connect key for a connected app themselves. Which it can because it knows what accounts belong to a user.

ChrisSchinnerl (changes requested):

n8mgr:
Got rid of 90% of this PR to just sign the request body and method since it increase security.

n8mgr:
@ChrisSchinnerl I'm honestly leaning towards dropping key rotation since the UX implementation is unclear and letting the apps handle recovery however they please (or most likely not).

ChrisSchinnerl (approved):
Looking good but the corresponding RFC is now outdated. Which makes me wonder if we should just delete them now since they were good guidelines but will only end up being outdated.

api/app/app.go
```diff
@@ -38,7 +39,12 @@ type (
 		ValidAppConnectKey(context.Context, string) (bool, error)
 		UseAppConnectKey(context.Context, string, types.PublicKey) error
 
-		HasAccount(ctx context.Context, ak types.PublicKey) (bool, error)
+		HasAccount(context.Context, types.PublicKey) (bool, error)
+	}
+
+	// Accounts defines the account management interface for the application API.
+	Accounts interface {
```
ChrisSchinnerl, line 46: (nit) to be perfectly consistent this should probably be the `ContractManager` interface

PR #329, opened 2025-08-12
Track account usage

### LINKED ISSUE #328: Track pinned data per account and enforce a limit

ChrisSchinnerl:
Tracks account usage and enforces a limit on how much data can be pinned per account

Closes https://github.com/SiaFoundation/indexd/issues/328


n8mgr (commented):
It's a bit awkward to define the storage limit for the connect key, but the limit then applies per account. I don't particularly mind this way since it's done, but it'd feel more correct to permanently associate the app connect key with the account then apply the limit per connect key rather than per app. An unfortunate side effect would be needing to track usage for the account and the connect key instead of just the account.

persist/postgres/init.sql
```diff
@@ -1,6 +1,8 @@
 CREATE TABLE accounts (
     id SERIAL PRIMARY KEY,
-    public_key BYTEA UNIQUE NOT NULL CHECK (LENGTH(public_key) = 32)
+    public_key BYTEA UNIQUE NOT NULL CHECK (LENGTH(public_key) = 32),
+    pinned_data BIGINT NOT NULL DEFAULT 0 CHECK (pinned_data >= 0), -- total pinned data in bytes
+    max_pinned_data BIGINT NOT NULL CHECK (max_pinned_data >= 0) -- max pinned data in bytes
```
n8mgr: Needs a migration

ChrisSchinnerl:
> It's a bit awkward to define the storage limit for the connect key, but the limit then applies per account. I don't particularly mind this way since it's done, but it'd feel more correct to permanently associate the app connect key with the account then apply the limit per connect key rather than per app. An unfortunate side effect would be needing to track usage for the account and the connect key instead of just the account.

Let's discuss this later today. I feel like it is a bit awkward indeed but I do like the isolation between apps. However, I understand that we need a solution for the SiaHub use case where multiple accounts share a limit. But I'm also not sure associating app keys with accounts fixes this since I'd expect SiaHub to generate a new app key on demand for each app to link.

n8mgr (approved):

PR #334, opened 2025-08-12
Add sector stats

ChrisSchinnerl:
This PR is an alternative to https://github.com/SiaFoundation/indexd/pull/319

Since it's a bit more work to get the same stats, this PR starts with the number of slabs and the API endpoint.


persist/postgres/stats_test.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+	}
+
+	// we start with 0 slabs
+	assertStats(0)
+
+	// pin some slabs
+	var slabIDs []slabs.SlabID
+	for i := byte(0); i < 10; i++ {
+		slabID, err := store.PinSlab(context.Background(), account, time.Now(), newSlab(1))
```
n8mgr:
```
for i := range byte(10)
```

n8mgr (dismissed):

PR #335, opened 2025-08-12
Add pin manager to API

ChrisSchinnerl:
Seems like we created the pin manger upon startup but we never added it to the API.

### DIFF
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -59,6 +59,11 @@ type (
 		ScanHost(ctx context.Context, hk types.PublicKey) (hosts.Host, error)
 	}
 
+	// PinManager defines an interface for managing pinned settings.
+	PinManager interface {
+		UpdatePinnedSettings(ctx context.Context, ps pins.PinnedSettings) error
+	}
+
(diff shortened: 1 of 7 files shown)


api/admin/admin.go
```diff
@@ -131,11 +137,12 @@ type (
 // API exposes endpoints to manage accounts, hosts, settings and the wallet.
 // This is different from the application API, which users, or rather their
 // applications, can use to pin slabs.
-func NewAPI(chain ChainManager, contracts ContractManager, hosts HostManager, syncer Syncer, wallet Wallet, store Store, opts ...Option) http.Handler {
+func NewAPI(chain ChainManager, contracts ContractManager, hosts HostManager, pm PinManager, syncer Syncer, wallet Wallet, store Store, opts ...Option) http.Handler {
 	a := &admin{
 		chain:     chain,
 		contracts: contracts,
 		hosts:     hosts,
+		pm:        pm,
```
n8mgr: nit: field name should be `pins` or something useful

n8mgr (approved):

PR #360, opened 2025-08-19
SDK: Add support for client side encryption

### LINKED ISSUE #359: (SDK): Add support for client side encryption
Right now every slab is encrypted on upload, but the reader we pass when uploading should also be wrapped in a cipher similar to how `renterd` does it. Otherwise, data is only protected from spying hosts but not the indexer..

chris124567:
Fixes https://github.com/SiaFoundation/indexd/issues/359

### DIFF
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -239,7 +239,14 @@ top:
 // Upload uploads the data to hosts and pins it to the indexer.
 //
 // Returns the metadata of the slabs that were pinned
-func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) ([]Slab, error) {
+func (s *SDK) Upload(ctx context.Context, key *[32]byte, r io.Reader, opts ...UploadOption) ([]Slab, error) {
+	if key != nil {
+		var err error
+		if r, err = encrypt(key, r, 0); err != nil {
+			return nil, fmt.Errorf("failed to get stream cipher: %w", err)
+		}
+	}
+
 	uo := uploadOption{
 		dataShards:   10,
 		parityShards: 20,
@@ -334,7 +341,11 @@ func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) ([]
 // Download downloads object metadata
 //
 // TODO: support seeks
-func (s *SDK) Download(ctx context.Context, w io.Writer, metadata []Slab, opts ...DownloadOption) error {
+func (s *SDK) Download(ctx context.Context, key *[32]byte, w io.Writer, metadata []Slab, opts ...DownloadOption) error {
+	if key != nil {
+		w = decrypt(key, w, 0)
+	}
+
 	if len(metadata) == 0 {
 		return errors.New("no slabs to download")
 	}
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -255,6 +255,8 @@ func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) (Ob
 
 	if (uo.parityShards+uo.dataShards)/uo.dataShards < 2 {
 		return Object{}, errors.New("redundancy must be at least 2x")
+	} else if uo.disableEncryption && uo.customKey != nil {
+		return Object{}, errors.New("custom key provided but encryption disabled ")
 	}
 
 	var obj Object
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -29,6 +29,9 @@ type (
 		parityShards uint8
 		hostTimeout  time.Duration
 		maxInflight  int
+
+		customKey         *[32]byte
+		disableEncryption bool
 	}
 
 	downloadOption struct {
@@ -76,7 +79,7 @@ type (
 	// An Object represents a collection of slabs that are associated with a
 	// specific key.
 	Object struct {
-		Key   string
+		Key   *[32]byte
 		Slabs []Slab
 	}
 
@@ -239,14 +242,7 @@ top:
 // Upload uploads the data to hosts and pins it to the indexer.
 //
 // Returns the metadata of the slabs that were pinned
-func (s *SDK) Upload(ctx context.Context, key *[32]byte, r io.Reader, opts ...UploadOption) ([]Slab, error) {
-	if key != nil {
-		var err error
-		if r, err = encrypt(key, r, 0); err != nil {
-			return nil, fmt.Errorf("failed to get stream cipher: %w", err)
-		}
-	}
-
+func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) (Object, error) {
 	uo := uploadOption{
 		dataShards:   10,
 		parityShards: 20,
@@ -258,7 +254,22 @@ func (s *SDK) Upload(ctx context.Context, key *[32]byte, r io.Reader, opts ...Up
 	}
 
 	if (uo.parityShards+uo.dataShards)/uo.dataShards < 2 {
-		return nil, errors.New("redundancy must be at least 2x")
+		return Object{}, errors.New("redundancy must be at least 2x")
+	}
+
+	var obj Object
+	if !uo.disableEncryption {
+		if uo.customKey != nil {
+			obj.Key = uo.customKey
+		} else {
+			obj.Key = new([32]byte)
+			frand.Read(obj.Key[:])
+		}
+
+		var err error
+		if r, err = encrypt(obj.Key, r, 0); err != nil {
+			return Object{}, fmt.Errorf("failed to get stream cipher: %w", err)
+		}
 	}
 
 	type work struct {
@@ -304,11 +315,10 @@ func (s *SDK) Upload(ctx context.Context, key *[32]byte, r io.Reader, opts ...Up
 	}()
 
 	// TODO: cleanup on failure
-	var pinned []Slab
 	for i := 0; ; i++ {
 		select {
 		case <-ctx.Done():
-			return nil, ctx.Err()
+			return Object{}, ctx.Err()
 		case work := <-workCh:
 			err := work.err
 			shards := work.shards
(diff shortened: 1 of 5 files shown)


sdk/encrypt.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (rs *rekeyStream) XORKeyStream(dst, src []byte) {
+	rs.counter += uint64(len(src))
+	if rs.counter < 64*math.MaxUint32 {
+		rs.c.XORKeyStream(dst, src)
+		return
+	}
+	// counter overflow; xor remaining bytes, then increment nonce and xor again
+	rem := 64*math.MaxUint32 - (rs.counter - uint64(len(src)))
+	rs.counter -= 64 * math.MaxUint32
```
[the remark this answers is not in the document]
chris124567: resovled

sdk/encrypt.go
```diff
@@ -0,0 +1,73 @@
...
+	rs.c.XORKeyStream(dst[:rem], src[:rem])
+	// NOTE: we increment the last 8 bytes because XChaCha uses the
+	// first 16 bytes to derive a new key; leaving them alone means
+	// the key will be stable, which might be useful.
+	rs.nonce++
+	nonce := make([]byte, 24)
+	binary.LittleEndian.PutUint64(nonce[16:], rs.nonce)
+	rs.c, _ = chacha20.NewUnauthenticatedCipher(rs.key, nonce)
```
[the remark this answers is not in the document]
chris124567: error only returned when len(nonce) != 24 or len(key) != 32

sdk/encrypt.go
```diff
@@ -0,0 +1,73 @@
...
+
+// decrypt returns a cipher.StreamWriter that decrypts w with k, starting at the
+// specified offset.
+func decrypt(key *[32]byte, w io.Writer, offset uint64) cipher.StreamWriter {
+	nonce64 := offset / (64 * math.MaxUint32)
+	offset %= 64 * math.MaxUint32
+
+	nonce := make([]byte, 24)
+	binary.LittleEndian.PutUint64(nonce[16:], nonce64)
+	c, _ := chacha20.NewUnauthenticatedCipher(key[:], nonce)
```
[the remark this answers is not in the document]
chris124567: error only returned when len(nonce) != 24 or len(key) != 32

sdk/encrypt.go
```diff
@@ -0,0 +1,73 @@
...
+func encrypt(key *[32]byte, r io.Reader, offset uint64) (cipher.StreamReader, error) {
+	if offset%64 != 0 {
+		return cipher.StreamReader{}, fmt.Errorf("offset must be a multiple of 64, got %v", offset)
+	}
+	nonce64 := offset / (64 * math.MaxUint32)
+	offset %= 64 * math.MaxUint32
+
+	nonce := make([]byte, 24)
+	binary.LittleEndian.PutUint64(nonce[16:], nonce64)
+	c, _ := chacha20.NewUnauthenticatedCipher(key[:], nonce)
```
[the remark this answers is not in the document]
chris124567: error only returned when len(nonce) != 24 or len(key) != 32

sdk/client.go
```diff
@@ -239,7 +239,14 @@ top:
 // Upload uploads the data to hosts and pins it to the indexer.
 //
 // Returns the metadata of the slabs that were pinned
-func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) ([]Slab, error) {
+func (s *SDK) Upload(ctx context.Context, key *[32]byte, r io.Reader, opts ...UploadOption) ([]Slab, error) {
```
ChrisSchinnerl: We should probably default to encrypt rather than accepting `nil` as a sentinel value for not encrypting.
So I'd pass the key via `uploadOption` and update the result from `[]Slab` to a new `Object` type that contains both the slabs as well as the key. This should give us a super simple SDK allowing users to upload and download without having to worry about encryption at all. For advanced users we can expose a `WithDisabledEncryption` or `WithXChaCha20Secret` to disable encryption or set a custom key.

-------------

An additional feature to consider is allowing to pass in a `cipher.Stream` or `cipher.AEAD` for encryption. Initially I liked that idea but now I'm leaning against it because it hides the key from us which makes it awkward to use. So I think a developer that wants to roll their own encryption is better off explicitly disabling it in the SDK and wrapping the input `io.Reader` on their end.

sdk/encrypt.go
```diff
@@ -0,0 +1,77 @@
...
+type rekeyStream struct {
+	key []byte
+	c   *chacha20.Cipher
+
+	counter uint64
+	nonce   uint64
+}
+
+const (
+	maxBytes = 64 * math.MaxUint32
```
ChrisSchinnerl: Can you leave a comment here explaining that this is the maximum amount of data we can encrypt with a single nonce since the cipher's counter is a uint32 and every tick of the counter can encrypt 64 bytes?
Maybe also rename the constant accordingly.

sdk/encrypt.go
```diff
@@ -0,0 +1,77 @@
...
+	return cipher.StreamReader{S: rs, R: r}, nil
+}
+
+// decrypt returns a cipher.StreamWriter that decrypts w with k, starting at the
+// specified offset.
+func decrypt(key *[32]byte, w io.Writer, offset uint64) cipher.StreamWriter {
+	nonce64 := offset / (maxBytes)
+	offset %= maxBytes
+
+	nonce := make([]byte, 24)
```
ChrisSchinnerl: I know this code has done its magic in renterd for a year now but maybe we can clean this up a bit.

e.g. these 4 lines can probably be extracted into a helper.

sdk/encrypt.go
```diff
@@ -0,0 +1,77 @@
...
+	rs.nonce++
+	nonce := make([]byte, 24)
+	binary.LittleEndian.PutUint64(nonce[16:], rs.nonce)
+	rs.c, _ = chacha20.NewUnauthenticatedCipher(rs.key, nonce)
+	rs.c.XORKeyStream(dst[rem:], src[rem:])
+}
+
+// encrypt returns a cipher.StreamReader that encrypts r with k starting at the
+// given offset.
+func encrypt(key *[32]byte, r io.Reader, offset uint64) (cipher.StreamReader, error) {
```
ChrisSchinnerl, line 55: Could you add a simple test for encrypting and decrypting data and making sure it ends up being the same? Ideally starting at various offsets below and beyond the `maxBytes`.

sdk/encrypt.go
```diff
@@ -0,0 +1,77 @@
...
+	nonce64 := offset / (maxBytes)
+	offset %= maxBytes
+
+	nonce := make([]byte, 24)
+	binary.LittleEndian.PutUint64(nonce[16:], nonce64)
+	c, _ := chacha20.NewUnauthenticatedCipher(key[:], nonce)
+	c.SetCounter(uint32(offset / 64))
+
+	var buf [64]byte
+	c.XORKeyStream(buf[:offset%64], buf[:offset%64])
```
ChrisSchinnerl, line 79: Could you check if we can avoid this "hack" by simply adding `+1` when setting the counter?

ChrisSchinnerl (changes requested):

sdk/encrypt.go
chris124567:
```
    c.SetCounter(uint32(offset/64) + 1)

    // var buf [64]byte
    // c.XORKeyStream(buf[:offset%64], buf[:offset%64])
```

data mismatch

sdk/client.go
```diff
@@ -251,7 +254,22 @@ func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) ([]
 	}
 
 	if (uo.parityShards+uo.dataShards)/uo.dataShards < 2 {
-		return nil, errors.New("redundancy must be at least 2x")
+		return Object{}, errors.New("redundancy must be at least 2x")
+	}
+
```
ChrisSchinnerl, line 261: Probably best to be super explicit here and check if both a custom key and disableEncryption are provided and return an error if that's the case.

sdk/client_test.go
```diff
@@ -24,18 +24,55 @@ func TestRoundtrip(t *testing.T) {
...
+	buf = bytes.NewBuffer(nil)
+	if err := s.Download(context.Background(), buf, obj); err != nil {
+		t.Fatal(err)
+	}
+
+	if !bytes.Equal(buf.Bytes(), data) {
+		t.Fatal("data mismatch")
+	}
+
+	// with client side encryption, custom key
```
ChrisSchinnerl, line 79: Make sure to check that the slab id is different after every upload to verify that the encryption is actually applied. Uploading the same data twice should result in the same slab id otherwise

sdk/encrypt_test.go
```diff
@@ -0,0 +1,69 @@
...
+func TestEncryptRoundtrip(t *testing.T) {
+	var data [4096]byte
+	frand.Read(data[:])
+
+	var key [32]byte
+	frand.Read(key[:])
+
+	for _, offset := range []uint64{0, 64, 128, 2048, 4096} {
+		t.Run(fmt.Sprint(offset), func(t *testing.T) {
+			expected := data[offset:]
```
ChrisSchinnerl: Is there any benefit to slicing here? If not, it seems like you can merge both loops.

sdk/encrypt_test.go
```diff
@@ -0,0 +1,69 @@
...
+			}
+
+			if !bytes.Equal(expected, buf.Bytes()) {
+				t.Fatalf("data mismatch: expected %v, got %v", expected, buf.Bytes())
+			}
+		})
+	}
+
+	// test offset beyond max bytes
+	for _, offset := range []uint64{maxBytesPerNonce, 2 * maxBytesPerNonce} {
```
ChrisSchinnerl: Also test for offsets just before `maxBytesPerNonce` such as `maxBytesPerNonce-64`, `maxBytesPerNonce-128` etc.

Right now the overflow handling code in `XORKeyStream` is probably never called.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (dismissed):

n8mgr:
@chris124567 can you resolve the conflict so we can merge?

chris124567:
yeah one sec

n8mgr (approved):

PR #364, opened 2025-08-21
Add binary encoding to the Object type returned by the SDK for uploaded objects

### LINKED ISSUE #362: Add binary encoding to the `Object` type returned by the `SDK` for uploaded objects
Fore more convenient SDK usage we should add `MarshalSia` and `UnmarshalSia` methods to the `Object` that encode the object using the Sia binary encoding. 

That way people can easily turn an object into a compact binary format for persistence rather than being forced into writing something custom.

chris124567:
Fixes #362

### DIFF
--- a/sdk/encoding.go
+++ b/sdk/encoding.go
@@ -5,33 +5,33 @@ import (
 )
 
 // EncodeTo implements types.EncoderTo.
-func (x Slab) EncodeTo(e *types.Encoder) {
-	e.Write(x.ID[:])
-	e.WriteUint64(uint64(x.Offset))
-	e.WriteUint64(uint64(x.Length))
+func (s Slab) EncodeTo(e *types.Encoder) {
+	e.Write(s.ID[:])
+	e.WriteUint64(uint64(s.Offset))
+	e.WriteUint64(uint64(s.Length))
 }
 
 // DecodeFrom implements types.DecoderFrom.
-func (x *Slab) DecodeFrom(d *types.Decoder) {
-	d.Read(x.ID[:])
-	x.Offset = uint32(d.ReadUint64())
-	x.Length = uint32(d.ReadUint64())
+func (s *Slab) DecodeFrom(d *types.Decoder) {
+	d.Read(s.ID[:])
+	s.Offset = uint32(d.ReadUint64())
+	s.Length = uint32(d.ReadUint64())
 }
 
 // EncodeTo implements types.EncoderTo.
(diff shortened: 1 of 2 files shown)


n8mgr (dismissed):

sdk/encoding.go
```diff
@@ -0,0 +1,37 @@
...
+
+// DecodeFrom implements types.DecoderFrom.
+func (x *Slab) DecodeFrom(d *types.Decoder) {
+	d.Read(x.ID[:])
+	x.Offset = uint32(d.ReadUint64())
+	x.Length = uint32(d.ReadUint64())
+}
+
+// EncodeTo implements types.EncoderTo.
+func (x Object) EncodeTo(e *types.Encoder) {
```
ChrisSchinnerl: I'd also add helper methods on the `Object` that let you encode/decode it for you like the JSON marshaller.

e.g.
```go
var obj Object
b, err := obj.MarshalSia()
if err != nil {
  ...
}

var decoded Object
if err := decoded.Unmarshal(b); err != nil {
  ...
}

```

Then I'd add a description to the docstring of those methods saying something along the lines of "This is a convenience method to encode the object metadata into bytes using the Sia encoding. This is equivalent to: <insert code example using encoder/decoder here>"

That way we get a nice convenience method but also tell devs how they can be more efficient by using the Sia encoder/decoder directly.

sdk/encoding.go
```diff
@@ -0,0 +1,37 @@
...
+	e.Write(x.ID[:])
+	e.WriteUint64(uint64(x.Offset))
+	e.WriteUint64(uint64(x.Length))
+}
+
+// DecodeFrom implements types.DecoderFrom.
+func (x *Slab) DecodeFrom(d *types.Decoder) {
+	d.Read(x.ID[:])
+	x.Offset = uint32(d.ReadUint64())
+	x.Length = uint32(d.ReadUint64())
```
ChrisSchinnerl: @n8maninger do you think it's worthwhile to attempt to save 8 bytes here? That would reduce the size of the encoded slab by almost 20%. 

We could also say the Sia encoding is far from optimal anyway and have people implement their own compact encoding to save some extra space.

ChrisSchinnerl (changes requested):

sdk/encoding.go
n8mgr: It's easy enough to encode both into a single u64, might as well.

ChrisSchinnerl (approved):

PR #369, opened 2025-08-22
Update `handleGETSlab` and `handleGETSlabs` to support binary encoding

### LINKED ISSUE #363: Update `handleGETSlab` and `handleGETSlabs` to support binary encoding
Go's JSON encoding is very slow and when downloading every millisecond matters.

A good HTTP compliant solution to this is having the client set the `Content-Type` header. Right now we always set `application/json` in `signedRequest`. For fetching slabs we should support setting it to `application/octet-stream`. 

The server currently calls `Encode` to respond which also sets the `Content-Type` header to json. So if the client requests binary data, we need to call a separate `EncodeBinary` helper that then sets the content type to `application/octet-stream` as well.

chris124567:
Fix #363

### DIFF
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -116,7 +116,7 @@ var (
 )
 
 const (
-	acceptHeader = "Content-Type"
+	acceptHeader = "Accept"
 	octetStream  = "application/octet-stream"
 )
 
@@ -173,7 +173,7 @@ func (a *app) handleGETSlab(jc jape.Context, pk types.PublicKey) {
 		return
 	}
 
-	if contentType := jc.Request.Header.Get(acceptHeader); contentType == octetStream {
+	if accept := jc.Request.Header.Get(acceptHeader); accept == octetStream {
 		e := types.NewEncoder(jc.ResponseWriter)
 		slab.EncodeTo(e)
 		jc.Check("failed to flush", e.Flush())
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -168,7 +168,15 @@ func (a *app) handleGETSlab(jc jape.Context, pk types.PublicKey) {
 		return
 	}
 
-	jc.Encode(slab)
+	if contentType := jc.Request.Header.Get("Content-Type"); contentType == "application/json" {
+		jc.Encode(slab)
+	} else if contentType == "application/octet-stream" {
+		e := types.NewEncoder(jc.ResponseWriter)
+		slab.EncodeTo(e)
+		if jc.Check("failed to flush", e.Flush()) != nil {
+			return
+		}
+	}
 }
 
 func (a *app) handleGETSlabs(jc jape.Context, pk types.PublicKey) {
@@ -182,7 +190,15 @@ func (a *app) handleGETSlabs(jc jape.Context, pk types.PublicKey) {
 		return
 	}
 
-	jc.Encode(slabIDs)
+	if contentType := jc.Request.Header.Get("Content-Type"); contentType == "application/json" {
+		jc.Encode(slabIDs)
+	} else if contentType == "application/octet-stream" {
+		e := types.NewEncoder(jc.ResponseWriter)
+		types.EncodeSlice(e, slabIDs)
+		if jc.Check("failed to flush", e.Flush()) != nil {
+			return
+		}
+	}
 }
 
 func (a *app) handleDELETESlab(jc jape.Context, pk types.PublicKey) {
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -117,7 +117,9 @@ var (
 
 const (
 	acceptHeader = "Accept"
-	octetStream  = "application/octet-stream"
+
+	applicationJSON        = "application/json"
+	applicationOctetStream = "application/octet-stream"
 )
 
 // WithLogger sets the logger for application API.
@@ -173,7 +175,7 @@ func (a *app) handleGETSlab(jc jape.Context, pk types.PublicKey) {
 		return
 	}
 
-	if accept := jc.Request.Header.Get(acceptHeader); accept == octetStream {
+	if accept := jc.Request.Header.Get(acceptHeader); accept == applicationOctetStream {
 		e := types.NewEncoder(jc.ResponseWriter)
 		slab.EncodeTo(e)
 		jc.Check("failed to flush", e.Flush())
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -116,9 +116,8 @@ var (
 )
 
 const (
-	contentType            = "Content-Type"
-	contentTypeJSON        = "application/json"
-	contentTypeOctetStream = "application/octet-stream"
+	acceptHeader = "Content-Type"
+	octetStream  = "application/octet-stream"
 )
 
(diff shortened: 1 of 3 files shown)


api/app/app.go
```diff
@@ -168,7 +174,15 @@ func (a *app) handleGETSlab(jc jape.Context, pk types.PublicKey) {
 		return
 	}
 
-	jc.Encode(slab)
+	if contentType := jc.Request.Header.Get(contentType); contentType == contentTypeJSON {
```
n8mgr: return bad request for unknown types

api/app/client.go
```diff
@@ -72,30 +73,67 @@ func sign(appKey types.PrivateKey, method, endpointURL string, req any) (string,
 	return u.String(), body, nil
 }
 
-func (c *Client) signedRequest(ctx context.Context, method, route string, data any, resp any) error {
+func (c *Client) signedRequestCustom(ctx context.Context, setHeaders func(http.Header), method, route string, data any) (io.ReadCloser, error) {
```
n8mgr: No benefit to making this generic, might as well just pass in the content type directly.

api/app/client.go
```diff
@@ -72,30 +73,67 @@ func sign(appKey types.PrivateKey, method, endpointURL string, req any) (string,
...
+		defer io.Copy(io.Discard, r.Body)
+		defer r.Body.Close()
+		b, _ := io.ReadAll(r.Body)
+		return nil, errors.New(strings.TrimSpace(string(b)))
+	}
+
+	return r.Body, nil
+}
+
+func (c *Client) signedRequest(ctx context.Context, method, route string, data, resp any) error {
```
n8mgr: nit: probably call this `signedRequestJSON` and rename `signedRequestCustom` to `signedRequest` for consistency with `signedRequestBinary`

slabs/encoding.go
```diff
@@ -0,0 +1,43 @@
+package slabs
+
+import (
+	"go.sia.tech/core/types"
+)
+
+// EncodeTo implements types.EncoderTo.
+func (x SlabID) EncodeTo(e *types.Encoder) {
```
n8mgr: nit: s

slabs/encoding.go
```diff
@@ -0,0 +1,43 @@
...
+	e.Write(x[:])
+}
+
+// DecodeFrom implements types.DecoderFrom.
+func (x *SlabID) DecodeFrom(d *types.Decoder) {
+	d.Read(x[:])
+}
+
+// EncodeTo implements types.EncoderTo.
+func (x PinnedSector) EncodeTo(e *types.Encoder) {
```
n8mgr: nit: ps

api/app/client.go
```diff
@@ -121,19 +159,28 @@ func (c *Client) UnpinSlab(ctx context.Context, slabID slabs.SlabID) error {
...
 // Slab retrieves a slab from the indexer by its ID.
 func (c *Client) Slab(ctx context.Context, slabID slabs.SlabID) (s slabs.PinnedSlab, err error) {
-	err = c.signedRequest(ctx, http.MethodGet, fmt.Sprintf("/slabs/%s", slabID), nil, &s)
+	err = c.signedRequestBinary(ctx, http.MethodGet, fmt.Sprintf("/slabs/%s", slabID), nil, &s)
 	return
 }
 
+// slabIDs is a wrapper type to allow us to call DecodeFrom on a slice of slab
+// IDs.
+type slabIDs []slabs.SlabID
```
n8mgr: I would consider not supporting the slab IDs endpoint just to get rid of this 🙃

n8mgr (changes requested):

api/app/app.go
```diff
@@ -115,6 +115,12 @@ var (
 	ErrUserRejected = errors.New("user rejected connection request")
 )
 
+const (
+	contentType            = "Content-Type"
```
ChrisSchinnerl: Ok so I did some more reading and I think this is actually not correct. The "Content-Type" specifies the content type of the request body. Not the requested response type.

The "Accept" header is instead used to negotiate the expected response type

e.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept#configuring_accept_request_headers_for_json_responses

api/app/app.go
ChrisSchinnerl: Related to my comment above (about Content-Type probably being wrong and Accept being the right header), [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-accept) states that it's an optional preference that a client can specify for a preferred response type. So I'd lean towards just defaulting to json to keep curling simple.

ChrisSchinnerl (changes requested):

api/app/client.go
```diff
@@ -72,30 +73,57 @@ func sign(appKey types.PrivateKey, method, endpointURL string, req any) (string,
...
 	}
 	req, err := http.NewRequestWithContext(ctx, method, u, body)
 	if err != nil {
-		return fmt.Errorf("failed to create request: %w", err)
+		return nil, fmt.Errorf("failed to create request: %w", err)
 	}
-	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Accept", accept)
+
 	r, err := http.DefaultClient.Do(req)
```
n8mgr, line 87: Probably best to check Content-Type is set correctly on the response.

api/app/client.go
```diff
@@ -73,34 +73,34 @@ func sign(appKey types.PrivateKey, method, endpointURL string, req any) (string,
...
 	if !(200 <= r.StatusCode && r.StatusCode < 300) {
 		defer io.Copy(io.Discard, r.Body)
 		defer r.Body.Close()
 		b, _ := io.ReadAll(r.Body)
-		return nil, errors.New(strings.TrimSpace(string(b)))
+		return nil, "", errors.New(strings.TrimSpace(string(b)))
 	}
 
-	return r.Body, nil
+	return r.Body, r.Header.Get("Content-Type"), nil
```
n8mgr: You already pass in the desired content type, no need to return it and check in the caller.

api/app/client.go
```diff
@@ -73,34 +73,36 @@ func sign(appKey types.PrivateKey, method, endpointURL string, req any) (string,
...
+		return nil, err
 	}
 
 	if !(200 <= r.StatusCode && r.StatusCode < 300) {
 		defer io.Copy(io.Discard, r.Body)
 		defer r.Body.Close()
 		b, _ := io.ReadAll(r.Body)
-		return nil, "", errors.New(strings.TrimSpace(string(b)))
+		return nil, errors.New(strings.TrimSpace(string(b)))
+	} else if contentType := r.Header.Get("Content-Type"); resp != nil && accept != contentType {
```
n8mgr: You should be able to add a carve-out for `r.StatusCode != http.StatusNoContent && accept != contentType` rather than pass in response. You could consider returning `(nil, nil)` for `http.StatusNoContent` to make it explicitly fail if the client is not expecting no content.

n8mgr (approved):

ChrisSchinnerl (approved):

Issue #367, opened 2025-08-21
Hosts filter by geolocation


n8mgr:
Add a filter to the App API to filter hosts by country code.

Depends on https://github.com/SiaFoundation/indexd/issues/366

PR #370, opened 2025-08-22
Filter usable hosts by protocol

### LINKED ISSUE #365: Filter usable hosts by protocol
The app API currently returns all usable hosts. We should add a query parameter that filters this list based on if a host has support for a protocol (`siamux` or `quic`).

chris124567:
Fix #365

Add URL query parameters `protocol` to app API endpoint `/hosts`, so users can query for siamux or quic hosts.


api/app/app.go
```diff
@@ -128,7 +129,19 @@ func (a *app) handleGETHosts(jc jape.Context, _ types.PublicKey) {
...
+	var p string
+	if err := jc.DecodeForm("protocol", &p); err != nil {
+		jc.Error(err, http.StatusBadRequest)
+		return
+	}
+
+	var protocol *chain.Protocol
+	if p != "" {
+		protocol = new(chain.Protocol)
+		*protocol = chain.Protocol(p)
```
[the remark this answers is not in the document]
chris124567: the usablehosts call will error then in sqlNetworkProtocol.Value()

persist/postgres/hosts.go
```diff
@@ -524,7 +524,7 @@ WHERE hosts.id = computed.id RETURNING hosts.id`,
 
 // UsableHosts returns a list of hosts that are not blocked, usable and have an
 // active contract. It returns only the host's public key and addresses.
-func (s *Store) UsableHosts(ctx context.Context, offset, limit int) ([]hosts.HostInfo, error) {
+func (s *Store) UsableHosts(ctx context.Context, protocol *chain.Protocol, offset, limit int) ([]hosts.HostInfo, error) {
```
n8mgr: This is fine for now, but I would consider passing in a struct or using the Option pattern for filters since #367 adds another potential filter parameter.

api/app/app.go
```diff
@@ -128,7 +129,19 @@ func (a *app) handleGETHosts(jc jape.Context, _ types.PublicKey) {
...
-	hosts, err := a.store.UsableHosts(jc.Request.Context(), offset, limit)
+	var p string
+	if err := jc.DecodeForm("protocol", &p); err != nil {
+		jc.Error(err, http.StatusBadRequest)
+		return
+	}
+
+	var protocol *chain.Protocol
+	if p != "" {
+		protocol = new(chain.Protocol)
```
ChrisSchinnerl: I'd check if the protocol is valid and return a bad request status code if not. Makes the DX slightly nicer.

persist/postgres/hosts.go
ChrisSchinnerl: +1

ChrisSchinnerl (changes requested):

n8mgr (dismissed):

ChrisSchinnerl (dismissed):

ChrisSchinnerl:
@chris124567 looking good but got conflicts

n8mgr (approved):

PR #372, opened 2025-08-25
Remove requirement of 64 byte alignment in encrypt

### LINKED ISSUE #371: Remove requirement of 64 byte alignment in `encrypt`
In theory it should be possible to add a `skip` field to `rekeyStream` which indicates how many bytes to skip when the offset that the stream was created with was not 64 byte aligned.

When `skip > 0`, we need to create a 64 byte buffer and copy the beginning of `src` to `but[skip:]`, apply the encryption on that `buf`, copy the encrypted bytes into `src`, decrement `skip` and then continue like we usually would.

After this is implemented and tested, we should port it back to `renterd`.

chris124567:
Fix #371

### DIFF
--- a/sdk/encrypt.go
+++ b/sdk/encrypt.go
@@ -31,10 +31,7 @@ func (rs *rekeyStream) XORKeyStream(dst, src []byte) {
 
 	if rs.skip > 0 {
 		// determine how many bytes we can process from the first block.
-		n := 64 - rs.skip
-		if n > len(src) {
-			n = len(src)
-		}
+		n := min(64-rs.skip, len(src))
 
 		// generate the full 64-byte keystream for the initial block.
 		var keyStream [64]byte
--- a/sdk/encrypt.go
+++ b/sdk/encrypt.go
@@ -3,7 +3,6 @@ package sdk
 import (
 	"crypto/cipher"
 	"encoding/binary"
-	"fmt"
 	"io"
 	"math"
 
@@ -11,8 +10,9 @@ import (
 )
 
 type rekeyStream struct {
-	key []byte
-	c   *chacha20.Cipher
+	key  []byte
+	c    *chacha20.Cipher
+	skip int
 
 	counter uint64
 	nonce   uint64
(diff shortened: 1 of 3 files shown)


sdk/encrypt.go
```diff
@@ -25,58 +25,87 @@ const (
...
+		var keyStream [64]byte
+		rs.c.XORKeyStream(keyStream[:], keyStream[:])
+
+		// XOR the relevant part of the keystream with the source data.
+		for i := 0; i < n; i++ {
+			dst[i] = src[i] ^ keyStream[rs.skip+i]
+		}
+
+		// update state and slice pointers for the rest of the operation
+		rs.counter += uint64(n)
```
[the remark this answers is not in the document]
chris124567:
```
	if rs.skip > 0 {
...
		src = src[n:]
...
```

so it's not the full len(src)

sdk/encrypt.go
```diff
@@ -25,58 +25,87 @@ const (
...
 func (rs *rekeyStream) XORKeyStream(dst, src []byte) {
+	if len(src) == 0 {
+		return
+	}
+
+	if rs.skip > 0 {
+		// determine how many bytes we can process from the first block.
+		n := 64 - rs.skip
+		if n > len(src) {
+			n = len(src)
```
ChrisSchinnerl: (nit) could use the new `min` builtin

ChrisSchinnerl (dismissed):

ChrisSchinnerl (approved):

sdk/encrypt.go
```diff
@@ -25,58 +25,84 @@ const (
...
+	if rs.skip > 0 {
+		// determine how many bytes we can process from the first block.
+		n := min(64-rs.skip, len(src))
+
+		// generate the full 64-byte keystream for the initial block.
+		var keyStream [64]byte
+		rs.c.XORKeyStream(keyStream[:], keyStream[:])
+
+		// XOR the relevant part of the keystream with the source data.
+		for i := 0; i < n; i++ {
```
n8mgr, line 41: for i := range n

n8mgr (approved):

PR #377, opened 2025-08-27
Add location to host scans

### LINKED ISSUE #366: Add location to host scans
Similar to `explored`, detect location when scanning hosts and store it in the database.

chris124567:
Fix #366
Logic largely copied from explored


hosts/manager.go
```diff
@@ -328,7 +337,16 @@ func (m *HostManager) ScanHost(ctx context.Context, hk types.PublicKey) (Host, e
 		scanExponentialBackoffMaxHours,
 	)
 
-	err = m.store.UpdateHost(ctx, hk, networks, settings, success, nextScan)
+	var loc geoip.Location
```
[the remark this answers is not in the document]
chris124567: UpdateHost noops if networks is empty.

hosts/manager.go
```diff
@@ -328,7 +337,16 @@ func (m *HostManager) ScanHost(ctx context.Context, hk types.PublicKey) (Host, e
...
+	for _, cidr := range networks {
+		loc, err = m.locator.Locate(cidr.IP)
+		if err != nil {
+			return Host{}, fmt.Errorf("failed to geolocate host: %w", err)
+		}
+		break
+	}
```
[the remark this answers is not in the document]
chris124567: Errors are only returned if there is a problem reading from the database like a closed file.  An error will not be returned just because there was no result: https://github.com/oschwald/maxminddb-golang/blob/8e01df50851f97b9f9fffa2064b05308d5f8acc3/reader.go#L380

So if it errors on one iteration of the loop, it will error on the next one.

persist/postgres/init.sql
```diff
@@ -27,6 +27,9 @@ CREATE TABLE hosts (
     last_successful_scan TIMESTAMP WITH TIME ZONE,
     last_announcement TIMESTAMP WITH TIME ZONE NOT NULL,
     next_scan TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    country_code TEXT NOT NULL DEFAULT '',
+    latitude DOUBLE PRECISION NOT NULL DEFAULT 0.0,
```
n8mgr:
```
coordinates POINT NOT NULL
```

There's probably a pgx helper for scanning, but I can't remember it off the top of my head.

This should work too:
```go
type Point [2]float64

func (p Point) Value() (driver.Value, error) {
	return fmt.Sprintf("(%f,%f)", p[0], p[1]), nil
}

func (p *Point) Scan(src interface{}) error {
	_, err := fmt.Sscanf(string(src.([]byte)), "(%f,%f)", &p[0], &p[1])
	return err
}
```

hosts/host.go
```diff
@@ -93,6 +93,9 @@ type (
 		RecentUptime           float64             `json:"recentUptime"`
 		Addresses              []chain.NetAddress  `json:"addresses"`
 		Networks               []net.IPNet         `json:"networks"`
+		CountryCode            string              `json:"countryCode"`
```
n8mgr, line 96: Add these to the App API response too

n8mgr (changes requested):

hosts/host.go
ChrisSchinnerl: Also make sure to update the openapi spec.

hosts/manager.go
```diff
@@ -328,7 +337,16 @@ func (m *HostManager) ScanHost(ctx context.Context, hk types.PublicKey) (Host, e
 		scanExponentialBackoffMaxHours,
 	)
 
-	err = m.store.UpdateHost(ctx, hk, networks, settings, success, nextScan)
+	var loc geoip.Location
+	for _, cidr := range networks {
+		loc, err = m.locator.Locate(cidr.IP)
```
ChrisSchinnerl: I'd move this into `resolveHost` and locate using the IPs returned by `LookupIPAddr`.
That should give us the most accurate results.

Also does this lookup fail if an IP address can't be associated with a city? If so, we might not want to fail here and just ignore the location for a given host.

ChrisSchinnerl:
@n8maninger depending on how reliable these lookups are, we might be able to replace our more complicated subnet logic with it right?

Instead of checking separate subnet masks for IPv4 and IPv6 we could just check whether two hosts are in the same city and whether all of a host's IPs come from the same city.

hosts/manager.go
```diff
@@ -567,8 +577,16 @@ func resolveHost(ctx context.Context, resolver Resolver, addresses []chain.NetAd
 			networks = append(networks, ipnets...)
 			filtered = append(filtered, na)
 		}
+
+		if len(ips) > 0 {
+			loc, err = locator.Locate(ips[0].IP)
+			if err != nil {
+				log.Debug("failed to locate IP address", zap.Error(err))
+				continue
```
ChrisSchinnerl: (nit) no need to `continue` here. Chances are that if we add more code after this `if`, we want it to execute even if the `Locate` failed.

persist/postgres/migrations.go
```diff
@@ -61,4 +61,16 @@ var migrations = []func(context.Context, *txn, *zap.Logger) error{
 		}
 		return nil
 	},
+	// adds the "country_code", "latitude", "longitude" columns
```
ChrisSchinnerl: outdated docstring

n8mgr (approved):

ChrisSchinnerl (approved):

n8mgr:
@alexfreska you can remove explored for this now

alexfreska:
@n8mgr woah native geo support, thats awesome! 🚀 I'll update the app.

PR #378, opened 2025-08-28
Add dockerfile for junkd

### LINKED ISSUE #221: Deploy a chunk uploader
Best way to test scalability is to upload data. As soon as the basic functionality has integration testing coverage, we should deploy a indexd on a server (e.g. iris) and start uploading to it. 

Maybe around 1TB per account for now to simulate an increasing number of users storing data.

ChrisSchinnerl:
This is the Dockerfile of the image currently deployed on Zeus. 
I manually pushed it to `ghcr.io/siafoundation/junkd:master`

Fixes https://github.com/SiaFoundation/indexd/issues/221

### DIFF
--- a/cmd/junkd/Dockerfile
+++ b/cmd/junkd/Dockerfile
@@ -1,0 +1,27 @@
+FROM golang:1.24 AS builder
+
+WORKDIR /junkd
+
+# get dependencies
+COPY go.mod go.sum ./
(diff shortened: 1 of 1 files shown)


n8mgr:
Can we consider moving junkd out of here? It seems silly to add a dockerfile for something no one should use.

PR #379, opened 2025-08-28
Add description, logoURL and serviceURL to account

### LINKED ISSUE #316: Persist the additional fields of the `RegisterAppRequest` like the `Logo` link alongside the account
We need those to give some more information about accounts in the UI.

To do so we can use the `req` in `handlePOSTAuthConnect` to pass additional information to `UseAppConnectKey` for account creation. Through the `req.Request` we have access to the original `RegisterAppRequest`. So we can take the `Description`, `LogoURL` and `ServiceURL` and persist them in the `accounts` table. 

To prevent people from uploading a whole library in those strings, we should probably also validate the request when we first receive it to limit the size of each of these fields to 1000 characters or so.

Also don't forget to update the service accounts we add to also have a description. Right now there is no telling which account is a service account. e.g. "Data Integrity Service Account"

Requires #311

ChrisSchinnerl:
Fixes https://github.com/SiaFoundation/indexd/issues/316


api/app/app.go
```diff
@@ -43,7 +43,7 @@ type (
 	// Accounts defines the account management interface for the application API.
 	Accounts interface {
 		ValidAppConnectKey(context.Context, string) (bool, error)
-		UseAppConnectKey(context.Context, string, types.PublicKey) error
+		UseAppConnectKey(ctx context.Context, connectKey, description, logoURL, serviceURL string, appKey types.PublicKey) error
```
n8mgr: Add a struct for the additional meta in case we want to extend it

persist/postgres/apps.go
```diff
@@ -149,7 +149,12 @@ func (s *Store) UseAppConnectKey(ctx context.Context, connectKey string, appKey
 			return accounts.ErrKeyExhausted
 		}
 
-		if err := addAccount(ctx, tx, appKey, false, accounts.WithMaxPinnedData(storageLimit)); err != nil {
+		if err := addAccount(ctx, tx, appKey, false,
+			accounts.WithDescription(description),
```
n8mgr: Using options seems out of place when we always pass them in.

n8mgr (changes requested):

n8mgr (approved):

PR #395, opened 2025-09-03
Only return hosts that have funded accounts

### LINKED ISSUE #385: Only return hosts that have funded accounts
When using the App API, currently all usable hosts are returned. After first registering the account, this causes lots of failures with `not enough funds` until enough time has passed for the account to be funded. The response should be changed to filter hosts to only those that have had their account funded.

chris124567:
Fix #385


api/app/app.go
```diff
@@ -133,6 +133,11 @@ func WithLogger(log *zap.Logger) Option {
 }
 
 func (a *app) handleGETHosts(jc jape.Context, _ types.PublicKey) {
+	var ak types.PublicKey
```
n8mgr: Use the ignored public key parameter. That’s the account key

api/app/client.go
```diff
@@ -128,7 +128,7 @@ func (c *Client) signedRequestBinary(ctx context.Context, method, route string,
 }
 
 // Hosts returns all usable hosts.
-func (c *Client) Hosts(ctx context.Context, opts ...api.URLQueryParameterOption) (hosts []hosts.HostInfo, err error) {
+func (c *Client) Hosts(ctx context.Context, accountKey types.PublicKey, opts ...api.URLQueryParameterOption) (hosts []hosts.HostInfo, err error) {
```
n8mgr: Extra param is unnecessary

api/app/client.go
```diff
@@ -237,6 +237,11 @@ func (c *Client) CheckAppAuth(ctx context.Context) (bool, error) {
 	}
 }
 
+// PublicKey returns the public key associated with the app key of the client.
+func (c *Client) PublicKey() types.PublicKey {
```
n8mgr: Unnecessary

persist/postgres/hosts.go
```diff
@@ -622,8 +627,10 @@ WHERE
 	-- active contracts
 	EXISTS (SELECT 1 FROM contracts WHERE host_id = hosts.id AND state <= 1) AND
 	-- protocol filter
-	($4::smallint IS NULL OR EXISTS (SELECT 1 FROM host_addresses WHERE host_id = hosts.id AND protocol = $4::smallint))
-LIMIT $1 OFFSET $2;`, limit, offset, queryOpts.CountryCode, (*sqlNetworkProtocol)(queryOpts.Protocol))
+	($4::smallint IS NULL OR EXISTS (SELECT 1 FROM host_addresses WHERE host_id = hosts.id AND protocol = $4::smallint)) AND
+	-- filter out hosts we don't have a balance with
+	EXISTS (SELECT 1 FROM service_accounts WHERE balance > 0 AND host_id = hosts.id AND account_id = $5)
```
n8mgr: Service accounts are internal to `indexd`. Normal accounts won't have a record here. I'd add a `funded` field to `account_hosts` that is set to true whenever funding succeeds, then check if `acccount_hosts.funded=true`

n8mgr (changes requested):

internal/testutils/cluster.go
```diff
@@ -168,6 +169,17 @@ func (c *Cluster) FundHosts(ctx context.Context, t testing.TB, hosts ...*Host) {
 	}
 }
 
+// FundHostAccounts funds the service account for that host.
+func (c *Cluster) FundHostAccounts(ctx context.Context, t testing.TB, pk types.PublicKey, hks ...types.PublicKey) {
+	t.Helper()
+
+	for _, hk := range hks {
+		if err := c.Indexer.store.UpdateServiceAccountBalance(ctx, hk, proto.Account(pk), types.Siacoins(100)); err != nil {
```
ChrisSchinnerl, line 177: This should only be used for service accounts and not regular user accounts

persist/postgres/hosts.go
ChrisSchinnerl: Also make sure to add/run a benchmark. You might want to try adding a benchmark over `(hostd_id, account_id) WHERE funded = true` to faster join hosts with funded accounts.

ChrisSchinnerl (changes requested):

persist/postgres/hosts.go
chris124567: What is a good way to tell if a regular account is funded?  `account_hosts.consecutive_failed_funds = 0`?  If there hasn't been a funding attempt that will still be 0.

n8mgr: I'd add a funded field to account_hosts that is set to true whenever funding succeeds, then check if `acccount_hosts.funded=true`

chris124567:
```
goos: linux
goarch: amd64
BenchmarkUsableHosts
BenchmarkUsableHosts-4               301           3993947 ns/op
PASS
```

Issue #392, opened 2025-09-02
Additional Metrics


n8mgr:
- [x] [Pinned Sectors](https://github.com/SiaFoundation/indexd/pull/406)
- [x] [Unpinned Sectors](https://github.com/SiaFoundation/indexd/pull/433)
- [x] [Number of times a sector was migrated](https://github.com/SiaFoundation/indexd/pull/431)
- [x] [Unpinnable Sectors](https://github.com/SiaFoundation/indexd/pull/434)
- [x] [Funds spent on a host](https://github.com/SiaFoundation/indexd/pull/435)

peterjan:
done

PR #406, opened 2025-09-09
Metrics: add `num_pinned_sectors`

peterjan:
This PR adds number of pinned sectors to the sector statistics. I've added a separate test to ensure pinning sectors as well as marking sectors as failed or lost update the count properly. I added a benchmark and ensured there's no regression in the relevant benchmarks. `BenchmarkUpdatePinnedSectors` runs in under 1ms.

references https://github.com/SiaFoundation/indexd/issues/392


persist/postgres/stats.go
```diff
@@ -16,13 +16,18 @@ func (s *Store) initSectorStats(ctx context.Context, tx *txn) error {
 	return err
 }
 
+func (s *Store) updatePinnedSectors(ctx context.Context, tx *txn) error {
```
n8mgr: Gotta increment manually. COUNT will never be fast at scale.

peterjan: Yeah I was hoping it would be fast enough if we add a partial index, because keeping a count manually is a little bit tricky. There's a couple of spots where we might count an unpin twice causing a corrupted count. I'll see if that's the case and what can be done about that.

persist/postgres/sectors.go
```diff
@@ -448,30 +519,33 @@ func (s *Store) PinSectors(ctx context.Context, contractID types.FileContractID,
...
-		} else if resp.RowsAffected() == 0 {
-			// if no sectors were updated, check if the contract exists
-			var exists bool
-			if err := tx.QueryRow(ctx, "SELECT EXISTS (SELECT 1 FROM contracts WHERE contracts.contract_id = $1)", sqlHash256(contractID)).
-				Scan(&exists); err != nil {
-				return fmt.Errorf("failed to check if contract exists: %w", err)
-			} else if !exists {
-				return contracts.ErrNotFound
+		} else if resp.RowsAffected() > 0 {
+			if err := s.incrementPinnedSectors(ctx, tx, resp.RowsAffected()); err != nil {
```
ChrisSchinnerl: This might increment the count incorrectly if `PinSectors` was called on a sector that was already pinned it seems.

persist/postgres/sectors_test.go
```diff
@@ -1511,6 +1630,82 @@ func BenchmarkUnpinnedSectors(b *testing.B) {
 	}
 }
 
+// BenchmarkUpdatePinnedSectors benchmarks UpdatePinnedSectors.
```
ChrisSchinnerl: leftover benchmark?

ChrisSchinnerl (commented):
Did you check that none of the queries degraded too much in performance at the latest iteration of your PR?

persist/postgres/sectors.go
peterjan: Hm this is a little odd and you are right, it's something I overlooked because we do `OR contract_sectors_map_id != $2`. I'm not sure when that would happen though because we never repin a sector that was already pinned, unless it was unpinned for some reason in which case we do want to increment the count when it was repinned. The query is oddly specific though so perhaps it's an edge case you had in mind when writing it?

peterjan: I updated `PinSectors` to fetch the `id` and `contract_sectors_map_id` of the sectors that are getting pinned. I then only count the sectors that go from `NULL` to the new `contract_sectors_map_id` as the ones being pinned.

peterjan:
> Did you check that none of the queries degraded too much in performance at the latest iteration of your PR?

Yes

persist/postgres/sectors.go
```diff
@@ -533,15 +533,37 @@ func (s *Store) PinSectors(ctx context.Context, contractID types.FileContractID,
...
+			return fmt.Errorf("failed to query sectors: %w", err)
+		}
+		defer rows.Close()
+
+		var sectorIDs []int64
+		var pinned int64
+		for rows.Next() {
+			var sectorID int64
+			var contractMapID sql.NullInt64
+			if err := rows.Scan(&sectorID, &contractMapID); err != nil {
```
ChrisSchinnerl: I think we got this exact loop in quite a few places now. Might be good to isolate this in a helper that returns a slice of sector ids and a number indicating how many of these sectors are pinned.

ChrisSchinnerl:
Might also want to change the base branch to `master` since this PR looks pretty standalone anyway.

persist/postgres/sectors.go
peterjan: I looked into this and unfortunately the behaviour is slightly different every time:
`WHERE host_id = $1 AND sector_root = ANY($2)`
`WHERE sector_root = ANY($1) AND (contract_sectors_map_id IS NULL OR contract_sectors_map_id != $2)`
`WHERE host_id = $1 AND consecutive_failed_checks >= $2 LIMIT $3`
`WHERE sector_root = $1`

The helper would have to take optional parameters and use it to construct a different query every time. It's not worth it in my opinion. I wrote it out for the basic case, but when I tried working that helper into the other spots it became too messy.

peterjan: I'll dedupe the scanning and pass the rows

ChrisSchinnerl (approved):

n8mgr (approved):

PR #413, opened 2025-09-10
Simplify hosts in dialer

chris124567:
Remove `ActiveHosts` method from interface and instead just ensure the list of public keys returned by `Hosts` begins with the active hosts' keys.

### DIFF
--- a/sdk/dialer.go
+++ b/sdk/dialer.go
@@ -193,25 +193,30 @@ func (d *Dialer) clearHostConnection(hostKey types.PublicKey) {
 	entry.clear()
 }
 
-// ActiveHosts implements the [HostDialer] interface.
-func (d *Dialer) ActiveHosts() (hks []types.PublicKey) {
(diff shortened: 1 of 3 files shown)


sdk/dialer.go
```diff
@@ -193,25 +193,30 @@ func (d *Dialer) clearHostConnection(hostKey types.PublicKey) {
...
+	seen := make(map[types.PublicKey]struct{})
 	for hk, entry := range d.conns {
 		if entry.tc != nil {
 			hks = append(hks, hk)
+			seen[hk] = struct{}{}
 		}
 	}
-	return
-}
+	shuffle(hks)
```
peterjan: We probably don't need to shuffle here since ranging over `d.conns` will return them in random order. Not too sure how random it really is though but probably good enough. Perhaps rename `length` to `active` to make it immediately obvious we are returning active hosts first followed by the inactive ones.

n8mgr (approved):

PR #417, opened 2025-09-10
Fix race

peterjan:
The stacktrace is from when we had `ActiveHosts` but the race is still relevant.

```
==================
WARNING: DATA RACE
Read at 0x00c002c06030 by goroutine 16:
  go.sia.tech/indexd/sdk.(*Dialer).ActiveHosts()
...
Previous write at 0x00c002c06030 by goroutine 65:
  go.sia.tech/indexd/sdk.(*connEntry).setTransport()
```

### DIFF
--- a/sdk/dialer.go
+++ b/sdk/dialer.go
@@ -54,6 +54,12 @@ func (c *connEntry) clear() {
 	}
 }
 
+func (c *connEntry) isConnected() bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.tc != nil
+}
+
 func (c *connEntry) setTransport(tc rhp.TransportClient, assign bool) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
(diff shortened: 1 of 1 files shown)


chris124567 (dismissed):

sdk/dialer.go
```diff
@@ -195,19 +201,23 @@ func (d *Dialer) clearHostConnection(hostKey types.PublicKey) {
 
 // Hosts implements the [HostDialer] interface.
 func (d *Dialer) Hosts() (hks []types.PublicKey) {
+	// grab current state
 	d.mu.Lock()
-	defer d.mu.Unlock()
+	addrs := d.addrs
+	conns := d.conns
```
ChrisSchinnerl: probably need to deep copy here as copilot suggested

peterjan: figured it wasn't necessary because it's a map and not a slice, updated regardless because it doesn't hurt

chris124567 (approved):

ChrisSchinnerl (approved):

PR #418, opened 2025-09-11
Update SDK upload

peterjan:
This PR improves the upload performance of the SDK by uploading data shards before we encode the shards since those don't change anyway. It also starts reading consecutive slabs while we potentially await the previous one. I did some tests on zeus and uploading 400MB files using our default 10-20 redundancy makes single thread uploads roughly 80% faster. Single threaded upload speeds go from 275mbps to 500mpbs and that scales up until roughly 6 threads (so ~3Gbps). 

I did some more tests on zeus, uploading 4 slabs at 2–4 redundancy (so 32MB/96MB redundant)

Config | Threads | Throughput | Throughput (slow encode, 500ms) | Improvement vs Baseline
-- | -- | -- | -- | --
Baseline | 1 | 75.16 Mbps | 72 Mbps | –
  | 4 | 123.48 Mbps | 96 Mbps | –
  | 8 | 362 Mbps | 308 Mbps | –
Branch | 1 | 110 Mbps | 107 Mbps | +46% / +49%
  | 4 | 418.76 Mbps | 297 Mbps | +239% / +209%
  | 8 | 523.45 Mbps | 502 Mbps | +45% / +63%


sdk/upload.go
```diff
@@ -0,0 +1,258 @@
...
+			}
+			shards := make([][]byte, totalShards)
+			for i := range shards {
+				shards[i] = make([]byte, proto4.SectorSize)
+			}
+			stripedSplit(buffer, shards[:dataShards])
+			encryptionKey := types.HashBytes(append(s.appKey[:], buffer[:n]...))
+
+			// since data shards do not change during encoding, we can launch
+			// these ahead of time and not wait for encoding to finish first
```
ChrisSchinnerl: I'm surprised that this would have that much of an impact on a server that takes only 20ms to encode the shards. Did you try running it without any encoding before this refactor for a comparison?

It would also be interesting to see how `BenchmarkUpload` performs before and after your change.

peterjan: That's not the only change though, the PR description was a little misleading indeed, I updated it. On `master` we upload one slab at a time, this PR changes that and uses a queue to upload shards, that queue might contain shards from more than one slab. E.g. we now read a slab while we're waiting on the slow shards of the previous one. 

I think `BenchmarkUpload` changed over time, it prints a matrix and it takes forever. It doesn't seem all that useful. I ran it with the following parameters: 30 | 0 | 30. So that's essentially 30 hosts that are equal in upload speed, and all slab shards can launch in parallel:

```
BenchmarkUpload/slow_30_timeout_0_inflight_30-10                       1        8744002500 ns/op          29.28 MB/s
BenchmarkUpload/slow_30_timeout_0_inflight_30-10                       1        7627275584 ns/op          33.56 MB/s
```

It's a little hard for me to properly compare these benchmarks because zeus is such a beast and locally I'm stuck with the slow hotel WiFi. I figured it's clearly faster uploading large files and the algorithm is unlikely to hurt performance.

peterjan: I'll move it to `DRAFT` until I can provide you with a better benchmarks.

peterjan: I extended the PR description with some more measurements.

sdk/upload.go
```diff
@@ -0,0 +1,258 @@
...
+
+		encryptionKey [32]byte
+		index         int
+		sector        []byte
+	}
+
+	slabUpload struct {
+		encryptionKey [32]byte
+
+		minShards   uint
```
ChrisSchinnerl: Don't need this field. You got `uo` in scope where you need it.

sdk/upload.go
```diff
@@ -0,0 +1,258 @@
...
+		totalShards int
```
ChrisSchinnerl: Same as `minShards`

sdk/upload.go
```diff
@@ -0,0 +1,258 @@
...
+		totalLength uint32
```
ChrisSchinnerl:
```suggestion
		length uint32
```

sdk/client.go
```diff
@@ -257,83 +168,70 @@ func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) (Ob
...
-				workCh <- work{err: fmt.Errorf("failed to encode slab %d shards: %w", i, err)}
-				return
-			}
-			encryptionKey := types.HashBytes(append(s.appKey[:], slabBuf[:n]...))
-			workCh <- work{length: n, encryptionKey: encryptionKey, shards: shards}
-		}
-	}()
+
+	// start an uploader
+	queue := uploader(ctx, s.dialer, uo.maxInflight, uo.hostTimeout)
```
ChrisSchinnerl: There is no reason for `uploader` to be called outside of `uploadSlabs` since you don't use its result outside of it.

sdk/upload.go
```diff
@@ -0,0 +1,258 @@
...
+			}()
+		}
+	}()
+
+	return queue
+}
+
+func (s *SDK) uploadSlabs(ctx context.Context, r io.Reader, enc reedsolomon.Encoder, queue chan shardUpload, dataShards, parityShards int) chan slabUpload {
+	slabsCh := make(chan slabUpload, 4)
+
```
ChrisSchinnerl, line 82: The `queue` should be created in here and then passed to `uploader`. Then `uploader` can keep pulling from the `queue` until `uploadSlabs` closes it.

That way you get a nicer worker pattern since the `uploader` doesn't know how much work there is and can therefore not close the channel.

sdk/upload.go
```diff
@@ -0,0 +1,258 @@
...
+				}
+			}()
+		}
+	}()
+
+	return queue
+}
+
+func (s *SDK) uploadSlabs(ctx context.Context, r io.Reader, enc reedsolomon.Encoder, queue chan shardUpload, dataShards, parityShards int) chan slabUpload {
+	slabsCh := make(chan slabUpload, 4)
```
ChrisSchinnerl: Random little `4` without comment

ChrisSchinnerl (changes requested):

sdk/client.go
peterjan: Yeah I knew that, I liked it better over passing all the arguments to uploadSlabs. I'll change it though I was on the fence myself to change it couple of times when writing it.

sdk/client.go
ChrisSchinnerl, line 1: Reminder to update `junkd` after merging this.

ChrisSchinnerl (dismissed):

ChrisSchinnerl (dismissed):

peterjan:
@n8mgr do we still want this change? We have it in the rust sdk already I see so we might as well mirror it in go? I'm pining because I have some merge conflicts incoming I believe.

ChrisSchinnerl (approved):

PR #419, opened 2025-09-11
Add path to signed url

n8mgr:
Currently signed URLs can be used on any request matching the HTTP host and method. Sharing an object URL GET request would allow the "sharee" to access any other GET endpoint using the url query parameters.

### DIFF
--- a/api/app/client.go
+++ b/api/app/client.go
@@ -33,9 +33,9 @@ type Client struct {
 	validity time.Duration
 }
 
-// Sign signs the request with the appropriate headers and returns the signed URL
+// sign signs the request with the appropriate headers and returns the signed URL
 // and request body.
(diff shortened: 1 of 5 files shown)


api/app/client.go
```diff
@@ -200,6 +199,16 @@ func (c *Client) DeleteObject(ctx context.Context, key types.Hash256) (err error
 	return
 }
 
+// ObjectShareURL generates a signed URL for accessing the object with the given
+// key. The URL is valid until the specified validUntil time.
+func (c *Client) ObjectShareURL(ctx context.Context, key types.Hash256, validUntil time.Time) (string, error) {
```
ChrisSchinnerl, line 204: Maybe test?

ChrisSchinnerl (approved):

PR #420, opened 2025-09-11
Add shared objects

n8mgr:


api/app/app.go
```diff
@@ -196,6 +197,24 @@ func (a *app) handleGETObject(jc jape.Context, pk types.PublicKey) {
 	jc.Encode(obj)
 }
 
+func (a *app) handleGETObjectShared(jc jape.Context, _ types.PublicKey) {
```
ChrisSchinnerl, line 200: Reminder to update the spec

api/app/app_test.go
```diff
@@ -475,3 +475,180 @@ func TestAppConnect(t *testing.T) {
...
+	buf2, err := json.MarshalIndent(sharedObj, "", "  ")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !bytes.Equal(buf1, buf2) {
+		t.Log("expected:", string(buf1))
+		t.Log("got:     ", string(buf2))
+		t.Fatal("shared object does not match expected")
+	}
+}
```
ChrisSchinnerl, line 661: Sleep 1s and try again?

persist/postgres/objects_test.go
```diff
@@ -224,3 +228,115 @@ func TestListObjectsRegression(t *testing.T) {
...
+	}
+
+	sharedObj, err := store.SharedObject(t.Context(), obj.Key)
+	if err != nil {
+		t.Fatal(err)
+	} else if !reflect.DeepEqual(expectedSharedObj, sharedObj) {
+		t.Fatalf("shared objects not equal: expected %+v, got %+v", expectedSharedObj, sharedObj)
+	}
+
+	// pin the slabs to the second account
```
ChrisSchinnerl, line 322: Could you try having the other account save the object before pinning the slabs which should fail, then pin the slabs, then save the object again and see if it works now?

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

PR #424, opened 2025-09-11
Extend SDK with `DownloadSharedObject`

peterjan:
This PR adds `DownloadSharedObject` to the SDK. I have not added any unit tests for this yet, because I think it doesn't make sense to add those. Since we re-use the download algorithm and we re-use it for shared objects, there shouldn't be any issues anyway. Adding tests is a bit of a nuisance because of some of our types, I'll open another PR to try and straighten those out a bit after which mocking this in testing should be easier.


chris124567 (approved):

chris124567 (changes requested):
add DownloadSharedObject test

peterjan:
> add DownloadSharedObject test

I want to sort out some of the types first. Adding a unit test was annoying because I had to mock saving an object and then I had to transform the various types, which would only be removed soon anyway.

api/app/client.go
```diff
@@ -217,37 +240,24 @@ func (c *Client) CreateSharedObjectURL(ctx context.Context, objectKey types.Hash
...
-	}
-	fields := strings.Split(u.Fragment, "=")
-	if len(fields) != 2 || fields[0] != "encryption_key" {
-		return slabs.SharedObject{}, fmt.Errorf("invalid shared URL: missing encryption key")
-	}
-	var encryptionKey [32]byte
-	if n, err := hex.Decode(encryptionKey[:], []byte(fields[1])); err != nil {
-		return slabs.SharedObject{}, fmt.Errorf("invalid shared URL: invalid encryption key: %w", err)
-	} else if n != 32 {
-		return slabs.SharedObject{}, fmt.Errorf("invalid shared URL: invalid encryption key length: expected 32 bytes, got %d", n)
```
ChrisSchinnerl, line 237: This seems like an unnecessary refactor considering the code wasn't duplicated.

sdk/client.go
```diff
@@ -338,13 +340,46 @@ func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) (Ob
...
+			return slabs.PinnedSlab{}, 0, fmt.Errorf("failed to get slab %d metadata: %w", curr, err)
+		}
+		return pinned, int(slab.Length), nil
+	})
+}
 
-	if metadata.Key != nil {
-		w = decrypt(metadata.Key, w, 0)
+// DownloadSharedObject downloads a shared object from a shared URL
+func (s *SDK) DownloadSharedObject(ctx context.Context, w io.Writer, url string, opts ...DownloadOption) error {
```
ChrisSchinnerl, line 381: Reminder to add a test before merging this

sdk/client.go
```diff
@@ -355,38 +390,67 @@ func (s *SDK) Download(ctx context.Context, w io.Writer, metadata Object, opts .
...
+		if curr >= len(obj.Slabs) {
+			return slabs.PinnedSlab{}, 0, nil
+		}
+		slab := obj.Slabs[curr]
+		curr++
+		return slab.PinnedSlab, int(slab.Length), nil
+	})
+}
+
+type slabIterFn func() (slabs.PinnedSlab, int, error)
```
ChrisSchinnerl: This could return a `SharedObjectSlab` since we will need the `Offset` eventually anyway.

api/app/client.go
peterjan: It was at some point, removed it.

ChrisSchinnerl (approved):

chris124567 (approved):

PR #426, opened 2025-09-11
Remove `sdk.Object` and `sdk.Slab`

peterjan:
It's a little hard to see through the changes here, but it essentially removes the types we defined in the `sdk` package. 

It boils down to this: upload in the SDK now takes a `cipher.StreamReader` which moves the encryption key to the caller side entirely. It will also return `slabs.SlabSlice`s. Obviously whatever upload returns can be directly fed into download, where I made the same changes.

I also want to change how we save an object, which is closely tied to this change in my opinion.
Instead of doing `POST /objects` and sending it a `slabs.Object` (which contains timestamps that were ignored), I now `PUT /objects/{:key}` and send only the slabs and the metadata. I like that better because if SDK users want to save an object they  only need to think of an object key, some object metadata and combine that with the slabs they received from the upload method.

I considered returning a `slabs.Object` from the upload method and save it, but that would require passing in an object key (and metadata) which is kind of a nuisance if you don't care about objects.


n8mgr (changes requested):
Break this up into more focused changes

peterjan:
Closed, see https://github.com/SiaFoundation/indexd/pull/429#discussion_r2348039690

PR #427, opened 2025-09-11
Oneshot for pinning shared objects

### LINKED ISSUE #422: Oneshot for pinning shared objects
Add a new endpoint to pin all of a shared object's slabs and save the object. This should be atomic, so will need a new db method.

chris124567:
Close #422


api/app/app.go
ChrisSchinnerl, line 1: Reminder to update the openapi spec

persist/postgres/sectors.go
```diff
@@ -250,6 +250,130 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
...
+		var pinnedData, maxPinnedData uint64
+		err := tx.QueryRow(ctx, "SELECT id, pinned_data, max_pinned_data FROM accounts WHERE public_key = $1", sqlPublicKey(account)).Scan(&accountID, &pinnedData, &maxPinnedData)
+		if errors.Is(err, sql.ErrNoRows) {
+			return accounts.ErrNotFound
+		} else if err != nil {
+			return err
+		}
+
+		for _, slab := range shared.Slabs {
+			// insert slab
```
ChrisSchinnerl, line 216: Is this a copy paste from `PinSlab`? If so, I'd probably isolate it into a `pinSlab` that we can call in both places just like you did for `saveObject`.

ChrisSchinnerl (changes requested):

persist/postgres/sectors.go
```diff
@@ -250,6 +250,130 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
 	})
 }
 
+// PinSharedObject pins slabs from the shared object to the provided account
+// and saves the resultant object.
+func (s *Store) PinSharedObject(ctx context.Context, account proto.Account, shared slabs.SharedObject) error {
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		var accountID int64
```
peterjan, line 200: These 6 lines are duplicated between here and `PinSlab`, it's only 6 lines but having a helper for this would also improve readability.

persist/postgres/sectors.go
```diff
@@ -250,6 +250,130 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
...
+			var slabID int64
+			var existingSlab bool
+			err = tx.QueryRow(ctx, `
+			INSERT INTO slabs (digest, encryption_key, min_shards)
+			VALUES ($1, $2, $3)
+			ON CONFLICT (digest) DO UPDATE SET pinned_at = NOW()
+			RETURNING id
+			`, sqlHash256(slab.ID), sqlHash256(slab.EncryptionKey), slab.MinShards).Scan(&slabID)
+			if errors.Is(err, sql.ErrNoRows) {
+				// slab already exists, fetch its slab id
```
peterjan: This looks fishy because you do `ON CONFLICT (digest) DO UPDATE SET pinned_at = NOW()`. Is there a unit test that asserts this case is being handled properly?

persist/postgres/sectors.go
```diff
@@ -250,6 +250,130 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
...
+			// and if not, update the account's pinned data NOTE: we perform this
+			// check here since we need to know if the slab is a new slab or whether
+			// it was just repinned.
+			if res.RowsAffected() > 0 {
+				newPinnedData := pinnedData + proto.SectorSize*uint64(len(slab.Sectors))
+				if newPinnedData > maxPinnedData {
+					return accounts.ErrStorageLimitExceeded
+				}
+
+				_, err := tx.Exec(ctx, `UPDATE accounts SET pinned_data = $1 WHERE id = $2`, newPinnedData, accountID)
```
peterjan: Probably want to either extend `TestPinSlabs` that asserts `pinned_data` is updated correctly when you pin shared objects. That or add a dedicated `TestPinnedData` test that pins both regular and shared objects to ensure pinned data is updated as expected.

chris124567:
@n8mgr like this? https://github.com/SiaFoundation/indexd/pull/427/commits/8fa62be3053101d1cd96e525f9beea3f6884d8ca

n8mgr:
@chris124567 merge conflicts

n8mgr (dismissed):

ChrisSchinnerl (commented):
Got a few leftover comments that need to be addressed/discussed

chris124567:
CI failure seems to be some sort of docker issue...

persist/postgres/sectors.go
chris124567: Since the separate `PinSharedObject` store function with logic copypasted from `PinSlab` was removed and I just adapted `PinSlab` to pin multiple slabs, I think this is no longer necessary.

api/app/app.go
```diff
@@ -300,12 +300,67 @@ func (a *app) handlePOSTSlabs(jc jape.Context, pk types.PublicKey) {
...
+	jc.Encode(slabIDs[0])
+}
+
+func (a *app) handlePOSTObjectsShared(jc jape.Context, pk types.PublicKey) {
+	var shared slabs.SharedObject
+	if jc.Decode(&shared) != nil {
+		return
+	}
+
+	var toPin []slabs.SlabPinParams
```
peterjan: need to call `Validate` on the pin params after building them

api/app/app.go
```diff
@@ -300,12 +300,67 @@ func (a *app) handlePOSTSlabs(jc jape.Context, pk types.PublicKey) {
...
+		}
+
+		toPin = append(toPin, slabs.SlabPinParams{
+			EncryptionKey: slab.EncryptionKey,
+			MinShards:     slab.MinShards,
+			Sectors:       sectors,
+		})
+	}
+
+	if _, err := a.slabs.PinSlabs(jc.Request.Context(), proto.Account(pk), time.Now(), toPin...); err != nil {
```
peterjan: maybe we should check the slab digests here to be 100% sure we pinned them correctly

api/app/app.go
```diff
@@ -300,12 +300,67 @@ func (a *app) handlePOSTSlabs(jc jape.Context, pk types.PublicKey) {
...
 		return
+	} else if len(slabIDs) == 0 {
+		jc.Error(fmt.Errorf("PinSlabs did not return any slab IDs"), http.StatusInternalServerError)
 	}
 
-	jc.Encode(slabID)
+	jc.Encode(slabIDs[0])
+}
+
+func (a *app) handlePOSTObjectsShared(jc jape.Context, pk types.PublicKey) {
```
peterjan: I'm not sure how the others feel about this but I feel this is quite a lot of logic for the API handler. Seeing as we have managers, and currently most of them do nothing more than wrap store methods, we might as well move some of the business logic over to the managers. In this case we could add `ShareObject` for instance that wraps both the pinning and the saving of the object. This would also facilitate unit testing.

persist/postgres/sectors.go
```diff
@@ -192,129 +192,130 @@ func (s *Store) markFailingSectorsLostBatch(ctx context.Context, hostKey types.P
...
-// the provided account.  The last used timestamp of the account is updated.
-func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrityCheck time.Time, slab slabs.SlabPinParams) (slabs.SlabID, error) {
-	digest, err := slab.Digest()
-	if err != nil {
-		return slabs.SlabID{}, fmt.Errorf("failed to calculate slab digest: %w", err)
-	}
-	return digest, s.transaction(ctx, func(ctx context.Context, tx *txn) (err error) {
+// PinSlabs adds slabs to the database for pinning. The slab are associated
+// with the provided account.
+func (s *Store) PinSlabs(ctx context.Context, account proto.Account, nextIntegrityCheck time.Time, toPin ...slabs.SlabPinParams) ([]slabs.SlabID, error) {
```
peterjan, line 197: I did a quick check and I think we never call `PinSlabs` with something other than `time.Now()` for `nextIntegrityCheck`. If that is the case we might as well drop it and cleanup the method signature a bit.

peterjan: I was a little surprised actually that we never pass a future time. Shouldn't we pick a time between now and a week from now or something to smear them out a bit.

peterjan (changes requested):

persist/postgres/sectors.go
ChrisSchinnerl: Out of scope for this PR

api/app/app.go
ChrisSchinnerl: Yeah we could add a `PinSharedObject` method to the slabs manager and call that instead here. 

But we need to make sure we still check whether the validation of the params failed and then return a `BadRequest` status code.

api/app/app.go
```diff
@@ -300,12 +300,67 @@ func (a *app) handlePOSTSlabs(jc jape.Context, pk types.PublicKey) {
...
+
+		toPin = append(toPin, slabs.SlabPinParams{
+			EncryptionKey: slab.EncryptionKey,
+			MinShards:     slab.MinShards,
+			Sectors:       sectors,
+		})
+	}
+
+	if _, err := a.slabs.PinSlabs(jc.Request.Context(), proto.Account(pk), time.Now(), toPin...); err != nil {
+		jc.Error(err, http.StatusInternalServerError)
```
ChrisSchinnerl: I'd use `jc.Check` here.

api/app/app.go
```diff
@@ -300,12 +300,67 @@ func (a *app) handlePOSTSlabs(jc jape.Context, pk types.PublicKey) {
...
+		Slabs: objSlabs,
+		Meta:  shared.Meta,
+	}
+
+	err := a.slabs.SaveObject(jc.Request.Context(), proto.Account(pk), obj)
+	if errors.Is(err, slabs.ErrObjectMetadataLimitExceeded) || errors.Is(err, slabs.ErrObjectMinimumSlabs) {
+		jc.Error(err, http.StatusBadRequest)
+		return
+	} else if err != nil {
+		jc.Error(err, http.StatusInternalServerError)
```
ChrisSchinnerl: Same here. 
```go
else if jc.Check("failed to save object", err) != nil {
  return
}```

**Review by ChrisSchinnerl — CHANGES_REQUESTED** [review]

File: api/app/app.go (hunk 6 above)
Comment (chris124567) [reply api/app/app.go]: Do you mean in the test?

File: api/app/app.go [hunk 11]
```diff
@@ -300,12 +302,30 @@ func (a *app) handlePOSTSlabs(jc jape.Context, pk types.PublicKey) {
...
+}
+
+func (a *app) handlePOSTObjectsShared(jc jape.Context, pk types.PublicKey) {
+	var shared slabs.SharedObject
+	if jc.Decode(&shared) != nil {
+		return
+	}
+
+	err := a.slabs.PinSharedObject(jc.Request.Context(), proto.Account(pk), shared)
+	if err != nil && (strings.Contains(err.Error(), "failed to validate slab") || errors.Is(err, slabs.ErrObjectMetadataLimitExceeded) || errors.Is(err, slabs.ErrObjectMinimumSlabs)) {
```
peterjan: `slabs.ErrInvalidSlab` would be nice

slabs/objects.go
```diff
@@ -107,3 +107,51 @@ func (m *SlabManager) ListObjects(ctx context.Context, account proto.Account, cu
...
 	return m.store.SharedObject(ctx, key)
 }
+
+// PinSharedObject is a helper to pin all the slabs in a shared object and save
+// the object to the account.
+func (m *SlabManager) PinSharedObject(ctx context.Context, account proto.Account, shared SharedObject) error {
+	var toPin []SlabPinParams
+	for _, slab := range shared.Slabs {
+		sectors := make([]PinnedSector, 0, len(slab.Sectors))
+		for _, sector := range slab.Sectors {
```
peterjan: Do we really need to deep copy the sectors here? Seems ok to pass them to `PinSlabs` as is.

slabs/objects.go
```diff
@@ -107,3 +107,51 @@ func (m *SlabManager) ListObjects(ctx context.Context, account proto.Account, cu
...
+			MinShards:     slab.MinShards,
+			Sectors:       sectors,
+		}
+		if err := s.Validate(); err != nil {
+			return fmt.Errorf("failed to validate slab: %w", err)
+		}
+		toPin = append(toPin, s)
+	}
+
+	if _, err := m.PinSlabs(ctx, account, time.Now(), toPin...); err != nil {
```
peterjan: We could assert the `SlabIDs` that are being returned after pinning here match the slab IDs of the shared object. Actually I think it's really unnecessary for `PinSlabs` to return the IDs in the first place since we calculate them using the `SlabPinParams`. 

Probably what we should do is calculate the `Digest` in the `Validate` to ensure the digest can be calculated and then just have `PinSlabs` return an error. If you want to slab IDs you can call digest on the parameters you just passed to `PinSlabs`. This is out of scope for this PR but maybe we can make it into an issue, and at the same time get rid of `nextIntegrityCheck time.Time`.

peterjan (dismissed):

slabs/objects.go
chris124567: Nice catch the reason for this was that they were not the same type at the time the PR was made

api/app/app.go
```diff
@@ -300,12 +301,30 @@ func (a *app) handlePOSTSlabs(jc jape.Context, pk types.PublicKey) {
...
+	var shared slabs.SharedObject
+	if jc.Decode(&shared) != nil {
+		return
+	}
+
+	err := a.slabs.PinSharedObject(jc.Request.Context(), proto.Account(pk), shared)
+	if errors.Is(err, slabs.ErrInvalidSlab) || errors.Is(err, slabs.ErrObjectMetadataLimitExceeded) || errors.Is(err, slabs.ErrObjectMinimumSlabs) {
+		jc.Error(err, http.StatusBadRequest)
+		return
+	} else if jc.Check("failed to save object", err) != nil {
```
peterjan:
```suggestion
	} else if jc.Check("failed to pin shared object", err) != nil {
```

peterjan (dismissed):

slabs/objects.go
```diff
@@ -107,3 +110,52 @@ func (m *SlabManager) ListObjects(ctx context.Context, account proto.Account, cu
...
+	digests, err := m.PinSlabs(ctx, account, time.Now(), toPin...)
+	if err != nil {
+		return fmt.Errorf("failed to pin slabs: %w", err)
+	} else if len(digests) != len(shared.Slabs) {
+		return fmt.Errorf("some slabs were not pinned: %w", err)
+	}
+
+	for i, slab := range shared.Slabs {
+		if slab.ID != digests[i] {
+			return fmt.Errorf("unexpected slab was pinned: expected %v, got %v", slab.ID, digests[i])
```
ChrisSchinnerl: We already know that `PinSlabs` is going to produce the right digests due to testing. So if this assertion fails it will already fail before calling `PinSlabs`. So check that the digest of each `SlabPinParams` matches `slab.ID` after `s.Validate`. We shouldn't pin slabs if we know that we are going to return an error anyway and then leave the user with slabs that don't match their expectations pinned to their account.

If we don't have that already we should also extend the test where we pin a shared object to corrupt an id, make sure pinning fails with a `StatusBadRequest` and then check no slabs were pinned.

ChrisSchinnerl (changes requested):

slabs/objects.go
```diff
@@ -107,3 +110,50 @@ func (m *SlabManager) ListObjects(ctx context.Context, account proto.Account, cu
...
+		s := SlabPinParams{
+			EncryptionKey: slab.EncryptionKey,
+			MinShards:     slab.MinShards,
+			Sectors:       slab.Sectors,
+		}
+		if err := s.Validate(); err != nil {
+			return fmt.Errorf("%w: %w", ErrInvalidSlab, err)
+		}
+
+		if digest, err := s.Digest(); err != nil {
```
peterjan: This belongs in the `Validate` method imo

slabs/objects.go
```diff
@@ -73,6 +73,9 @@ var (
 	// ErrObjectUnpinnedSlab is returned when an user attempts to save an
 	// object containing a slab that is not pinned to their account.
 	ErrObjectUnpinnedSlab = errors.New("object contains unpinned slab")
+
+	// ErrInvalidSlab is returned when a slab in the object contains an invalid slab.
+	ErrInvalidSlab = errors.New("object contains invalid slab")
```
peterjan: I wouldn't mention "objects" in the invalid slab error. We return this to users when they pin slabs for instance, that might be confusing.

peterjan (changes requested):
Has some conflicts but other than that this looks good to merge. Left two comments though, since it's a user facing error message we should probably update it.

n8mgr (approved):

n8mgr (changes requested):
Since we also want to validate the signature when pinning, let's remove the shared object endpoint and have the user use the existing `SaveObject` endpoint. The only necessary change would be to make pin slabs accept a slice of `[]SlabPinParams`

openapi/app.yml
```diff
@@ -226,6 +226,26 @@ paths:
         "404":
           description: Object not found
 
+  /objects/shared:
```
peterjan: This should be removed now that we got rid of the endpoint.

peterjan (changes requested):
Looks like there's still a `PinSharedObject` on the mock store that's now unused.

n8mgr (dismissed):

ChrisSchinnerl (dismissed):

slabs/objects.go
```diff
@@ -80,6 +80,9 @@ var (
 	// ErrObjectUnpinnedSlab is returned when an user attempts to save an
 	// object containing a slab that is not pinned to their account.
 	ErrObjectUnpinnedSlab = errors.New("object contains unpinned slab")
+
+	// ErrInvalidSlab is returned when a slab in the object contains an invalid slab.
```
peterjan:
```suggestion
	// ErrInvalidSlab is returned when an invalid slab is being pinned
```

peterjan (dismissed):

ChrisSchinnerl (approved):

n8mgr (approved):

PR #429, opened 2025-09-12
Update SDK Upload/Download signature

peterjan:
This PR updates the SDK to have upload/download receive a cipher reader/writer and return slabs. This sets us up to get rid of the `Object` and `Slab` types in the `sdk` package.

### DIFF
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"crypto/cipher"
 	"errors"
 	"fmt"
 	"io"
@@ -20,7 +21,6 @@ import (
 	"go.sia.tech/indexd/slabs"
 	"go.uber.org/zap"
 	"golang.org/x/crypto/chacha20"
-	"lukechampine.com/frand"
 )
 
 type (
@@ -29,9 +29,6 @@ type (
 		parityShards uint8
 		hostTimeout  time.Duration
 		maxInflight  int
-
-		customKey         *[32]byte
-		disableEncryption bool
 	}
 
 	downloadOption struct {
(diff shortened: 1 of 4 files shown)


sdk/client.go
```diff
@@ -236,21 +233,7 @@ func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) (Ob
...
-	if !uo.disableEncryption {
-		if uo.customKey != nil {
-			obj.Key = uo.customKey
-		} else {
-			obj.Key = new([32]byte)
-			frand.Read(obj.Key[:])
-		}
-
-		r = encrypt(obj.Key, r, 0)
+		return nil, errors.New("redundancy must be at least 2x")
```
chris124567: we should make this 3x like in https://github.com/SiaFoundation/indexd/blob/155feafa61bf47ea4bcce835101e56d84a86e5b4/slabs/slabs.go#L128

chris124567 (dismissed):

sdk/client.go
```diff
@@ -224,7 +221,7 @@ top:
 // Upload uploads the data to hosts and pins it to the indexer.
 //
 // Returns the metadata of the slabs that were pinned
-func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) (Object, error) {
+func (s *SDK) Upload(ctx context.Context, r cipher.StreamReader, opts ...UploadOption) (uploaded []Slab, _ error) {
```
ChrisSchinnerl, line 224: To me this seems like strictly worse DX. We went from a one-liner for uploading an object that defaults to sane behavior (generating a new random key for each object) to a minimum of 2 calls and the potential for key-reuse. 
Iirc that was also the reason we didn't go for the more "elegant" `StreamReader` solution in the initial PR and instead went with the safer API design.  

I'm also not sure why we would need to get rid of the `Object` type. One `Object` is the API type and the other the SDK type. Both serve different purposes which seems alright to me if it makes for a nicer SDK.

peterjan: Ok, if you are fine with the `Object` in the sdk package I will close both PRs.

PR #431, opened 2025-09-12
Metrics: add `num_migrated`

peterjan:
references https://github.com/SiaFoundation/indexd/issues/392


chris124567 (dismissed):

persist/postgres/init.sql
```diff
@@ -306,7 +306,10 @@ CREATE TABLE sectors (
 
     -- data integrity
     next_integrity_check TIMESTAMP WITH TIME ZONE NOT NULL,
-    consecutive_failed_checks SMALLINT NOT NULL DEFAULT 0
+    consecutive_failed_checks SMALLINT NOT NULL DEFAULT 0,
+
+    -- statistics
+    num_migrated INTEGER NOT NULL DEFAULT 0
```
ChrisSchinnerl, line 312: Probably want a `DESC` index on this and probably also the total number of sector migrations on `sectors_stats` as a value we can expose as an overall metric.

peterjan: I thought about this actually but wasn't 100% sure whether we would want to add a field `num_migrated_sectors` to the `sector_stats` table which we increment by one very time we call `MigrateSector`, or if we would want to aggregate it manually across all sectors on the fly perhaps...

Do we need that index? I figured we would never return the sectors in an ordered fashion anyway, unless maybe for debugging purposes, so I didn't add it. It's a pretty big table.. so if we don't need the index we should probably hold off on adding it for the time being?

peterjan: I added `num_migrated_sectors` to the `sector_stats` table. I don't think we need the index for the time being. Let me know though if you want me to add it anyway.

ChrisSchinnerl: We can add it in whatever PR we expose it too. Once we know how we are going to expose it.

ChrisSchinnerl (dismissed):

ChrisSchinnerl (approved):

n8mgr (approved):

PR #435, opened 2025-09-12
Metrics: add `funds_spent` to `hosts`

peterjan:
references #392


chris124567 (dismissed):

n8mgr (dismissed):

client/client.go
```diff
@@ -350,6 +353,8 @@ func (c *HostClient) withRevision(ctx context.Context, contractID types.FileCont
 	if revised.Revision.RevisionNumber > contract.Revision.RevisionNumber {
 		if err := c.store.UpdateContractRevision(ctx, revised); err != nil {
 			c.log.Error("failed to update contract revision", zap.Error(err))
+		} else if err := c.store.UpdateHostUsage(ctx, c.hostKey, usage); err != nil {
```
ChrisSchinnerl, line 356: Should we also call this when we sync a revision with the host after a mismatch? We won't be able to recover what we spent it on exactly but maybe we can at least account for it in a metric. Maybe not for this PR but as an idea for a f/u.

ChrisSchinnerl (dismissed):

client/client.go
peterjan: So a `drift` field 😁? Yeah good catch I guess, we should probably track that indeed.

ChrisSchinnerl (approved):

PR #436, opened 2025-09-13
Increase initial contract funding based on number of active accounts

### LINKED ISSUE #298: Increase initial contract funding based on number of accounts
The current contract funding mechanism will immediately drain a contract if there are more than 10 accounts. The renter funds should increase based on the number of accounts. 

Bonus points: consider active accounts only

chris124567:
Close #298

With 100k accounts:
```
BenchmarkActiveAccounts-4            565           2087113 ns/op
```


persist/postgres/accounts.go
```diff
@@ -97,6 +97,15 @@ func (s *Store) HasAccount(ctx context.Context, ak types.PublicKey) (bool, error
 	return exists, nil
 }
 
+// ActiveAccounts returns the number of accounts active since the threshold
+// time.
+func (s *Store) ActiveAccounts(ctx context.Context, threshold time.Time) (count uint64, err error) {
+	err = s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		return tx.QueryRow(ctx, `SELECT COUNT(*) FROM accounts;`).Scan(&count)
```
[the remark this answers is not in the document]
chris124567: I don't want to have to change signature in future

accounts/manager.go
```diff
@@ -17,6 +17,17 @@ const (
 	accountFundBatch            = proto.MaxAccountBatchSize // equals max batch size used in replenish RPC
 	accountFundInterval         = time.Hour
 	accountExpBackoffMaxMinutes = 128
+
+	// We multiply the target amount of funds per host by the number of active
+	// accounts up to a certain point.  clampAccounts is the highest number of
+	// active accounts we will use for the purposes of giving hosts extra funding.
+	clampAccounts = 1000
```
ChrisSchinnerl: I don't mind clamping accounts for now but I think I remember @n8mgr disagreeing with this approach.

persist/postgres/accounts.go
```diff
@@ -97,6 +97,15 @@ func (s *Store) HasAccount(ctx context.Context, ak types.PublicKey) (bool, error
 	return exists, nil
 }
 
+// ActiveAccounts returns the number of accounts active since the threshold
+// time.
+func (s *Store) ActiveAccounts(ctx context.Context, threshold time.Time) (count uint64, err error) {
+	err = s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		return tx.QueryRow(ctx, `SELECT COUNT(*) FROM accounts WHERE last_used >= $1;`, threshold).Scan(&count)
```
ChrisSchinnerl, line 109: I wanted to suggest a materialized view at first but considering that it only takes 2ms for 100k accounts in the benchmark and we only call this once every full moon anyway it might be fine.

persist/postgres/sectors.go
```diff
@@ -149,7 +149,7 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
 	return digest, s.transaction(ctx, func(ctx context.Context, tx *txn) error {
 		var accountID int64
 		var pinnedData, maxPinnedData uint64
-		err := tx.QueryRow(ctx, "SELECT id, pinned_data, max_pinned_data FROM accounts WHERE public_key = $1", sqlPublicKey(account)).Scan(&accountID, &pinnedData, &maxPinnedData)
+		err := tx.QueryRow(ctx, "UPDATE accounts SET last_used = NOW() WHERE public_key = $1 RETURNING id, pinned_data, max_pinned_data", sqlPublicKey(account)).Scan(&accountID, &pinnedData, &maxPinnedData)
```
ChrisSchinnerl: I know that @n8mgr suggested this at the summit and it probably works well enough but I'd like to propose an alternative approach for him to weigh in on.

We could also check how many accounts we funded with a deposit > 0 in `FundAccounts` and then update the `last_used` for these accounts.

The advantage would be that it includes accounts that are only used for downloading (and never pinning) as well.

n8mgr: If you also update `last_used` when grabbing slab metadata you track downloads without relying on hosts. I would probably be okay with removing clamping if we track `last_used` on both ends.

slabs/slabs.go
```diff
@@ -146,10 +146,39 @@ func (s SlabPinParams) Validate() error {
...
+	}
 	return m.store.PinSlab(ctx, account, nextIntegrityCheck, slab)
 }
 
+// UnpinSlab removes the association between the account and the given slab. If
+// this slab was only referenced by the given account, it will also be deleted.
+// The sectors are potentially orphaned and will be removed by a background
+// process.
+func (m *SlabManager) UnpinSlab(ctx context.Context, account proto.Account, slabID SlabID) error {
+	if err := m.store.UpdateLastUsed(ctx, account); err != nil {
```
peterjan: This feels a little awkward to me, definitely something we'll overlook when we extend our API with new routes. Since it appears we want to update the `last_used` field on some routes, but not others, it makes more sense to add it closer to where we define routes. Would it not be better to pass a boolean to `wrapSignedAuth`? We could call it `updateLastUsed` or `trackAccountUsage` or something. If someone then adds a route he'll be forced to consider whether we want to update the timestamp yes or no.

Something like this:
```go
		"GET /slabs":            wrapCORS(wrapSignedAuth(a.handleGETSlabs, false)),
		"POST /slabs":           wrapCORS(wrapSignedAuth(a.handlePOSTSlabs, true)),
		"GET /slabs/:slabid":    wrapCORS(wrapSignedAuth(a.handleGETSlab, true)),
		"DELETE /slabs/:slabid": wrapCORS(wrapSignedAuth(a.handleDELETESlab, true)),
```

ChrisSchinnerl: I don't think we want it for all routes anyway. Adding it all the way up there in the API seems quite intrusive for something we probably want to be limited to whenever a user uploads and downloads. I'd go with what Nate suggested which is updating it exclusively on pinning slabs and fetching a slab for downloading.

accounts/manager.go
```diff
@@ -19,6 +19,12 @@ const (
 	accountExpBackoffMaxMinutes = 128
 )
 
+var (
+	// defaultFundTarget is the target amount of funds per host per active
+	// account.
+	defaultFundTarget = types.Siacoins(1).Div64(10)
```
ChrisSchinnerl: The fund target for accounts should still be 1SC.

The funding that we want to multiply by the number of active accounts is the contract funding. e.g. if we have 10 active accounts we want to fund contracts with at least 10SC to make sure we can fund 10 active accounts at once.

Doesn't make sense to pour more money into each user account just be cause we have more users.

accounts/manager.go
```diff
@@ -94,6 +101,15 @@ func (m *AccountManager) FundAccounts(ctx context.Context, host hosts.Host, cont
...
 	}
 
+	// target m.fundTarget, per account, per host
+	count, err := m.store.ActiveAccounts(ctx, time.Now().Add(-7*24*time.Hour))
+	if err != nil {
+		return fmt.Errorf("failed to get number of active accounts: %w", err)
+	} else if count == 0 {
+		count = 1
+	}
+	fundTarget := m.fundTarget.Mul64(count)
```
ChrisSchinnerl: As mentioned in the other comment. The contract funding should increase as we add more accounts. Not the per-account funding. So you want to look at `contractFunding` and the `minAllowance` clamping we do in there.

ChrisSchinnerl (changes requested):

accounts/manager.go
```diff
@@ -19,6 +19,12 @@ const (
 	accountExpBackoffMaxMinutes = 128
 )
 
+var (
+	// defaultFundTarget is the target amount of funds per host per active
```
ChrisSchinnerl: (nit) strictly speaking it's not active accounts but all accounts. Even inactive accounts will be kept at that target.

slabs/slabs.go
ChrisSchinnerl, line 1: We miss tests to make sure `PinSlab` and `PinnedSlab` update the account timestamp.

persist/postgres/accounts.go
```diff
@@ -97,6 +97,23 @@ func (s *Store) HasAccount(ctx context.Context, ak types.PublicKey) (bool, error
...
+func (s *Store) UpdateLastUsed(ctx context.Context, account proto.Account) error {
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `UPDATE accounts SET last_used = NOW() WHERE public_key = $1`, sqlPublicKey(account))
+		return err
+	})
+}
+
+// ActiveAccounts returns the number of accounts active since the threshold
+// time.
+func (s *Store) ActiveAccounts(ctx context.Context, threshold time.Time) (count uint64, err error) {
```
ChrisSchinnerl, line 107: Missing a test for this

persist/postgres/accounts.go
```diff
@@ -97,6 +97,23 @@ func (s *Store) HasAccount(ctx context.Context, ak types.PublicKey) (bool, error
...
 
+// UpdateLastUsed updates the time the account was last used to now.
+func (s *Store) UpdateLastUsed(ctx context.Context, account proto.Account) error {
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `UPDATE accounts SET last_used = NOW() WHERE public_key = $1`, sqlPublicKey(account))
+		return err
+	})
+}
+
+// ActiveAccounts returns the number of accounts active since the threshold
```
peterjan:
```suggestion
// ActiveAccounts returns the number of accounts that have been used since the threshold
```

slabs/slabs.go
```diff
@@ -140,10 +140,36 @@ func (s SlabPinParams) Validate() error {
...
+
 // Slabs returns the slabs with the given IDs from the database.
-func (m *SlabManager) Slabs(ctx context.Context, accountID proto.Account, slabIDs []SlabID) ([]Slab, error) {
-	return m.store.Slabs(ctx, accountID, slabIDs)
+func (m *SlabManager) Slabs(ctx context.Context, account proto.Account, slabIDs []SlabID) ([]Slab, error) {
+	return m.store.Slabs(ctx, account, slabIDs)
+}
+
+// PinnedSlab retrieves a pinned slab from the database by its ID.
+func (m *SlabManager) PinnedSlab(ctx context.Context, account proto.Account, slabID SlabID) (PinnedSlab, error) {
```
peterjan, line 166: I guess the manager needs a unit test now to assert `last_used` was updated for pinning a slab or fetching a pinned slab and not for its other methods.

slabs/slabs.go
```diff
@@ -145,10 +145,36 @@ func (s SlabPinParams) Validate() error {
 
 // PinSlab pins the given slab and associates it with the given account.
 func (m *SlabManager) PinSlab(ctx context.Context, account proto.Account, nextIntegrityCheck time.Time, slab SlabPinParams) (SlabID, error) {
+	if err := m.store.UpdateLastUsed(ctx, account); err != nil {
```
n8mgr: This should be atomic in `PinSlab`

slabs/slabs.go
```diff
@@ -145,10 +145,36 @@ func (s SlabPinParams) Validate() error {
...
 // Slabs returns the slabs with the given IDs from the database.
-func (m *SlabManager) Slabs(ctx context.Context, accountID proto.Account, slabIDs []SlabID) ([]Slab, error) {
-	return m.store.Slabs(ctx, accountID, slabIDs)
+func (m *SlabManager) Slabs(ctx context.Context, account proto.Account, slabIDs []SlabID) ([]Slab, error) {
+	return m.store.Slabs(ctx, account, slabIDs)
+}
+
+// PinnedSlab retrieves a pinned slab from the database by its ID.
+func (m *SlabManager) PinnedSlab(ctx context.Context, account proto.Account, slabID SlabID) (PinnedSlab, error) {
+	if err := m.store.UpdateLastUsed(ctx, account); err != nil {
```
n8mgr: This should be atomic in `PinnedSlab`

n8mgr (dismissed):

contracts/form.go
```diff
@@ -239,6 +239,13 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
 		addHost(host)
 	}
 
+	activeAccounts, err := cm.store.ActiveAccounts(ctx, time.Now().Add(-24*7*time.Hour))
```
peterjan: `nit` `const accountActivityThreshold = -24*7*time.Hour` or something along those lines

persist/postgres/accounts_test.go
```diff
@@ -719,6 +719,36 @@ func TestServiceAccounts(t *testing.T) {
...
 }
 
+func TestActiveAccounts(t *testing.T) {
+	// prepare database
+	store := initPostgres(t, zaptest.NewLogger(t).Named("postgres"))
+
+	now := time.Now()
+	const day = 24 * 7 * time.Hour
+
+	durations := []time.Duration{day, 2 * day, 3 * day, 7 * day, 10 * day, 14 * day, 30 * day}
```
peterjan: Verifying if a value exceeds a threshold is generally done by setting `last_used` to the exact `threshold` and then querying the active accounts using `threshold` plus or minus one second. This is also fine it's just a little excessive and harder to read imo.

openapi/components.yml
```diff
@@ -415,6 +415,10 @@ components:
           type: string
           format: uri
           description: The homepage of the app
+        lastUsed:
+          type: string
+          format: date-time
+          description: The last time the account was used
```
peterjan: Maybe we should specify we only update it when pinning a slab or fetching a pinned slab. Not too sure myself actually but it came to mind when I read it.

slabs/slabs.go
```diff
@@ -148,7 +148,28 @@ func (m *SlabManager) PinSlab(ctx context.Context, account proto.Account, nextIn
...
 // Slabs returns the slabs with the given IDs from the database.
-func (m *SlabManager) Slabs(ctx context.Context, accountID proto.Account, slabIDs []SlabID) ([]Slab, error) {
-	return m.store.Slabs(ctx, accountID, slabIDs)
+func (m *SlabManager) Slabs(ctx context.Context, account proto.Account, slabIDs []SlabID) ([]Slab, error) {
+	return m.store.Slabs(ctx, account, slabIDs)
+}
+
+// PinnedSlab retrieves a pinned slab from the database by its ID.  If account
+// is not nil, the last used field of that account will be updated.
+func (m *SlabManager) PinnedSlab(ctx context.Context, account *proto.Account, slabID SlabID) (PinnedSlab, error) {
```
peterjan: The pointer is fishy, why is that necessary? If it's necessary for testing/benchmarking purposes only we should work around it and avoid passing a pointer.

peterjan (changes requested):

persist/postgres/slabs.go
```diff
@@ -59,9 +60,15 @@ func (s *Store) Slab(ctx context.Context, slabID slabs.SlabID) (slab slabs.Slab,
 }
 
 // PinnedSlab retrieves a pinned slab from the database by its ID.
-func (s *Store) PinnedSlab(ctx context.Context, slabID slabs.SlabID) (slab slabs.PinnedSlab, err error) {
+func (s *Store) PinnedSlab(ctx context.Context, account *proto.Account, slabID slabs.SlabID) (slab slabs.PinnedSlab, err error) {
```
ChrisSchinnerl: Shouldn't be a pointer. We always have that information in the app api.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

n8mgr (approved):

PR #440, opened 2025-09-15
Allow sorting usable hosts by distance

### LINKED ISSUE #368: Hosts sort by distance
if we store lat/long in the database with #366, we can use postgis to sort by proximity to our coordinates which may let apps improve speed/distribution.

peterjan:
Benchmark results:

| Benchmark                                    | Iterations | ns/op      |
|----------------------------------------------|------------|------------|
| UsableHosts_DefaultLimit                     | 823        | 1,779,516  |
| UsableHosts_WithProtocol_DefaultLimit        | 145        | 7,833,027  |
| UsableHosts_WithCountry_DefaultLimit         | 562        | 1,979,408  |
| UsableHosts_WithProximity_DefaultLimit       | 74         | 14,477,565 |
| UsableHosts_MaxLimit                         | 356        | 3,237,296  |
| UsableHosts_WithProtocol_MaxLimit            | 45         | 24,678,305 |
| UsableHosts_WithCountry_MaxLimit             | 309        | 3,772,112  |
| UsableHosts_WithProximity_MaxLimit           | 74         | 15,136,980 |

Fixes https://github.com/SiaFoundation/indexd/issues/368


persist/postgres/hosts.go
```diff
@@ -629,8 +642,17 @@ WHERE
 	-- active contracts
 	EXISTS (SELECT 1 FROM contracts WHERE host_id = hosts.id AND state <= 1) AND
 	-- protocol filter
-	($4::smallint IS NULL OR EXISTS (SELECT 1 FROM host_addresses WHERE host_id = hosts.id AND protocol = $4::smallint))
-LIMIT $1 OFFSET $2;`, limit, offset, queryOpts.CountryCode, (*sqlNetworkProtocol)(queryOpts.Protocol))
+	($4::smallint IS NULL OR EXISTS (SELECT 1 FROM host_addresses WHERE host_id = hosts.id AND protocol = $4::smallint))`
+		args := []any{limit, offset, queryOpts.CountryCode, (*sqlNetworkProtocol)(queryOpts.Protocol)}
+
+		if queryOpts.SortOptions != nil {
+			baseQuery += `ORDER BY location <-> point($5, $6) `
```
peterjan, line 636: It would be more performant to change the usable hosts query to use a `JOIN LATERAL` and k-NN sort the hosts there to then afterwards apply the more expensive joins and various filters we have. The downside is that we have to limit that because otherwise it would sort all hosts by location, before filtering them.

I haven't done that because it complicates the query and because we should probably limit the inner select to use a pool of hosts that get ordered, which then gets filtered down according to the other filters. Limiting the inner select could theoretically mess up paging. I added a benchmark, I feel it's fast enough to keep the query we have.

api/options.go
```diff
@@ -23,10 +24,29 @@ var (
 	ErrInvalidLimit = fmt.Errorf("limit must between 1 and %d", maxLimit)
 )
 
+type (
+	// SortOptions configures sorting of query results.
+	SortOptions struct {
+		SortBy  string
+		SortDir string
+		SortCtx any
```
n8mgr: I would prefer a more descriptive parameter like `location` over a catch-all for every sort and preferably avoid the type assertion.

peterjan: It was an attempt to make sorting feel more generic across our API, which is not a bad thing imo. Having to pass the location as context is indeed unfortunate. I'm pretty sure we'll be adding more generic sorting at one point though. I don't mind getting rid of it if you all don't like it.

ChrisSchinnerl: +1 on strong types. I'd rather add one sorting type per endpoint and one strongly typed field per value over type assertions. Makes it easier to see and enforce what my options are as a dev over a generic type.

The client will still be very nice with the `SortByDistance` pattern.

peterjan: I think of this more from an API design perspective, e.g. we'll have `/hosts?location=(lat,lng)` to sort hosts by distance, and then something like `/hosts?sortby=uptime&sortdir=desc`. Right now the sort direction is obvious but we might run into a case where we want to sort on something that's not necessary a column and still want to pass a sort direction. The type assertion is definitely unfortunate, I'll change it. 

edit: removing it felt nice so probably better indeed

hosts/options.go
```diff
@@ -49,23 +50,34 @@ type (
 	// UsableHostsQueryOpts holds optional parameters for querying usable
 	// hosts.
 	UsableHostsQueryOpts struct {
-		Protocol    *chain.Protocol
 		CountryCode *string
+		Protocol    *chain.Protocol
+
+		// sorting
+		Location *pgtype.Point
```
chris124567: nit: This sort of assumes that the underlying store is Postgres but it is outside of the store.  I think we could use a `[2]float64` here and not lose any clarity.

peterjan: nice

ChrisSchinnerl (dismissed):

persist/postgres/hosts.go
```diff
@@ -632,9 +632,9 @@ WHERE
 	($4::smallint IS NULL OR EXISTS (SELECT 1 FROM host_addresses WHERE host_id = hosts.id AND protocol = $4::smallint)) `
 		args := []any{limit, offset, queryOpts.CountryCode, (*sqlNetworkProtocol)(queryOpts.Protocol)}
 
-		if queryOpts.Location != nil {
+		if len(queryOpts.Location) == 2 {
```
chris124567: `len(queryOpts.Location)` will always be 2 right?  It could be `*[2]float64` instead and then you could check if it's `!= nil`.

api/app/app.go
```diff
@@ -165,13 +165,28 @@ func (a *app) handleGETHosts(jc jape.Context, _ types.PublicKey) {
...
 	var opts []hosts.UsableHostQueryOpt
-	if p != "" {
-		opts = append(opts, hosts.WithProtocol(chain.Protocol(p)))
+	if protocol != "" {
+		opts = append(opts, hosts.WithProtocol(chain.Protocol(protocol)))
 	}
 	if countryCode != "" {
 		opts = append(opts, hosts.WithCountry(countryCode))
 	}
+	if locationStr != "" {
```
n8mgr: This check is duplicated from above. You can move `opts` up and build it while parsing the options to avoid duplicating the checks here.

ChrisSchinnerl (dismissed):

n8mgr (approved):

ChrisSchinnerl (approved):

PR #442, opened 2025-09-16
Fix integrity checks

ChrisSchinnerl:
Almost all integrity checks failed on zeus due to expired prices. This PR fixes that and a few other things by changing the following:

- `WithScannedHost` doesn't execute the closure when the prices are obviously outdated
- `WithScannedHost` doesn't execute the closure twice upon successful execution
- `performIntegrityChecksForHost` calls `WithScannedHost` per batch to guarantee every batch has up-to-date prices

After deploying this on Zeus, things are looking a lot better
```
indexd-1  | DEBUG	slabs.integrity	performed integrity checks	{"hostKey": "ed25519:c6e4c0b5d05b561c0d6b7718baf2841c6ecd3c5e2bcdb2c6a40bdc7a48e88a0a", "lost": 0, "failed": 0, "successful": 1000}
indexd-1  | DEBUG	slabs.integrity	performed integrity checks	{"hostKey": "ed25519:92282dfa2cb8e962b07f6d2f78ff0a78f336b75d810844c5104bb054595a21ff", "lost": 0, "failed": 0, "successful": 1000}
```

### DIFF
--- a/hosts/manager.go
+++ b/hosts/manager.go
@@ -278,9 +278,15 @@ func (m *HostManager) WithScannedHost(ctx context.Context, hk types.PublicKey, f
 		return fmt.Errorf("%w: blocked=%t, usable=%t, networks=%d", ErrBadHost, host.Blocked, host.Usability.Usable(), len(host.Networks))
 	}
 
-	// optimistically call the function
-	if err := fn(host); err != nil && !errors.Is(err, proto4.ErrPricesExpired) {
-		return err
+	// optimistically call the function if the prices are still valid for a
+	// bit
+	const validPriceBuf = 5 * time.Minute
+	if host.Settings.Prices.ValidUntil.After(time.Now().Add(validPriceBuf)) {
+		if err := fn(host); err != nil && !errors.Is(err, proto4.ErrPricesExpired) {
+			return err
+		} else if err == nil {
+			return nil // 'fn' succeeded so we're done
+		}
 	}
 	logger.Debug("host has outdated prices, rescan")
 
--- a/slabs/dataintegrity.go
+++ b/slabs/dataintegrity.go
@@ -12,13 +12,13 @@ import (
 )
 
 func (m *SlabManager) performIntegrityChecksForHost(ctx context.Context, hostKey types.PublicKey, logger *zap.Logger) {
-	hostLogger := logger.With(zap.Stringer("hostKey", hostKey))
+	logger = logger.With(zap.Stringer("hostKey", hostKey))
 
 	const batchSize = 1000 // batch size for sector retrieval
 	for interrupt := false; !interrupt; {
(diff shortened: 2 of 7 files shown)


slabs/dataintegrity.go
```diff
@@ -11,12 +11,12 @@ import (
 	"go.uber.org/zap"
 )
 
-func (m *SlabManager) performIntegrityChecksForHost(ctx context.Context, host hosts.Host, logger *zap.Logger) {
-	hostLogger := logger.With(zap.Stringer("hostKey", host.PublicKey))
+func (m *SlabManager) performIntegrityChecksForHost(ctx context.Context, hostKey types.PublicKey, logger *zap.Logger) {
+	hostLogger := logger.With(zap.Stringer("hostKey", hostKey))
```
n8mgr: Override `logger` here. You don't need two named references.

slabs/dataintegrity.go
```diff
@@ -26,7 +26,11 @@ func (m *SlabManager) performIntegrityChecksForHost(ctx context.Context, host ho
 		interrupt = len(toCheck) < batchSize
 
 		// perform integrity checks
-		results, err := m.verifier.VerifySectors(ctx, host, toCheck)
+		var results []CheckSectorsResult
+		err = m.hm.WithScannedHost(ctx, hostKey, func(host hosts.Host) error {
+			results, err = m.verifier.VerifySectors(ctx, host, toCheck)
```
n8mgr: This function is still a potential problem, no? All calls in the batch still share the same price table and there's no way to refresh it without retrying the whole batch. Ideally it won't take 30m to verify 1000 sectors, but it theoretically could. Reverifying 1000 sectors because 1 failed half an hour in is not ideal.

ChrisSchinnerl: Yeah it could still fail if checking 1000 sectors takes more than 5 minutes.
I added a loop to check prices in `VerifySectors` right before performing the RPC. That should fix it for all hosts.

hosts/manager.go
```diff
@@ -278,9 +278,15 @@ func (m *HostManager) WithScannedHost(ctx context.Context, hk types.PublicKey, f
 		return fmt.Errorf("%w: blocked=%t, usable=%t, networks=%d", ErrBadHost, host.Blocked, host.Usability.Usable(), len(host.Networks))
 	}
 
-	// optimistically call the function
-	if err := fn(host); err != nil && !errors.Is(err, proto4.ErrPricesExpired) {
-		return err
+	// optimistically call the function if the prices are still valid for a
+	// bit
+	const validPriceBuf = 5 * time.Minute
```
peterjan, line 283: We should probably make this a minimum requirement for the price table validity when scanning hosts.

peterjan (approved):

hosts/manager.go
ChrisSchinnerl: The minimum requirement we enforce is actually higher. 15 minutes I think.

PR #445, opened 2025-09-16
Add slab pruning endpoint

### LINKED ISSUE #407: Add slab pruning endpoint to objects API
We should provide an endpoint that prunes all pinned slabs of a user not currently connected to an object to allow easy cleanup of orphaned slabs.

chris124567:
Close #407 

With 50,000 accounts, each having 5 objects (containing 3 slabs each) and 15 slabs attached to no objects
```
BenchmarkPruneSlabs-4                 24          48661396 ns/op
```


persist/postgres/sectors.go
```diff
@@ -299,74 +299,122 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
...
+		return fmt.Errorf("failed to disable seqscan: %w", err)
+	}
+
+	var args []sqlHash256
+	for _, slabID := range slabIDs {
+		args = append(args, sqlHash256(slabID))
+	}
+
+	// delete the association between the account and the slab
+	rows, err := tx.Query(ctx, `DELETE FROM account_slabs
```
peterjan: This should be better for the query planner:

```postgres
DELETE FROM account_slabs as
USING slabs s
WHERE as.account_id = $1
  AND s.digest = ANY($2)
  AND as.slab_id = s.id
RETURNING as.slab_id
```

persist/postgres/sectors.go
```diff
@@ -299,74 +299,122 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
...
+	for rows.Next() {
 		var sID int64
-		err := tx.QueryRow(ctx, `
-			DELETE FROM account_slabs
-			WHERE
-				account_id = (SELECT id FROM accounts WHERE public_key = $1) AND
-				slab_id = (SELECT id FROM slabs WHERE digest = $2)
-			RETURNING slab_id`, sqlPublicKey(accountID), sqlHash256(slabID)).Scan(&sID)
-		if errors.Is(err, sql.ErrNoRows) {
-			return slabs.ErrSlabNotFound
```
peterjan, line 335: This is an odd error to be returning here, returning `nil` is probably fine.

persist/postgres/sectors.go
```diff
@@ -299,74 +299,122 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
...
 
-// UnpinSlab removes the association between the account and the given slab. If
-// this slab was only referenced by the given account, it will also be deleted.
-// The sectors are potentially orphaned and will be removed by a background
-// process.
-func (s *Store) UnpinSlab(ctx context.Context, accountID proto.Account, slabID slabs.SlabID) error {
-	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
-		// delete the association between the account and the slab
+func (s *Store) unpinSlabs(ctx context.Context, tx *txn, accountID int64, slabIDs []slabs.SlabID) error {
+	if _, err := tx.Exec(ctx, `SET LOCAL enable_seqscan=FALSE`); err != nil {
```
peterjan: We should definitely avoid this at all costs. Is this the only way to make the query planner choose the optimal plan?

persist/postgres/sectors.go
```diff
@@ -299,74 +299,122 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
...
+	}
+
+	// ignore the slabs that are pinned by another account
+	rows, err = tx.Query(ctx, `SELECT slab_id FROM account_slabs WHERE slab_id = ANY($1)`, sIDs)
+	if err != nil {
+		return fmt.Errorf("failed to check if slab was pinned: %w", err)
+	}
+	defer rows.Close()
+
+	seen := make(map[int64]struct{})
```
peterjan, line 375: We can probably replace the `seen` map with a `DISTINCT` in the query.

persist/postgres/sectors.go
```diff
@@ -299,74 +299,122 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
...
 
-		// return early if the slab is pinned by another account
-		var pinned bool
-		err = tx.QueryRow(ctx, `SELECT EXISTS (SELECT 1 FROM account_slabs WHERE slab_id = $1)`, sID).Scan(&pinned)
-		if err != nil {
-			return fmt.Errorf("failed to check if slab was pinned: %w", err)
-		} else if pinned {
-			return nil
+	// get all of the slabs that are not pinned by another account
+	var notPinned []int64
```
peterjan: `nit` renaming this to `toDelete` or `toRemove` makes the call to `incrementNumSlabs` a little easier to understand

persist/postgres/sectors.go
```diff
@@ -299,74 +299,122 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
...
+		return fmt.Errorf("failed to decrement number of slabs: %w", err)
+	}
 
-		return nil
+	return nil
+}
+
+// UnpinSlab removes the association between the account and the given slab. If
+// this slab was only referenced by the given account, it will also be deleted.
+// The sectors are potentially orphaned and will be removed by a background
```
peterjan: We're deleting the sectors immediately right?

persist/postgres/sectors.go
```diff
@@ -299,74 +299,122 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
...
 }
 
-// UnpinSlab removes the association between the account and the given slab. If
-// this slab was only referenced by the given account, it will also be deleted.
-// The sectors are potentially orphaned and will be removed by a background
-// process.
-func (s *Store) UnpinSlab(ctx context.Context, accountID proto.Account, slabID slabs.SlabID) error {
-	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
-		// delete the association between the account and the slab
+func (s *Store) unpinSlabs(ctx context.Context, tx *txn, accountID int64, slabIDs []slabs.SlabID) error {
```
peterjan, line 320: I gave this to Chat GPT and it spit out this glorious multi-stage CTE. I haven't tested it but it looks almost perfect and it's very pretty. Note the `EXCEPT`, which I didn't know existed. It needs some work though, things like using `GREATEST` when it comes to updating pinned_data for example is def. fishy.

```sql
WITH
input_digests(digest) AS (
  SELECT UNNEST($2::bytea[])
),
slabs_to_unpin AS (
  SELECT s.id
  FROM slabs s
  JOIN input_digests i ON i.digest = s.digest
),
-- remove this account's pins, remember which slabs we touched
unlinked AS (
  DELETE FROM account_slabs a
  USING slabs_to_unpin su
  WHERE a.account_id = $1
    AND a.slab_id = su.id
  RETURNING a.slab_id
),
-- slabs still pinned by *any* other account
still_pinned AS (
  SELECT DISTINCT a.slab_id
  FROM account_slabs a
  JOIN unlinked u ON u.slab_id = a.slab_id
),
-- slabs no longer pinned by anyone
fully_unpinned AS (
  SELECT slab_id FROM unlinked
  EXCEPT
  SELECT slab_id FROM still_pinned
),
-- decrement this account's pinned bytes for everything it just unlinked
updated_acct AS (
  UPDATE accounts acc
  SET pinned_data = GREATEST(
      acc.pinned_data - (
        SELECT COALESCE(COUNT(*), 0) * $3
        FROM slab_sectors ss
        JOIN unlinked u ON u.slab_id = ss.slab_id
      ),
      0
  )
  WHERE acc.id = $1
  RETURNING 1
),
-- delete sectors that belong only to the fully-unpinned slabs and to no other slab
deleted_sectors AS (
  DELETE FROM sectors s
  USING slab_sectors ss
  WHERE ss.slab_id IN (SELECT slab_id FROM fully_unpinned)
    AND s.id = ss.sector_id
    AND NOT EXISTS (
      SELECT 1
      FROM slab_sectors ss2
      WHERE ss2.sector_id = ss.sector_id
        AND ss2.slab_id <> ss.slab_id
    )
  RETURNING s.id
),
deleted_slabs AS (
  DELETE FROM slabs s
  WHERE s.id IN (SELECT slab_id FROM fully_unpinned)
  RETURNING s.id
)
SELECT
  (SELECT COUNT(*) FROM unlinked)               AS unlinked_count,
  (SELECT COUNT(*) FROM fully_unpinned)         AS fully_unpinned_count,
  (SELECT COUNT(*) FROM deleted_sectors)        AS deleted_sectors_count,
  (SELECT COUNT(*) FROM deleted_slabs)          AS deleted_slabs_count;
```

persist/postgres/slabs.go
```diff
@@ -98,3 +99,51 @@ ORDER BY ss.slab_index ASC`, dbID)
...
+// object.
+func (s *Store) PruneSlabs(ctx context.Context, account proto.Account) error {
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		id, err := accountID(ctx, tx, account)
+		if err != nil {
+			return fmt.Errorf("failed to get account ID: %w", err)
+		}
+
+		rows, err := tx.Query(ctx, `SELECT s.digest
+FROM slabs s
```
peterjan, line 125: You could select from `account_slabs` and join on `slabs` - that way you select from the smallest most selective set. The query planner might do that for you but I don't think it does. I ran it through the benchmark and it made no difference but I think in production it probably would be slightly better.

persist/postgres/slabs_test.go
```diff
@@ -183,3 +186,175 @@ func TestPinnedSlab(t *testing.T) {
...
+	// prune slabs for acc2
+	if err := store.PruneSlabs(context.Background(), acc2); err != nil {
+		t.Fatal(err)
+	}
+
+	assertSlabs(acc1, slab2ID)
+	assertSlabs(acc2)
+}
+
+func BenchmarkPruneSlabs(b *testing.B) {
```
peterjan, line 312: I ran this on 5k accounts with 1k objects each and got 4s/op. That's actually not too bad I'd say but we will have to eventually get this down further.

persist/postgres/slabs.go
```diff
@@ -98,3 +99,51 @@ ORDER BY ss.slab_index ASC`, dbID)
 	})
 	return
 }
+
+// PruneSlabs prunes all pinned slabs of a user not currently connected to an
+// object.
+func (s *Store) PruneSlabs(ctx context.Context, account proto.Account) error {
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
```
ChrisSchinnerl: I really don't expect us to be able to quickly prune all slabs within a single transaction. For all we know a user might store 1TB worth of slabs that they want to prune at once.

So I'd start with a for-loop containing a transaction that fetches 1 slab to prune and then prunes it. 
We can test higher limits like 5 or 10 but really the most important part is that the individual transactions are fast. `PruneSlabs` taking a minute because a user prunes 1TB of data isn't that big of a deal imo.

persist/postgres/slabs_test.go
```diff
@@ -183,3 +186,175 @@ func TestPinnedSlab(t *testing.T) {
...
+	const (
+		numAccounts          = 50_000
+		numObjectsPerAccount = 10
```
ChrisSchinnerl: I'd tweak the setup a little to have fewer accounts but more slabs that get pruned per account. e.g. 1000 slabs per account.

persist/postgres/slabs_test.go
```diff
@@ -183,3 +186,175 @@ func TestPinnedSlab(t *testing.T) {
...
+			}
+		}
+	}
+	batch.Queue(`UPDATE sectors_stats SET num_slabs = $1`, slabID)
+	if err := store.pool.SendBatch(b.Context(), batch).Close(); err != nil {
+		b.Fatal(err)
+	}
+
+	for b.Loop() {
+		if err := store.PruneSlabs(b.Context(), accs[frand.Intn(len(accs))]); err != nil {
```
ChrisSchinnerl, line 360: Can you report the number of pruned slabs here using `ReportMetric`? Right now it's not quite obvious what the prune rate is.

persist/postgres/sectors.go
ChrisSchinnerl: I'd hold off on hard to understand vibe queries right now.

persist/postgres/sectors.go
chris124567: Without:
```
BenchmarkPruneSlabs-4                 14          73091129 ns/op
```

It's not a small difference but it's not huge either.

persist/postgres/sectors.go
chris124567: This is slower

```
BenchmarkPruneSlabs
BenchmarkPruneSlabs-4                 12          94574858 ns/op
```

persist/postgres/slabs.go
chris124567: Why not do it in one transaction?  Before I rewrote the unpin slab logic to take multiple slabs I was using a loop and it was about twice as slow in the benchmark as the logic without a loop.

persist/postgres/slabs.go
chris124567: This is slower for me:

```
SELECT s.digest
FROM account_slabs a
JOIN slabs s ON s.id = a.slab_id
WHERE a.account_id = $1
    AND NOT EXISTS (
    SELECT 1
    FROM objects o
    JOIN object_slabs os ON o.id = os.object_id
    WHERE o.account_id = a.account_id AND
    os.slab_digest = s.digest
```

```
BenchmarkPruneSlabs-4                  8         125426815 ns/op                15.00 slabs/op
```

chris124567:
```
        numAccounts       = 1000
        objectsPerAccount = 500
        slabsPerObject    = 3
```

```
BenchmarkPruneSlabs-4                 10         104622979 ns/op               750.0 slabs/op
```

n8mgr:
@chris124567 merge conflicts

api/app/app.go
```diff
@@ -296,6 +298,15 @@ func (a *app) handlePOSTSlabs(jc jape.Context, pk types.PublicKey) {
 	jc.Encode(slabID)
 }
 
+func (a *app) handlePOSTSlabsPrune(jc jape.Context, pk types.PublicKey) {
+	err := a.slabs.PruneSlabs(jc.Request.Context(), proto.Account(pk))
```
n8mgr, line 313: Can you remind me why this is an endpoint instead of something we just do periodically?

chris124567: The issue said endpoint.  Even if we make improvements it's still going to be a somewhat expensive operation so I think it makes sense for it to be manual.  Also a user could want to pin slabs through the indexer but implement their own object management.

ChrisSchinnerl: > Can you remind me why this is an endpoint instead of something we just do periodically?

As Christopher already pointed out users might not want to rely on objects. But even if they did, the current endpoint for pinning a slab would become obsolete unless we add a timeout to the pruning code since a user could not rely on slabs randomly disappearing otherwise.
But at the same time if you are just a madlad uploading a 100GB object and it takes you a day, you shouldn't really have to worry about finishing that upload before the slabs get potentially pruned.

persist/postgres/slabs.go
ChrisSchinnerl: A single transaction is always going to be faster than multiple but for pruning operations which require exclusive locks on rows that tradeoff can be worth it.
e.g. if pruning takes 60s for an account, that means other background tasks such as migrations, data integrity checks etc. which potentially need to update the same rows get blocked for the full duration of that transaction as well.
If you perform the pruning in many smaller transactions that take milliseconds or at least less than a second each, these background tasks aren't blocked for too long even if the pruning operation appears to take longer for that individual user.

persist/postgres/sectors.go
```diff
@@ -317,73 +317,132 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
...
+		return fmt.Errorf("failed to disable seqscan: %w", err)
+	}
+
+	var args []sqlHash256
+	for _, slabID := range slabIDs {
+		args = append(args, sqlHash256(slabID))
+	}
+
+	// delete the association between the account and the slab
+	rows, err := tx.Query(ctx, `DELETE FROM account_slabs
```
n8mgr:
```
index 2ef2b97..4e0d543 100644
@@ -318,21 +318,18 @@ func (s *Store) PinSlab(ctx context.Context, account proto.Account, nextIntegrit
 }
 
 func (s *Store) unpinSlabs(ctx context.Context, tx *txn, accountID int64, slabIDs []slabs.SlabID) error {
-	if _, err := tx.Exec(ctx, `SET LOCAL enable_seqscan=FALSE`); err != nil {
-		return fmt.Errorf("failed to disable seqscan: %w", err)
-	}
-
 	var args []sqlHash256
 	for _, slabID := range slabIDs {
 		args = append(args, sqlHash256(slabID))
 	}
 
 	// delete the association between the account and the slab
-	rows, err := tx.Query(ctx, `DELETE FROM account_slabs
-WHERE
-	account_id = $1 AND
-	slab_id = ANY(SELECT id FROM slabs WHERE digest = ANY($2))
-RETURNING slab_id`, accountID, args)
+	rows, err := tx.Query(ctx, `DELETE FROM account_slabs a
+USING slabs s
+WHERE a.account_id = $1
+  AND a.slab_id = s.id
+  AND s.digest = ANY($2)
+RETURNING a.slab_id;`, accountID, args)
 	if err != nil {
 		return fmt.Errorf("failed to unpin slab: %w", err)
 	}
```

```
Before:
goos: darwin
goarch: arm64
cpu: Apple M4 Max
BenchmarkPruneSlabs-16    	      21	  49777052 ns/op	       750.0 slabs/op	  657168 B/op	   10218 allocs/op
PASS
ok  	go.sia.tech/indexd/persist/postgres	39.275s


---
After:
goos: darwin
goarch: arm64
cpu: Apple M4 Max
BenchmarkPruneSlabs-16    	      25	  40114210 ns/op	       750.0 slabs/op	  688840 B/op	   10723 allocs/op
PASS
ok  	go.sia.tech/indexd/persist/postgres	39.043s
```

chris124567: Thank you

n8mgr (dismissed):

persist/postgres/slabs_test.go
peterjan: This seems to have fallen through the cracks?

persist/postgres/sectors.go
ChrisSchinnerl: I'd be careful about that. The other day on one of Christopher's PRs a query ended up quite a bit faster by removing logic from the query and performing it in Go.
So I'd say only do that if the benchmark doesn't suffer.

api/app/app.go
ChrisSchinnerl, line 1: Missing an update to the openapi spec

ChrisSchinnerl (changes requested):

persist/postgres/slabs_test.go
chris124567: The latest benchmark output has `slabs/op`

chris124567:
OK, I changed it to use multiple transactions.  Tried different  batch sizes but it didn't seem to make a huge difference so I  went with 20.  I still don't really like losing atomicity here because who's to say that the slab won't be in use by the time the transaction runs but I see that a really long transaction might be worse.

This change does sadly make the performance worse:
```
BenchmarkPruneSlabs-4                  4         386214540 ns/op               750.0 slabs/op
```

persist/postgres/slabs.go
```diff
@@ -103,3 +103,67 @@ ORDER BY ss.slab_index ASC`, dbID)
...
+
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	const batchSize = 20
+	for i := 0; i < len(slabIDs); i += batchSize {
+		err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
```
ChrisSchinnerl, line 161: The batch needs to be fetched within the same transaction that unpins it. As you already said yourself, if you don't, you end up unpinning a slab that might be in use. So the same transaction needs to fetch a small batch and unpin it.

There are some more improvements we can make further down the road but that should be enough for now.

When picking a batch size also check the duration it takes for a single batch to finish. That's really the speed we care about. The duration of the whole thing doesn't matter too much for now.

ChrisSchinnerl (changes requested):

persist/postgres/slabs.go
chris124567:
```
    var exhausted bool
    const batchSize = 100
    for !exhausted {
        t1 := time.Now()
            slabIDs, err := getSlabs(tx, batchSize)
            if err != nil {
                return fmt.Errorf("failed to get slabs to unpin: %w", err)
            } else if len(slabIDs) < batchSize {
                exhausted = true
            }
            if err := s.unpinSlabs(ctx, tx, id, slabIDs); err != nil {
                return fmt.Errorf("failed to unpin slabs: %w", err)
            }
            return nil
        })
        if err != nil {
            return err
        }
        log.Println(time.Now().Sub(t1))
    }
```

When running the benchmark:
```
2025/09/25 11:33:14 393.996848ms
2025/09/25 11:33:15 304.21125ms
2025/09/25 11:33:15 301.620867ms
2025/09/25 11:33:15 300.378468ms
2025/09/25 11:33:16 306.759765ms
2025/09/25 11:33:16 68.814816ms
2025/09/25 11:33:16 70.712346ms
2025/09/25 11:33:16 80.084732ms
```

chris124567: forgot to add this context:
```
BenchmarkPruneSlabs
BenchmarkPruneSlabs-4                  1        2130341416 ns/op               750.0 slabs/op
```

```
	const (
		numAccounts       = 1000
		objectsPerAccount = 500
		slabsPerObject    = 3
	)
```

n8mgr (dismissed):

ChrisSchinnerl (dismissed):

peterjan (dismissed):

ChrisSchinnerl (approved):

persist/postgres/migrations.go
```diff
@@ -237,6 +237,18 @@ CREATE INDEX object_slabs_object_id_slab_index_idx ON object_slabs(object_id, sl
 		}
 		return nil
 	},
+	// reset registered accounts
```
peterjan, line 240: new migrations should go at the end

openapi/app.yml
```diff
@@ -445,6 +445,16 @@ paths:
               // ...
             }
 
+  /slabs/prune:
+    post:
+      tags:
+        - slabs
+      summary: Unpin all slabs not used by an object currently pinned by the user
```
peterjan, line 452:
```suggestion
      summary: Unpin all slabs not attached to an object currently pinned by the user
```

peterjan (changes requested):

persist/postgres/migrations.go
ChrisSchinnerl: Isn't it? The one at the bottom is `// add indexes to speed up unpinning slabs`.

peterjan: heh that's so odd, GH showed me this in a diff where it was above the removal of the resolved cidrs, which is isn't even in here :/

peterjan (approved):

PR #447, opened 2025-09-17
Fix ScanHost not updating store when context is timed out

ChrisSchinnerl:
At least one of the hosts on Zeus keeps timing out during scans with the following error.
```
indexd-1  | DEBUG	contracts.maintenance.contractpruning	failed to prune contracts	{"hostKey": "ed25519:389258107290842af1f14d29f853695d71d621a51022e3322d5673e6c6519888", "error": "failed to scan host, failed to update host, failed to begin transaction: context deadline exceeded"}
```

This PR fixes this by putting the `UpdateHost` call on a separate context.

### DIFF
--- a/contracts/pruning.go
+++ b/contracts/pruning.go
@@ -61,6 +61,7 @@ loop:
 			}()
 
 			if err := cm.hosts.WithScannedHost(ctx, hostKey, func(host hosts.Host) error {
+				hostLog = hostLog.With(zap.Stringer("protocolVersion", host.Settings.ProtocolVersion))
 				return cm.performContractPruningOnHost(ctx, host, hostLog)
 			}); err != nil {
 				hostLog.Debug("failed to prune contracts", zap.Error(err))
--- a/hosts/manager.go
+++ b/hosts/manager.go
@@ -337,7 +337,12 @@ func (m *HostManager) ScanHost(ctx context.Context, hk types.PublicKey) (Host, e
 		scanExponentialBackoffMaxHours,
 	)
 
-	err = m.store.UpdateHost(ctx, hk, networks, settings, loc, success, nextScan)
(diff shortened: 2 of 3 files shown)


contracts/pruning.go
```diff
@@ -61,6 +61,7 @@ loop:
 			}()
 
 			if err := cm.hosts.WithScannedHost(ctx, hostKey, func(host hosts.Host) error {
+				hostLog = hostLog.With(zap.Stringer("protocolVersion", host.Settings.ProtocolVersion))
```
ChrisSchinnerl, line 64: This hopefully proves my suspicion about another error related to pruning being the fault of outdated hosts.

The error is:
```
indexd-1  | DEBUG	contracts.maintenance.contractpruning	failed to prune contract	{"hostKey": "ed25519:5b34cd8e05a47fbce60778a328baa0265124f48d15ed9ed3775ae8f88a6ca8de", "error": "failed to prune contract sectors: failed to fetch free sectors: failed to free sectors: failed to read response: failed to read request: EOF (4)"}
```

chris124567 (approved):
If a host is timing out, presumably it is due to slow requests over the network, so wouldn't `ScanHost` have returned earlier than https://github.com/SiaFoundation/indexd/pull/447/files#diff-25167c9080ef6b617a869214b8eac94e23a88959a0cbc50cb75efc3bb1d15d71R345?  All of the other function calls in `ScanHost` take a `context.Context` too.  How likely is it that a host would make it through `store.Host`, `resolveHost`, `fetchSettings` in slightly under a minute (`scanTimeout`) but not have enough time left to call `UpdateHost`?


But we should try it if you think it will fix a bug, because it doesn't really add much complexity.

n8mgr (approved):

peterjan (approved):

PR #453, opened 2025-09-18
Random encryption key

ChrisSchinnerl:
Right now the upload code derives the slab encryption key from the slab data. Which unnecessarily introduces determinism in our key derivation and since we use the shard index as the nonce, also introduces a security concern for any slabs that are uploaded multiple times.

e.g. when the same shard is uploaded to the same host twice with only a subset of it changing, a host will be able to tell since the remaining data will still match. By xor'ing the the part that differs they can then get access to the key and even decrypt the changed part. 

This PR changes the SDK to use randomly generated keys for each slab which makes this impossible.

NOTE: This also matches our Rust SDK now.

### DIFF
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -113,11 +113,8 @@ func (s *SDK) uploadSlab(ctx context.Context, shards [][]byte, dataShards uint8,
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	var encryptionKey [32]byte
-	frand.Read(encryptionKey[:])
-
 	slab := slabs.SlabPinParams{
-		EncryptionKey: encryptionKey,
+		EncryptionKey: frand.Entropy256(),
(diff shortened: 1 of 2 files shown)


peterjan (dismissed):

chris124567 (dismissed):

n8mgr (approved):

sdk/client.go
```diff
@@ -113,6 +113,9 @@ func (s *SDK) uploadSlab(ctx context.Context, encryptionKey [32]byte, shards [][
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	var encryptionKey [32]byte
```
n8mgr: nit: `frand.Entropy256()`

n8mgr (changes requested):

n8mgr (approved):

peterjan (approved):

PR #457, opened 2025-09-18
Fix SDK download

peterjan:
The SDK was decrypting after reconstructing the erasure coded data, it should do that before to match/mirror the behaviour on upload. We actually had a unit test that would've surfaced this issue, but we neglected to check whether the downloaded bytes matched the ones we uploaded in `TestDownload`.

### DIFF
--- a/sdk/client_test.go
+++ b/sdk/client_test.go
@@ -10,7 +10,10 @@ import (
 	"testing"
 	"time"
 
+	proto "go.sia.tech/core/rhp/v4"
 	"go.sia.tech/core/types"
+	"go.sia.tech/indexd/accounts"
+	"go.sia.tech/indexd/internal/testutils"
 	"go.sia.tech/indexd/slabs"
 	"lukechampine.com/frand"
 )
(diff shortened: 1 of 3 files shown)


ChrisSchinnerl (dismissed):

sdk/client_test.go
```diff
@@ -132,6 +135,46 @@ func TestRoundtrip(t *testing.T) {
 	}
 }
 
+func TestRoundtripCluster(t *testing.T) {
+	logger := testutils.NewLogger(false)
+	cluster := testutils.NewCluster(t, testutils.WithLogger(logger), testutils.WithHosts(6), testutils.WithIndexer(testutils.WithSlabOptions(slabs.WithHealthCheckInterval(time.Second))))
```
ChrisSchinnerl: Since you mentioned that the issue was partially hidden by the mock dialer using the data shard hosts first, it might be a good idea to use the mock dialer here but delete half the sectors to force using both data shards and parity shards.

Using a real cluster slows things down and won't guarantee that we never exclusively use data shards.

ChrisSchinnerl (approved):

chris124567 (approved):

PR #458, opened 2025-09-18
Fix migrations

peterjan:
Turns out we weren't decrypting the shards before reconstructing the data when migrating slabs. We have quite a bit of testing around migrations but the mistake was always the same, we were either mocking hosts and returning unencrypted data, or we weren't setting the slab's encryption key. Removing the decrypt/re-encrypt steps in the migration code makes the test fail now.

### DIFF
--- a/slabs/migrations.go
+++ b/slabs/migrations.go
@@ -95,9 +95,11 @@ func (m *SlabManager) migrateSlab(ctx context.Context, slabID SlabID, allHosts [
 	// decrypt the shards
 	nonce := make([]byte, 24)
 	for i := range shards {
-		nonce[0] = byte(i)
-		c, _ := chacha20.NewUnauthenticatedCipher(slab.EncryptionKey[:], nonce)
-		c.XORKeyStream(shards[i], shards[i])
+		if len(shards[i]) > 0 {
+			nonce[0] = byte(i)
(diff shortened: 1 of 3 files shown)


slabs/migrations.go
```diff
@@ -91,6 +92,14 @@ func (m *SlabManager) migrateSlab(ctx context.Context, slabID SlabID, allHosts [
 		return fmt.Errorf("failed to download slab %s: %w", slab.ID, err)
 	}
 
+	// decrypt the shards
+	nonce := make([]byte, 24)
+	for i := range shards {
```
n8mgr, line 97: only need to decrypt the ones that were actually recovered. Pretty sure it's a no-op in XORKeyStream, but better to be explicit.

slabs/migrations.go
```diff
@@ -112,6 +121,13 @@ func (m *SlabManager) migrateSlab(ctx context.Context, slabID SlabID, allHosts [
 		}
 	}
 
+	// re-encrypt the shards
+	for i := range shards {
```
n8mgr: this can be combined with the loop above. Only need to re-encrypt the ones we intend to migrate.

n8mgr (changes requested):

ChrisSchinnerl (approved):

chris124567 (approved):

n8mgr (approved):

Issue #291, opened 2025-07-30
Prometheus Metrics


ChrisSchinnerl:
This issue serves as a collection of metrics we need to implement prometheus support for to monitor them.

- [x] Wallet Balance
- [x] Uptime
- [x] Synced Height
- [x] Scanned Height
- [x] Pinned Data
- [x] Unpinned Data

PR #460, opened 2025-09-18
Add Prometheus metrics for Indexd state, wallet and sector stats

ChrisSchinnerl:
Related to https://github.com/SiaFoundation/indexd/issues/291

Finally allows us to hook indexd up to our dashboard.

Not sure what the best place to document this is yet. I don't want to pollute the API docs since this is quite niche but we should probably mention it somewhere.

### DIFF
--- a/api/admin/prometheus.go
+++ b/api/admin/prometheus.go
@@ -26,11 +26,7 @@ func (s State) PrometheusMetric() (metrics []prometheus.Metric) {
 		"os":         s.OS,
 		"build_time": s.BuildTime.String(),
 
-		"network":          s.Network,
-		"start_time":       s.StartTime.String(),
-		"explorer_enabled": s.Explorer.Enabled,
-		"explorer_url":     s.Explorer.URL,
-		"synced":           s.Synced,
+		"network": s.Network,
 	}
 	return []prometheus.Metric{
 		{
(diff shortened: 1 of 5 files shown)


api/admin/prometheus.go
```diff
@@ -0,0 +1,72 @@
...
+// PrometheusMetric implements the prometheus.Marshaller interface for the
+// sector stats response.
+func (s SectorsStatsResponse) PrometheusMetric() (metrics []prometheus.Metric) {
+	return []prometheus.Metric{
+		{
+			Name:  "indexd_num_slabs",
+			Value: float64(s.NumSlabs),
+		},
+		{
+			Name:  "indexd_num_pinned_sectors",
```
peterjan, line 14: I'll make sure to update these after metrics get merged, no need to keep updating these should some of those PRs get into master before this one.

internal/prometheus/encoder_test.go
```diff
@@ -0,0 +1,62 @@
...
+		Test: 1.5,
+	}
+
+	var b bytes.Buffer
+	e := NewEncoder(&b)
+	if err := e.Append(&v); err != nil {
+		t.Fatal(err)
+	}
+
+	got := string(b.Bytes())
```
peterjan: `nit` could use `b.String()`

peterjan (dismissed):

PR #462, opened 2025-09-19
Fix TestApplicationAPI NDF

peterjan:
We have an NDF in `TestApplicationAPI` when we try and filter hosts by location. This happens because we scan the hosts in the background and we the locator can't locate our test hosts which have the loopback IP, so there's a race in where it `NULL`s the location. 

Obviously there's a couple of ways to fix this, e.g. disable host scans or use a custom locator. The latter seemed like a good option but that would require a change to the interface where we pass the host key so we can really control it from testing. Figured instead of changing the interface only for testing was a worse option over not updating the host's location when we pass it an empty location.

### DIFF
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -49,4 +49,4 @@ jobs:
       - name: Test
         uses: n8maninger/action-golang-test@v2
         with:
-          args: -race;-failfast;-count=10
+          args: -race;-failfast;-count=10;-timeout=4h
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -49,4 +49,4 @@ jobs:
       - name: Test
         uses: n8maninger/action-golang-test@v2
         with:
-          args: -race;-failfast;-run=TestApplicationAPI;-count=100
+          args: -race;-failfast;-count=10
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -49,4 +49,4 @@ jobs:
       - name: Test
         uses: n8maninger/action-golang-test@v2
         with:
-          args: -race;-failfast;-count=10;-timeout=4h
+          args: -race;-failfast
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -49,4 +49,4 @@ jobs:
       - name: Test
         uses: n8maninger/action-golang-test@v2
         with:
-          args: -race;-failfast
+          args: -race;-failfast;-run=TestApplicationAPI;-count=100
--- a/persist/postgres/hosts.go
+++ b/persist/postgres/hosts.go
@@ -462,7 +462,7 @@ SET
 	next_scan = $3,
 
 	country_code = CASE WHEN $4 <> '' THEN $4 ELSE country_code END,
-	location = CASE WHEN $4 <> '' THEN $5 ELSE location END,
+	location = CASE WHEN $5 <> POINT(0.0, 0.0) THEN $5 ELSE location END,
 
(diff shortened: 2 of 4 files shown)


.github/workflows/go-test.yml
```diff
@@ -49,4 +49,4 @@ jobs:
       - name: Test
         uses: n8maninger/action-golang-test@v2
         with:
-          args: -race;-failfast
+          args: -race;-failfast;-count=10;-timeout=4h
```
chris124567: This results in a 20 minute CI time instead of 2 minutes.  Could this be lower?

persist/postgres/hosts.go
```diff
@@ -461,8 +461,8 @@ SET
 	last_successful_scan = NOW(),
 	next_scan = $3,
 
-	country_code = $4,
-	location = $5,
+	country_code = CASE WHEN $4 <> '' THEN $4 ELSE country_code END,
+	location = CASE WHEN $4 <> '' THEN $5 ELSE location END,
```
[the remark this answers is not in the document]
chris124567: I think Copilot is right here but also for a default value for the location I think you would need something like `POINT(0, 0)`

.github/workflows/go-test.yml
n8mgr: Why are we running 10x at all?

peterjan: Heh that's weird :/ I thought I checkout out master for this file. This is just a leftover from tracking the NDF on the CI

persist/postgres/hosts.go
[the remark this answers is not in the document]
peterjan: This is how I intended to write it. I figured if `Locate` returns an empty `Country` we can discard the location as well. So in other words: if we have a valid country we update, if not we don't update both fields. (I was a little surprised to see location is `NOT NULL` and defaults to (0,0) - this might at one point give issues for our users in Ghana 😛 - not exactly 0,0 but close enough)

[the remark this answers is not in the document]
ChrisSchinnerl: I'd listen to Copilot here. I'm not sure how `Locate` behaves but if it gives us coordinates we should use them.
If there is no country code we set the country code to the empty string.

Due to https://github.com/SiaFoundation/indexd/issues/473, having valid coordinates is a lot more important than having a country code. 

Regarding the 0,0 coordinate I think that's fine for now. Seems like it's 600km from Ghana. But I'm definitely curious to see if we ever end up with 0,0 hosts and if so, how many.

ChrisSchinnerl (changes requested):

persist/postgres/hosts.go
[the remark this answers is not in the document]
peterjan: We own the interface and the country code would never be empty if the location is set, but I don't mind to check for the `POINT(0.0)` instead and update if it's different from that as long as we do the conditional update to avoid the NDF.

ChrisSchinnerl (approved):

persist/postgres/hosts_test.go
```diff
@@ -1223,6 +1223,23 @@ func TestUpdateHost(t *testing.T) {
 	} else if h.Longitude != location.Longitude {
 		t.Fatal("unexpected Longitude", h.Longitude)
 	}
+
+	// assert updating with an empty location doesn't affect the host's
```
ChrisSchinnerl, line 1227: (nit) this is a bit outdated now since the location update no longer depends on the country code.

peterjan: Is it? It doesn't mention the country code right, just mentions an empty `geoip.Location{}`

n8mgr (approved):

PR #467, opened 2025-09-19
Add contracts metrics

### LINKED ISSUE #464: Contract metrics
- [ ] Number of active contracts
- [ ] Total size of all active contracts
- [ ] Total capacity of all active contracts
- [ ] Number of bad active contracts
- [ ] Number of contracts that are up for renew

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/indexd/issues/464


persist/postgres/contracts.go
```diff
@@ -30,6 +31,57 @@ func (s *Store) ContractRevision(ctx context.Context, contractID types.FileContr
...
+    			SELECT scanned_height FROM global_settings
+       		)
+			SELECT
+				COUNT(*),       -- all contracts
+				COALESCE(SUM(good::int), 0), -- good contracts
+				COALESCE(SUM(capacity), 0),  -- total capacity
+				COALESCE(SUM(size), 0)      -- total size
+			FROM contracts
+			CROSS JOIN globals
+			WHERE
```
peterjan, line 50: Instead of using the `expiration_height`, why did you not use the contract's `state` field? It makes the docs somewhat misleading imo.

persist/postgres/contracts.go
```diff
@@ -30,6 +31,57 @@ func (s *Store) ContractRevision(ctx context.Context, contractID types.FileContr
 	return rhp.ContractRevision{ID: contractID, Revision: types.V2FileContract(revision)}, renewed, nil
 }
 
+// ContractsStats returns statistics about the contracts in the database.
+func (s *Store) ContractsStats(ctx context.Context) (resp admin.ContractsStatsResponse, _ error) {
```
peterjan, line 36: Could've added a benchmark. I see why you didn't but it's still a `COUNT` across a large table, which we generally want to avoid, so perhaps an accompanying benchmark is not the worst thing.

api/admin/admin_test.go
```diff
@@ -772,6 +772,27 @@ func TestWalletAPI(t *testing.T) {
 	}
 }
 
+func TestContractsStatsAPI(t *testing.T) {
```
peterjan, line 783: `nit` since it's a very minimal check anyway I would create the cluster like this `testutils.NewCluster(t, testutils.WithHosts(1))`

n8mgr:
@ChrisSchinnerl merge conflicts

persist/postgres/contracts.go
ChrisSchinnerl: I added that as well and changed it to proof_start because after the proof start we can't use the contract anymore anyway.

The problem with only using only the state is that contracts become no longe revisable when they enter the proof window but a contract might resolve at any time in or even after the proof window. Using a hard cutoff means that all contracts are treated the same and we don't rely on the host resolving them for our stats.

persist/postgres/contracts_test.go
```diff
@@ -1543,6 +1607,71 @@ func BenchmarkContracts(b *testing.B) {
 	})
 }
 
+func BenchmarkContractsStats(b *testing.B) {
```
ChrisSchinnerl, line 1610: For 100k contracts (100 active contracts)

```
BenchmarkContractsStats
BenchmarkContractsStats-12    	    1772	    676798 ns/op
PASS
```

n8mgr (approved):

peterjan (approved):

Issue #470, opened 2025-09-22
Add metric for the number of registered and active accounts


ChrisSchinnerl:
That way we can visualize the number of users an instance of indexd is serving.
This includes:

- [x] Database support via a cached row that gets updated whenever an account is created/deleted
- [x] Prometheus support for registered accounts
- [ ] Prometheus support for active accounts

PR #475, opened 2025-09-22
Add registered accounts metric

chris124567:
Partially implement #470

Will add active accounts in separate PR once #436 is merged.


api/admin/admin_test.go
```diff
@@ -838,6 +838,52 @@ func TestSectorStatsAPI(t *testing.T) {
 	}
 }
 
+func TestContractsStatsAPI(t *testing.T) {
```
n8mgr: Should this not be `AccountsStats`?

persist/postgres/init.go
```diff
@@ -49,6 +49,8 @@ func (s *Store) initNewDatabase(ctx context.Context, target int64, ms contracts.
 			return err
 		} else if err := initSettings(ctx, tx, ms, us); err != nil {
 			return fmt.Errorf("failed to init settings: %w", err)
+		} else if err := s.initAccountStats(ctx, tx); err != nil {
```
n8mgr: Instead of adding a new method every time we add a new type of stat, combine it with the existing method.

persist/postgres/init.sql
```diff
@@ -9,6 +9,11 @@ CREATE TABLE accounts (
     service_url TEXT NOT NULL DEFAULT ''
 );
 
+CREATE TABLE account_stats (
```
n8mgr: We don't really benefit from another table here. Could combine with `sector_stats` and rename it to `stats`.

n8mgr (changes requested):

n8mgr (dismissed):

ChrisSchinnerl (approved):

Issue #473, opened 2025-09-22
Remove CIDR checks and use distance instead


ChrisSchinnerl:
Now that we have coordinates for hosts we don't need the CIDR check anymore. Instead we can now pick hosts that are a certain distance away from each other. So we should drop CIDR checks and instead make sure hosts are further away from each other than 10km.

While we are at it we should also update the GeoIP we use.

NOTE: Hosts without a location are considered to have the same location.

PR #479, opened 2025-09-23
Ensure minimum distance between hosts using location

peterjan:
This PR replaces our CIDR checks with a min. distance check using the host's location. I'm planning to F/U with a PR that gets rid of `host_resolved_cidrs` in the database. Note that I updated the geoip database to the latest version.

References https://github.com/SiaFoundation/indexd/issues/473

### DIFF
--- a/contracts/form.go
+++ b/contracts/form.go
@@ -83,7 +83,7 @@ func (s *formContractSigner) SignV2Inputs(txn *types.V2Transaction, toSign []int
 }
 
 // performContractFormation makes sure that we have at least 'wanted' good
-// contracts with good hosts in unique CIDRs.
+// contracts with good hosts that are sufficiently spaced apart.
 func (cm *ContractManager) performContractFormation(ctx context.Context, period uint64, wanted int64, log *zap.Logger) error {
 	formationLog := log.Named("formation")
 	formationLog.Debug("started", zap.Uint64("period", period), zap.Int64("wanted", wanted))
@@ -118,7 +118,7 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period 
 	formationLog.Debug("found candidates", zap.Uint64("n", uint64(len(candidates))))
 
 	// forceFormation is a map of hosts that we will always form a contract with
-	// regardless of how many we already have or what CIDR they are on
+	// regardless of how many we already have or where the host is located
 	forceFormation := make(map[types.PublicKey]bool)
 
 	// determine which hosts are 'full', meaning they have exclusively full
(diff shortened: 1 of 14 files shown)


geoip/geoip.go
```diff
@@ -11,6 +12,69 @@ import (
 //go:embed GeoLite2-City.mmdb
 var maxMindCityDB []byte
 
+const radiusKm = 6371.0088
+
+type distanceUnit int
+
+const (
+	// Kilometers is a distance unit in kilometers.
+	Kilometers distanceUnit = iota
```
n8mgr: Do we really need to support both? Should be fine to enforce KM.

hosts/distance.go
```diff
@@ -0,0 +1,55 @@
...
+		if distance.LessThan(s.minDistance) {
+			return false
+		}
+	}
+
+	return true
+}
+
+func (s *SpacedSet) contains(h Host) bool {
+	for _, other := range s.selected {
```
n8mgr: Could add a map of public keys to make this `O(1)`.

Either a separate map or `map[PubKey][2]float64`. There's no real need to store the whole host since it's never exposed after adding.

contracts/form.go
```diff
@@ -161,48 +161,16 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
...
-		for _, cidr := range host.Networks {
-			if hk, known := usedCidrs[cidr]; known {
-				return hk, true
-			}
-		}
-		return types.PublicKey{}, false
-	}
-
 	// helper to check if a host is good to form a contract with
+	set := hosts.NewSpacedSet(cm.minHostDistance)
```
n8mgr: A good follow-up might be selecting the farthest "good" candidate first.

n8mgr (changes requested):

geoip/geoip.go
peterjan: Heh, I did this for you in particular. I'm fine with removing it but I think it's quite nice DX for just a couple of lines. It also tied in nicely with `HaversineDistance` on the `Location` struct which otherwise returns a float64 in km. 

```go
contracts.WithMinHostDistance(geoip.Km(10))
contracts.WithMinHostDistance(geoip.Miles(6))
```

contracts/form.go
peterjan: Yeah I thought about that too actually when I was implementing it. For this PR though I was really aiming to replace CIDR checks with a location check in the least amount of lines possible. I personally feel the contractor is already getting away from us, again. It's not trivial to follow along with the host selection when forming contracts, most complexity coming from `forceFormation`.

geoip/geoip.go
ChrisSchinnerl: I'd stick to km

slabs/migrations.go
```diff
@@ -195,47 +197,21 @@ func sectorsToMigrate(slab Slab, allHosts []hosts.Host, goodContracts []contract
...
-		host := hostsMap[contract.HostKey]
-		for _, network := range host.Networks {
-			if _, ok := usedCIDRs[network]; ok {
-				continue LOOP
-			}
-		}
-
-		for _, network := range host.Networks {
-			usedCIDRs[network] = struct{}{}
+		if host, ok := hostsMap[contract.HostKey]; ok && set.CanAddHost(host) {
```
ChrisSchinnerl: `Add` returns a boolean right? So this might be slightly nicer

```suggestion
		if host, ok := hostsMap[contract.HostKey]; ok && set.Add(host) {
			candidates = append(candidates, host)
```

ChrisSchinnerl (changes requested):
This cleans up some code quite nicely.

ChrisSchinnerl (approved):

n8mgr (approved):

PR #480, opened 2025-09-23
Remove `host_resolved_cidrs` table

### LINKED ISSUE #473: Remove CIDR checks and use distance instead

peterjan:
We can get rid of the table in the database now that we distribute shards to hosts based on their location.

Fixes https://github.com/SiaFoundation/indexd/issues/473

### DIFF
--- a/hosts/host.go
+++ b/hosts/host.go
@@ -21,10 +21,6 @@ var (
 	// ErrNotFound is returned by database operations that fail due to a host
 	// not being found.
 	ErrNotFound = errors.New("host not found")
-
-	// ErrNoNetworks is returned when a host has no networks even though it
-	// should.
-	ErrNoNetworks = errors.New("host has no networks")
 )
 var (
 	// DefaultHostsQueryOpts re the default options applied when querying hosts. By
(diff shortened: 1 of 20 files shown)


n8mgr:
Conflicts + squash

hosts/host.go
```diff
@@ -92,7 +88,6 @@ type (
 		ConsecutiveFailedScans int                 `json:"consecutiveFailedScans"`
 		RecentUptime           float64             `json:"recentUptime"`
 		Addresses              []chain.NetAddress  `json:"addresses"`
-		Networks               []string            `json:"networks"`
```
ChrisSchinnerl, line 95: Requires openapi spec update

ChrisSchinnerl (changes requested):

hosts/host.go
peterjan: Nice catch

ChrisSchinnerl (approved):

PR #484, opened 2025-09-24
Update `SectorStats` prometheus metric and JSON 

### LINKED ISSUE #481: Extend sector stats prometheus encoding
We got new fields which are not yet encoded.

peterjan:
Fixes https://github.com/SiaFoundation/indexd/issues/481

### DIFF
--- a/api/admin/types.go
+++ b/api/admin/types.go
@@ -30,11 +30,11 @@ type (
 
 	// SectorsStatsResponse is the response body for the [GET] /stats/sectors
 	SectorsStatsResponse struct {
-		NumSlabs             int64 `json:"numSlabs"`
-		NumMigratedSectors   int64 `json:"numMigratedSectors"`
-		NumPinnedSectors     int64 `json:"numPinnedSectors"`
-		NumUnpinnableSectors int64 `json:"numUnpinnableSectors"`
-		NumUnpinnedSectors   int64 `json:"numUnpinnedSectors"`
+		NumSlabs             int64 `json:"slabs"`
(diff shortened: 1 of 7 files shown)


api/admin/types.go
```diff
@@ -30,11 +30,11 @@ type (
 
 	// SectorsStatsResponse is the response body for the [GET] /stats/sectors
 	SectorsStatsResponse struct {
-		NumSlabs             int64 `json:"numSlabs"`
-		NumMigratedSectors   int64 `json:"numMigratedSectors"`
-		NumPinnedSectors     int64 `json:"numPinnedSectors"`
-		NumUnpinnableSectors int64 `json:"numUnpinnableSectors"`
-		NumUnpinnedSectors   int64 `json:"numUnpinnedSectors"`
+		NumSlabs             int64 `json:"slabs"`
```
peterjan: I was on the fence on whether to make this change or not, I think the shorter names are indeed better and it's nice to have the full descriptive names in the prometheus labels. This makes it consistent with `AccountStatsResponse`, which was added later but has the better naming imo. I don't mind reverting because this'll require a UI change, I just don't like that our stats endpoints now have inconsistent naming conventions.

ChrisSchinnerl: I don't mind changing the fields but we should update the Go fields to match the new json tags. 

Also needs an update to the openapi spec

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

chris124567 (approved):

PR #488, opened 2025-09-24
Add active accounts to account stats

### LINKED ISSUE #459: Add accounts stats endpoint
It should expose the number of total accounts as well as active accounts.

### LINKED ISSUE #470: Add metric for the number of registered and active accounts

chris124567:
Close #459 
Close #470

This will count service accounts as active.  Do we want to do that?  It could result in confusion if the active number is greater than the registered number.

### DIFF
--- a/openapi/components.yml
+++ b/openapi/components.yml
@@ -775,8 +775,12 @@ components:
       properties:
         registered:
           type: integer
-          format: int64
+          format: uint64
           description: Number of accounts currently registered on the indexer
+        active:
+          type: integer
+          format: uint64
+          description: Number of accounts active in the last week.
 
     ContractsStatsResponse:
       type: object
--- a/persist/postgres/stats.go
+++ b/persist/postgres/stats.go
@@ -62,7 +62,9 @@ func (s *Store) AccountStats(ctx context.Context) (admin.AccountStatsResponse, e
 		if err != nil {
 			return fmt.Errorf("failed to get number of registered accounts: %w", err)
 		}
-		stats.Active, err = activeAccounts(ctx, tx, time.Now().Add(-24*7*time.Hour))
+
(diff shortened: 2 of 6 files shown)


persist/postgres/stats.go
```diff
@@ -56,7 +58,17 @@ func (s *Store) incrementNumAccounts(ctx context.Context, tx *txn, delta int64)
...
-		return tx.QueryRow(ctx, "SELECT num_accounts_registered FROM stats").Scan(&stats.Registered)
+		err := tx.QueryRow(ctx, "SELECT num_accounts_registered FROM stats").Scan(&stats.Registered)
+		if err != nil {
+			return fmt.Errorf("failed to get number of registered accounts: %w", err)
+		}
+
+		const activeAccountThreshold = 7 * 24 * time.Hour
+		stats.Active, err = activeAccounts(ctx, tx, time.Now().Add(-activeAccountThreshold))
+		if err != nil {
+			return fmt.Errorf("failed to get active accounts: %w", err)
```
peterjan, line 69:
```suggestion
			return fmt.Errorf("failed to get number of active accounts: %w", err)
```

openapi/components.yml
```diff
@@ -775,8 +775,12 @@ components:
       properties:
         registered:
           type: integer
-          format: int64
+          format: uint64
           description: Number of accounts currently registered on the indexer
+        active:
```
peterjan, line 780: `nit` alphabetizing fields where possible is nice

peterjan (approved):
left two nits but feel free to resolve, they're very minor

ChrisSchinnerl (approved):

PR #489, opened 2025-09-24
Add master key to objects

### LINKED ISSUE #487: Deterministically generates unique object keys using a per-object seed
Then users only need one master key and can more easily back up that secret.

n8mgr:
Fixes #487 

If the key is derived deterministically using a random seed and app key, sharing breaks. The receiver would need to re-encrypt the raw data using their app key or store the derived encryption key somewhere anyway. Instead we store the master key encrypted using the user's app key.

- Introduces the concept of a `SealedObject` within the `indexd` backend and an `Object` in the SDK to clarify usage.

- Object key is now deterministic based on the slab slices.

- Adds a new `EncryptedMasterKey` field to the object store. This key is guaranteed to be unique when using the SDK and encrypted using the user's app key. This enables data recovery using only the user's app key. The master key is used to derive the metadata encryption key.

- Adds BIP39 recovery phrase generation to the SDK. This should be the preferred way to generate app keys.

- Removes `sdk.Slab` in favor of `slabs.SlabSlice`.

- Adds signature to the object to attest to the validity of the encrypted data.


api/app/client.go
```diff
@@ -215,36 +215,40 @@ func (c *Client) Account(ctx context.Context) (resp accounts.Account, err error)
...
 	}
 
-	var encryptionKey [32]byte
-	if n, err := hex.Decode(encryptionKey[:], []byte(fields[1])); err != nil {
+	encryptionKey, err := hex.DecodeString(keyStr)
+	if err != nil {
 		return slabs.SharedObject{}, nil, fmt.Errorf("invalid encryption key: %w", err)
-	} else if n != 32 {
-		return slabs.SharedObject{}, nil, fmt.Errorf("invalid encryption key length: expected 32 bytes, got %d", n)
 	}
```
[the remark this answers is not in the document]
n8mgr: The master key is decrypted here

n8mgr:
No idea what's up with the tests:
```
Error response from daemon: unauthorized: authentication required
```

Probably a rate limit on DockerHub, but I don't know how to login when using a GithubCI service...

sdk/object.go
```diff
@@ -0,0 +1,172 @@
...
+// UpdatedAt returns the time the object was last updated.
+func (o *Object) UpdatedAt() time.Time {
+	return o.updatedAt
+}
+
+// Lock returns a LockedObject that can be safely serialized and shared.
+func (o *Object) Lock(appKey types.PrivateKey) slabs.LockedObject {
+	keyCipher := masterKeyCipher(appKey)
+	nonce := frand.Bytes(keyCipher.NonceSize())
+	encryptedMasterKey := keyCipher.Seal(nonce, nonce, o.masterKey, nil)
```
ChrisSchinnerl, line 55: Instead of encrypting the master key and storing 72 bytes alongside the object, you could create a `masterKeySalt` of 32 bytes which you combine with the `appKey` to get the key like this
```
objectKey := hkdf.New(blake2b.New256, appKey, masterKeySalt, []byte("object"))
metaKey := hkdf.New(blake2b.New256, appKey, masterKeySalt, []byte("metadata"))
```
like we do in `keys/keys.go` -> `DeriveKey`

This also makes the key derivation very nice and symmetrical. 

Then you include the `masaterKeySalt` in the `Object.ID()` to make sure the indexer can't change it without you knowing. We can't do that for the metadata since updating it shouldn't change the object id but the salt should stay the same forever.

Saves us 40 bytes per object.

sdk/object.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	nonce := encryptedMeta[:metadataCipher.NonceSize()]
+	metadata, err := metadataCipher.Open(nil, nonce, encryptedMeta[metadataCipher.NonceSize():], nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unlock metadata: %w", err)
+	}
+	return metadata, nil
+}
+
+func newObjectFromLockedObject(lo slabs.LockedObject, appKey types.PrivateKey) (Object, error) {
```
ChrisSchinnerl: This function or its callers should check that the ID of the returned object matches the requested ID. I think right now we only do that in the Rust SDK.

persist/postgres/migrations.go
```diff
@@ -237,4 +237,45 @@ CREATE INDEX object_slabs_object_id_slab_index_idx ON object_slabs(object_id, sl
 		}
 		return nil
 	},
+	func(ctx context.Context, tx *txn, _ *zap.Logger) error {
+		// note: it is not practical to migrate the existing object data since
+		// the master key column does not exist.
+		if _, err := tx.Exec(ctx, `DROP TABLE IF EXISTS object_slabs; DROP TABLE IF EXISTS objects;`); err != nil {
```
ChrisSchinnerl: If you do `TRUNCATE objects CASCADE;` the migration is probably also very fast and then you only need to alter the 2 columns you changed instead of recreating 2 tables and all the indices.

ChrisSchinnerl (changes requested):

persist/postgres/objects.go
```diff
@@ -212,7 +212,7 @@ func (s *Store) DeleteObject(ctx context.Context, account proto.Account, objectK
 
 // SaveObject saves the given object for the given account. If an object with
 // the given key exists for an account, it is overwritten.
-func (s *Store) SaveObject(ctx context.Context, account proto.Account, obj slabs.Object) error {
+func (s *Store) SaveObject(ctx context.Context, account proto.Account, obj slabs.LockedObject) error {
```
peterjan: The docstring could be improved here since it's a bit misleading. Perhaps something like "If an object with identical id already exists for the given account, it will be overwritten"

api/app/client.go
```diff
@@ -177,14 +177,14 @@ func (c *Client) SlabIDs(ctx context.Context, opts ...api.URLQueryParameterOptio
 }
 
 // Object retrieves the object with the given key for the given account.
-func (c *Client) Object(ctx context.Context, key types.Hash256) (resp slabs.Object, err error) {
+func (c *Client) Object(ctx context.Context, key types.Hash256) (resp slabs.LockedObject, err error) {
```
peterjan: `key` is becoming an overloaded term imo, should we consider calling it an object id?

persist/postgres/init.sql
```diff
@@ -275,10 +275,11 @@ CREATE INDEX slabs_id_last_repair_attempt_idx ON slabs(last_repair_attempt ASC);
 CREATE TABLE objects (
     id BIGSERIAL PRIMARY KEY,
     object_key BYTEA NOT NULL CHECK(LENGTH(object_key) = 32), -- user provided, object identifier
```
peterjan: `nit` no longer applies `-- user provided, object identifier`

persist/postgres/init.sql
```diff
@@ -275,10 +275,11 @@ CREATE INDEX slabs_id_last_repair_attempt_idx ON slabs(last_repair_attempt ASC);
...
+    encrypted_master_key BYTEA NOT NULL CHECK(LENGTH(encrypted_master_key) = 72), -- user provided, master encryption key (xchacha20 nonce + key + tag)
```
peterjan: uniqueness is enforced through the SDK, so we should probably enforce it on database level too? we can always remove it afterwards

sdk/object.go
peterjan: I think this should be a method on the `LockedObject` - `ToObject` or perhaps we can call it `Unlock` even

sdk/object.go
```diff
@@ -0,0 +1,172 @@
...
+	obj := Object{
+		slabs:     slices.Clone(lo.Slabs),
+		createdAt: lo.CreatedAt,
+		updatedAt: lo.UpdatedAt,
+	}
+	keyCipher := masterKeyCipher(appKey)
+	if len(lo.EncryptedMasterKey) < keyCipher.NonceSize() {
+		return Object{}, fmt.Errorf("encrypted master key too short")
```
peterjan, line 171: We should probably validate this when persisting the object right? Or is this just a sanity check

persist/postgres/objects.go
```diff
@@ -221,10 +221,10 @@ func (s *Store) SaveObject(ctx context.Context, account proto.Account, obj slabs
 
 		var objectID int64
 		err = tx.QueryRow(ctx, `
-			INSERT INTO objects (object_key, account_id, meta) VALUES ($1, $2, $3)
-			ON CONFLICT (account_id, object_key) DO UPDATE SET meta = EXCLUDED.meta, updated_at = NOW()
+			INSERT INTO objects (object_key, account_id, encrypted_master_key, encrypted_metadata) VALUES ($1, $2, $3, $4)
+			ON CONFLICT (account_id, object_key) DO UPDATE SET (encrypted_master_key, encrypted_metadata, updated_at) = (EXCLUDED.encrypted_master_key, EXCLUDED.encrypted_metadata, NOW())
```
peterjan: ` DO UPDATE SET (encrypted_master_key, encrypted_metadata,` seems like a good candidate to cover with a unit test

peterjan (commented):
`SealedObject` covers the load slightly better imo - nice PR 🔥

sdk/object.go
ChrisSchinnerl: I thought so as well at first but the downside is that you then split up the encryption logic from the decryption logic and end up with SDK encryption logic in indexd itself.

sdk/object.go
n8mgr: > If the key is derived deterministically using a random seed and app key, sharing breaks. The receiver would need to re-encrypt the raw data using their app key or store the derived encryption key somewhere anyway. Instead we store the master key encrypted using the user's app key.

that would make it impossible for a different user to pin the shared object without re-encrypting the raw data or saving the key somewhere.

persist/postgres/objects.go
n8mgr: Out of scope

persist/postgres/objects.go
n8mgr: It is? https://github.com/SiaFoundation/indexd/pull/489/files#diff-8cd5ca928d5110cbf2d5ac6f7fe87bc0a32e4e1de30b9c22989da1007543d662R149

sdk/object.go
n8mgr: It is validated in the database

persist/postgres/objects.go
peterjan: It doesn't cover the master key. Might as well add `obj1.EncryptedMasterKey = frand.Bytes(72)` and cover all fields that require an update on save.

persist/postgres/init.sql
```diff
@@ -274,11 +274,13 @@ CREATE INDEX slabs_id_last_repair_attempt_idx ON slabs(last_repair_attempt ASC);
...
     id BIGSERIAL PRIMARY KEY,
-    object_key BYTEA NOT NULL CHECK(LENGTH(object_key) = 32), -- user provided, object identifier
+    object_key BYTEA NOT NULL CHECK(LENGTH(object_key) = 32),
+    encrypted_master_key BYTEA UNIQUE NOT NULL CHECK(LENGTH(encrypted_master_key) = 72), -- user provided, master encryption key (xchacha20 nonce + key + tag)
     account_id INTEGER REFERENCES accounts(id) NOT NULL, -- account that owns object
     created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
     updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(), -- allow sorting by update time
-    meta BYTEA -- user provided, encrypted metadata
+    encrypted_metadata BYTEA, -- user provided, encrypted metadata
+    signature BYTEA UNIQUE NOT NULL CHECK(LENGTH(signature) = 64) -- signature of blake2b(object_key || encrypted_master_key || encrypted_metadata)
```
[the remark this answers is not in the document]
n8mgr: The signature covers the encrypted key, which we are enforcing must be unique since it should be unique per object and encrypted using the account's private key. That should force signatures to also be unique per account, which is desired behavior.

n8mgr:
@ChrisSchinnerl we talked about removing `Signature` and including the encrypted key in the object ID. However, I think it's preferable to keep signature and object ID separate. We now have a signed attestation that the object ID, encryption key, and metadata have all been verified by the owner of the app key. If we remove `Signature`, a user cannot tell whether an indexer has modified the objects when recovering since they may not know the original object ID. An indexer can still delete or remove objects, but it cannot modify them without alerting the user. 

We can remove the HMAC from the encryption key and metadata to save some bytes since the signature can attest to the validity, but I don't really think it's necessary.

peterjan (dismissed):

sdk/object.go
```diff
@@ -0,0 +1,181 @@
...
+	encryptedMeta := metaCipher.Seal(nonce, nonce, o.metadata, nil)
+
+	so := slabs.SealedObject{
+		EncryptedMasterKey: encryptedMasterKey,
+		Slabs:              o.slabs,
+		EncryptedMetadata:  encryptedMeta,
+		CreatedAt:          o.createdAt,
+		UpdatedAt:          o.updatedAt,
+	}
+	so.Signature = appKey.SignHash(so.SigHash())
```
ChrisSchinnerl, line 68: How about deriving a dedicated signing key similar to the master key encryption key and metadata encryption key. 
Right now, if I come across a shared object online, I can use a previously obtained public account key to figure out if that object was shared by that account.
If we derive another key that link is broken.

sdk/object.go
```diff
@@ -0,0 +1,181 @@
...
+// UpdatedAt returns the time the object was last updated.
+func (o *Object) UpdatedAt() time.Time {
+	return o.updatedAt
+}
+
+// Seal returns a SealedObject that can be safely serialized and shared.
+func (o *Object) Seal(appKey types.PrivateKey) slabs.SealedObject {
+	keyCipher := masterKeyCipher(appKey)
+	nonce := frand.Bytes(keyCipher.NonceSize())
+	encryptedMasterKey := keyCipher.Seal(nonce, nonce, o.masterKey, nil)
```
ChrisSchinnerl, line 55: (nit) one thing that confused me when reading this is calling the encryption key of the actual data the `masterKey`. I kept thinking some other keys must be derived from it since it's the master when `appKey` is really the master key we derive from.

sdk/object.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// UpdatedAt returns the time the object was last updated.
+func (o *Object) UpdatedAt() time.Time {
+	return o.updatedAt
+}
+
+// Seal returns a SealedObject that can be safely serialized and shared.
+func (o *Object) Seal(appKey types.PrivateKey) slabs.SealedObject {
+	keyCipher := masterKeyCipher(appKey)
```
ChrisSchinnerl: If we use the object id as a salt to the key derivation, we get a unique key per object. That won't break sharing and we will not have to be worried about getting pigeon hole'd on the nonce ever.

sdk/object.go
```diff
@@ -0,0 +1,181 @@
...
+func (s *SDK) CreateSharedObjectURL(ctx context.Context, objectKey types.Hash256, validUntil time.Time) (string, error) {
+	obj, err := s.Object(ctx, objectKey)
+	if err != nil {
+		return "", fmt.Errorf("failed to get object: %w", err)
+	}
+	return s.client.CreateSharedObjectURL(ctx, obj.ID(), obj.masterKey, validUntil)
+}
+
+func masterKeyCipher(appKey types.PrivateKey) cipher.AEAD {
+	h := types.HashBytes(appKey[:])
```
ChrisSchinnerl: If Rust has a good implementation for it as well I'd use our `DeriveKey` helper or write a new one since it uses `hkdf` which provides proven domain separation compared to the simple hash approach.

I did some more research and at least Blake2b is not susceptible to hash extension attacks. Meaning that given `Hash(m)` you can't create a `Hash(m || "foo")`. If you use it in keyed some it is even considered equivalent to an HMAC. 

However append + hash still introduces avoidable correlation between derived keys during computation due to the nature of hashing being performed on blocks.

sdk/object.go
n8mgr: We don’t derive anything from the app key. It's used to encrypt the object key. The object key is random. We do derive the metadata encryption key from the object key. We can call it object key if you want. I don't really care.

sdk/object.go
n8mgr: Sure, but shared objects don’t expose the signature so there’s no leak risk from sharing.

sdk/object.go
n8mgr: Rust will require us to replace blake2b_simd with rustcrypto blake2 and hkdf. Not great, but not the end of the world.

sdk/object.go
n8mgr: Are you wanting to derive the object key encryption key from the app key instead of using the app key directly? It's unclear what you're asking for here.

sdk/object.go
ChrisSchinnerl: Ah right we return a different type when sharing

sdk/object.go
n8mgr: If you're wanting to derive the object encryption key from the object ID, we can't since we require the key to encrypt the object before uploading. We could randomly generate the object ID again, but that opens us back up to accidentally sending the encryption key to the server.

ChrisSchinnerl: No, the object encryption key should be random indeed.

I mean we can pass the object id to `masterKeyCipher` and then do

`HKDF(app_key, object_id, "master")` where `app_key` is the secret we derive from, `object_id` the salt and `master` the additional domain separator.

ChrisSchinnerl (dismissed):
If possible I'd implement the remaining comment about passing the object id to `masterKeyCipher` and `metadataCipher` as a salt but otherwise lgtm.

ChrisSchinnerl (dismissed):

openapi/app.yml
```diff
@@ -131,7 +131,7 @@ paths:
                   slabs:
                     type: array
                     items:
-                      $ref: "#/components/schemas/Object"
+                      $ref: "#/components/schemas/LockedObject"
```
peterjan:
```suggestion
                      $ref: "#/components/schemas/SealedObject"
```

peterjan: couple of these in openapi spec, won't tag them all

ChrisSchinnerl (approved):

api/app/app.go
ChrisSchinnerl, line 1: Blocker to only merge after updating the Rust SDK

PR #495, opened 2025-09-25
Add host metrics

### LINKED ISSUE #471: Host metrics
This is a tracking issue for a list of host metrics that would be useful to have to detect outlier in our hosts which might highlight malicious behavior. The goal for this endpoint is to return 1 stats object for every used host containing the following metrics:

- [x] Total Spent
- [x] Account funding
- [x] Total size of all active contracts with that host
- [x] Lost sectors of the host

A good indicator for whether a host is used could be the spending metrics and size of contracts. If all of those are 0, the host isn't used.

peterjan:
This PR adds hosts stats. Took me quite some time to get the benchmark down to a reasonable performance, considering the amount of hosts and contracts I feel it's actually quite fast. This is the query plan for 10k hosts (all usable) and 25 contracts per host (~half are active).

```
Sort  (cost=4783.43..4808.47 rows=10017 width=81) (actual time=20.362..20.730 rows=10000 loops=1)
  Sort Key: h.usage_total_spent DESC
  Sort Method: quicksort  Memory: 1166kB
  ->  Hash Right Join  (cost=667.28..4117.79 rows=10017 width=81) (actual time=3.344..18.117 rows=10000 loops=1)
        Hash Cond: (contracts.host_id = h.id)
        ->  GroupAggregate  (cost=0.29..3324.32 rows=10018 width=36) (actual time=0.023..13.333 rows=10000 loops=1)
              Group Key: contracts.host_id
              ->  Index Only Scan using contracts_active_host_size_idx on contracts  
                    (cost=0.29..2746.39 rows=90540 width=12) 
                    (actual time=0.015..6.453 rows=99770 loops=1)
                    Heap Fetches: 0
        ->  Hash  (cost=542.00..542.00 rows=9999 width=53) (actual time=3.316..3.316 rows=10000 loops=1)
              Buckets: 16384  Batches: 1  Memory Usage: 1026kB
              ->  Seq Scan on hosts h  
                    (cost=0.00..542.00 rows=9999 width=53) 
                    (actual time=0.006..2.231 rows=10000 loops=1)
                    Filter: (usage_total_spent > '0'::numeric)
Planning Time: 0.541 ms
Execution Time: 20.970 ms
```

Fixes #471

### DIFF
--- a/api/admin/types.go
+++ b/api/admin/types.go
@@ -5,6 +5,7 @@ import (
 
 	"go.sia.tech/core/types"
 	"go.sia.tech/coreutils/wallet"
+	"go.sia.tech/indexd/hosts"
 )
 
 type (
@@ -52,6 +53,14 @@ type (
 		Registered uint64 `json:"registered"`
 		Active     uint64 `json:"active"`
 	}
+
+	// HostStatsResponse is the response body for the [GET] /stats/hosts.
+	HostStatsResponse struct {
+		Hosts []HostStats `json:"hosts"`
+	}
+
+	// HostStats wraps hosts.HostStats to provide a custom PrometheusMetric method.
+	HostStats hosts.HostStats
 
 	// State is the response body for the [GET] /state endpoint.
 	State struct {
--- a/persist/postgres/stats.go
+++ b/persist/postgres/stats.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"go.sia.tech/indexd/api/admin"
+	"go.sia.tech/indexd/hosts"
 )
 
 func incrementNumAccounts(ctx context.Context, tx *txn, delta int64) error {
@@ -69,6 +70,57 @@ func (s *Store) AccountStats(ctx context.Context) (admin.AccountStatsResponse, e
 			return fmt.Errorf("failed to get active accounts: %w", err)
 		}
 		return nil
+	})
+	return stats, err
+}
+
+// HostStats reports statistics about used hosts. We consider a host to be used
+// as soon as we spent any funds on it.
+func (s *Store) HostStats(ctx context.Context, offset, limit int) ([]hosts.HostStats, error) {
+	var stats []hosts.HostStats
+	err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		rows, err := tx.Query(ctx, `
+			WITH globals AS (
+				SELECT scanned_height FROM global_settings
+			),
+			selected_hosts AS (
+				SELECT id, public_key, lost_sectors, usage_account_funding, usage_total_spent
+				FROM hosts
+				WHERE usage_total_spent > 0
+				ORDER BY usage_total_spent DESC
+				OFFSET $1
+				LIMIT $2
+			)
+			SELECT
+				h.public_key,
+				h.lost_sectors,
+				COALESCE(cs.total_contracts_size, 0) AS total_contracts_size,
+				h.usage_account_funding,
+				h.usage_total_spent
+			FROM selected_hosts h
+			LEFT JOIN LATERAL (
+			SELECT SUM(size) AS total_contracts_size
+			FROM contracts
+			WHERE host_id = h.id
+				AND (state = 0 OR state = 1)
+				AND renewed_to IS NULL
+				AND proof_height > (SELECT scanned_height FROM globals)
+			) cs ON TRUE
+			ORDER BY h.usage_total_spent DESC;
+		`, offset, limit)
+		if err != nil {
+			return err
+		}
+		defer rows.Close()
+
+		for rows.Next() {
+			var hs hosts.HostStats
+			if err := rows.Scan((*sqlPublicKey)(&hs.PublicKey), &hs.LostSectors, &hs.ActiveContractsSize, (*sqlCurrency)(&hs.AccountUsage), (*sqlCurrency)(&hs.TotalUsage)); err != nil {
+				return err
+			}
+			stats = append(stats, hs)
+		}
+		return rows.Err()
 	})
 	return stats, err
 }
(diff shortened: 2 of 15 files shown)


persist/postgres/stats.go
```diff
@@ -72,3 +72,43 @@ func (s *Store) AccountStats(ctx context.Context) (admin.AccountStatsResponse, e
...
+	err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		rows, err := tx.Query(ctx, `
+			SELECT
+				h.public_key,
+				h.lost_sectors,
+				COALESCE(cs.total_contracts_size, 0) AS total_contracts_size,
+				h.usage_account_funding,
+				h.usage_total_spent
+			FROM hosts h
+			LEFT JOIN (
```
n8mgr: Add a column to track total usable contract size, like we do for lost sectors, to remove this join. That would make the query very fast.

persist/postgres/stats.go
```diff
@@ -72,3 +72,43 @@ func (s *Store) AccountStats(ctx context.Context) (admin.AccountStatsResponse, e
...
+				SELECT host_id, SUM(size) AS total_contracts_size
+				FROM contracts
+				WHERE state = 0 OR state = 1
+				GROUP BY host_id
+				ORDER BY host_id -- this is free because of the index and ensures the query planner uses a GroupAggregate
+			) cs ON cs.host_id = h.id
+			WHERE h.usage_total_spent > 0
+			ORDER BY h.usage_total_spent DESC;
```
n8mgr, line 109: Should add pagination

peterjan: Good point sir.

persist/postgres/stats.go
peterjan: It's a lot faster now though with the updated query and the paging. I was really really hoping we could avoid that column. I thought about that when I wrote it and tried to do everything in my power to convince you all that it's more than fast enough as is.

```
BenchmarkHosts/HostStats_10       579     2097536 ns/op     11574 B/op      386 allocs/op
BenchmarkHosts/HostStats_100      516     2727935 ns/op     99687 B/op     3539 allocs/op
BenchmarkHosts/HostStats_1000     351     3336142 ns/op    944227 B/op    35046 allocs/op
```

Is it really worth all that added complexity?

api/admin/types.go
```diff
@@ -53,6 +54,12 @@ type (
 		Active     uint64 `json:"active"`
 	}
 
+	// HostStatsResponse is the response body for the [GET] /stats/hosts.
+	HostStatsResponse []HostStats
```
ChrisSchinnerl: I'd make this a struct with a `hosts` field for the slice. There is a good chance we'll add global host stats to this endpoint and this will allow us to not break the endpoint in the process.

persist/postgres/stats.go
peterjan: I updated the benchmark to use 10k hosts, each host has 25 contracts of which roughly 10 are active. With those new parameters the host benchmarks are now

## Benchmark Results

| Benchmark                  | Iterations | ns/op      | B/op    | allocs/op |
|-----------------------------|------------|------------|---------|-----------|
| BenchmarkHosts/Host-10      | 1928       |   621,214  |  12,725 |       222 |
| BenchmarkHosts/Hosts_10     |  260       | 5,194,501  |  72,934 |     1,264 |
| BenchmarkHosts/Hosts_100    |  156       | 7,742,187  | 624,948 |    11,527 |
| BenchmarkHosts/Hosts_1000   |  110       |10,631,756  |6,500,825|   114,110 |
| BenchmarkHosts/HostStats_100|  236       | 5,066,039  | 103,123 |     3,880 |
| BenchmarkHosts/HostStats_500|  184       | 6,854,192  | 495,127 |    19,248 |
| BenchmarkHosts/UpdateHost   | 2010       |   583,955  |   3,855 |       144 |

Adding the column would needlessly introduce a lot of complexity. It's different for things like number of sectors or number of slabs, there's quite a few spots where we would need to call this from, all scattered and relatively high frequency updates. At 5 and 6ms respectively for 100/500 default paging parameters, it should be fast enough imo.

persist/postgres/stats.go
```diff
@@ -72,3 +73,45 @@ func (s *Store) AccountStats(ctx context.Context) (admin.AccountStatsResponse, e
...
+				h.public_key,
+				h.lost_sectors,
+				COALESCE(cs.total_contracts_size, 0) AS total_contracts_size,
+				h.usage_account_funding,
+				h.usage_total_spent
+			FROM hosts h
+			LEFT JOIN (
+				SELECT host_id, SUM(size) AS total_contracts_size
+				FROM contracts
+				WHERE (state = 0 OR state = 1) AND renewed_to IS NULL
```
ChrisSchinnerl: This should match the contract stats endpoint query so we should also check the proof height here.

persist/postgres/stats.go
```diff
@@ -72,3 +73,45 @@ func (s *Store) AccountStats(ctx context.Context) (admin.AccountStatsResponse, e
 	})
 	return stats, err
 }
+
+// HostStats reports statistics about used hosts. We consider a host to be used
+// as soon as we spent any funds on it.
+func (s *Store) HostStats(ctx context.Context, offset, limit int) ([]hosts.HostStats, error) {
```
ChrisSchinnerl, line 79: This is missing a test

ChrisSchinnerl (changes requested):

api/admin/types.go
peterjan: Updated, not sure how to handle prometheus though. I've removed the prometheus metrics for the host stats endpoint for now. I think you could argue the host level stats are probably not _that_ useful in prometheus anyway. I was thinking perhaps we can make it into a prometheus metric once we add those global host stats and then ignore the individual host stats.

ChrisSchinnerl: The host level stats are going to be very useful to detect outliers. e.g. if we fund one host a lot more over a certain period of time it might mean they are cheating.

I'd keep the prometheus response the same for now until we figure out a better way. Whether that be a query arg, a separate endpoint or something else.

n8mgr (approved):

ChrisSchinnerl (approved):

PR #500, opened 2025-09-25
Account for active accounts when refreshing/renewing contracts

### LINKED ISSUE #497: Account for active accounts when refreshing/renewing contracts
Right now we account for active accounts when forming new contracts but we don't do that when we refresh/renew contracts yet.

chris124567:
Fix #497


contracts/form.go
```diff
@@ -88,6 +88,18 @@ func (s *formContractSigner) SignV2Inputs(txn *types.V2Transaction, toSign []int
 	s.w.SignV2Inputs(txn, toSign)
 }
 
+// activeAccounts returns the number of accounts that have been active within
+// the accountActivityThreshold, floored at 1
+func (cm *ContractManager) activeAccounts(ctx context.Context) (uint64, error) {
+	activeAccounts, err := cm.store.ActiveAccounts(ctx, time.Now().Add(-accountActivityThreshold))
+	if err != nil {
+		return 0, err
+	} else if activeAccounts == 0 {
```
peterjan: could return `return max(activeAccounts, 1), nil` - actually I'd probably get rid of this helper and call contractFunding with `max(activeAccounts,1)`

peterjan (dismissed):

peterjan (changes requested):
I was on the fence at first but we should probably add some 🧪 here. It's not _that_ tricky to add one, if we start off with no hosts, then add some accounts, then form a contract and then add more accounts, it should be relatively straightforward to assert the allowance was properly scaled up

ChrisSchinnerl:
> I was on the fence at first but we should probably add some 🧪 here. It's not _that_ tricky to add one, if we start off with no hosts, then add some accounts, then form a contract and then add more accounts, it should be relatively straightforward to assert the allowance was properly scaled up

I think we already have pretty accurate testing for that in `TestPerformContractRefreshes` and `TestPerformContractRenewals`. We could probably refactor that test to run multiple times with various numbers of active accounts.

contracts/refresh.go
```diff
@@ -59,14 +59,20 @@ func (cm *ContractManager) refreshContract(ctx context.Context, contract Contrac
...
+
 		// TODO: Right now this isn't quite correct since allowance and
 		// collateral are added on top of the existing one. This should fix
 		// itself once the new version of the refresh RPC is used. The problems
 		// are:
 		// 1. If the contract is out of funds we also add collateral and vice versa
 		// 2. The total collateral might exceed the host's maximum collateral
 		//    since 'contractFunding' doesn't take into account existing collateral
-		allowance, collateral := contractFunding(host.Settings, contract.Size, minAllowance, minHostCollateral, period)
+		allowance, collateral := contractFunding(host.Settings, contract.Size, minAllowance.Mul64(activeAccounts), minHostCollateral, period)
```
ChrisSchinnerl: I missed it in the other PR unfortunately but these should be 
```
MAX(minAllowance, accountFundTarget * activeAccounts)
```

The `ContractManager` already has access to an instance of `AccountManager` so we can just expose the configured account fund target there and fetch it.

So the result is that for less than 10 accounts, we use 10SC (`minAllowance`) and for more than 10 accounts we effectively use 1SC per account for funding since the target is 1SC right now.

ChrisSchinnerl (changes requested):

peterjan:
> > I was on the fence at first but we should probably add some 🧪 here. It's not _that_ tricky to add one, if we start off with no hosts, then add some accounts, then form a contract and then add more accounts, it should be relatively straightforward to assert the allowance was properly scaled up
> 
> I think we already have pretty accurate testing for that in `TestPerformContractRefreshes` and `TestPerformContractRenewals`. We could probably refactor that test to run multiple times with various numbers of active accounts.

We're definitely missing testing somewhere because we weren't taking into account the active accounts on renewals and refreshes. 

edit: I'm fine with it if you are, it just feels weird to merge this and chalk it up as an oversight, we generally add a small test when this happens

ChrisSchinnerl:
@chris124567 This needs an update

ChrisSchinnerl (dismissed):

accounts/manager.go
```diff
@@ -167,6 +167,11 @@ func (m *AccountManager) DeleteAccount(ctx context.Context, ak types.PublicKey)
 	return m.store.DeleteAccount(ctx, ak)
 }
 
+// FundTarget returns the configured fund target of the account manager.
+func (am *AccountManager) FundTarget() types.Currency {
```
peterjan: Instead of adding the `maxCurrency` helper I would make `FundTarget` on the manager smarter. We can probably even get away with fetching active accounts in the manager. That makes sense because it's the account manager after all. This is a good example of where the manager can hold some business logic.

`target := am.FundTarget(minAllowance)` would be sweet and we can get rid of the `maxCurrency` helper and clean up all call sites.

peterjan (changes requested):

accounts/manager.go
```diff
@@ -167,6 +175,23 @@ func (m *AccountManager) DeleteAccount(ctx context.Context, ak types.PublicKey)
...
+func (am *AccountManager) FundTarget(ctx context.Context, minAllowance types.Currency) (types.Currency, error) {
+	activeAccounts, err := am.store.ActiveAccounts(ctx, time.Now().Add(-accountActivityThreshold))
+	if err != nil {
+		return types.ZeroCurrency, err
+	}
+
+	// get the greater of the min allowance or the funding target multiplied by
+	// the number of active accounts
+	target := am.fundTarget.Mul64(activeAccounts)
+	if minAllowance.Cmp(target) == 1 {
```
peterjan, line 188: 👀

chris124567: Do you like this better?
```
	allowance := minAllowance
	if v := am.fundTarget.Mul64(activeAccounts); v.Cmp(allowance) == 1 {
		allowance = v
	}
```

peterjan: Oh brain fart sorry. I didn't see the `1`. I always write `.Cmp() > 0`

peterjan (approved):

n8mgr (approved):

PR #501, opened 2025-09-25
Remove `AddAccount`

### LINKED ISSUE #493: Cleanup: Remove `AddAccount`
`AddAccount` is not used anymore apart from testing. So we can get rid of it and call `addAccount` in our database tests directly and add a helper method to `Indexer` for our integration tests which first creates an app key and then uses that key.

chris124567:
Close #493


internal/testutils/indexer.go
```diff
@@ -314,6 +315,25 @@ func (idx *Indexer) WalletAddr() types.Address {
 	return idx.wallet.Address()
 }
 
+// AddAccount creates a test app connect key if it does not already exist and
+// creates an account using it.
+func (idx *Indexer) AddAccount(ctx context.Context, ak types.PublicKey, meta accounts.AccountMeta) error {
```
peterjan: If we pass `t *testing.T` instead of the context we can drop the `error` and use `t.Context()`. I was considering `AddAppAccount` maybe but I think it's a little clunky, `AddAccount` is probably fine.

internal/testutils/indexer.go
```diff
@@ -314,6 +315,25 @@ func (idx *Indexer) WalletAddr() types.Address {
...
+	store := idx.Store()
+
+	const connectKey = "test"
+	_, err := store.AddAppConnectKey(ctx, accounts.UpdateAppConnectKey{
+		Key:           connectKey,
+		MaxPinnedData: 1e10,
+		RemainingUses: 10000,
+	})
+	// ignore if key was already inserted
+	if err != nil && !strings.Contains(err.Error(), "duplicate") {
```
peterjan: The store should handle this more gracefully. Look for `pgerrcode.UniqueViolation` and have `AddAppConnectKey` return `ErrAppConnectKeyAlreadyExists` so we can catch it here and ignore it.  Those store methods have an incorrect usage of `fmt.Errorf` where they check max pinned data. I'd add a decent error for that too while we're at it.

peterjan (changes requested):

internal/testutils/indexer.go
ChrisSchinnerl: (nit) `ErrAppConnectKeyExists` to be consistent

peterjan (commented):
Looks good to me. Linter complained about always passing`accounts.AccountMeta{}` so probably ok to remove the parameter.

internal/testutils/indexer.go
ChrisSchinnerl: Looks like this was not updated? So I'd like to add to pass `testing.TB` rather than the concrete type

ChrisSchinnerl (changes requested):

peterjan (dismissed):

ChrisSchinnerl (commented):
New merge conflicts

n8mgr (approved):

peterjan (approved):

ChrisSchinnerl (approved):

Issue #469, opened 2025-09-22
Account for unknown spending when syncing a revision


ChrisSchinnerl:
It's possible that we have an outdated revision when revising a contract gets interrupted. 
When that happens we should also track that sort of spending in an "unknown" or similar category.

peterjan:
Just checking to see if we're talking about the same thing here, when we revise a contract it's possible a failure occurs. It's possible however the usage is not nil and we currently ignore that. We sync and then try and revise the contract again, overriding the usage. _That_ is what you mean when you say "that sort of spending" right? Why should we track that under an `unknown` category? It'll be usage so we can track it as normal?  @ChrisSchinnerl

ChrisSchinnerl:
Yeah that spending. We can add it to the total spending for this issue. If we ever break the spending down further it might be better to be explicit since we don't know what we spent it on.

peterjan:
Done in https://github.com/SiaFoundation/indexd/pull/510 - GH can't handle `Fixes X & Y` apparently.

PR #510, opened 2025-09-29
Atomically update host usage

### LINKED ISSUE #491: Update host usage atomically
Right now updating a contract in the database and increasing host usage is not atomic. So if we fail to persist the host usage after revising a contract, the revision in the database gets updated while the usage remains the same.

We should change that by searching for all calls to `UpdateHostUsage` and merging that behavior into whatever method we call before that to add or update a contract.

peterjan:
Fixes  https://github.com/SiaFoundation/indexd/issues/491 &  https://github.com/SiaFoundation/indexd/issues/469

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -56,8 +56,7 @@ type (
 	// contract's revision.
 	RevisionStore interface {
 		ContractRevision(ctx context.Context, contractID types.FileContractID) (rhp.ContractRevision, bool, error)
-		UpdateContractRevision(ctx context.Context, contract rhp.ContractRevision) error
-		UpdateHostUsage(ctx context.Context, hostKey types.PublicKey, usage proto.Usage) error
+		UpdateContractRevision(ctx context.Context, contract rhp.ContractRevision, usage proto.Usage) error
 	}
 )
 
@@ -141,8 +140,6 @@ func (c *HostClient) FormContract(ctx context.Context, settings proto.HostSettin
 	res, err := rhp.RPCFormContract(ctx, c.client, c.cm, c.signer, c.cm.TipState(), settings.Prices, c.hostKey, settings.WalletAddress, params)
 	if err != nil {
 		return rhp.RPCFormContractResult{}, fmt.Errorf("failed to form contract: %w", err)
-	} else if err := c.store.UpdateHostUsage(ctx, c.hostKey, res.Usage); err != nil {
-		return rhp.RPCFormContractResult{}, fmt.Errorf("failed to update host usage: %w", err)
 	}
 	return res, nil
 }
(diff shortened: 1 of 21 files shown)


client/client.go
```diff
@@ -282,7 +279,7 @@ func (c *HostClient) WriteSector(ctx context.Context, settings proto.HostPrices,
 	return res, nil
 }
 
-func (c *HostClient) syncRevision(ctx context.Context, contractID types.FileContractID, revision types.V2FileContract) (types.V2FileContract, bool, error) {
+func (c *HostClient) syncRevision(ctx context.Context, contractID types.FileContractID, revision types.V2FileContract, usage proto.Usage) (types.V2FileContract, bool, error) {
```
peterjan: This little change essentially tackles https://github.com/SiaFoundation/indexd/issues/469.

I want to draw attention to it because currently we still would not track usage in the event we fail to revise the contract, accrue usage, but then afterwards fail to fetch the latest revision from the host (or if our local version is newer for some reason). I think that's a bit of an edge case and maybe it is fine to ignore that usage?  

If we really want to track that, we would have to add a separate method on the store only to update the usage and that's kind of sad. This whole PR is about atomically updating usage when we update or persist the revision.

Updating usage currently only happens in `AddFormedContract`, `AddRenewedContract` and `UpdateContractRevision`, which all make sense. I even briefly considered merging everything into an `UpsertContract` but it gets messy.

How do you all feel about that?

persist/postgres/contracts.go
```diff
@@ -139,14 +134,24 @@ func (s *Store) AddFormedContract(ctx context.Context, hostKey types.PublicKey,
 		} else if resp.RowsAffected() != 1 {
 			return fmt.Errorf("expected 1 row to be affected, got %d", resp.RowsAffected())
 		}
+
+		// update host usage
+		query := `UPDATE hosts SET usage_account_funding = usage_account_funding + $1, usage_total_spent = usage_total_spent + $2 WHERE id = $3`
```
ChrisSchinnerl: If you use the public key for the host here you can extract this and other duplicated calls into a helper function.

client/client.go
ChrisSchinnerl: This doesn't seem quite right. When `withRevision` is used, we execute an RPC that is paid for by contract. Considering that we call `syncRevision` after the RPC failed, `usage` should be 0 since the contract wasn't revised.

On the other hand I'm missing something like the following in here 
```go
missedUsage := revision.RemainingAllowance().Sub(resp.Contract.RemainingAllowance())
```
before calling `UpdateContractRevision` with `missedUsage` to catch up on any contract spending we missed.

ChrisSchinnerl (changes requested):

client/client.go
peterjan: Ah I misunderstood. Thanks, will update PR accordingly

persist/postgres/contracts.go
peterjan: Thanks for catching that

chris124567 (approved):

ChrisSchinnerl (approved):

n8mgr (approved):

Issue #478, opened 2025-09-23
Update `contracts.State` indices and queries


ChrisSchinnerl:
Something interesting I came across is that 

```
CREATE INDEX contracts_state_active_idx ON contracts(state) WHERE state = 0 OR state = 1;
```
combined with
```
WHERE state = 0 OR state = 1
```
in a query is about twice as fast in the contracts stats query as switching to `state <= 1` for both index and query.

Considering that, we should go through our queries and indices, try to update them, re-run the benchmarks to confirm the improvement and then use that instead of `<=`.

PR #511, opened 2025-09-29
Efficiently pin sectors across contracts

n8mgr:
Changes the pin sectors function to determine the maximum number of sectors that can be pinned given the current revision of a contract instead of attempting to pin all sectors and failing to the next contract. Also changes the sort to prefer contracts with more empty capacity to "refill" contracts that have been pruned.

This should be a more efficient use of space when pinning large numbers of sectors or for contracts that have been pruned.

```
BenchmarkContracts/contracts_for_pinning-16           	    3490	    325186 ns/op	    5589 B/op	     132 allocs/op
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -113,7 +113,7 @@ func (c *HostClient) AccountBalance(ctx context.Context, account proto.Account) 
 // The integer returned does not indicate the number of sectors that were
 // appended, but rather the number of sectors that were attempted. Check the
 // result for the actual number of sectors that were appended.
-func (c *HostClient) AppendSectors(ctx context.Context, hostPrices proto.HostPrices, contractID types.FileContractID, sectors []types.Hash256) (res rhp.RPCAppendSectorsResult, remaining int, err error) {
+func (c *HostClient) AppendSectors(ctx context.Context, hostPrices proto.HostPrices, contractID types.FileContractID, sectors []types.Hash256) (res rhp.RPCAppendSectorsResult, attempted int, err error) {
 	// sanity check
 	if len(sectors) > proto.MaxSectorBatchSize {
 		return rhp.RPCAppendSectorsResult{}, 0, fmt.Errorf("too many sectors, %d > %d", len(sectors), proto.MaxSectorBatchSize) // developer error
@@ -142,7 +142,7 @@ func (c *HostClient) AppendSectors(ctx context.Context, hostPrices proto.HostPri
 		// only attempt to append up to the calculated maximum number of sectors
 		if uint64(len(sectors)) > maxAppendSectors {
 			sectors = sectors[:maxAppendSectors]
-			remaining = len(sectors)
+			attempted = len(sectors)
 		}
 
 		res, err = rhp.RPCAppendSectors(ctx, c.client, c.signer, c.cm.TipState(), hostPrices, contract, sectors)
--- a/client/client.go
+++ b/client/client.go
@@ -106,29 +106,53 @@ func (c *HostClient) AccountBalance(ctx context.Context, account proto.Account) 
 	return rhp.RPCAccountBalance(ctx, c.client, account)
 }
 
-// AppendSectors appends the given sectors to the contract.
-func (c *HostClient) AppendSectors(ctx context.Context, hostPrices proto.HostPrices, contractID types.FileContractID, sectors []types.Hash256) (rhp.RPCAppendSectorsResult, error) {
+// AppendSectors appends the given sectors to the contract. If the contract
+// cannot fit all sectors, as many as possible will be appended and the number of
+// sectors attempted will be returned.
+//
+// The integer returned does not indicate the number of sectors that were
+// appended, but rather the number of sectors that were attempted. Check the
+// result for the actual number of sectors that were appended.
+func (c *HostClient) AppendSectors(ctx context.Context, hostPrices proto.HostPrices, contractID types.FileContractID, sectors []types.Hash256) (res rhp.RPCAppendSectorsResult, remaining int, err error) {
 	// sanity check
 	if len(sectors) > proto.MaxSectorBatchSize {
-		return rhp.RPCAppendSectorsResult{}, fmt.Errorf("too many sectors, %d > %d", len(sectors), proto.MaxSectorBatchSize) // developer error
+		return rhp.RPCAppendSectorsResult{}, 0, fmt.Errorf("too many sectors, %d > %d", len(sectors), proto.MaxSectorBatchSize) // developer error
 	}
 
 	// append sectors
-	var res rhp.RPCAppendSectorsResult
-	if err := c.withRevision(ctx, contractID, func(contract rhp.ContractRevision) (_ rhp.ContractRevision, _ proto.Usage, err error) {
+	err = c.withRevision(ctx, contractID, func(contract rhp.ContractRevision) (_ rhp.ContractRevision, _ proto.Usage, err error) {
 		if contract.Revision.Filesize > maxContractSize {
 			return rhp.ContractRevision{}, proto.Usage{}, fmt.Errorf("contract is too large, %d > %d", contract.Revision.Filesize, maxContractSize)
 		}
+		// calculate the maximum number of sectors we can append based on the
+		// contract's remaining capacity and collateral
+		maxRemainingSectors := (maxContractSize - contract.Revision.Filesize) / proto.SectorSize
+		var maxAppendSectors uint64
+		if contract.Revision.Filesize < contract.Revision.Capacity {
+			maxAppendSectors = (contract.Revision.Capacity - contract.Revision.Filesize) / proto.SectorSize
+		}
+		sectorCollateralCost := hostPrices.RPCAppendSectorsCost(1, contract.Revision.ExpirationHeight-hostPrices.TipHeight).RiskedCollateral
+		if sectorCollateralCost.IsZero() {
+			sectorCollateralCost = types.NewCurrency64(1) // avoid division by zero
+		}
+		maxAppendSectors += contract.Revision.RemainingCollateral().Div(sectorCollateralCost).Big().Uint64()
+		// ensure the maximum contract size is not exceeded
+		maxAppendSectors = min(maxAppendSectors, maxRemainingSectors)
+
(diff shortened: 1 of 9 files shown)


client/client.go
```diff
@@ -106,29 +106,53 @@ func (c *HostClient) AccountBalance(ctx context.Context, account proto.Account)
...
+		var maxAppendSectors uint64
+		if contract.Revision.Filesize < contract.Revision.Capacity {
+			maxAppendSectors = (contract.Revision.Capacity - contract.Revision.Filesize) / proto.SectorSize
+		}
+		sectorCollateralCost := hostPrices.RPCAppendSectorsCost(1, contract.Revision.ExpirationHeight-hostPrices.TipHeight).RiskedCollateral
+		if sectorCollateralCost.IsZero() {
+			sectorCollateralCost = types.NewCurrency64(1) // avoid division by zero
+		}
+		maxAppendSectors += contract.Revision.RemainingCollateral().Div(sectorCollateralCost).Big().Uint64()
+		// ensure the maximum contract size is not exceeded
```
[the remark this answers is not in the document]
n8mgr: The existing free capacity is already paid for, adding the collateral-based constraint to the capacity-constraint is correct

client/client.go
```diff
@@ -106,29 +106,52 @@ func (c *HostClient) AccountBalance(ctx context.Context, account proto.Account)
...
 	// append sectors
-	var res rhp.RPCAppendSectorsResult
-	if err := c.withRevision(ctx, contractID, func(contract rhp.ContractRevision) (_ rhp.ContractRevision, _ proto.Usage, err error) {
+	err = c.withRevision(ctx, contractID, func(contract rhp.ContractRevision) (_ rhp.ContractRevision, _ proto.Usage, err error) {
 		if contract.Revision.Filesize > maxContractSize {
 			return rhp.ContractRevision{}, proto.Usage{}, fmt.Errorf("contract is too large, %d > %d", contract.Revision.Filesize, maxContractSize)
 		}
+		// calculate the maximum number of sectors we can append based on the
+		// contract's remaining capacity and collateral
+		maxRemainingSectors := (maxContractSize - contract.Revision.Filesize) / proto.SectorSize
```
peterjan, line 131: `nit` should bail early if size matches the `maxContractSize`, so on line 124 we should do `contract.Revision.Filesize >= maxContractSize`

client/client.go
```diff
@@ -106,29 +106,52 @@ func (c *HostClient) AccountBalance(ctx context.Context, account proto.Account)
...
+		var maxAppendSectors uint64
+		if contract.Revision.Filesize < contract.Revision.Capacity {
+			maxAppendSectors = (contract.Revision.Capacity - contract.Revision.Filesize) / proto.SectorSize
+		}
+		sectorCollateralCost := hostPrices.RPCAppendSectorsCost(1, contract.Revision.ExpirationHeight-hostPrices.TipHeight).RiskedCollateral
```
peterjan: sanity check `contract.Revision.ExpirationHeight` is greater than the `hostPrices.TipHeight` to avoid an underflow

client/client.go
```diff
@@ -106,29 +106,52 @@ func (c *HostClient) AccountBalance(ctx context.Context, account proto.Account)
...
+		if sectorCollateralCost.IsZero() {
+			sectorCollateralCost = types.NewCurrency64(1) // avoid division by zero
+		}
+		maxAppendSectors += contract.Revision.RemainingCollateral().Div(sectorCollateralCost).Big().Uint64()
+		// ensure the maximum contract size is not exceeded
+		maxAppendSectors = min(maxAppendSectors, maxRemainingSectors)
```
peterjan, line 140: can `maxAppendSectors` ever be zero here? It doesn't hurt to add a sanity check before executing the RPC here to ensure we're effectively trying to append sectors

client/client.go
```diff
@@ -106,29 +106,52 @@ func (c *HostClient) AccountBalance(ctx context.Context, account proto.Account)
...
+		maxRemainingSectors := (maxContractSize - contract.Revision.Filesize) / proto.SectorSize
+		var maxAppendSectors uint64
+		if contract.Revision.Filesize < contract.Revision.Capacity {
+			maxAppendSectors = (contract.Revision.Capacity - contract.Revision.Filesize) / proto.SectorSize
+		}
+		sectorCollateralCost := hostPrices.RPCAppendSectorsCost(1, contract.Revision.ExpirationHeight-hostPrices.TipHeight).RiskedCollateral
+		if sectorCollateralCost.IsZero() {
+			sectorCollateralCost = types.NewCurrency64(1) // avoid division by zero
+		}
+		maxAppendSectors += contract.Revision.RemainingCollateral().Div(sectorCollateralCost).Big().Uint64()
```
peterjan, line 138: `nit` we would benefit from more descriptive variable names here

```
withinCurrentCapacity := (rev.FS - rev.Cap) / proto.SS
additionalFromCollateral := remainingColl / sectorCollCost
allowedByProtocol := maxCS / proto.SS

maxAppend := withinCurrentCapacity + additionalFromCollateral
if maxAppend > allowedByProtocol {
   maxAppend = allowedByProtocol
}
```

client/client.go
```diff
@@ -106,29 +106,52 @@ func (c *HostClient) AccountBalance(ctx context.Context, account proto.Account)
...
-	if err := c.withRevision(ctx, contractID, func(contract rhp.ContractRevision) (_ rhp.ContractRevision, _ proto.Usage, err error) {
+	err = c.withRevision(ctx, contractID, func(contract rhp.ContractRevision) (_ rhp.ContractRevision, _ proto.Usage, err error) {
 		if contract.Revision.Filesize > maxContractSize {
 			return rhp.ContractRevision{}, proto.Usage{}, fmt.Errorf("contract is too large, %d > %d", contract.Revision.Filesize, maxContractSize)
 		}
+		// calculate the maximum number of sectors we can append based on the
+		// contract's remaining capacity and collateral
+		maxRemainingSectors := (maxContractSize - contract.Revision.Filesize) / proto.SectorSize
+		var maxAppendSectors uint64
+		if contract.Revision.Filesize < contract.Revision.Capacity {
```
peterjan: `nit` this check is redundant because the capacity is always greater or equal, so we can safely do cap - size / sector size

persist/postgres/contracts.go
```diff
@@ -321,7 +321,8 @@ SELECT c.contract_id
 FROM contracts c
 INNER JOIN hosts h ON c.host_id = h.id
 WHERE h.public_key = $1 AND c.good = TRUE AND c.state <= $2 AND c.remaining_allowance > 0 AND c.size < $3
-ORDER BY c.capacity DESC, c.size DESC`, sqlPublicKey(hk), sqlContractState(contracts.ContractStateActive), maxContractSize)
+-- sort by empty capacity desc to prefer heavily pruned contracts 
+ORDER BY (c.capacity - c.size) DESC`, sqlPublicKey(hk), sqlContractState(contracts.ContractStateActive), maxContractSize)
```
ChrisSchinnerl, line 328: Might want an index for this since right now we are doing a quicksort according to `explain analyze`.

peterjan (changes requested):

contracts/pinning.go
```diff
@@ -104,82 +91,65 @@ func (cm *ContractManager) performSectorPinningOnHost(ctx context.Context, host
...
-			contractLog.Debug("no sectors were pinned")
+			log.Debug("no sectors were pinned")
 			continue
 		}
 
-		// figure out which sectors were missing if necessary
-		if len(res.Sectors) != len(sectors) {
-			lookup := make(map[types.Hash256]struct{}, len(sectors))
-			for _, sector := range sectors {
+		// TODO: this should be atomic with the contract revision. Any time a
```
ChrisSchinnerl: We can do that but it won't guarantee consistency in case the transaction gets rolled back after successfully appending the sectors to the contract. We will then end up with an inconsistent state after syncing the revision with the host because we get the latest revision but not the latest sectors in the contract. Which is the same state we end up in right now if the first transaction goes through but the second one fails.

contracts/pinning.go
```diff
@@ -104,82 +91,65 @@ func (cm *ContractManager) performSectorPinningOnHost(ctx context.Context, host
...
+		// revision is saved the sectors that were added or removed should
+		// be updated in the same transaction for consistency.
+		if err := cm.store.PinSectors(ctx, contractID, res.Sectors); err != nil {
+			return fmt.Errorf("failed to pin sectors: %w", err)
+		}
+
+		// Only the sectors that were attempted should be marked
+		// as missing. So sectors that were not part of the append
+		// call can be pinned to other contracts.
+		if len(res.Sectors) != len(sectors[:attempted]) {
```
ChrisSchinnerl:
```suggestion
		if len(res.Sectors) != attempted {
```

persist/postgres/init.sql
```diff
@@ -223,7 +223,7 @@ CREATE INDEX contracts_state_formation_idx ON contracts(state, formation); -- fo
 CREATE INDEX contracts_state_good_idx ON contracts(state) WHERE state <= 1 AND good; -- for filtering contracts
 CREATE INDEX contracts_last_broadcast_attempt_contract_id_idx ON contracts (last_broadcast_attempt ASC, contract_id) WHERE renewed_to IS NULL; -- for fetching contracts for broadcasting
 CREATE INDEX contracts_host_id_remaining_allowance_contract_id_idx ON contracts (host_id, remaining_allowance DESC, contract_id) WHERE good = true AND remaining_allowance > 0; -- for fetching contracts for funding
-CREATE INDEX contracts_capacity_size_contract_id_idx ON contracts (capacity DESC, size DESC, contract_id) WHERE good = true AND remaining_allowance > 0; -- for fetching contracts for pinning
+CREATE INDEX contracts_capacity_size_contract_id_idx ON contracts (host_id, (capacity - size) DESC, size) WHERE good = true AND state <= 1 AND remaining_allowance > 0; -- for fetching contracts for pinning
```
peterjan, line 227: I'll take it along in https://github.com/SiaFoundation/indexd/issues/478

peterjan (dismissed):

ChrisSchinnerl (dismissed):

PR #518, opened 2025-10-02
Improve contract query performance

### LINKED ISSUE #478: Update `contracts.State` indices and queries

peterjan:
I went over every index on our contracts table and used our benchmarks to analyze every query plan individually. Our benchmarks don't always reflect improvements, but I did manage to get quite a lot of performance gains. 

I made tiny changes to our test setup:
- random contracts more accurately reflect production usage (used zeus)
- vacuum contracts and hosts table more accurately reflects a running instance

The obvious wins are:

| Benchmark | Before (plan) | After (plan) | Best time (before → after) | Change |
| -- | -- | -- | -- | --- |
| contracts_for_broadcasting | BitmapHeapScan + Top-N sort | Index-only scan on contracts_last_broadcast_attempt_active_idx (ORDER BY covered) | 7.02 ms → 0.86 ms | ~88% faster |
| contracts_for_funding (limit) | Nested loop + sort on remaining_allowance | Index-only scan on contracts_host_id_remaining_allowance_active_idx (ORDER BY covered) | 0.46 ms → 0.41 ms | ~11% faster |
| reject_expired_contracts | Heap scan of pending + time check | Partial index contracts_formation_pending_idx | 10.11 ms → 5.22 ms | ~48% faster |
| contracts_stats (active slice) | Broader scans | Index-only scan on contracts_active_host_size_idx (INCLUDEs) | 2.45 ms → 0.98 ms | ~60% faster |

The query plan definitely improved for the main `store.Contracts` query, but it's not clear in the benchmarks. I dumped all query plans pre and post change in Chat GPT though and this is the result:

Benchmark | Before (plan) | After (plan) | Change
-- | -- | -- | --- |
contracts_1000 | Merge Join on host_id; Index Scan contracts_host_id_idx + hosts_pkey | Merge Join on host_id; Index Scan contracts_host_id_idx + hosts_pkey (unchanged) | +2.3%
contracts_revisable_good_limit_1000 | Merge Join; Index Scan contracts_host_active_idx with Filter (good AND renewed_to IS NULL) | Merge Join; Index Scan contracts_host_id_active_good_idx (predicate-matching partial) | +16.0%
contracts_revisable_bad_limit_1000 | Merge Join; Index Scan contracts_host_active_idx with Filter (!good AND renewed_to IS NULL) | Merge Join; Index Scan contracts_host_id_active_bad_idx (predicate-matching partial) | +39.3%
contracts_nonrevisable_good_limit_1000 | Nested Loop; Seq Scan contracts (filter good AND state>1) + Memoize → hosts_pkey | Merge Join; Index Scan contracts_host_id_inactive_good_idx (partial) | +19.4%
contracts_nonrevisable_bad_limit_1000 | Nested Loop; Seq Scan contracts (filter !good AND state>1) + Memoize → hosts_pkey | Merge Join; Index Scan contracts_host_id_inactive_bad_idx (partial) | +45.0%

There is no regression, if the benchmark ended up being slightly slower (<5%) I went in and verified it manually. All query plans are either the same or have improved. Unfortunately that doesn't always show in the benchmarks at every possible limit.

I can pull these out into separate PRs if you want, I did not do so because it's 200 lines, because I want to avoid the endless migration conflicts, but most importantly because I wrote some custom code to easily compare my changes before and after making changes to the schema and print a query plan, it was just easier to do everything in one go.

This should resolve https://github.com/SiaFoundation/indexd/issues/478 - I also rewrote our queries to be more consistent over all, but kept it focused on contracts only.

### DIFF
--- a/persist/postgres/contracts.go
+++ b/persist/postgres/contracts.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/jackc/pgx/v5"
@@ -38,8 +39,8 @@ func (s *Store) ContractsStats(ctx context.Context) (resp admin.ContractsStatsRe
 		var numContracts, numGood, totalCapacity, totalSize uint64
 		err := tx.QueryRow(ctx, `
 			WITH globals AS (
-    			SELECT scanned_height FROM global_settings
-       		)
+			SELECT scanned_height FROM global_settings
+			)
 			SELECT
 				COUNT(*),       				-- non-expired contracts
 				COALESCE(SUM(good::int), 0), 	-- good contracts
@@ -48,11 +49,10 @@ func (s *Store) ContractsStats(ctx context.Context) (resp admin.ContractsStatsRe
 			FROM contracts
 			CROSS JOIN globals
 			WHERE
-				proof_height > globals.scanned_height AND
+				state IN (0,1) AND
 				renewed_to IS NULL AND
-				(state = $1 OR state = $2)
-		`, contracts.ContractStatePending, contracts.ContractStateActive).
-			Scan(&numContracts, &numGood, &totalCapacity, &totalSize)
+				proof_height > globals.scanned_height
+		`).Scan(&numContracts, &numGood, &totalCapacity, &totalSize)
 		if err != nil {
 			return err
 		}
@@ -66,12 +66,11 @@ func (s *Store) ContractsStats(ctx context.Context) (resp admin.ContractsStatsRe
 			FROM contracts
 			CROSS JOIN globals
 			WHERE
-				proof_height > globals.scanned_height AND
+				state IN (0,1) AND
 				renewed_to IS NULL AND
-				(state = $1 OR state = $2) AND
+				proof_height > globals.scanned_height AND
 				globals.scanned_height + globals.contracts_renew_window >= proof_height
-		`, contracts.ContractStatePending, contracts.ContractStateActive).
-			Scan(&numRenewing)
+		`).Scan(&numRenewing)
 		if err != nil {
 			return err
 		}
@@ -216,6 +215,7 @@ func (s *Store) Contracts(ctx context.Context, offset, limit int, queryOpts ...c
 	var contracts []contracts.Contract
 	if err := s.transaction(ctx, func(ctx context.Context, tx *txn) (err error) {
 		rows, err := tx.Query(ctx, `
+EXPLAIN (ANALYZE, BUFFERS, TIMING ON)
 SELECT c.contract_id, c.formation, h.public_key, c.proof_height, c.expiration_height, c.renewed_from, c.renewed_to, c.revision_number, c.state, c.capacity, c.size, c.contract_price, c.initial_allowance, c.remaining_allowance, c.miner_fee, c.used_collateral, c.total_collateral, c.good, c.append_sector_spending, c.free_sector_spending, c.fund_account_spending, c.sector_roots_spending, c.next_prune, c.last_broadcast_attempt
 FROM contracts c
 INNER JOIN hosts h ON c.host_id = h.id
@@ -225,8 +225,8 @@ WHERE
 	-- active filter
 	(
 		$2::boolean IS NULL OR
-		($2::boolean = TRUE AND c.state <= 1 AND c.renewed_to IS NULL) OR
-		($2::boolean = FALSE AND c.state > 1)
+		($2::boolean = TRUE AND c.state IN (0,1) AND c.renewed_to IS NULL) OR
+		($2::boolean = FALSE AND c.state IN (2,3,4))
 	)
 LIMIT $3 OFFSET $4`, opts.Good, opts.Revisable, limit, offset)
 		if err != nil {
@@ -234,13 +234,20 @@ LIMIT $3 OFFSET $4`, opts.Good, opts.Revisable, limit, offset)
 		}
 		defer rows.Close()
 
+		var sb strings.Builder
 		for rows.Next() {
+			var line string
+			rows.Scan(&line)
+			sb.WriteString(line + "\n")
+			continue
+
 			contract, err := scanContract(rows)
 			if err != nil {
 				return fmt.Errorf("failed to scan contract: %w", err)
 			}
 			contracts = append(contracts, contract)
 		}
+		fmt.Println(sb.String())
 		return rows.Err()
 	}); err != nil {
 		return nil, err
@@ -257,11 +264,11 @@ func (s *Store) ContractsForBroadcasting(ctx context.Context, minBroadcast time.
 	var fcids []types.FileContractID
 	err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
 		rows, err := tx.Query(ctx, `
-SELECT c.contract_id
-FROM contracts c
-WHERE c.renewed_to IS NULL AND c.state <= $1 AND c.last_broadcast_attempt < $2
-ORDER BY c.last_broadcast_attempt ASC
-LIMIT $3`, sqlContractState(contracts.ContractStateActive), minBroadcast, limit)
+			SELECT c.contract_id
+			FROM contracts c
+			WHERE c.state IN (0, 1) AND c.renewed_to IS NULL AND c.last_broadcast_attempt < $1
+			ORDER BY c.last_broadcast_attempt ASC
+			LIMIT $2`, minBroadcast, limit)
 		if err != nil {
 			return fmt.Errorf("failed to fetch contracts for broadcasting: %w", err)
 		}
@@ -289,11 +296,15 @@ func (s *Store) ContractsForFunding(ctx context.Context, hk types.PublicKey, lim
 		rows, err := tx.Query(ctx, `
 SELECT c.contract_id
 FROM contracts c
-INNER JOIN hosts h ON c.host_id = h.id
-WHERE h.public_key = $1 AND c.good = TRUE AND c.state <= $2 AND c.remaining_allowance > 0
+WHERE
+	c.host_id = (SELECT id FROM hosts WHERE public_key = $1) AND
+	c.state IN (0, 1) AND
+	c.renewed_to IS NULL AND
+	c.good AND
+	c.remaining_allowance > 0
 ORDER BY c.remaining_allowance DESC
-LIMIT $3
-`, sqlPublicKey(hk), sqlContractState(contracts.ContractStateActive), limit)
+LIMIT $2
+`, sqlPublicKey(hk), limit)
 		if err != nil {
 			return fmt.Errorf("failed to fetch contracts for funding: %w", err)
 		}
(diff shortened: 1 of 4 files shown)


persist/postgres/init.sql
```diff
@@ -220,29 +220,40 @@ CREATE TABLE contracts (
   -- raw contract revision data
   raw_revision BYTEA NOT NULL
 );
-CREATE INDEX contracts_state_formation_idx ON contracts(state, formation); -- for rejecting expired contracts
-CREATE INDEX contracts_state_good_idx ON contracts(state) WHERE state <= 1 AND good; -- for filtering contracts
-CREATE INDEX contracts_last_broadcast_attempt_contract_id_idx ON contracts (last_broadcast_attempt ASC, contract_id) WHERE renewed_to IS NULL; -- for fetching contracts for broadcasting
-CREATE INDEX contracts_host_id_remaining_allowance_contract_id_idx ON contracts (host_id, remaining_allowance DESC, contract_id) WHERE good = true AND remaining_allowance > 0; -- for fetching contracts for funding
-CREATE INDEX contracts_capacity_size_contract_id_idx ON contracts (host_id, (capacity - size) DESC, size) WHERE good = true AND state <= 1 AND remaining_allowance > 0; -- for fetching contracts for pinning
```
peterjan, line 227: @n8mgr I ran into a little hick-up here, no matter how hard I nudge the planner to use this index, it always prefers `contracts_host_id_active_good_idx`. 

On master:
```
Sort  (cost=48.22..48.23 rows=4 width=40) (actual time=0.050..0.051 rows=16 loops=1)
  Sort Key: ((c.capacity - c.size)) DESC
  Sort Method: quicksort  Memory: 26kB
  Buffers: shared hit=12
  ->  Nested Loop  (cost=4.84..48.18 rows=4 width=40) (actual time=0.034..0.044 rows=16 loops=1)
        Buffers: shared hit=12
        ->  Index Scan using hosts_public_key_key on hosts h  (cost=0.27..8.29 rows=1 width=4) (actual time=0.016..0.016 rows=1 loops=1)
              Index Cond: (public_key = '\x84a80a7b8766527355ee4d6be9bacb643bfb7bc50bf58be3442572ef03de2873'::bytea)
              Buffers: shared hit=3
        ->  Bitmap Heap Scan on contracts c  (cost=4.57..39.79 rows=9 width=52) (actual time=0.017..0.024 rows=16 loops=1)
              Recheck Cond: ((h.id = host_id) AND (size < '10995116277760'::bigint) AND good AND (state <= 1) AND (remaining_allowance > '0'::numeric))
              Heap Blocks: exact=7
              Buffers: shared hit=9
              ->  Bitmap Index Scan on contracts_capacity_size_contract_id_idx  (cost=0.00..4.57 rows=9 width=0) (actual time=0.007..0.007 rows=16 loops=1)
                    Index Cond: ((host_id = h.id) AND (size < '10995116277760'::bigint))
                    Buffers: shared hit=2
Planning:
  Buffers: shared hit=143
Planning Time: 0.302 ms
Execution Time: 0.094 ms
```

On branch:

```
Sort  (cost=44.36..44.38 rows=10 width=41) (actual time=0.042..0.042 rows=8 loops=1)
  Sort Key: ((c.capacity - c.size)) DESC
  Sort Method: quicksort  Memory: 25kB
  Buffers: shared hit=12
  InitPlan 1
    ->  Index Scan using hosts_public_key_key on hosts  (cost=0.28..8.29 rows=1 width=4) (actual time=0.005..0.006 rows=1 loops=1)
          Index Cond: (public_key = '\x19642ba0534f5bf6aa7085e14f031c2835ef2ac3a32bb1cfee1815e77501dcdd'::bytea)
          Buffers: shared hit=3
  ->  Index Scan using contracts_host_id_active_good_idx on contracts c  (cost=0.29..35.90 rows=10 width=41) (actual time=0.012..0.019 rows=8 loops=1)
        Index Cond: (host_id = (InitPlan 1).col1)
        Filter: ((remaining_allowance > '0'::numeric) AND (size < '10995116277760'::bigint))
        Buffers: shared hit=12
Planning:
  Buffers: shared hit=244
Planning Time: 0.522 ms
Execution Time: 0.061 ms
```

As you can see it's a better path and it's faster, but I was a little annoyed I couldn't force it to use that dedicated index. I ended up removing the index all together. How do you feel about that? Less indices less problems? I tried various things, like `INCLUDE` size and adding tie breakers, even disabled sorting in the hopes it would then be forced to select it.

persist/postgres/contracts.go
```diff
@@ -580,8 +590,7 @@ func (s *Store) MarkUnrenewableContractsBad(ctx context.Context, minProofHeight
 // pending and have a formation height older than 'maxFormation'.
 func (s *Store) RejectPendingContracts(ctx context.Context, maxFormation time.Time) error {
 	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
-		_, err := tx.Exec(ctx, `UPDATE contracts SET state = $1 WHERE state = $2 AND formation < $3`,
-			sqlContractState(contracts.ContractStateRejected), sqlContractState(contracts.ContractStatePending), maxFormation)
+		_, err := tx.Exec(ctx, `UPDATE contracts SET state = 4 WHERE state = 0 AND formation < $1`, maxFormation)
```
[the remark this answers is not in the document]
peterjan: I would generally agree but in this case I don't and prefer hardcoding the states in the query. `sqlContractState(contracts.ContractStateX),` bloats the query and the parameters make copy pasting annoying when copy pasting the query to run it manually against a database.

persist/postgres/migrations.go
```diff
@@ -310,4 +310,36 @@ CREATE INDEX contracts_capacity_size_contract_id_idx ON contracts (host_id, (cap
...
+			CREATE INDEX contracts_host_id_active_good_idx ON contracts(host_id) WHERE state IN (0,1) AND renewed_to IS NULL AND good;
+			CREATE INDEX contracts_host_id_active_bad_idx ON contracts(host_id) WHERE state IN (0,1) AND renewed_to IS NULL AND NOT good;
+			CREATE INDEX contracts_host_id_inactive_good_idx ON contracts (host_id) WHERE state IN (2,3,4) AND good;
+			CREATE INDEX contracts_host_id_inactive_bad_idx ON contracts (host_id) WHERE state IN (2,3,4) AND NOT good;
+			CREATE INDEX contracts_last_broadcast_attempt_active_idx ON contracts (last_broadcast_attempt ASC, contract_id) WHERE state IN (0,1) AND renewed_to IS NULL;
+			CREATE INDEX contracts_host_id_remaining_allowance_active_idx ON contracts (host_id, remaining_allowance DESC, contract_id) WHERE state IN (0,1) AND renewed_to IS NULL AND good AND remaining_allowance > 0;
+			CREATE INDEX contracts_size_contract_id_idx ON contracts (host_id, size DESC, contract_id) INCLUDE(next_prune) WHERE state IN (0,1) AND renewed_to IS NULL AND good AND remaining_allowance > 0;
+			CREATE INDEX contracts_formation_pending_idx ON contracts(formation) WHERE state = 0;
```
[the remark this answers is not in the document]
peterjan: Thank you for pointing that out. I already addressed it in my comment [here](https://github.com/SiaFoundation/indexd/pull/518#discussion_r2398253475)

persist/postgres/migrations.go
```diff
@@ -310,4 +310,36 @@ CREATE INDEX contracts_capacity_size_contract_id_idx ON contracts (host_id, (cap
...
 		return err
 	},
+	// recreate all contracts indices
+	func(ctx context.Context, tx *txn, l *zap.Logger) error {
+		if _, err := tx.Exec(ctx, `
+			DROP INDEX contracts_state_formation_idx;
+			DROP INDEX contracts_state_good_idx;
+			DROP INDEX contracts_last_broadcast_attempt_contract_id_idx;
+			DROP INDEX contracts_host_id_remaining_allowance_contract_id_idx;
+			DROP INDEX contracts_capacity_size_contract_id_idx;
```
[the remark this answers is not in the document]
peterjan: Already addressed [here](https://github.com/SiaFoundation/indexd/pull/518#discussion_r2398253475)

persist/postgres/contracts.go
```diff
@@ -38,8 +38,8 @@ func (s *Store) ContractsStats(ctx context.Context) (resp admin.ContractsStatsRe
 		var numContracts, numGood, totalCapacity, totalSize uint64
 		err := tx.QueryRow(ctx, `
 			WITH globals AS (
-    			SELECT scanned_height FROM global_settings
-       		)
+			SELECT scanned_height FROM global_settings
```
ChrisSchinnerl: The formatting broke here for the line that contains the `SELECT`.

persist/postgres/contracts.go
```diff
@@ -48,11 +48,10 @@ func (s *Store) ContractsStats(ctx context.Context) (resp admin.ContractsStatsRe
 			FROM contracts
 			CROSS JOIN globals
 			WHERE
-				proof_height > globals.scanned_height AND
+				state IN (0,1) AND
```
ChrisSchinnerl, line 51: Why hardcode the states? I did like the extra readability you got from looking at the vars.

persist/postgres/contracts_test.go
```diff
@@ -22,6 +22,15 @@ import (
 	"lukechampine.com/frand"
 )
 
+const (
+	defaultLimit = 100
+	maxLimit     = 500
+)
+
+var (
+	apiLimits = []int{defaultLimit, maxLimit, 2 * maxLimit}
```
ChrisSchinnerl: Should be in `BenchmarkContracts` since it's only used there.

persist/postgres/contracts.go
```diff
@@ -257,11 +255,11 @@ func (s *Store) ContractsForBroadcasting(ctx context.Context, minBroadcast time.
...
 	err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
 		rows, err := tx.Query(ctx, `
-SELECT c.contract_id
-FROM contracts c
-WHERE c.renewed_to IS NULL AND c.state <= $1 AND c.last_broadcast_attempt < $2
-ORDER BY c.last_broadcast_attempt ASC
-LIMIT $3`, sqlContractState(contracts.ContractStateActive), minBroadcast, limit)
+			SELECT c.contract_id
+			FROM contracts c
+			WHERE c.state IN (0, 1) AND c.renewed_to IS NULL AND c.last_broadcast_attempt < $1
```
ChrisSchinnerl, line 248: Is there a reason for always swapping the order of conditions to check the state first? Afaik Postgres shouldn't care since it analyzes the whole thing every time to figure out which condition to go for first.

peterjan: There is no reason other than consistency, no. In my head state has a higher weight, it's the same with ordering function parameters, I would also never put a contract id before a host id.

persist/postgres/contracts.go
peterjan: I answered that [here](https://github.com/SiaFoundation/indexd/pull/518#discussion_r2398265376). I would generally agree, but not in the case of contract state. It only adds bloat in my opinion and makes the query less readable over all. It's also annoying for copy-pasting, the fewer $ arguments the better. 

I don't mind reverting, I did carefully consider it before suggesting it because I was expecting the comment. I like it better 🤷‍♂️ we all know the contract states and if most of the time it's very clear from the context.

peterjan: Oh I have another reason, probably what triggered me is that because we move away from <= and > it would require us to pass in all the valid contract states. That is what made passing the states feel very tedious.

persist/postgres/init.sql
n8mgr: If it's not willing to use the index, no point in keeping it.

n8mgr:
@peterjan merge conflicts

n8mgr (dismissed):

PR #526, opened 2025-10-03
SDK: upload slab to hosts that aren't too close to each other

### LINKED ISSUE #474: SDK: Don't upload slabs to hosts that are too close to each other
Related to https://github.com/SiaFoundation/indexd/issues/473

peterjan:
This updates the dialer to return hosts that are at least 10km removed from one another first. Hosts that are in too close proximity of a selected host gets pushed to the end of the list. Couple of approaches here, in the future we probably want to be smarter about this. I kind of liked this minimum LoC approach that satisfies the issue imo, reusing the spaced set we also use in the contractor.

Fixes https://github.com/SiaFoundation/indexd/issues/474

### DIFF
--- a/hosts/host.go
+++ b/hosts/host.go
@@ -104,12 +104,11 @@ type (
 	// key and addresses. It is used for listing usable hosts in the
 	// application API.
 	HostInfo struct {
-		PublicKey types.PublicKey    `json:"publicKey"`
-		Addresses []chain.NetAddress `json:"addresses"`
-
-		CountryCode string  `json:"countryCode"`
-		Latitude    float64 `json:"latitude"`
-		Longitude   float64 `json:"longitude"`
+		PublicKey   types.PublicKey    `json:"publicKey"`
+		Addresses   []chain.NetAddress `json:"addresses"`
+		CountryCode string             `json:"countryCode"`
+		Latitude    float64            `json:"latitude"`
+		Longitude   float64            `json:"longitude"`
 	}
 
 	// Usability represents a series of host checks that can be used to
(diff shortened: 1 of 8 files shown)


sdk/dialer.go
```diff
@@ -216,18 +218,30 @@ func (d *dialer) Hosts() (hks []types.PublicKey) {
...
 	}
 
-	return
+	// enforce minimum distance between hosts
+	set := hosts.NewSpacedSet(d.minHostDistanceKm)
+	spaced := make([]types.PublicKey, 0, len(hks))
+	others := make([]types.PublicKey, 0, len(hks))
+	for _, hk := range hks {
+		if set.Add(d.hosts[hk]) {
+			spaced = append(spaced, hk)
```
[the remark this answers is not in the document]
peterjan: That's embarrassing

chris124567 (approved):

hosts/host.go
```diff
@@ -160,17 +159,28 @@ type (
 	}
 )
 
+// Info returns the HostInfo of the host.
+func (h *Host) Info() HostInfo {
```
n8mgr, line 163: Why not get rid of `Host` or `HostInfo`?

peterjan: I feel they both have their place? The (full) `Host` is being returned by the admin api and a subset of the information is contained in the `HostInfo` which is returned by the app API. I feel that getting rid of `HostInfo` would result in needlessly passing around a lot of useless information.

n8mgr: Yeah.. Not sure wha I was thinking here.

PR #527, opened 2025-10-03
Extend `Dialer` with QUIC support

### LINKED ISSUE #521: Extend `Dialer` with Quic support
Right now we have a potential edge case for hosts announcing themselves on QUIC but not Siamux. The scans will pass but we will always attempt to reach out via Siamux in `DialHost` (`client/dialer.go`).

One way to fix this would be to change the `DialHost` interface method to accept a `chain.NetAddress` and then have the dialer dial using either the Siamux protocol, quic or return an error if the protocol is unknown.

We will also need to update `Host.SiamuxAddr()` (`hosts/manager.go`) to `RHP4Addrs` or so and return all of the host's addresses. The dialer can then dial them one-by-one until one of them succeeds. Prioritizing the siamux protocol.

chris124567:
Close #521

### DIFF
--- a/client/dialer.go
+++ b/client/dialer.go
@@ -3,6 +3,7 @@ package client
 import (
 	"context"
 	"errors"
+	"fmt"
 	"time"
 
 	"go.sia.tech/core/types"
@@ -15,7 +16,7 @@ import (
 
 const dialTimeout = 10 * time.Second
 
-// Dialer can be used to dial a host using the SiaMux protocol.
+// Dialer can be used to dial a host using SiaMux or QUIC.
 type Dialer struct {
 	cm                       ChainManager
 	revisionSubmissionBuffer uint64
@@ -61,6 +62,7 @@ func (d *Dialer) DialHost(ctx context.Context, hk types.PublicKey, addrs []chain
 	ctx, cancel := context.WithTimeout(ctx, dialTimeout)
 	defer cancel()
 
+	var lastErr error
 	tryDial := func(proto chain.Protocol, dial func(addr string) (rhp.TransportClient, error)) rhp.TransportClient {
 		for _, addr := range addrs {
 			if addr.Protocol != proto {
@@ -74,6 +76,7 @@ func (d *Dialer) DialHost(ctx context.Context, hk types.PublicKey, addrs []chain
 					zap.String("addr", addr.Address),
 					zap.Error(err),
 				)
+				lastErr = err
 				continue
 			}
 			return tc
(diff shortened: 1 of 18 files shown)


client/dialer.go
```diff
@@ -2,44 +2,46 @@ package client
...
+	"go.sia.tech/coreutils/rhp/v4/quic"
 	"go.sia.tech/coreutils/rhp/v4/siamux"
 	"go.uber.org/zap"
 )
 
 const dialTimeout = 10 * time.Second
 
-// SiamuxDialer can be used to dial a host using the SiaMux protocol.
-type SiamuxDialer struct {
+// Dialer can be used to dial a host using the SiaMux protocol.
```
peterjan: `nit` update docstring

client/dialer.go
```diff
@@ -53,15 +55,43 @@ func NewSiamuxDialer(cm ChainManager, signer rhp.FormContractSigner, store Revis
...
+// with the host. It tries SiaMux first, then QUIC, and returns a host client
+// that exposes the RPC methods defined in the RHP.
+func (d *Dialer) DialHost(ctx context.Context, hk types.PublicKey, addrs []chain.NetAddress) (*HostClient, error) {
 	ctx, cancel := context.WithTimeout(ctx, dialTimeout)
 	defer cancel()
 
-	tc, err := siamux.Dial(ctx, addr, hk)
-	if err != nil {
-		return nil, fmt.Errorf("failed to dial host: %w", err)
+	tryDial := func(proto chain.Protocol, dial func(addr string) (rhp.TransportClient, error)) rhp.TransportClient {
```
peterjan, line 66: Too bad the signatures of `siamux.Dial` and `quic.Dial` aren't identical so we could easily pass the function around. Another approach would've been to stable sort the addresses using a priority map (`map[chain.Protocol]int`). I kind of like that because it's immediately obvious what we're doing, comes with issues of its own though. 

Decided to leave a comment because perhaps we want to consider:
- handling the context differently? 
- improving the error we return

I think `errors.New("host has no valid addresses")` is really not all that useful. We could at least return number of dial attempts. Or collect the errors or log the last error for context. 

```go
if len(addrs) == 0 {
	return nil, errors.New("no addresses provided")
}
if lastErr != nil {
	return nil, fmt.Errorf("all dials failed: %w", lastErr)
}
return nil, errors.New("no supported protocols in address list")
```

I think sanitizing `addrs` and return a useful error is probably worth adding.

peterjan (changes requested):

peterjan (approved):

n8mgr (approved):

PR #528, opened 2025-10-06
broadcast formations, refreshes, and renewals

n8mgr:


contracts/form.go
```diff
@@ -85,8 +85,7 @@ func (s *formContractSigner) SignV2Inputs(txn *types.V2Transaction, toSign []int
 // performContractFormation makes sure that we have at least 'wanted' good
 // contracts with good hosts that are sufficiently spaced apart.
 func (cm *ContractManager) performContractFormation(ctx context.Context, period uint64, wanted int64, log *zap.Logger) error {
-	formationLog := log.Named("formation")
-	formationLog.Debug("started", zap.Uint64("period", period), zap.Int64("wanted", wanted))
+	log.Debug("started", zap.Uint64("period", period), zap.Int64("wanted", wanted))
```
peterjan, line 88: What's the reason for removing the named loggers? I know Chris likes them a lot. I personally don't really use them that often to work my way through larger log files.

contracts/renew.go
```diff
@@ -74,29 +63,45 @@ func (cm *ContractManager) renewContract(ctx context.Context, contract Contract,
...
 		}
 
-		contractLog.Info("successfully renewed contract",
+		log.Info("successfully renewed contract",
 			zap.Stringer("computedAllowance", allowance),
 			zap.Stringer("computedCollateral", collateral),
 			zap.Stringer("newRemainingAllowance", renewed.Revision.RemainingAllowance()),
-			zap.Stringer("newRemainingCollateral", renewed.Revision.RemainingCollateral()),
-			zap.Stringer("newContractID", renewed.ID))
+			zap.Stringer("newRemainingCollateral", renewed.Revision.RemainingCollateral()))
```
peterjan, line 104: Why did we no longer log the ID of the renewal?

peterjan (approved):

contracts/form.go
n8mgr: Having multiple named logs in scope with various fields increases the likelihood of picking the wrong one. In formation, there are at least 3 loggers in scope at various points. Which do you use: `log`, `formationLog`, `contractLog`, `hostLog`? There are even a few instances where `formationLog` was used where `hostLog` was probably the correct choice. This gets rid of that possibility. 

There are valid reasons to have multiple named logs in the same scope, but in this case the desire is to always use the most specific one.

contracts/renew.go
n8mgr: It's added to the log fields on line 84. No reason to duplicate it here.

Issue #522, opened 2025-10-03
Additional filtering on hosts and contracts endpoints


ChrisSchinnerl:
I am quite reluctant to add as many filtering options as possible to our endpoints and UI but the following ones should be implemented since they would've made some debugging so far a lot more enjoyable.

- [x] Allow for filtering hosts using a list of public keys
    - [x] Backend + docs
    - [ ] UI 
- [x] Allow for filtering contracts using a list of public keys
    - [x] Backend + docs
    - [ ] UI 
- [x] Allow for filtering contracts using a list of contract ids
    - [x] Backend + docs
    - [ ] UI

chris124567:
Unassigning myself because the backend part is done

n8mgr:
Closing in favor of the UI specific issue #570

PR #530, opened 2025-10-07
Filter hosts by pubkeys

chris124567:
Re: #522

Add options to filter hosts by host public keys to the store and the admin API.  We add a "pk" query parameter containing a string encoded public key which can be repeated multiple times if there are multiple pubkeys.

### DIFF
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -666,7 +666,7 @@ func (a *admin) handleGETHosts(jc jape.Context) {
 		}
 		opts = append(opts, hosts.WithActiveContracts(activeContracts))
 	}
-	if strs := jc.Request.Form["pk"]; len(strs) > 0 {
+	if strs := jc.Request.Form["hostkey"]; len(strs) > 0 {
 		hks := make([]types.PublicKey, len(strs))
 		for i := range strs {
 			var hk types.PublicKey
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/hex"
 	"net/http"
+	"strings"
 
 	"errors"
 	"fmt"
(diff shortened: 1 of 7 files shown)


api/admin/admin.go
```diff
@@ -666,6 +667,24 @@ func (a *admin) handleGETHosts(jc jape.Context) {
 		}
 		opts = append(opts, hosts.WithActiveContracts(activeContracts))
 	}
+	if jc.Request.FormValue("publickeys") != "" {
+		var s string
+		if jc.DecodeForm("publickeys", &s) != nil {
```
n8mgr: URL query can be a list of values already. Prefer to use standards over custom parsing.

`pk=foo&pk=bar` is perfectly valid and should return `["foo", "bar"]` when using url.Values or parsing in just about any language.

api/admin/admin.go
```diff
@@ -666,6 +666,18 @@ func (a *admin) handleGETHosts(jc jape.Context) {
 		}
 		opts = append(opts, hosts.WithActiveContracts(activeContracts))
 	}
+	if strs := jc.Request.Form["pk"]; len(strs) > 0 {
```
peterjan: since this is user facing I would call this `hostkey` instead of `pk`

openapi/admin.yml
```diff
@@ -737,6 +737,12 @@ paths:
           required: false
           schema:
             type: boolean
+        - name: publickeys
+          in: query
+          description: Only include hosts that have public keys within this comma separated set of public keys
```
peterjan: outdated description

api/admin/options.go
```diff
@@ -83,6 +84,17 @@ func WithActiveContracts(activeContracts bool) HostQueryParameterOption {
 	}
 }
 
+// WithPublicKeys sets the 'publickeys' parameter.
```
peterjan: outdated docstring (?)

peterjan: I think this is the OpenAPI way of defining this convention (explode: true):

```
parameters:
  - name: foo
    in: query
    schema:
      type: array
      items: { type: string }
    style: form
    explode: true
```

peterjan (changes requested):

peterjan (approved):

n8mgr (approved):

PR #536, opened 2025-10-09
Use mock locator in testing

peterjan:
We can't run our e2e tests locally because of the geo database being missing. I think we recently changed that to use git attributes but that probably broke the tests locally. It's fine to not have locations in testing because the local IPv6 address all our hosts use map to a NULL location anyway.

```
=== RUN   TestMigrations
    indexer.go:132: error opening database: invalid MaxMind DB file
```

### DIFF
--- a/internal/testutils/indexer.go
+++ b/internal/testutils/indexer.go
@@ -75,6 +75,16 @@ type (
 	}
 )
 
+type mockLocator struct{}
+
+// NewMockLocator returns a new instance of a mock Locator.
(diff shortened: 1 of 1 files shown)


chris124567 (changes requested):
This is fine but I can run the tests locally without any problems.  What happens when you run `git lfs checkout`?

n8mgr:
Closing as tests run fine locally after grabbing the db.

PR #543, opened 2025-10-11
Smarter slab validation

n8mgr:
Changes the hardcoded redundancy of 3× to a probability of recovery. This rejects slabs with parameters that are more likely to result in data loss, even if they meet the 3× requirement (e.g., 1/3 or 2/6), and enables more exotic combinations, such as 30/60 for 2×.

### DIFF
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -138,8 +138,9 @@ func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) (Ob
 		opt(&uo)
 	}
 
-	if (uo.parityShards+uo.dataShards)/uo.dataShards < 2 {
-		return Object{}, errors.New("redundancy must be at least 2x")
+	totalShards := int(uo.dataShards) + int(uo.parityShards)
+	if err := slabs.ValidateECParams(int(uo.dataShards), totalShards); err != nil {
+		return Object{}, err
 	}
 
 	obj := Object{masterKey: frand.Bytes(32)}
--- a/slabs/slabs.go
+++ b/slabs/slabs.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"time"
 
 	proto "go.sia.tech/core/rhp/v4"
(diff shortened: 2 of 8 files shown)


slabs/slabs.go
```diff
@@ -180,3 +178,44 @@ func (m *SlabManager) SlabIDs(ctx context.Context, account proto.Account, offset
...
+// ValidateECParams checks the erasure coding parameters are
+// acceptable and ensure sufficient durability of the data. If they
+// are not, an error is returned.
+//
+// It does this by calculating the probability of being able to recover
+// the original data. The calculation assumes that each shard has a
+// fixed probability of being available, and uses the binomial
+// distribution to calculate the probability of having at least
+// `n` data shards available out of `m` total shards.
+func ValidateECParams(dataShards, totalShards int) error {
```
peterjan, line 191: This is a lot better than what we had, that's great. Did you consider moving it to the slab manager and make it configurable so we can lower the threshold in testing? There's little benefit to running all our e2e tests with these redundancy settings, and there's a considerable downside to needing at least 7 hosts - I believe in testing you even set it to 10.

sdk/client.go
```diff
@@ -138,8 +138,9 @@ func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) (Ob
 		opt(&uo)
 	}
 
-	if (uo.parityShards+uo.dataShards)/uo.dataShards < 2 {
-		return Object{}, errors.New("redundancy must be at least 2x")
```
peterjan, line 142: you missed the docstring in `WithRedundancy`

peterjan (approved):

slabs/slabs.go
ChrisSchinnerl: +1 on making `recoveryProbability` configurable. It would be nice to default to 100% in testing to make sure we can get away with fewer hosts to speed things up.

ChrisSchinnerl (approved):

slabs/slabs.go
n8mgr: I didn't feel the need to make it configurable. One of our goals with `indexd` was to limit configuration options and default to sane values. The test workflow time did not increase significantly with the increase in host count and we should probably be testing with values closer to prod in e2e tests anyway.

PR #544, opened 2025-10-11
Increase maintenance frequency

n8mgr:
Decreases the maintenance frequency from 10 minutes to one to hopefully reduce out of funds errors on the service accounts. Also switches the ticker for a timer. The ticker can fire immediately if maintenance takes longer than the frequency. This change means there is always the same amount of time between maintenances regardless of how long they run for. Also refactors the maintenance loop to run each task regardless of the success of the previous task.

### DIFF
--- a/contracts/maintenance.go
+++ b/contracts/maintenance.go
@@ -13,7 +13,6 @@ import (
 
 func (cm *ContractManager) performAccountFunding(ctx context.Context, force bool, log *zap.Logger) error {
 	start := time.Now()
-	log = log.Named("accounts")
 
 	// fund accounts on usable hosts with active contracts
 	opts := []hosts.HostQueryOpt{
@@ -69,8 +68,6 @@ func (cm *ContractManager) performAccountFunding(ctx context.Context, force bool
 }
 
 func (cm *ContractManager) performContractMaintenance(ctx context.Context, log *zap.Logger) error {
-	log.Debug("performing contract maintenance")
-
 	// fetch settings and determine if maintenance is supposed to run
 	settings, err := cm.store.MaintenanceSettings(ctx)
 	if err != nil {
@@ -119,9 +116,8 @@ func (cm *ContractManager) performContractMaintenance(ctx context.Context, log *
 // to perform on contracts
 func (cm *ContractManager) maintenanceLoop(ctx context.Context) {
 	log := cm.log.Named("maintenance")
-
-	ticker := time.NewTicker(cm.maintenanceFrequency)
-	defer ticker.Stop()
+	t := time.NewTimer(cm.maintenanceFrequency)
+	defer t.Stop()
 
 	for {
 		if !cm.waitUntilSynced(ctx, log) {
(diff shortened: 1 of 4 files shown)


chris124567 (approved):

contracts/maintenance.go
```diff
@@ -142,54 +138,33 @@ func (cm *ContractManager) maintenanceLoop(ctx context.Context) {
 			if err := cm.performContractPruning(ctx, true, log); err != nil {
 				log.Error("contract pruning failed", zap.Error(err))
 			}
+			continue
 		case <-cm.triggerMaintenanceChan:
-			// reset ticker
```
peterjan, line 146: we should probably reset the timer when maintenance is triggered

contracts/maintenance.go
```diff
@@ -142,54 +138,33 @@ func (cm *ContractManager) maintenanceLoop(ctx context.Context) {
...
-		}, log) {
-			return
-		}
-
-		if !performMaintenanceJob("sector pinning", func() error {
-			return cm.performSectorPinning(ctx, log)
-		}, log) {
-			return
-		}
+		contractMaintenanceLog := log.Named("contracts")
```
peterjan, line 148: `nit` inline the logs, no need for the extra variables

contracts/manager.go
```diff
@@ -428,7 +428,7 @@ func newContractManager(renterKey types.PublicKey, accounts AccountManager, chai
 		expiredContractBroadcastBuffer:    144,           // 144 block after expiration
 		expiredContractPruneBuffer:        144,           // 144 blocks after broadcast
 		expiredContractSectorsPruneBuffer: 36,            // 36 blocks (~6 hours) after expiration
-		maintenanceFrequency:              10 * time.Minute,
+		maintenanceFrequency:              time.Minute,
```
peterjan, line 431: Probably fine to set this to one minute but it doesn't sit right with me 100%. We have two types of maintenance jobs it feels like, the ones related to contract lifecycle management, and then the others that need a contract lock to perform things like account funding and sector pinning. Should we consider splitting those up, that would obviously increase complexity but it feels weird to form/renew/refresh/prune every minute. Even more so because for the other jobs it's probably fine to just continuously check if they need doing.

contracts/maintenance.go
ChrisSchinnerl: Inlining is probably pretty nasty but maybe we can add a `name` argument to `logError` and then get rid of the log variables.

contracts/maintenance.go
ChrisSchinnerl: Seems like it is

contracts/manager.go
ChrisSchinnerl: Could you elaborate on why you think it's "weird"? It is pretty nice to have that guarantee of maintenance being complete before starting any dependent work. I'd expect it to be a no-op most of the time.

ChrisSchinnerl (approved):

contracts/maintenance.go
peterjan: missed that

contracts/maintenance.go
peterjan: fine with resolving, I didn't see we pass it twice

contracts/manager.go
peterjan: That's true for the most part but I'm a little worried about contract formations. That maintenance is already quite verbose and running it every minute feels excessive. Checking whether contracts need to be renewed is indeed a no-op but that too feels excessive to check at a frequency higher than the block time.

I guess I wanted to point out that it's indeed a no-op in the happy case, but I'm not too sure it won't cause log spam and any failed refresh/renews will be retried every minute now, which might also cause unnecessary actions and/or log spam.

peterjan: I feel we set it to 10 minutes for a reason, and then now we lower it to 1 minute to ensure service accounts are filled. Should we consider alternative approaches that accomplish the same thing, e.g. triggering refreshes + account funding more frequently but keep doing the other maintenance jobs at a 10 minute interval.

peterjan (approved):

PR #551, opened 2025-10-14
Add sorting to `/hosts`

peterjan:
The UI is capable of supporting it if we extend the API with sort options. Pretty low effort and would be useful debugging the indexer. If this gets approved I'll add support for `/contracts` too.


persist/postgres/hosts.go
```diff
@@ -906,3 +912,38 @@ func (s *Store) HostsWithLostSectors(ctx context.Context) ([]types.PublicKey, er
...
+		"settings.prices.collateral":      "settings_collateral",
+		"settings.prices.storagePrice":    "settings_storage_price",
+		"settings.prices.ingressPrice":    "settings_ingress_price",
+		"settings.prices.egressPrice":     "settings_egress_price",
+		"settings.prices.freeSectorPrice": "settings_free_sector_price",
+	}
+
+	parts := make([]string, 0, len(sorts))
+	for _, sort := range sorts {
+		// validate column, note: we don't validate direction here, the API
```
n8mgr: Absolutely should validate here too to guarantee safe handling of inputs. Consider making `sortdir` a bool to avoid validating at all, i.e. descending ? "DESC" : "ASC".

persist/postgres/hosts.go
```diff
@@ -202,9 +210,7 @@ WHERE
 	-- active contracts filter
 	AND (($5::boolean IS NULL) OR ($5::boolean = EXISTS (SELECT 1 FROM contracts WHERE host_id = hosts.id AND state >= $6 AND state <= $7)))
 	-- public key filter
-	AND ((CARDINALITY($8::bytea[]) = 0) OR (public_key = ANY($8)))
-	LIMIT $1 OFFSET $2
-;`, limit, offset, opts.Good, opts.Blocked, opts.ActiveContracts, contracts.ContractStatePending, contracts.ContractStateActive, hks)
+	AND ((CARDINALITY($8::bytea[]) = 0) OR (public_key = ANY($8)))`+"\n"+orderClause+"\n"+`LIMIT $1 OFFSET $2`, limit, offset, opts.Good, opts.Blocked, opts.ActiveContracts, contracts.ContractStatePending, contracts.ContractStateActive, hks)
```
ChrisSchinnerl: It's probably nicer to wrap the query in a `Sprintf`. Mixing two types of strings like this and having to squeeze in `\n` harms readability quite a bit.

api/options.go
```diff
@@ -15,12 +16,27 @@ const (
 	maxLimit     = 500
 )
 
+const (
+	// SortDirectionAsc represents ascending ordering.
+	SortDirectionAsc = "ASC"
```
n8mgr: These do not need to be exported

api/options.go
```diff
@@ -85,3 +101,39 @@ func ParseOffsetLimit(jc jape.Context) (offset int, limit int, ok bool) {
...
+}
+
+// ParseSortOptions parses 'sortby' and 'sortdir' query parameters from the
+// request context. It returns the parsed sort options and a boolean indicating
+// whether parsing succeeded. If invalid parameters are provided an error is
+// written to the response and false is returned.
+func ParseSortOptions(jc jape.Context) (sorts []SortOption, ok bool) {
+	sortBy := jc.Request.Form["sortby"]
+	sortDir := jc.Request.Form["sortdir"]
+	if len(sortBy)+len(sortDir) == 0 {
```
n8mgr: Why not make this match the book instead of parsing a string?

peterjan: Didn't consider it. Updated it though, thanks.

persist/postgres/hosts.go
```diff
@@ -202,9 +210,7 @@ WHERE
 	-- active contracts filter
 	AND (($5::boolean IS NULL) OR ($5::boolean = EXISTS (SELECT 1 FROM contracts WHERE host_id = hosts.id AND state >= $6 AND state <= $7)))
 	-- public key filter
-	AND ((CARDINALITY($8::bytea[]) = 0) OR (public_key = ANY($8)))
-	LIMIT $1 OFFSET $2
-;`, limit, offset, opts.Good, opts.Blocked, opts.ActiveContracts, contracts.ContractStatePending, contracts.ContractStateActive, hks)
+	AND ((CARDINALITY($8::bytea[]) = 0) OR (public_key = ANY($8))) %s LIMIT $1 OFFSET $2`, orderClause), limit, offset, opts.Good, opts.Blocked, opts.ActiveContracts, contracts.ContractStatePending, contracts.ContractStateActive, hks)
```
ChrisSchinnerl: Does the following work?
```
AND ((CARDINALITY($8::bytea[]) = 0) OR (public_key = ANY($8)))
%s -- orderClause
LIMIT $1 OFFSET $2`
```

ChrisSchinnerl (dismissed):

ChrisSchinnerl (dismissed):

peterjan:
@ChrisSchinnerl if this gets merged should I go ahead and copy paste it over to `/contracts` too? It's not tracked on the board but we'll need it eventually anyway.

chris124567 (dismissed):

ChrisSchinnerl (approved):

PR #555, opened 2025-10-15
Fix TestMigrations NDF

### LINKED ISSUE #554: Fix NDF in `TestMigrations`
=== RUN   TestMigrations
      e2e_test.go:90: expected sector f353b4ef346049435deafd4c1e27a0b024f91f7d62f27a34d56bead4c518c922 on host ed25519:7062912231070ad067fad4f98227a34c327298dbeb1edf190b0616d73d1e43ff, got f353b4ef346049435deafd4c1e27a0b024f91f7d62f27a34d56bead4c518c922 on host ed25519:067b50bcde2e632114add2b24a5c104735dbfb6d022b0c51050ec36559cbc0d0

peterjan:
I found out that when we block a host that has unpinned sectors, those sectors end up in a somewhat awkward unpinnable state. They get out of this state though, but only after a certain period of time (the unpinnable sector threshold). For a slab to be unhealthy, its sectors have to be unpinnable, or linked to a contract that's inactive or bad.

I figured it's probably a good idea to mark the sectors as unpinnable in `BlockHosts`... I was on the fence about that one though, because it requires updating the sector stats. The alternative is to update the health query, or live with the fact that it takes some time to pick up on this issue.

Other findings:

We should open an issue to look into our hosts methods to see if we treat the blocklist correctly. For instance, I think `HostsWithUnpinnableSectors` should either not return blocked hosts, or we should update the call site. In the contract manager we don't use the `host.IsGood` but only check usability... decided to fix that separately though, we have to go through all of them anyway. 

Fixes https://github.com/SiaFoundation/indexd/issues/554 (passed 100 times on CI)

### DIFF
--- a/persist/postgres/hosts.go
+++ b/persist/postgres/hosts.go
@@ -280,16 +280,32 @@ func (s *Store) BlockHosts(ctx context.Context, hks []types.PublicKey, reason st
 			if err != nil {
 				return fmt.Errorf("failed to add host %q to blocklist: %w", hk, err)
 			}
-			_, err = tx.Exec(ctx, `
-				UPDATE contracts
-				SET good = FALSE
-				FROM contracts c
-				INNER JOIN hosts h ON h.id = c.host_id
-				INNER JOIN hosts_blocklist hb ON hb.public_key = h.public_key
-				WHERE contracts.id = c.id AND h.public_key = $1
-				`, sqlPublicKey(hk))
+
+			var hostID int64
+			err = tx.QueryRow(ctx, "SELECT id FROM hosts WHERE public_key = $1", sqlPublicKey(hk)).Scan(&hostID)
+			if errors.Is(err, sql.ErrNoRows) {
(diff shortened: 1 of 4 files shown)


n8mgr (approved):

ChrisSchinnerl (approved):

peterjan:
I verified sector stats are exact now. If that changes we'll know where to look.

| total_sectors | total_checksum | db_pinned | stats_pinned | delta_pinned | db_unpinned | stats_unpinned | delta_unpinned | db_unpinnable | stats_unpinnable | delta_unpinnable |
|----------------|----------------|------------|---------------|---------------|---------------|----------------|----------------|-----------------|------------------|------------------|
| 18505454 | 18505454 | 16140993 | 16140993 | 0 | 74916 | 74916 | 0 | 2289545 | 2289545 | 0 |

[data-1760690128597.csv](https://github.com/user-attachments/files/22966765/data-1760690128597.csv)

PR #546, opened 2025-10-14
Rremove mock store from tests in `pins`

chris124567:
Part of #350

PR #560, opened 2025-10-15
Remove mock store from `accounts` tests

chris124567:
#546


accounts/funder_test.go
```diff
@@ -1,4 +1,4 @@
-package accounts
+package accounts_test
```
ChrisSchinnerl: We don't need to change the package name. These are all unit tests anyway. 
Then you can also unexport `Dialer` again.

ChrisSchinnerl (changes requested):

accounts/funder_test.go
chris124567: This caused circular import errors because accounts imports postgres which imports accounts.

peterjan: That's only true for `manager_test.go` - for this file you can indeed use `package accounts` and unexport the `Dialer`

accounts/manager_test.go
```diff
@@ -37,262 +42,119 @@ var (
...
-
-func (s *mockStore) ActiveAccounts(ctx context.Context, threshold time.Time) (uint64, error) {
-	return uint64(len(s.accounts)), nil
+	return len(accounts), 1, nil
 }
 
-func (s *mockStore) HasAccount(ctx context.Context, pk types.PublicKey) (bool, error) {
-	_, ok := s.accounts[pk]
-	return ok, nil
+type testStore struct {
```
peterjan, line 67: This already uses the `Store` but it's a little unfortunate it keeps track of `eas` separately. That feels like it defeats its purpose almost. I think methods like `addTestX` are not an issue but I feel slightly uneasy about`resetNextFund` for instance.

This is just a suggestion but what if we added a `testutil` in our `internals` package and did something like:

```go
type TestStore struct {
	pool *pgxpool.Pool
	*postgres.Store
}

func NewTestStore(t testing.TB, maintenanceSettings contracts.MaintenanceSettings, log *zap.Logger) TestStore {
	// parse connection info from env vars
    ...

	// create database
	dbName := t.Name()
	pool, err := pgxpool.New(context.Background(), ci.String())
	...
	ci.Database = dbName

	// create store
	store, err := postgres.NewStore(context.Background(), ci, maintenanceSettings, hosts.DefaultUsabilitySettings, log)
	if err != nil {
		t.Fatalf("failed to create store: %v", err)
	}

	return TestStore{
		pool:  pool,
		Store: store,
	}
}

func (ts TestStore) Foo(t testing.TB) []any {
	rows, err := ts.pool.Query(t.Context(), `SELECT * FROM bar;`)
	if err != nil {
		t.Fatal(err)
	}
	defer rows.Close()

	// process rows...
	return nil
}
```

That way it can be reused across packages, we can access the internal database if need be and expose methods otherwise not exposed on the `Store`. Other than that though it's a normal Store and we're not mocking any of its methods.

peterjan (approved):

ChrisSchinnerl (approved):

PR #562, opened 2025-10-16
Refill maintenance accounts

### LINKED ISSUE #541: Service Account Funding
Service accounts are spending too quickly and constantly fail with insufficient balance errors:

```
2025-10-11T18:06:15Z	DEBUG	slabs.integrity	integrity checks got interrupted	{"hostKey": "ed25519:c8100e62d9bf5de75ce874ef9b23e23da5afbab1d988e25dd7c528b09f1522fb", "error": "insufficient service account balance"}
```

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/indexd/issues/541

This PR raises the maintenance frequency slightly again due to https://github.com/SiaFoundation/indexd/pull/559 but introduces triggering account refills for service accounts instead.

While the usual funding frequency of 2SC per hour should be enough for uploads and downloads, integrity checks are a lot more likely to run out of funds. That's because verifying a sector costs as much as downloading one while at the same time being a much faster process. So while downloads are limited to approximately 1-10Gbps for very fast hosts, integrity checks might involve costs orders of magnitude higher than that.

This PR attempts to fix this by triggering a refill of the account on demand whenever the integrity check code thinks we are running low on funds. (IMPORTANT: Not when the host tells us we are out of funds).

Of course an alternative approach would be to pour more money into service accounts. However, I'm not very comfortable with potentially keeping hundreds or thousands of SC in an account.


persist/postgres/accounts.go
ChrisSchinnerl, line 1: Blocker since this is based on https://github.com/SiaFoundation/indexd/pull/560

persist/postgres/accounts.go
```diff
@@ -197,6 +197,20 @@ func (s *Store) ScheduleAccountsForFunding(ctx context.Context, hostKey types.Pu
 	})
 }
 
+// ScheduleAccountForFunding marks the given account for the given host key as
+// due for funding.
+func (s *Store) ScheduleAccountForFunding(ctx context.Context, hostKey types.PublicKey, account proto.Account) error {
+	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `
+			UPDATE account_hosts
+			SET next_fund = NOW()
```
peterjan: since we order on `next_fund` it might be worthwhile to keep `next_fund` to its value if it's before `NOW()`

persist/postgres/accounts.go
```diff
@@ -197,6 +197,20 @@ func (s *Store) ScheduleAccountsForFunding(ctx context.Context, hostKey types.Pu
 	})
 }
 
+// ScheduleAccountForFunding marks the given account for the given host key as
+// due for funding.
+func (s *Store) ScheduleAccountForFunding(ctx context.Context, hostKey types.PublicKey, account proto.Account) error {
```
peterjan, line 202: little benchmark? especially since we might throw in a compare on `next_fund` - fine if you don't feel it's worth adding one but it would be a shame for our coverage stats

peterjan (changes requested):

persist/postgres/accounts.go
ChrisSchinnerl: Good point. I set it to the unix epoch instead to make sure it always ends up in the next batch of accounts we attempt to fund. I don't expect funding to lag behind but the 1 service account we use this on should probably get preferential treatment.

persist/postgres/accounts.go
ChrisSchinnerl: I think this is simple enough of a query to be fine without one. Considering that we only update one row.

persist/postgres/accounts.go
peterjan: Kudos from the boss, nice.

peterjan (approved):

persist/postgres/accounts.go
chris124567: can resolve now

n8mgr (approved):

PR #563, opened 2025-10-16
Update `store.ContractElementsForBroadcast `

### LINKED ISSUE #548: Fix `file contract resolution parent has invalid Merkle Proof`
`Indexd` periodically prints a lot of errors like the following:
```
indexd-1  | 2025-10-14T13:28:10Z	ERROR	contracts	failed to retrieve txn set for broadcasting	{"contractID": "257494d9aa501d62d95d736141e0ecb4e385e7ade61f9fc077ec168206440f7f", "expirationHeight": 545177, "error": "failed to update transaction set basis: transaction 1c23f340e05c4113988a9ab9c9795fa1b06165fbf527411126f875c844b5139f is invalid: file contract resolution parent has invalid Merkle proof"}
```

A look at siascan reveals that the contract has been resolved already. In the database the state of the contract is also `2` which means `resolved`. The problem is that `ContractElementsForBroadcast` doesn't take into account the state of the contract.

peterjan:
This ensures we only try and broadcast elements for contracts that are active. I also added a benchmark, it's reasonably fast (26.3 ms/op) considering there's 100k contracts and ~5k elements are returned on average. There's not much more to optimise except for maybe a covering index on `contract_elements` but it didn't improve the benchmark and that index would be quite big.

Fixes https://github.com/SiaFoundation/indexd/issues/548

### DIFF
--- a/persist/postgres/contracts.go
+++ b/persist/postgres/contracts.go
@@ -407,20 +407,24 @@ func (s *Store) ContractElementsForBroadcast(ctx context.Context, maxBlocksSince
 	err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
 		rows, err := tx.Query(ctx, `
 WITH current_height AS (
-    SELECT scanned_height FROM global_settings
+	SELECT scanned_height FROM global_settings
 )
 SELECT
-    fces.contract_id,
-    fces.contract,
-    fces.leaf_index,
-    fces.merkle_proof
-FROM contract_elements fces
-INNER JOIN contracts c ON fces.contract_id = c.contract_id
+	fces.contract_id,
+	fces.contract,
+	fces.leaf_index,
+	fces.merkle_proof
+FROM contracts c
+INNER JOIN contract_elements fces ON fces.contract_id = c.contract_id
 CROSS JOIN current_height
-WHERE current_height.scanned_height >= c.expiration_height + $1;`, maxBlocksSinceExpiry)
+WHERE current_height.scanned_height >= c.expiration_height + $1
+	AND c.state = 1
+	AND c.renewed_to IS NULL;`, maxBlocksSinceExpiry)
 		if err != nil {
 			return err
 		}
+		defer rows.Close()
+
 		for rows.Next() {
 			fce, err := scanContractElement(rows)
 			if err != nil {
--- a/persist/postgres/contracts_test.go
+++ b/persist/postgres/contracts_test.go
@@ -1717,10 +1717,10 @@ func BenchmarkContracts(b *testing.B) {
 		}
 
 		for b.Loop() {
-			eles, err := store.ContractElementsForBroadcast(context.Background(), frand.Uint64n(10))
+			fces, err := store.ContractElementsForBroadcast(context.Background(), frand.Uint64n(10))
 			if err != nil {
 				b.Fatal(err)
-			} else if len(eles) == 0 {
+			} else if len(fces) == 0 {
 				b.Fatal("expected at least one element")
 			}
 		}
--- a/persist/postgres/contracts_test.go
+++ b/persist/postgres/contracts_test.go
@@ -223,7 +223,10 @@ func TestContractElementsForBroadcast(t *testing.T) {
 
 	// add contract element, 1 contract to broadcast
 	err = store.UpdateChainState(context.Background(), func(tx subscriber.UpdateTx) error {
-		return tx.UpdateContractElements(fce)
+		return errors.Join(
+			tx.UpdateContractElements(fce),
+			tx.UpdateContractState(fcid, contracts.ContractStateActive),
+		)
 	})
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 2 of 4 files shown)


chris124567 (dismissed):

persist/postgres/contracts.go
```diff
@@ -406,21 +406,21 @@ func (s *Store) ContractElementsForBroadcast(ctx context.Context, maxBlocksSince
...
-WHERE current_height.scanned_height >= c.expiration_height + $1;`, maxBlocksSinceExpiry)
+	fces.contract_id,
+	fces.contract,
+	fces.leaf_index,
+	fces.merkle_proof
+FROM contracts c
+JOIN contract_elements fces ON fces.contract_id = c.contract_id
+WHERE c.state = 1
+	AND c.renewed_to IS NULL
+	AND c.expiration_height <= (SELECT scanned_height FROM global_settings) - $1;`, maxBlocksSinceExpiry)
```
ChrisSchinnerl: Keep the original comparison order here

persist/postgres/contracts.go
```diff
@@ -406,21 +406,21 @@ func (s *Store) ContractElementsForBroadcast(ctx context.Context, maxBlocksSince
 	var fces []types.V2FileContractElement
 	err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
 		rows, err := tx.Query(ctx, `
-WITH current_height AS (
```
ChrisSchinnerl, line 409: Does this make it faster? Because changing this makes it inconsistent with our other queries.

persist/postgres/contracts_test.go
```diff
@@ -233,14 +236,24 @@ func TestContractElementsForBroadcast(t *testing.T) {
...
+	fces, err := store.ContractElementsForBroadcast(context.Background(), 10)
 	if err != nil {
 		t.Fatal(err)
 	} else if len(fces) != 1 {
 		t.Fatalf("expected 1 contract to broadcast, got %d", len(fces))
 	} else if !reflect.DeepEqual(fces[0], fce) {
 		t.Fatalf("mismatch: \n%+v\n%+v", fce, fces[0])
 	}
+
+	// assert contract state are taken into account
```
ChrisSchinnerl, line 248: Might as well test that `renewed_to` is taken into account too.

persist/postgres/contracts_test.go
```diff
@@ -1683,6 +1697,21 @@ func BenchmarkContracts(b *testing.B) {
 		})
 	}
 
+	b.Run(fmt.Sprintf("contracts_elements_for_broadcast"), func(b *testing.B) {
```
ChrisSchinnerl:
```suggestion
	b.Run("contract_elements_for_broadcast", func(b *testing.B) {
```

persist/postgres/contracts_test.go
```diff
@@ -1683,6 +1697,21 @@ func BenchmarkContracts(b *testing.B) {
...
+		if err := store.UpdateChainState(b.Context(), func(tx subscriber.UpdateTx) error {
+			return tx.UpdateLastScannedIndex(b.Context(), types.ChainIndex{Height: 700})
+		}); err != nil {
+			b.Fatal(err)
+		}
+
+		for b.Loop() {
+			_, err := store.ContractElementsForBroadcast(context.Background(), frand.Uint64n(10))
```
ChrisSchinnerl: This benchmark is getting a bit out of hand with all the sub benchmarks. The more we add here the less specific the benchmarks become and as a result less meaningful.

e.g. the test contracts that are added have an expiration height of 800.
You set the last scanned height to 700.

So none of the contracts should be expired due to reaching the expiration height right?
I'd double check that this returns >0.

ChrisSchinnerl (changes requested):

persist/postgres/contracts_test.go
peterjan: I did check that, but I'll add a check in the benchmark as well.

peterjan: The contracts that are added have an expiration height between 601 and 800. I set the scanned height to 700 and then just randomly change the `maxBlocksSinceExpiry` for some randomness.

peterjan: I somewhat disagree with the statement about it being less meaningful. The setup in a benchmark should generally try and match a (very large) production node as close as possible. I don't see any benefit in pulling these apart? 

Not really that big of a deal but running all of these together could also highlight regression in other benchmarks. E.g. adding an index that is then favoured by another query, making the other method less performant. 

I don't really mind though, if you want to repeat the setup and separate all of these benchmarks then I'll go ahead and do it.

persist/postgres/contracts.go
peterjan: It's a little slower after the rewrite, it's not really meaningful though. It went from ~25ms/op to 30ms/op

ChrisSchinnerl (dismissed):

ChrisSchinnerl (approved):

chris124567 (approved):

PR #564, opened 2025-10-17
Add missing client.Close calls

ChrisSchinnerl:


peterjan (dismissed):
We're missing some calls to `Close` in our test suite as well (`indexer.HostClient`).

PR #567, opened 2025-10-19
Only form contracts with hosts if they are 20% below our price gouging settings

### LINKED ISSUE #557: Only form contracts with hosts if they are 20% below our price gouging settings

chris124567:
Close #557

### DIFF
--- a/contracts/form.go
+++ b/contracts/form.go
@@ -32,10 +32,6 @@ const (
 	// refreshes due to how long it would take to reasonably upload
 	// that amount of data with a 10 Gbps connection.
 	maxContractGrowthRate = 256 << 30
-
-	// usabilityPriceLimit defines the maximum acceptable host prices as 80% of
-	// the price gouging maximum.
-	usabilityPriceLimit = 0.8
 )
 
 var (
@@ -86,9 +82,9 @@ func (s *formContractSigner) SignV2Inputs(txn *types.V2Transaction, toSign []int
 	s.w.SignV2Inputs(txn, toSign)
 }
 
-func mulUsabilityLimit(c types.Currency) types.Currency {
-	num, den := uint64(usabilityPriceLimit*10000), uint64(10000)
-	return c.Mul64(num).Div64(den)
+func withinGougingLeeway(cost, limit types.Currency) bool {
+	// cost must be 20% lower than the limit
+	return cost.Mul64(10).Cmp(limit.Mul64(8)) > 0
 }
 
 // performContractFormation makes sure that we have at least 'wanted' good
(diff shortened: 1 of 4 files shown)


contracts/form.go
```diff
@@ -32,6 +32,10 @@ const (
 	// refreshes due to how long it would take to reasonably upload
 	// that amount of data with a 10 Gbps connection.
 	maxContractGrowthRate = 256 << 30
+
+	// usabilityPriceLimit defines the maximum acceptable host prices as 80% of
+	// the price gouging maximum.
+	usabilityPriceLimit = 0.8
```
peterjan: @ChrisSchinnerl since we decided to perform host usability checks in the database, I think we should be consistent here and take this gouging leeway into account there as well as opposed to in code where we can forget to apply it and have to perform a set of extra checks. `host.Usability` should be our SSOT

contracts/form.go
```diff
@@ -82,6 +86,11 @@ func (s *formContractSigner) SignV2Inputs(txn *types.V2Transaction, toSign []int
 	s.w.SignV2Inputs(txn, toSign)
 }
 
+func mulUsabilityLimit(c types.Currency) types.Currency {
```
ChrisSchinnerl: If you turn this into a `isWithinGougingLeeway(cost, limit types.Currency) bool`, you can get around the precision issues by only multiplying. 

So instead of
`cost > limit * 0.8`
do
`10*cost > 8*limit`

Maybe even turn it into a method on the host to avoid passing the prices separately.

contracts/form.go
```diff
@@ -176,6 +190,17 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
 			log.Debug("host is not usable since host has less than 10GiB of storage left", zap.Uint64("remainingStorage", host.Settings.RemainingStorage))
 			return false
 		}
+
+		if host.Settings.Prices.StoragePrice.Cmp(mulUsabilityLimit(usabilitySettings.MaxStoragePrice)) > 0 {
```
ChrisSchinnerl: These checks should also be ignored if `force == true`. Once we use a host we don't care about the leeway anymore.

contracts/form.go
ChrisSchinnerl: Since this is not a gouging check itself I think it's fine for now. Even if we ended up adding it to the database I probably wouldn't add it to `host.Usability` since it doesn't affect the overall usability of a host.

ChrisSchinnerl (changes requested):

contracts/form.go
peterjan: Hm ok but it should be called something like `gougingLeeway`, be defined on the `ContractManager` and probably we want to only log when a host is not usable IFF it's otherwise usable but it doesn't pass this 20% leeway

ChrisSchinnerl (approved):

peterjan (approved):

Issue #537, opened 2025-10-10
Debug migration cause


peterjan:
Figure out why so many slabs are migrating

peterjan:
Migrations seem to be  working fine. 
Slabs are healthy again after we migrate them.

Some slab numbers:
- 0.6M unhealthy slabs
- 3.2M sectors that need to be migrated (13TB)
- that'll take ~7 weeks to migrate (not counting slabs being pinned and needing migrations in the mean time)

Some host numbers:
- 62 hosts are blocked
- 13 of those are usable again
- a lot of them aren't on protocol 5.0.1 yet, but a decent portion is 
- a good chunk of hosts are too expensive (storage, egress)

Should we:

1. I believe roughly half of the migrations would be gone (5 out of 11TB) if we were to unblock hosts that are usable again. Do we want to more actively manage the hosts manually from the UI? 

2. Storage is pinned at 2USD and egress at 5USD - should we consider increasing those values.
I have not yet investigated to what price we would have to set those values to include most of them.

3. We currently migrate 10 slabs in parallel, perhaps we should make that configurable and play around with that value?

4. We don't have good insights in total number of unhealthy slabs. We also don't have an estimate on when migrations will be finished. We could add a statistic or periodically count the number of unhealthy slabs. 

5. Our query to fetch the next unhealthy slab is a little naïve, to keep the query fast we don't give priority to slabs that are more unhealthy than others. We could update the query to take into account some health value, or query for slabs below a certain health percentage first, or manipulate the order for slabs that really need to be migrated asap.

n8mgr:
Block reason: "usability." Love it.

n8mgr:
Noticed some things while looking at this yesterday:

1. We have a lot of "legacy" slabs with very bad redundancy settings: 1/2, 2/2, 2/3, etc. They are spamming the migration logs and cannot be migrated.
2. Our check for 3x redundancy is probably not smart enough. 2/6 is technically 3x, but highly likely to have data loss unless integrity checks detect the loss very quickly.
3. 10 slabs at a time is clearly not enough. I pushed a version using `runtime.NumCPU()` and cleared through the migrations in a day. Making it configurable is a must.

I deleted everything with objectively bad redundancy settings on Zeus. We should keep an eye on data loss with 10/30 and I'll create some sub-issues for the other items.

peterjan:
> Block reason: "usability." Love it.

[...]

PR #568, opened 2025-10-20
Update block reasons

peterjan:
Currently our blocked hosts have as reason in the database "usability". That's not very helpful, the UI helps but it would be great if we could create more complex queries to see why hosts are blocked and thus also cause slabs to become unhealthy. 

Related to https://github.com/SiaFoundation/indexd/issues/537


peterjan:
@ChrisSchinnerl if this gets merged I think it would be nice to expose number of blocked hosts and a count per reason, maybe housing anything that's not usability under an `other` category or something. Would be nice for the dashboard.

n8mgr (approved):

ChrisSchinnerl (approved):

PR #572, opened 2025-10-21
Handle irreparable slabs

### LINKED ISSUE #322: Handle irreparable slabs
There are reasons why a slab might not be recoverable. In some cases the slab will never be recoverable because wrong metadata was submitted and sometimes it might just be temporarily unavailable.

Right now we effectively try to repair all unhealthy slabs every 30 minutes by default. A.k.a. `healthCheckInterval`.
That is a bit wasteful and considering that some slabs might not recover ever, they unnecessarily reduce our ability to migrate repairable slabs.

We should come up with a performant way to do the following:

- exponentially back off from trying to migrate slabs that keep failing multiple times in a raw
- after 10 or so attempts over a generous amount of time (e.g. 1 week) we should stop repairing the slab altogether

peterjan:
This PR handles irreparable slabs by adding an exponential backoff mechanism to slab repairs. After a certain amount of consecutive failures we give up on the slab. 

```
cpu: Apple M1 Max
BenchmarkUnhealthySlabs
BenchmarkUnhealthySlabs/50-10        429           2505358 ns/op           16140 B/op        397 allocs/op
BenchmarkUnhealthySlabs/100-10               319           6157442 ns/op           33124 B/op        749 allocs/op
BenchmarkUnhealthySlabs/250-10               112          10159576 ns/op           74086 B/op       1797 allocs/op
```

Fixes https://github.com/SiaFoundation/indexd/issues/322


persist/postgres/slabs.go
```diff
@@ -5,12 +5,139 @@ import (
...
+//
+// NOTE: Subsequent calls to this function do not return the same slabs because
+// a minimum of 1 hour must pass between consecutive migration attempts. The
+// caller is expected to update the slab with the repair result after the
+// migration was attempted.
+//
+// NOTE: For the sake of scalability, we don't prioritize slabs based on their
+// health but simply return the slabs that have been waiting the longest for a
+// repair first.
+func (s *Store) UnhealthySlabs(ctx context.Context, limit int) (unhealthy []slabs.Slab, err error) {
```
ChrisSchinnerl: Unless there is a good reason to I'd stick to returning ids if it's faster. We already moved to full slabs in one PR just to revert back to ids.

persist/postgres/slabs.go
```diff
@@ -5,12 +5,139 @@ import (
...
+			SELECT s.id, s.digest, s.encryption_key, s.min_shards, s.pinned_at 
+			FROM slabs s
+			WHERE s.next_repair_attempt_after < NOW()
+				AND s.consecutive_failed_repairs < $1
+				AND EXISTS (
+					SELECT 1
+					FROM slab_sectors ss
+					JOIN sectors sec ON sec.id = ss.sector_id
+					WHERE ss.slab_id = s.id
+						AND (sec.host_id IS NULL OR (sec.contract_sectors_map_id IN (SELECT id FROM bad_contracts)))
```
ChrisSchinnerl: using `EXISTS` might be a bit more efficient than `IN (SELECT id FROM bad_contracts)`.

persist/postgres/sectors_test.go
```diff
@@ -1049,36 +1049,26 @@ func TestPinSectors(t *testing.T) {
 func TestUnhealthySlabs(t *testing.T) {
```
ChrisSchinnerl, line 1113: You moved `UnhealthySlabs` to `slabs.go` but kept this in `sectors.go`.

persist/postgres/slabs.go
```diff
@@ -5,12 +5,139 @@ import (
...
+		}
+
+		return nil
+	})
+	return
+}
+
+// MarkSlabRepaired marks the slab as repaired or increments the failed repair count.
+// If the slab has failed repair 10 times consecutively, it is marked as unrecoverable.
+func (s *Store) MarkSlabRepaired(ctx context.Context, slabID slabs.SlabID, success bool) error {
```
ChrisSchinnerl, line 19: Missing test. The backoff mechanism is quite critical.

persist/postgres/slabs.go
```diff
@@ -5,12 +5,139 @@ import (
...
+		if errors.Is(err, sql.ErrNoRows) {
+			return slabs.ErrSlabNotFound
+		} else if err != nil {
+			return fmt.Errorf("failed to fetch repair state: %w", err)
+		}
+
+		nextRepairBackoff := min(minRepairBackoff*time.Duration(1<<(currentFailures)), maxRepairBackoff)
+		_, err = tx.Exec(ctx, `
+			UPDATE slabs
+			SET consecutive_failed_repairs = $2, last_repair_attempt = NOW(), next_repair_attempt_after = $3
```
ChrisSchinnerl: We can probably get rid of `last_repair_attempt` now.

persist/postgres/slabs.go
```diff
@@ -5,12 +5,139 @@ import (
...
+			WHERE s.next_repair_attempt_after < NOW()
+				AND s.consecutive_failed_repairs < $1
+				AND EXISTS (
+					SELECT 1
+					FROM slab_sectors ss
+					JOIN sectors sec ON sec.id = ss.sector_id
+					WHERE ss.slab_id = s.id
+						AND (sec.host_id IS NULL OR (sec.contract_sectors_map_id IN (SELECT id FROM bad_contracts)))
+				)
+			ORDER BY s.last_repair_attempt ASC
```
ChrisSchinnerl: I think we need to order by `s.next_repair_attempt_after` here. Two slabs might have a similar `last_repair_attempt` if we attempted to repair them at the same time but we want to prioritize the slab with the lower `next_repair_attempt` because that's the one that hasn't failed as often yet.

persist/postgres/init.sql
```diff
@@ -266,14 +269,18 @@ CREATE TABLE slabs (
     digest BYTEA UNIQUE NOT NULL CHECK(LENGTH(digest) = 32), -- unique identifier for the slab derived from sector roots
 
     encryption_key BYTEA NOT NULL,
+    min_shards SMALLINT NOT NULL CHECK(min_shards > 0),
+
+    consecutive_failed_repairs SMALLINT NOT NULL DEFAULT 0 CHECK (consecutive_failed_repairs >= 0),
     last_repair_attempt TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
-    min_shards SMALLINT NOT NULL CHECK(min_shards > 0)
+    next_repair_attempt_after TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
```
ChrisSchinnerl: No need for `after`? It's a timestamp, not a duration.

ChrisSchinnerl (changes requested):

persist/postgres/slabs.go
peterjan: Well it depends on how you look at it being faster.  Fetching 128 unhealthy slabs with all setors is 210ms, fetching 128 unhealthy slab IDs  is 170ms but we'll follow it up with 128 parallel requests for a full slab. I would argue that unless there is good reason to return ids we should probably return full slabs here. Why did we revert back to ids the last time?

persist/postgres/init.sql
peterjan: `after` means `not_before` - I don't mind removing it but I prefer `next_repair_attempt_after` over `next_repair_attempt`, does knowing what I meant change your stance on the matter 😛 ?

persist/postgres/slabs.go
peterjan: Resolving in favour of https://github.com/SiaFoundation/indexd/pull/572#discussion_r2450754037

persist/postgres/slabs.go
peterjan: Yeah that was a bit lazy 👍

persist/postgres/slabs.go
peterjan: Hm actually I don't really mind, I think last time I was the one advocating to return IDs even. The fact that we have to hydrate them after fetching unhealthy slabs is def. an argument for returning IDs. I was on the fence because I do feel it's kind of nice we avoid the 128 slab fetches.

persist/postgres/slabs.go
peterjan: It's the same, 170ms vs 168ms for batch size of 250. I changed it to using `EXISTS`.

persist/postgres/init.sql
peterjan: I'll just change it

persist/postgres/slabs.go
peterjan: My gut feeling is that this is incorrect. Are you sure about this? I'm having a hard time putting it in words but I think that the current approach maintains fairness, while getting rid of `last_repair_attempt` would actually get rid of an important piece of data.

We only pull slabs whose next_repair_attempt has already passed, so they’re all eligible right now. Ordering by last_repair_attempt keeps “oldest waiting first,” which prevents a slab from being starved. If two slabs share the same last attempt time, we break the tie with consecutive_failed_repairs ASC so the slab that has failed fewer times (lower backoff) gets priority.

ChrisSchinnerl: > We only pull slabs whose next_repair_attempt has already passed, so they’re all eligible right now.
That is pretty much the point. The question is which one was eligible first. That's the one that has been waiting for a longer time. 

`last_repair_attempt` is always the time of the last repair. So a slab that has been healthy for a year before it becoming unhealthy again taking precedence over a slab that has been in need of migration for a couple of hours means that we are trying to migrate the slab that has been unhealthy for less time first. Because 1 year is seemingly a longer wait time than a few hours. Which does not reflect reality. 

To fix this you'd need to set `last_repair_attempt` to a sentinel value like `NULL` after successfully migrating the slab and take that into account when you order them. But then you start undermining the point of your backoff mechanism and add unnecessary complexity again. Over just sorting by time since becoming eligible for migration.

persist/postgres/sectors_test.go
```diff
@@ -1046,192 +1046,6 @@ func TestPinSectors(t *testing.T) {
 	}
 }
 
-func TestUnhealthySlabs(t *testing.T) {
```
ChrisSchinnerl, line 1113: The benchmark is still in `sectors_test.go`

persist/postgres/init.sql
```diff
@@ -221,6 +221,9 @@ CREATE TABLE contracts (
   raw_revision BYTEA NOT NULL
 );
 
+-- speeds up lookup of unhealthy slabs
```
ChrisSchinnerl: As you said, the speedup part of this PR should have been a separate one.
The refactor into `slabs.go` too probably since it does a good job of hiding changes to the important parts of the PR.

persist/postgres/slabs.go
```diff
@@ -5,12 +5,133 @@ import (
...
+//
+// NOTE: Subsequent calls to this function do not return the same slabs because
+// a minimum of 1 hour must pass between consecutive migration attempts. The
+// caller is expected to update the slab with the repair result after the
+// migration was attempted.
+//
+// NOTE: For the sake of scalability, we don't prioritize slabs based on their
+// health but simply return the slabs that have been waiting the longest for a
+// repair first.
+func (s *Store) UnhealthySlabs(ctx context.Context, limit int) (unhealthy []slabs.SlabID, err error) {
```
ChrisSchinnerl: Make sure you test your speed improvements on the prod database as well. IIRC Nate explicitly mentioned that some of his optimizations seemingly slowed things down in our synthetic benchmarks but were faster on Zeus.

persist/postgres/init.sql
peterjan: I reverted the move from sectors to slabs. I'd like to keep the speedup in the same PR since the query changes and I would have to verify it again. If you disagree I'll split it up.

peterjan: Might end up splitting it off anyway

persist/postgres/sectors_test.go
```diff
@@ -1866,33 +1844,42 @@ func BenchmarkUnhealthySlabs(b *testing.B) {
...
 		if err != nil {
 			b.Fatal(err)
 		}
 	}
 
-	for _, batchSize := range []int{10, 25, 50} {
-		b.Run(fmt.Sprint(batchSize), func(b *testing.B) {
-			reset(b)
-			b.ResetTimer()
+	// analyze tables to ensure query planner has up-to-date statistics
```
peterjan, line 1913: This is what kills the benchmark performance on `master` currently.

persist/postgres/sectors.go
```diff
@@ -671,69 +677,68 @@ func (s *Store) UnpinnedSectors(ctx context.Context, hostKey types.PublicKey, li
...
-          AND (c.good = FALSE OR c.state NOT IN ($2, $3)))
-    )
-)
-ORDER BY s.last_repair_attempt ASC
-LIMIT $4;`
-		var selected []int64
-		rows, err := tx.Query(ctx, query, maxRepairAttempt, sqlContractState(contracts.ContractStateActive), sqlContractState(contracts.ContractStatePending), limit)
+			FROM slabs s
+			WHERE s.next_repair_attempt < NOW()
+				AND s.consecutive_failed_repairs < $1
```
n8mgr: I'm not sure we should ever stop trying to repair. The backoff will prevent slabs from clogging the queue more than once a day. Better to have manual intervention.

peterjan: We can always add this later idd. It would be nice to expose max `consecutive_failed_repairs` somewhere 🤔 just to see what our worst case slab is doing.

persist/postgres/init.sql
```diff
@@ -397,9 +400,6 @@ CREATE TABLE slab_sectors (
     PRIMARY KEY (slab_id, sector_id)
 );
 
--- speeds up lookup of unhealthy slabs
-CREATE INDEX slab_sectors_sector_id_idx ON slab_sectors(sector_id);
```
ChrisSchinnerl, line 401: Are you sure this index is unused? We still seem to join over `sector_id`.

ChrisSchinnerl (dismissed):

PR #573, opened 2025-10-22
Prevent slabs with bad hosts from being pinned

### LINKED ISSUE #566: Prevent slabs with bad hosts from being pinned
We already check that the redundancy settings are sane and that all the hosts for the slab actually exist, but we currently don't care about whether the hosts are good, bad or even blocked.

We can give some leeway but we should make sure that most of the shards are on good hosts at the time of pinning to make sure we can actually pin the sectors or at least migrate the missing ones to good hosts.

chris124567:
Close #566

Currently there is no leeawy/tolerance for bad hosts and PinSlabs will just error if one of the sectors is on a bad host.  I wanted to start with a simple version and wasn't sure it was necessary given that if the end user uses the app client hosts endpoint then they should only be working with usable hosts anyways.

### DIFF
--- a/api/admin/admin_test.go
+++ b/api/admin/admin_test.go
@@ -939,10 +939,7 @@ func TestSectorStatsAPI(t *testing.T) {
 	indexer := cluster.Indexer
 	adminClient := indexer.Admin
 
-	// form contracts with hosts so we can pin slabs
-	for _, h := range cluster.Hosts {
-		indexer.Store().AddTestContract(t, h.PublicKey(), types.FileContractID(h.PublicKey()))
-	}
+	testutils.WaitForContracts(t, adminClient, len(cluster.Hosts))
 
 	// assert 0 slabs
 	stats, err := adminClient.StatsSectors(context.Background())
--- a/api/admin/admin_test.go
+++ b/api/admin/admin_test.go
@@ -939,6 +939,11 @@ func TestSectorStatsAPI(t *testing.T) {
 	indexer := cluster.Indexer
 	adminClient := indexer.Admin
 
+	// form contracts with hosts so we can pin slabs
+	for _, h := range cluster.Hosts {
+		indexer.Store().AddTestContract(t, h.PublicKey(), types.FileContractID(h.PublicKey()))
+	}
+
 	// assert 0 slabs
 	stats, err := adminClient.StatsSectors(context.Background())
 	if err != nil {
--- a/api/admin/admin_test.go
+++ b/api/admin/admin_test.go
@@ -386,6 +386,38 @@ func TestContractsAPI(t *testing.T) {
 		t.Fatal("expected no contract", len(contracts))
 	}
 
+	assertErr := func(got, expected error) {
+		t.Helper()
+		if got == nil || !strings.Contains(got.Error(), expected.Error()) {
+			t.Fatalf("expected error %v, got %v", expected, got)
+		}
+	}
+
+	if _, err := adminClient.Contracts(t.Context(), admin.WithContractSort("formation", false)); err != nil {
+		t.Fatal(err)
+	}
+
+	var err error
+	_, err = adminClient.Contracts(t.Context(), func(q url.Values) {
+		q.Add("sortby", "formation")
+		q.Add("desc", "invalid")
+	})
+	assertErr(err, api.ErrInvalidSortPair)
+
+	_, err = adminClient.Contracts(t.Context(), func(q url.Values) {
+		q.Add("sortby", "")
+		q.Add("desc", "true")
+	})
+	assertErr(err, api.ErrInvalidSortPair)
+
+	_, err = adminClient.Contracts(t.Context(), func(q url.Values) {
+		q.Add("sortby", "formation")
+	})
+	assertErr(err, api.ErrMissingSortPair)
+
+	_, err = adminClient.Contracts(t.Context(), admin.WithContractSort("does.not.exist", false))
+	assertErr(err, contracts.ErrInvalidSortField)
+
 	// block host and assert it's not returned
 	if err := adminClient.HostsBlocklistAdd(context.Background(), []types.PublicKey{h.PublicKey()}, []string{t.Name()}); err != nil {
 		t.Fatal(err)
--- a/api/admin/admin_test.go
+++ b/api/admin/admin_test.go
@@ -951,9 +951,8 @@ func TestSectorStatsAPI(t *testing.T) {
 	account := types.GeneratePrivateKey()
 	indexer.Store().AddTestAccount(t, account.PublicKey())
 	slabIDs, err := indexer.App(account).PinSlabs(context.Background(), slabs.SlabPinParams{
-		IgnoreBadHosts: true,
-		EncryptionKey:  [32]byte{1},
-		MinShards:      1,
+		EncryptionKey: [32]byte{1},
+		MinShards:     1,
 		Sectors: func() (s []slabs.PinnedSector) {
 			for _, h := range cluster.Hosts {
 				s = append(s, slabs.PinnedSector{Root: frand.Entropy256(), HostKey: h.PublicKey()})
--- a/api/admin/admin_test.go
+++ b/api/admin/admin_test.go
@@ -952,8 +952,8 @@ func TestSectorStatsAPI(t *testing.T) {
 	indexer.Store().AddTestAccount(t, account.PublicKey())
 	slabIDs, err := indexer.App(account).PinSlabs(context.Background(), slabs.SlabPinParams{
 		IgnoreBadHosts: true,
-		EncryptionKey: [32]byte{1},
-		MinShards:     1,
+		EncryptionKey:  [32]byte{1},
+		MinShards:      1,
 		Sectors: func() (s []slabs.PinnedSector) {
 			for _, h := range cluster.Hosts {
 				s = append(s, slabs.PinnedSector{Root: frand.Entropy256(), HostKey: h.PublicKey()})
--- a/api/admin/admin_test.go
+++ b/api/admin/admin_test.go
@@ -951,6 +951,7 @@ func TestSectorStatsAPI(t *testing.T) {
 	account := types.GeneratePrivateKey()
 	indexer.Store().AddTestAccount(t, account.PublicKey())
 	slabIDs, err := indexer.App(account).PinSlabs(context.Background(), slabs.SlabPinParams{
+		IgnoreBadHosts: true,
 		EncryptionKey: [32]byte{1},
 		MinShards:     1,
 		Sectors: func() (s []slabs.PinnedSector) {
--- a/internal/testutils/db.go
+++ b/internal/testutils/db.go
@@ -105,6 +105,17 @@ func (ts TestStore) AddTestHost(t testing.TB, host hosts.Host) {
 	}
 }
 
+func (ts TestStore) AddTestContract(t testing.TB, hk types.PublicKey, fcid types.FileContractID) {
+	rev := types.V2FileContract{
+		HostPublicKey:  hk,
+		RevisionNumber: 1,
+	}
+	err := ts.AddFormedContract(t.Context(), hk, fcid, rev, types.ZeroCurrency, types.ZeroCurrency, types.ZeroCurrency, proto.Usage{})
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
 // AddTestAccount adds an account to the database for testing.
 func (ts TestStore) AddTestAccount(t testing.TB, ak types.PublicKey) {
 	t.Helper()
--- a/internal/testutils/db.go
+++ b/internal/testutils/db.go
@@ -105,18 +105,6 @@ func (ts TestStore) AddTestHost(t testing.TB, host hosts.Host) {
 	}
 }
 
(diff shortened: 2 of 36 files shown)


api/admin/admin_test.go
```diff
@@ -918,7 +918,10 @@ func TestSectorStatsAPI(t *testing.T) {
 	// pin a slab
 	account := types.GeneratePrivateKey()
 	indexer.Store().AddTestAccount(t, account.PublicKey())
-	slabIDs, err := indexer.App(account).PinSlabs(context.Background(), slabs.SlabPinParams{
+
+	app := indexer.App(account)
+	testutils.WaitForHosts(t, app, 10)
```
peterjan: Why was this necessary?

persist/postgres/hosts_test.go
```diff
@@ -2276,7 +2277,7 @@ func BenchmarkHostsWithUnpinnableSectors(b *testing.B) {
 	}
 }
 
-func (s *Store) addTestHost(t testing.TB, hks ...types.PublicKey) types.PublicKey {
+func (s *Store) addTestHost(t testing.TB, good bool, hks ...types.PublicKey) types.PublicKey {
```
peterjan: Can we change `addTestHost` to update the host with valid settings, and add `addBadHost` that callls `addTestHost` and then invalidates the settings? We generally want test hosts to be good and this way we avoid the extra parameter and having to think about what to pass every time we require a test host.

peterjan: In tests you seem to pass `good=false` for the most part. Is that because we don't care whether the host is good or not because the test passes anyway? I would expect that we want good hosts for the most part. If we add the `addBadHost` helper we get rid of any confusion around the matter and you immediately understand what's going on if you read through the test.

persist/postgres/sectors.go
```diff
@@ -263,6 +263,67 @@ func (s *Store) PinSlabs(ctx context.Context, account proto.Account, nextIntegri
 			// insert slab's sectors in a single batch
 			batch := &pgx.Batch{}
 			for _, sector := range slab.Sectors {
+				// check host query
+				batch.Queue(`
+WITH globals AS (
```
peterjan: This is a little unfortunate right. I'm already pained by the fact we repeat this query, or a version of it, three times already, and now we have to add it a fourth time. Would it not be better to fetch all hosts using `WithPublicKeys` instead? 

We'll also have to check our benchmarks to see if this is causing a meaningful regression.  Pretty important route here. We might have to come up with a way of caching these checks.

peterjan (changes requested):

persist/postgres/sectors.go
ChrisSchinnerl: A way around this is checking contracts instead of hosts directly.
A contract will have `good=FALSE` set if it belongs to a bad or blocked host.

So when we insert a sector, we can check if we have `>0` good contracts with that host with a state of pending or active. To do so we can fetch all hosts for which that condition is true before inserting the sectors and store them in a map. Then check how many sectors are stored on bad hosts. 

Of course we still need to check how it affects the benchmark.

persist/postgres/sectors.go
```diff
@@ -279,12 +340,24 @@ func (s *Store) PinSlabs(ctx context.Context, account proto.Account, nextIntegri
...
-				var inserted bool
-				if err := br.QueryRow().Scan(&sectorIDs[i], &inserted); err != nil {
+				var blocklisted, good bool
+				if err := br.QueryRow().Scan(&blocklisted, &good); err != nil {
 					br.Close()
 					if errors.Is(err, sql.ErrNoRows) {
 						return fmt.Errorf("unknown host %q for sector", sector.HostKey)
 					}
+					return fmt.Errorf("failed to check if host %q is bad: %w", sector.HostKey, err)
+				} else if blocklisted {
```
ChrisSchinnerl: These checks might be a bit too unforgiving. Clients are expected to cache the list of hosts they use and they don't necessarily know if one of those hosts has turned bad / has been blocked.

So we should give it a bit of a leeway. e.g. 20% of parity shards.

We also need to expose a clear error that we can check for in the api to return a `BadRequest` status code as well as allow client apps to potentially handle that error.

ChrisSchinnerl (changes requested):

api/admin/admin_test.go
chris124567: Because if hosts aren't scanned in time, the call to pin slabs will fail because it won't find settings for the host and thus won't be able to check if the collateral/prices are good.

persist/postgres/hosts_test.go
chris124567: good=false is when we don't care if it's good or bad (for instance if the test doesn't need to pin any slabs) or if we are in a host test and will handle updating the manually.  We could do something like `addGoodHost` and `addDefaultHost`

peterjan: why is good not the default?

chris124567: It could be.  It's just unnecessary for a lot of tests because they aren't pinning slabs but it also doesn't really matter.

persist/postgres/sectors.go
chris124567: Meh... and mind you this query is covered by `contracts_host_id_active_good_idx`.

```
goos: linux
goarch: amd64
BenchmarkSlabs/PinSlab_checkHosts-4                   62      17213165 ns/op    2436.68 MB/s
BenchmarkSlabs/PinSlab_noCheckHosts-4                 80      14471741 ns/op    2898.27 MB/s
BenchmarkSlabs/Slab-4                               1671        601889 ns/op    69685.65 MB/s
BenchmarkSlabs/PinnedSlab-4                          421       2571357 ns/op    16311.63 MB/s
BenchmarkSlabs/Slabs-40MiB-4                        1353        978346 ns/op    42871.40 MB/s
BenchmarkSlabs/Slabs-400MiB-4                        300       3984644 ns/op    10526.17 MB/s
BenchmarkSlabs/Slabs-4GiB-4                           30      38787837 ns/op    1081.35 MB/s
BenchmarkSlabs/SlabIDs-4                              26      40799924 ns/op
PASS
ok      go.sia.tech/indexd/persist/postgres 534.834s
```

n8mgr (dismissed):

persist/postgres/sectors.go
```diff
@@ -193,7 +193,7 @@ func (s *Store) markFailingSectorsLostBatch(ctx context.Context, hostKey types.P
 
 // PinSlabs adds slabs to the database for pinning. The slabs are associated
 // with the provided account.
-func (s *Store) PinSlabs(ctx context.Context, account proto.Account, nextIntegrityCheck time.Time, toPin ...slabs.SlabPinParams) ([]slabs.SlabID, error) {
+func (s *Store) PinSlabs(ctx context.Context, account proto.Account, nextIntegrityCheck time.Time, checkHosts bool, toPin ...slabs.SlabPinParams) ([]slabs.SlabID, error) {
```
peterjan: I don't think `checkHosts` is ever false in production. If that is the case it should not be an argument. I can see how it's probably for testing but we should solve that differently, having it in the method signature here only invites the reader to think there's an edge case or a spot in our codebase where we call `PinSlabs` without checking hosts.

peterjan (changes requested):

chris124567:
I think that test is failing due to https://github.com/SiaFoundation/indexd/commit/10220f9dbc3031eb2968bf11dc9eb976a5e15382 which randomized the download candidate order

api/admin/admin_test.go
```diff
@@ -951,8 +951,9 @@ func TestSectorStatsAPI(t *testing.T) {
 	account := types.GeneratePrivateKey()
 	indexer.Store().AddTestAccount(t, account.PublicKey())
 	slabIDs, err := indexer.App(account).PinSlabs(context.Background(), slabs.SlabPinParams{
-		EncryptionKey: [32]byte{1},
-		MinShards:     1,
+		IgnoreBadHosts: true,
```
ChrisSchinnerl: I don't think this really resolves PJ's comment. It just moves the argument to the parameters. We never really want to ignore bad hosts. So we should update whatever tests are failing to have good hosts that don't cause `PinSlabs` to fail.

api/admin/admin_test.go
```diff
@@ -939,6 +939,11 @@ func TestSectorStatsAPI(t *testing.T) {
 	indexer := cluster.Indexer
 	adminClient := indexer.Admin
 
+	// form contracts with hosts so we can pin slabs
+	for _, h := range cluster.Hosts {
+		indexer.Store().AddTestContract(t, h.PublicKey(), types.FileContractID(h.PublicKey()))
```
peterjan: I can see how this is confusing but whenever we deal with a test cluster we don't `addTestX` things. The cluster should be forming contracts with those 10 hosts on its own, the only thing that needs to be done here is to wait until contracts are formed.

internal/testutils/db.go
```diff
@@ -105,6 +105,17 @@ func (ts TestStore) AddTestHost(t testing.TB, host hosts.Host) {
 	}
 }
 
+func (ts TestStore) AddTestContract(t testing.TB, hk types.PublicKey, fcid types.FileContractID) {
```
peterjan: This should not be necessary

persist/postgres/hosts_test.go
```diff
@@ -2298,6 +2308,30 @@ func (s *Store) addTestHost(t testing.TB, hks ...types.PublicKey) types.PublicKe
...
+		return tx.AddHostAnnouncement(hk, chain.V2HostAnnouncement{ha}, time.Now())
+	}); err != nil {
+		t.Fatal(err)
+	}
+	s.addTestContract(t, hk, types.FileContractID(frand.Entropy256()))
+	return hk
+}
+
+// addHost simply adds a host to the database without adding a test contract
+func (s *Store) addHost(t testing.TB, hks ...types.PublicKey) types.PublicKey {
```
peterjan: There are numerous spots where we call `addTestHost` and then immediately afterwards call `addTestContract`, or `addHost` followed by `addTestContract`, so either defeating the point of the helper, or leaving us with two contracts for that host.

I would remove this and revert `addTestHost` back to simply adding a host without a contract. I know there's probably a bunch of tests where we have to take this into account now, and that sucks. But it's only one line, it's not worth adding all this mental load for developers to have to think about when reading and writing tests.

If you really want to save the extra line you could add `addTestHostWithContract` that wraps `addTestHost` and `addTestContract`, or alternatively `addActiveHost` but I'm not sure if that's necessary.

peterjan (changes requested):

internal/testutils/indexer.go
```diff
@@ -359,3 +359,21 @@ func WaitForHosts(t *testing.T, app *app.Client, n int) []hosts.HostInfo {
...
 	}
 }
+
+// WaitForContracts waits for the given number of contracts to be formed.
+func WaitForContracts(t *testing.T, admin *admin.Client, n int) {
+	t.Helper()
+	start := time.Now()
+	limit := 10 * time.Second
+	for {
+		contracts, err := admin.Contracts(context.Background())
```
peterjan: `nit` use `t.Context()`

peterjan (dismissed):
Thanks for implementing my feedback, it's very straightforward now imo 👍

persist/postgres/sectors.go
```diff
@@ -292,6 +315,12 @@ func (s *Store) PinSlabs(ctx context.Context, account proto.Account, nextIntegri
 			}
 			br.Close()
 
+			// if more than 20% of parity shards are on bad hosts, don't allow slab to be pinned
+			parityShards := max(0, len(slab.Sectors)-int(slab.MinShards))
```
ChrisSchinnerl: oh wow if you ran into this then that means we don't check if we have more than `MinShards` sectors. Which is quite severe. 

Can you add checks to the beginning of the loop to make sure
1. `MinShards > 0`
2. `len(slab.Sectors) >= MinShards`

Then return a custom error and extend `TestPinSlabs` to check for it?

That way you don't need the `max` anymore and you close the edge case for `badHosts > 0` for `badHosts == 0`.

While you are at it I'd also move `0.2` to a `const`.

persist/postgres/sectors_test.go
```diff
@@ -734,13 +741,62 @@ func TestPinSlabs(t *testing.T) {
...
+	// pin slabs
+	slab1ID, slab1 := newSlab(1, hk1)
+	if slabIDs, err := store.PinSlabs(context.Background(), proto.Account{1}, nextCheck, slab1); err != nil {
+		t.Fatal(err)
+	} else if slabIDs[0] != slab1ID {
+		t.Fatalf("expected slab ID %v, got %v", slab1ID, slabIDs[0])
+	}
+
+	_, slab2 := newSlab(1, hk2)
+	if _, err := store.PinSlabs(context.Background(), proto.Account{1}, nextCheck, slab2); err == nil || !strings.Contains(err.Error(), "bad host") {
```
ChrisSchinnerl: Why not use `errors.Is` with `ErrBadHosts` here?

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

n8mgr (approved):

PR #575, opened 2025-10-22
Extend host stats

### LINKED ISSUE #571: Metrics: Extend `HostStats` to include block reasons
Then we can update the grafana dashboard to transform that response into a chart showing us the number of blocked hosts and the reason they were blocked for.

NOTE: I don't think we need to extend the database code to return that breakdown. Grafana should be able to do that by itself. We wouldn't be able to attach a breakdown to the existing prometheus host stats anyway.

Related to https://github.com/SiaFoundation/indexd/pull/568

peterjan:
`BenchmarkHostStats-10                100          10764501 ns/op ` 
slightly slower than master (7ms) but still fine imo

Fixes https://github.com/SiaFoundation/indexd/issues/571

### DIFF
--- a/api/admin/prometheus.go
+++ b/api/admin/prometheus.go
@@ -85,29 +85,24 @@ func (h HostStats) PrometheusMetric() []prometheus.Metric {
 		},
 	}
 
-	if h.Blocked {
-		metrics = append(metrics, prometheus.Metric{
-			Name:   "indexd_host_blocked",
-			Labels: labels,
-			Value:  1,
-		})
-		if len(h.BlockedReasons) > 0 {
-			for _, reason := range h.BlockedReasons {
-				metrics = append(metrics, prometheus.Metric{
-					Name: "indexindexd_host_blocked_reason",
-					Labels: map[string]any{
-						"public_key": h.PublicKey.String(),
-						"reason":     reason,
-					},
-					Value: 1,
-				})
+	metrics = append(metrics, prometheus.Metric{
+		Name:   "indexd_host_blocked",
+		Labels: labels,
+		Value: func() float64 {
+			if h.Blocked {
+				return 1
 			}
-		}
-	} else {
+			return 0
+		}(),
+	})
+	for _, reason := range h.BlockedReasons {
 		metrics = append(metrics, prometheus.Metric{
-			Name:   "indexd_host_blocked",
-			Labels: labels,
-			Value:  0,
+			Name: "indexd_host_blocked_reason",
+			Labels: map[string]any{
+				"public_key": h.PublicKey.String(),
+				"reason":     reason,
+			},
+			Value: 1,
 		})
 	}
 
--- a/api/admin/prometheus.go
+++ b/api/admin/prometheus.go
@@ -57,11 +57,12 @@ func (h HostStatsResponse) PrometheusMetric() (metrics []prometheus.Metric) {
 
 // PrometheusMetric implements the prometheus.Marshaller interface for a single
 // host's stats.
-func (h HostStats) PrometheusMetric() (metrics []prometheus.Metric) {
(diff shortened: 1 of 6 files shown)


hosts/manager.go
```diff
@@ -69,6 +69,8 @@ type (
 		PublicKey           types.PublicKey `json:"publicKey"`
 		LostSectors         int64           `json:"lostSectors"`
 		TotalUsage          types.Currency  `json:"totalUsage"`
+		Blocked             bool            `json:"blocked"`
+		BlockedReasons      []string        `json:"blockedReasons"`
```
ChrisSchinnerl, line 73: Need to be added to `prometheus.go` as well.

ChrisSchinnerl (changes requested):

hosts/manager.go
peterjan: 🤦

api/admin/prometheus.go
```diff
@@ -83,6 +84,34 @@ func (h HostStats) PrometheusMetric() (metrics []prometheus.Metric) {
 			Value:  float64(h.ActiveContractsSize),
 		},
 	}
+
+	if h.Blocked {
```
ChrisSchinnerl: Are these conditions necessary for grafana? As opposed to just always attaching a 1 when a host is blocked and a 0 when it's not plus an empty list of reasons when it's not?

peterjan: I mean that's more or less what I'm doing right? But you are right, I was winging it here to be honest. I was going to see what comes out of it and see how to build up the query in Grafana. I'll remove if len(h.BlockedReasons) > 0 {` but other than that I think it's probably fine?

peterjan: I restructured it a bit

peterjan:
@ChrisSchinnerl we should probably look into https://github.com/prometheus/client_golang ? 

> Prefer prometheus/client_golang gauges (or const metrics via prometheus.MustNewConstMetric) over a custom struct. You’ll also get # HELP/# TYPE lines (Grafana/Prometheus best practice).

n8mgr:
> @ChrisSchinnerl we should probably look into https://github.com/prometheus/client_golang ?

Why? That looks more concerned with collecting metrics to export. We already collect them. Our implementation has a simpler API, supports all or at least most of the open metrics spec, and works without a dependency.

api/admin/prometheus.go
```diff
@@ -83,6 +84,29 @@ func (h HostStats) PrometheusMetric() (metrics []prometheus.Metric) {
 			Value:  float64(h.ActiveContractsSize),
 		},
 	}
+
+	metrics = append(metrics, prometheus.Metric{
```
n8mgr: Inline this with the definition above

n8mgr (changes requested):

api/admin/prometheus.go
peterjan: 🤦🏻‍♂️ refactor booboo

peterjan:
> Why? That looks more concerned with collecting metrics to export. We already collect them. Our implementation has a simpler API, supports all or at least most of the open metrics spec, and works without a dependency.

We don't need it per se because what we have works and we don't have an extra dependency. That said, it would be nice if eventually we have inlined `# HELP` and `# TYPE` annotations, something like

```
# HELP indexd_host_lost_sectors The total number of sectors this host has lost
# TYPE indexd_host_lost_sectors counter
indexd_host_lost_sectors 1
```

I'm sure you'd get a bunch of other nice things like proper escaping, label ordering and compatibility with various other tooling. Typing it out it's really not that important, I know Chris wrote the prometheus integration so I was also fishing why we didn't use it in the first place.

ChrisSchinnerl (approved):

n8mgr (approved):

PR #577, opened 2025-10-23
Reuse dialed connections during migrations

### LINKED ISSUE #565: Reuse dialed connections during migrations
Right now we dial a new tcp connection for each shard download and upload. Once done we immediately close it.

Instead, we should dial hosts on demand and keep those connections open for the duration of the migration. That means we would have some state object wrapping the dialer before starting migrations and then closing it and all of its underlying connections once the migrations are done.

When we dial and it fails since the connection was closed by the host in the meantime, we should retry once similar to `retry` in the Go SDK.

chris124567:
Close #565

### DIFF
--- a/slabs/dialer.go
+++ b/slabs/dialer.go
@@ -43,12 +43,6 @@ func (c *connEntry) clear() {
 	}
 }
 
-func (c *connEntry) isConnected() bool {
-	c.mu.Lock()
-	defer c.mu.Unlock()
-	return c.tc != nil
-}
-
 func (c *connEntry) setTransport(tc HostClient, assign bool) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
@@ -69,7 +63,7 @@ type dialer struct {
 	conns map[types.PublicKey]*connEntry
 }
 
-// newDialer returns a new dialer.
+// newDialer returns a new dialer that reuses existing connections if possible.
 func newDialer(d Dialer, log *zap.Logger) *dialer {
 	return &dialer{
 		log:    log.Named("dialer"),
--- a/slabs/dialer.go
+++ b/slabs/dialer.go
@@ -1,0 +1,179 @@
+package slabs
+
+import (
+	"context"
+	"fmt"
+	"maps"
+	"slices"
+	"sync"
+
+	proto "go.sia.tech/core/rhp/v4"
+	"go.sia.tech/core/types"
+	"go.sia.tech/coreutils/chain"
+	"go.uber.org/zap"
+)
+
+type connEntry struct {
+	hostKey types.PublicKey
+
+	mu   sync.Mutex
+	dial chan struct{} // signals dial completion
+	tc   HostClient
+}
+
+func (c *connEntry) close(log *zap.Logger) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if c.tc != nil {
+		if err := c.tc.Close(); err != nil {
+			log.Debug("Failed to close connection", zap.Stringer("pk", c.hostKey), zap.Error(err))
+		}
+	}
+	c.tc = nil
+}
+
+func (c *connEntry) clear() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if c.tc != nil {
+		c.tc.Close()
+		c.tc = nil
+	}
+}
+
+func (c *connEntry) isConnected() bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.tc != nil
+}
+
+func (c *connEntry) setTransport(tc HostClient, assign bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if assign {
+		c.tc = tc
+	}
+	close(c.dial)
+	c.dial = nil
+}
+
+// dialer implements the [HostDialer] interface.
+type dialer struct {
+	log    *zap.Logger
+	dialer Dialer
+
+	mu    sync.Mutex
+	conns map[types.PublicKey]*connEntry
+}
+
+// newDialer returns a new dialer.
+func newDialer(d Dialer, log *zap.Logger) *dialer {
+	return &dialer{
+		log:    log.Named("dialer"),
+		dialer: d,
+
+		conns: make(map[types.PublicKey]*connEntry),
+	}
+}
+
+// Close closes all the open connections on the dialer.
+func (d *dialer) Close() {
+	d.mu.Lock()
+	conns := slices.Collect(maps.Values(d.conns))
+	d.mu.Unlock()
+
+	for _, entry := range conns {
+		entry.close(d.log)
+	}
+	clear(d.conns)
+}
+
+// clearHostConnection clears the connection for a host
+func (d *dialer) clearHostConnection(hostKey types.PublicKey) {
+	d.mu.Lock()
+	entry, exists := d.conns[hostKey]
+	d.mu.Unlock()
+	if !exists {
+		return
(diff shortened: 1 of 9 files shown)


slabs/dialer.go
```diff
@@ -0,0 +1,173 @@
...
+	// Clear connection if we got transport error
+	d.clearHostConnection(hostKey)
+
+	// Second attempt
+	tc, err = d.dialHost(ctx, hostKey, addrs)
+	if err != nil {
+		return fmt.Errorf("failed to redial host: %w", err)
+	}
+	return fn(tc)
+}
```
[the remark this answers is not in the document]
chris124567: We are only calling `Settings`, `WriteSector`, and `ReadSector` which are basically just simple wrappers around the function in core so there should not be any issues

slabs/dialer.go
```diff
@@ -0,0 +1,173 @@
...
+	if err := fn(tc); err == nil {
+		return nil
+	} else if proto.ErrorCode(err) != proto.ErrorCodeTransport {
+		return err
+	}
+
+	// Clear connection if we got transport error
+	d.clearHostConnection(hostKey)
+
+	// Second attempt
```
ChrisSchinnerl: (nit) could you change these comments and their counterparts in the sdk to not start with an uppercase letter?  We do try to be consistent about that but it must have slipped.

slabs/dialer.go
```diff
@@ -0,0 +1,173 @@
...
+	defer c.mu.Unlock()
+
+	if assign {
+		c.tc = tc
+	}
+	close(c.dial)
+	c.dial = nil
+}
+
+// dialer implements the [HostDialer] interface.
```
ChrisSchinnerl: This is not accurate for the `slabs` package. I'd rename `dialer` here to `connPool` and change the comment to saying that a connection pool wraps a `[Dialer]` and caches its connections.

slabs/uploads.go
```diff
@@ -81,24 +82,25 @@ func (m *SlabManager) uploadShards(ctx context.Context, slab Slab, shards [][]by
...
-		return proto.Usage{}, types.Hash256{}, fmt.Errorf("failed to dial host: %w", err)
-	}
-	defer client.Close()
-
-	settings, err := client.Settings(ctx)
-	if err != nil {
-		return proto.Usage{}, types.Hash256{}, fmt.Errorf("failed to fetch host settings: %w", err)
-	}
+	var result rhp.RPCWriteSectorResult
+	err := d.retry(ctx, h.PublicKey, h.RHP4Addrs(), func(client HostClient) error {
```
ChrisSchinnerl: To avoid polluting `uploadShard` with the retry logic of the connection pool, I'd add a `WriteSector` method to the pool which dials a host, calls `Settings` and `WriteSector` while wrapping it in `retry`.

Then in `uploadShard` it just becomes 
```go
result, err := pool.WriteSector(...)
if err != nil {
  ...
}
```

slabs/downloads.go
```diff
@@ -142,25 +143,28 @@ outer:
...
 	defer cancel()
 
-	client, err := m.dialer.DialHost(ctx, h.PublicKey, h.RHP4Addrs())
-	if err != nil {
-		return proto.Usage{}, nil, fmt.Errorf("failed to dial host: %w", err)
-	}
-	defer client.Close()
+	var result rhp.RPCReadSectorResult
+	var buf bytes.Buffer
+	err := d.retry(ctx, h.PublicKey, h.RHP4Addrs(), func(client HostClient) error {
```
ChrisSchinnerl: Same as my comment below. I'd pack the retry logic away in a `pool.ReadSector` method to keep the download code clean.

ChrisSchinnerl (changes requested):

slabs/manager.go
```diff
@@ -381,8 +381,8 @@ func (m *SlabManager) performSlabMigrations(ctx context.Context) error {
 	log := m.log.Named("migrations")
 	log.Debug("starting slab migrations", zap.Time("start", start))
 
-	d := newDialer(m.dialer, log)
-	defer d.Close()
+	c := newConnPool(m.dialer, log)
```
peterjan: We recreate the pool here and defer a close after migrations are done. Should we consider keeping a connection pool on the Slab manager instead? It would get rid of the need to pass the `connPool` around. (Could we rename it to `pool`? The single character variable names are just really hard to spot if you're reading through code)

peterjan (dismissed):

slabs/downloads_test.go
```diff
@@ -143,6 +143,9 @@ func TestDownloadShards(t *testing.T) {
 		}
 	}
 
+	c := newConnPool(sm.dialer, zap.NewNop())
```
peterjan: this can be removed

slabs/migrations_test.go
```diff
@@ -119,6 +119,9 @@ func TestMigrateSlab(t *testing.T) {
 	}
 
 	// migrate the slab
+	c := newConnPool(dialer, zap.NewNop())
```
peterjan: this can be removed

slabs/uploads_test.go
```diff
@@ -51,6 +51,9 @@ func TestUploadShards(t *testing.T) {
 	}
 	sm.shardTimeout = 50 * time.Millisecond
 
+	c := newConnPool(sm.dialer, zap.NewNop())
```
peterjan: this can be removed

slabs/manager.go
ChrisSchinnerl: @chris124567 please revert this. We don't want connections to remain open forever so we should close them as soon as we are done with them.

chris124567: > @chris124567 please revert this. We don't want connections to remain open forever so we should close them as soon as we are done with them. 

There is still the call to "defer m.pool.Close()" in the perform migrations function so nothing should be left open.  I will revert it anyways because I wasn't a huge fan of doing it this way.

peterjan (approved):
Code looks fine but the code duplication is making me a little uneasy.

I looked into it a bit and there are really no meaningful differences between the dialer from the `slabs` package and the one from the `sdk` package? Perhaps I'm missing the one crucial difference though. 

To me it looks like the dialer from the `sdk` package should be refactored to use the new `pool` component, that will allow us to `defer` a close in the migration code to ensure all connections get closed out.

The `connEntry` is identical, the dial logic is somewhat the same, but they both support reading and writing a sector to and from a host with a built-in retry. It seems like a bit of a shame to duplicate all that logic in two places.

@ChrisSchinnerl roping you in on this conversation since you already reviewed this PR and didn't make the same remark. I want to avoid Christopher refactoring the pool into the dialer from the `sdk` package only to revert the changes because there was good enough reason to warrant the code duplications.

peterjan:
I dug a little deeper, I think most of the friction is caused because the `hostDialer` interface has a `Hosts` method. I admit it's a bit of a tricky one because the `Hosts` method uses the app client, but also the connection pool to decide what hosts to return there. I don't think it really belongs on that interface though, if we strip it out there, the dialer from the `sdk` package and the `slabs` package become essentially the same thing.

The reason why `Hosts` is the issue is because it needs an app client, so the constructor of the dialer looks like this:
`newDialer(c appClient, appKey types.PrivateKey, log *zap.Logger)`. If it wouldn't need an app client it would make a lot more sense, the `appKey` would be the migration account key and we would be able to read and write to and from hosts supported by an underlying connection pool that closes out connections and retries on transport errors.

Could be a cool F/U imo.

ChrisSchinnerl (approved):

PR #580, opened 2025-10-23
Reduce size of signed urls

n8mgr:
new: 187 characters
old: 295 characters

36% size reduction

### DIFF
--- a/api/app/auth.go
+++ b/api/app/auth.go
@@ -2,7 +2,7 @@ package app
 
 import (
 	"bytes"
-	"encoding/hex"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


chris124567 (approved):

peterjan (approved):
Consider `RawURLEncoding` to get rid of padding and have even cleaner URLs? We could fall back when decoding so we accept both padded and unpadded.

api/app/auth.go
```diff
@@ -120,28 +120,24 @@ func validateSignedURLAuth(jc jape.Context, hostname, path string, store Account
 	return pk, true
 }
 
-func parseCredential(req *http.Request) (types.PublicKey, error) {
-	credStr := req.URL.Query().Get(queryParamCredential)
-	var pk types.PublicKey
-	err := pk.UnmarshalText([]byte(credStr))
+func parseCredential(req *http.Request) (pk types.PublicKey, _ error) {
+	n, err := base64.URLEncoding.Decode(pk[:], []byte(req.URL.Query().Get(queryParamCredential)))
```
ChrisSchinnerl, line 124: I'll leave this up to you since we also need to update the Rust SDK but according to [RFC4648](https://www.rfc-editor.org/rfc/rfc4648.html#section-3.2) it might be fine to use the `RAWURLEncoding` since we know the expected size of the data on both ends.

ChrisSchinnerl (approved):

PR #584, opened 2025-10-24
Calculate account funding

### LINKED ISSUE #576: Calculate Account Funding
Change account funding from a hardcoded value to an amount of storage usage in a period of time. i.e. 128 GiB worth of calls to `WriteSector` and `ReadSector` using each host's current pricing. This will ensure that we can always upload at ~3Gbps per host before refilling regardless of their pricing.

peterjan:
Fixes https://github.com/SiaFoundation/indexd/issues/576


accounts/manager.go
```diff
@@ -27,15 +27,24 @@ const (
...
+	// minFundTarget and maxFundTarget define the minimum and maximum fund target
+	// per contract used when funding accounts.
+	minFundTarget = types.Siacoins(2)
+	maxFundTarget = types.Siacoins(5)
+
+	// fundTargetBytes is the number of bytes used to calculate the fund target
+	// per host. We fund accounts to cover this amount of read and write usage.
+	// It roughly comes down to uploading or downloading from a host at ~3gbps
+	// for a period of 5 minutes.
+	fundTargetBytes = uint64(128 << 30) // 128 GiB
```
n8mgr: const?

accounts/manager.go
```diff
@@ -100,16 +108,16 @@ func (m *AccountManager) Close() error {
 
 // FundAccounts attempts to fund all accounts for the given host key. It does so
 // using the provided contract IDs, which are used in the order they're given.
-func (m *AccountManager) FundAccounts(ctx context.Context, host hosts.Host, contractIDs []types.FileContractID, force bool, log *zap.Logger) error {
+func (m *AccountManager) FundAccounts(ctx context.Context, host hosts.Host, contractIDs []types.FileContractID, force bool, hostLog *zap.Logger) error {
```
n8mgr: Please leave this as is. I am not interested in perpetuating this pattern further. It should be `log` unless absolutely necessary to split the logging context (there is very rarely a reason).

accounts/manager.go
```diff
@@ -27,15 +27,24 @@ const (
...
 	// accountActivityThreshold is the threshold for determining whether an
 	// account has been active recently for the purposes of contract funding.
 	// An account is considered active if it has been used within the threshold
 	// period.  We multiply the funding per contract by the number of active
 	// accounts.
 	accountActivityThreshold = 24 * 7 * time.Hour
+
+	// minFundTarget and maxFundTarget define the minimum and maximum fund target
+	// per contract used when funding accounts.
+	minFundTarget = types.Siacoins(2)
```
n8mgr: Why? If they have acceptable pricing based on our gouging settings, fund it.

peterjan: This is the part I wasn't too sure of yet, and which was the reason why it was still a `DRAFT`. 

The [issue](https://github.com/SiaFoundation/indexd/issues/576)  mentions an amount of storage used in a certain period of time assuming 3gbps transfer speeds. Since we refill every 5 minutes that's roughly 128GB worth of data. The problem I ran into though is that using average host prices it's quite costly to read/write 128GB worth of sectors. I was going to calculate the exact value using production averages but I think it's somewhere around 150SC.

I was going to propose the following:
a) only use the `WriteSector` cost as a fund target
b) keep the fund target between a sane min and max

I felt more comfortable applying a sane minimum, mostly because we only look at the write sector cost for funding. There's no downside to funding at least 1SC imo. The same goes for a maximum, I would be a little uneasy funding EAs without a cap, especially because we multiply this number with `number of accounts` when we calculate the contract's funds when forming/refreshing/renewing. That's also why I log a notice if we exceed the min/max target.

persist/postgres/hosts.go
```diff
@@ -830,8 +830,43 @@ func (s *Store) HostsForIntegrityChecks(ctx context.Context, maxLastCheck time.T
 	return hosts, nil
 }
 
+// HostsForFunding returns a list of host keys that need accounts funded on
+// them. A host is eligible for funding if it is not blocked and has an active
+// contract.
+func (s *Store) HostsForFunding(ctx context.Context) ([]types.PublicKey, error) {
```
chris124567, line 836: Should we benchmark this?

accounts/manager.go
n8mgr: Min and max don't fix the problem. You would be better off raising the arbitrary fund amount to 10 SC over adding complexity to compute and enforcing a lower cap.

If the concern is overfunding, the correct response would be to tune the funding frequency or change the expected data burst rate relative to the number of hosts, not introducing more arbitrary caps.

Users should not be hitting out of funds errors.

>  especially because we multiply this number with number of accounts when we calculate the contract's funds when forming/refreshing/renewing

We have gouging settings for this purpose: to limit the amount we are spending. Introducing more arbitrary min/max funding logic adds another location with cost control. If gouging isn't granular enough, fix it, don't add another independent limit.

persist/postgres/hosts.go
peterjan: Haha yeah I considered it but it's literally a duplicate of all `HostsForX` methods, and this one should be faster. I'll add one though, you are right

accounts/manager.go
```diff
@@ -216,13 +234,27 @@ func UpdateFundedAccounts(accounts []HostAccount, n int) {
 	}
 }
 
+// HostFundTarget calculates the fund target for the given host. We fund
+// accounts to cover 128GB of write usage, capped between a given minimum and
+// maximum.
+func HostFundTarget(host hosts.Host) types.Currency {
+	fundTarget := host.Settings.Prices.RPCWriteSectorCost(fundTargetBytes).RenterCost()
+
```
ChrisSchinnerl: This is missing the `ReadSector` part of the issue.

contracts/form_test.go
```diff
@@ -130,6 +131,7 @@ func (c *hostClientMock) FormContract(ctx context.Context, settings proto.HostSe
 type hostManagerMock struct {
 	settings map[types.PublicKey]proto.HostSettings
 
+	mu    sync.Mutex // prevents race when calling WithScannedHost concurrently, the storeMock is not thread-safe
```
ChrisSchinnerl: Shouldn't this be added to the store mock to make it thread safe rather than the host manager?

contracts/maintenance.go
```diff
@@ -7,62 +7,61 @@ import (
...
 func (cm *ContractManager) performAccountFunding(ctx context.Context, force bool, log *zap.Logger) error {
 	start := time.Now()
 
-	// fund accounts on usable hosts with active contracts
-	opts := []hosts.HostQueryOpt{
-		hosts.WithUsable(true),
-		hosts.WithBlocked(false),
-		hosts.WithActiveContracts(true),
+	// fetch hosts that need accounts funded
+	hff, err := cm.hosts.HostsForFunding(ctx)
```
ChrisSchinnerl: Don't rename vars just for the sake of it. Nothing wrong with `hostsToFund`.

The whole `HostsForFunding` change could've been a separate PR since it's not related to the original issue.

accounts/manager.go
```diff
@@ -122,6 +134,9 @@ func (m *AccountManager) FundAccounts(ctx context.Context, host hosts.Host, cont
 		}
 	}
 
+	// calculate the fund target for this host
+	fundTarget := HostFundTarget(host)
```
ChrisSchinnerl, line 131: You can log and skip the rest of this method if this is `0`.

ChrisSchinnerl (changes requested):

accounts/manager.go
peterjan: I took this out on purpose but reworked it already locally. See https://github.com/SiaFoundation/indexd/pull/584#discussion_r2465431902

contracts/form_test.go
peterjan: Yeah that was intentional. I felt stupid making our mocks thread safe. By putting a little mutex here we avoid the race and the problem resolves itself since we're getting rid of our mocks anyway.

peterjan: Moved it anyway because it looks like Nate ran into the same issue

n8mgr (dismissed):

ChrisSchinnerl (commented):
Got conflicts now after merging the other PR

ChrisSchinnerl (dismissed):

n8mgr (dismissed):

PR #585, opened 2025-10-27
Detect changing seed 

### LINKED ISSUE #583: Detect changing seed and deletion/swapping of chain state.
We should double-check that we handle the following edge-cases. 

- [x] A user swaps out a consensus db for another
- [x] A user deletes the consensus db 
- [x] A user changed the wallet seed they are using

We already call `ResetChainState` within the `Sync` method of the `Subscriber` so we might handle the first two cases correctly already. Doesn't hurt to double-check though. 

For the seed verification we should draw inspiration from `hostd`, but return an error rather than implicitly resetting the state for `indexd` since it won't be able to handle the migration of the data anyway.

chris124567:
Close #583

Logic mostly copied code from hostd.

There is a test for the DB method but I also tested the command.  As expected, after reconfiguring indexd and getting a new seed:

```
$ ./indexd 
2025-10-27T00:05:24-04:00       INFO    postgres        connected       {"database": "indexd", "host": "127.0.0.1", "port": 5432}
daemon startup failed: wallet seed change detected, cannot proceed
```

### DIFF
--- a/cmd/indexd/run.go
+++ b/cmd/indexd/run.go
@@ -103,6 +103,13 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
 	go s.Run()
 	defer s.Close()
 
+	walletHash := types.HashBytes(walletKey[:])
+	if err := store.VerifyWalletKey(walletHash); errors.Is(err, wallet.ErrDifferentSeed) {
+		return errors.New("wallet seed change detected, cannot proceed")
+	} else if err != nil {
+		return fmt.Errorf("failed to verify wallet key: %w", err)
+	}
+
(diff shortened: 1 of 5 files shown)


ChrisSchinnerl (approved):

n8mgr (changes requested):
Probably don't want to merge this as-is since the contract key and service account keys are derived from the wallet seed. This works fine in hostd since the host's contract key is not derived from the wallet seed. In indexd and renterd we'd lose access to all our existing contracts and data.

chris124567:
> Probably don't want to merge this as-is since the contract key and service account keys are derived from the wallet seed. This works fine in hostd since the host's contract key is not derived from the wallet seed. In indexd and renterd we'd lose access to all our existing contracts and data.

I thought that was the reason for #583.  If we change the seed, we won't be able to access any of our data, so there's no point of starting indexd.  Should we just log a warning instead?

n8mgr:
Ignore me, it returns an error

PR #587, opened 2025-10-27
Upload shards in parallel to improve migration performance

peterjan:
Fixes https://github.com/SiaFoundation/indexd/issues/55

### DIFF
--- a/slabs/downloads.go
+++ b/slabs/downloads.go
@@ -70,8 +70,8 @@ func (dc *downloadCandidates) next() (hosts.Host, bool) {
 // downloadShards downloads at least the minimum number of shards required to
 // recover the slab.
 func (m *SlabManager) downloadShards(ctx context.Context, slab Slab, allHosts []hosts.Host, pool *connPool, logger *zap.Logger) ([][]byte, error) {
-	ctx, downloadCancel := context.WithCancel(ctx)
-	defer downloadCancel()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 
 	shards := make([][]byte, len(slab.Sectors))
 	var downloaded atomic.Uint32
@@ -126,15 +126,13 @@ outer:
 				return
 			}
 
-			debitCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
-			err = m.am.DebitServiceAccount(debitCtx, host.PublicKey, m.migrationAccount, usage.RenterCost())
+			err = m.am.DebitServiceAccount(context.Background(), host.PublicKey, m.migrationAccount, usage.RenterCost())
 			if err != nil {
 				logger.Debug("failed to debit service account for sector read", zap.Error(err))
 			}
-			cancel()
 
 			if n := downloaded.Add(1); n >= uint32(slab.MinShards) {
-				downloadCancel()
+				cancel()
 			}
 		}(host, candidates.indices[host.PublicKey], logger.With(zap.Stringer("hostKey", host.PublicKey)))
 	}
--- a/slabs/downloads.go
+++ b/slabs/downloads.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	proto "go.sia.tech/core/rhp/v4"
 	"go.sia.tech/core/types"
@@ -123,10 +124,12 @@ outer:
 				return
 			}
 
-			err = m.am.DebitServiceAccount(ctx, host.PublicKey, m.migrationAccount, usage.RenterCost())
+			debitCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			err = m.am.DebitServiceAccount(debitCtx, host.PublicKey, m.migrationAccount, usage.RenterCost())
 			if err != nil {
 				logger.Debug("failed to debit service account for sector read", zap.Error(err))
 			}
+			cancel()
 
 			if n := downloaded.Add(1); n >= uint32(slab.MinShards) {
 				cancel()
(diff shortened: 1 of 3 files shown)


slabs/uploads.go
```diff
@@ -29,53 +30,103 @@ type (
...
 // and should be tracked in the database. The given shards must not be nil and
-// the given hosts must all be good and be sufficiently spaced apart.
-func (m *SlabManager) uploadShards(ctx context.Context, slab Slab, shards [][]byte, uploadCandidates []hosts.Host, pool *connPool, logger *zap.Logger) ([]Shard, error) {
-	uploaded := make([]Shard, 0, len(shards))
-
+// the given hosts must all be good and be sufficiently spaced apart. Note that
+// the shards returned are not necessarily in the same order as the input
+// shards.
+func (m *SlabManager) uploadShards(ctx context.Context, slab Slab, shards [][]byte, candidates []hosts.Host, pool *connPool, logger *zap.Logger) ([]Shard, error) {
 	if len(slab.Sectors) != len(shards) {
```
ChrisSchinnerl, line 38: if you wrap `ctx` using `WithCancel`, you can cancel all ongoing uploads immediately in case you run out of hosts or know that the upload is going to fail for some other reason.

slabs/uploads.go
```diff
@@ -29,53 +30,103 @@ type (
...
+				uploadMu.Lock()
+				if root != slab.Sectors[i].Root {
+					uploadErr = fmt.Errorf("failed to upload shard %d: %w", shardIndex, errRootMismatch)
+					uploadMu.Unlock()
+					return
+				}
+				uploadMu.Unlock()
+
+				// debit service account
+				err = m.am.DebitServiceAccount(ctx, host.PublicKey, m.migrationAccount, usage.RenterCost())
```
ChrisSchinnerl: This should be on a separate context. We might need to fix the same thing in `downloadShards` as well.

slabs/uploads_test.go
```diff
@@ -91,8 +91,6 @@ func TestUploadShards(t *testing.T) {
 	uploaded, err = sm.uploadShards(context.Background(), corrupted, shards, []hosts.Host{h1, h2, h3, h4}, pool, zap.NewNop())
 	if !errors.Is(err, errRootMismatch) {
 		t.Fatalf("expected [errRootMismatch] got %v", err)
-	} else if len(uploaded) != 1 {
-		t.Fatalf("expected 1 uploaded shard, got %d", len(uploaded))
```
ChrisSchinnerl, line 95: This should still work right?

slabs/uploads.go
```diff
@@ -29,53 +30,103 @@ type (
...
+				// upload the shard
+				usage, root, err := m.uploadShard(ctx, host, bytes.NewReader(shard), pool)
+				if err != nil {
+					logger.Debug("failed to upload shard", zap.Stringer("hostKey", host.PublicKey), zap.Error(err))
+					continue
+				}
+
+				// handle root mismatch
+				uploadMu.Lock()
+				if root != slab.Sectors[i].Root {
```
ChrisSchinnerl: Pass the root as an argument to the goroutine. Then you don't need to cover `slab` with `uploadMu`.

slabs/uploads.go
```diff
@@ -29,53 +30,103 @@ type (
...
+// the shards returned are not necessarily in the same order as the input
+// shards.
+func (m *SlabManager) uploadShards(ctx context.Context, slab Slab, shards [][]byte, candidates []hosts.Host, pool *connPool, logger *zap.Logger) ([]Shard, error) {
 	if len(slab.Sectors) != len(shards) {
 		panic(fmt.Sprintf("slab %s has %d sectors but %d shards", slab.ID, len(slab.Sectors), len(shards))) // developer error
 	}
 
+	var uploadMu sync.Mutex
 	var uploadErr error
+	uploaded := make([]Shard, 0, len(shards))
```
ChrisSchinnerl: (nit) this preallocates around 2KB of capacity when usually we probably need closer to 1-3 shards.

ChrisSchinnerl (changes requested):

slabs/uploads_test.go
peterjan: Kind of, we used to hit it immediately because we tried the hosts in order. We'll now get two uploaded shards through but seeing as we break out early if we encounter a root mismatch it's technically possible to end up with no uploaded shards at all. That's very unlikely though, I ran it 100 times and it was always 2.

slabs/uploads.go
peterjan: That's better than checking the `uploadErr` 👍  - this made me rewrite the test quite a bit though. Calling cancel on error makes it very non-deterministic and racy, but that's fine.

slabs/uploads.go
```diff
@@ -30,58 +31,103 @@ type (
...
+					uploadCancel()
+					return
+				}
+				host := candidates[0]
+				candidates = candidates[1:]
+				uploadMu.Unlock()
+
+				// upload the shard
+				start := time.Now()
+				usage, root, err := m.uploadShard(ctx, host, shard, pool)
```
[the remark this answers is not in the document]
peterjan: nice

slabs/downloads.go
```diff
@@ -126,13 +126,15 @@ outer:
 				return
 			}
 
-			err = m.am.DebitServiceAccount(ctx, host.PublicKey, m.migrationAccount, usage.RenterCost())
+			debitCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
n8mgr: Be consistent. You've got one callsite overloading `ctx` with a named `downloadCancel` and the other call site as `debitCtx, cancel`.

n8mgr: Also begs the question, do we want this to time out? What are the benefits and consequences of arbitrarily timing out after 10s? This is a database method for keeping our account balances somewhat synchronized, I would probably argue we should use context.Background.

peterjan: I used `context.Background` 🤷🏻‍♂️ - generally not a big fan though but I don't want to drag in `shutdownCtx` either 🤣

ChrisSchinnerl (dismissed):

chris124567 (dismissed):
I ran tests with `-race` and everything seems to be good

ChrisSchinnerl (approved):

chris124567 (approved):

n8mgr (approved):

PR #591, opened 2025-10-28
Fix master

### LINKED ISSUE #590: Fix TestDownloadCandidates NDF
https://github.com/SiaFoundation/indexd/actions/runs/18849211346/job/53781287543#step:6:255

peterjan:
Fixes master after merging https://github.com/SiaFoundation/indexd/pull/588 

Fixes https://github.com/SiaFoundation/indexd/issues/590

### DIFF
--- a/slabs/downloads_test.go
+++ b/slabs/downloads_test.go
@@ -1,6 +1,7 @@
 package slabs
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"reflect"
@@ -43,14 +44,22 @@ func TestDownloadCandidates(t *testing.T) {
 	candidates := newDownloadCandidates([]hosts.Host{h1, h2, h3, h4}, slab)
 
 	// test next() method
-	if host, ok := candidates.next(); !ok || host.PublicKey != h3.PublicKey {
-		t.Fatalf("expected host %v, got %v", h3.PublicKey, host.PublicKey)
-	} else if host, ok := candidates.next(); !ok || host.PublicKey != h4.PublicKey {
-		t.Fatalf("expected host %v, got %v", h4.PublicKey, host.PublicKey)
-	} else if host, ok := candidates.next(); !ok || host.PublicKey != h1.PublicKey {
-		t.Fatalf("expected host %v, got %v", h1.PublicKey, host.PublicKey)
-	} else if _, ok := candidates.next(); ok {
+	selected := make(map[types.PublicKey]struct{})
+	for range 3 {
+		if host, ok := candidates.next(); ok {
+			selected[host.PublicKey] = struct{}{}
+		}
+	}
+	if _, ok := candidates.next(); ok {
 		t.Fatal("expected no more hosts")
+	} else if len(selected) != 3 {
+		t.Fatalf("expected 3 selected hosts, got %d", len(selected))
+	} else if _, ok := selected[h1.PublicKey]; !ok {
+		t.Fatal("expected host 1 to be selected")
+	} else if _, ok := selected[h3.PublicKey]; !ok {
+		t.Fatal("expected host 3 to be selected")
+	} else if _, ok := selected[h4.PublicKey]; !ok {
+		t.Fatal("expected host 4 to be selected")
 	}
 
 	// assert indices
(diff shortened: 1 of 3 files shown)


ChrisSchinnerl (approved):

ChrisSchinnerl (changes requested):
Seems like we got one more NDF that's possibly related

peterjan:
> Seems like we got one more NDF that's possibly related

one does not simply randomize hosts in the download code 😅

slabs/downloads_test.go
```diff
@@ -197,15 +213,21 @@ func TestDownloadShards(t *testing.T) {
 			t.Fatal(err)
 		} else if !reflect.DeepEqual(sectors, [][]byte{nil, sector2[:], sector3[:]}) {
 			t.Fatal("downloaded sectors do not match expected sectors")
-		} else if sectors := store.lostSectors[hk1]; len(sectors) != 1 {
-			t.Fatalf("expected 1 lost sector for host %v, got %d", hk1, len(store.lostSectors[hk1]))
-		} else if _, ok := sectors[proto.SectorRoot(&sector1)]; !ok {
-			t.Fatalf("expected sector %v to be marked as lost, but it wasn't", proto.SectorRoot(&sector1))
+		}
+
+		if len(store.lostSectors) > 0 {
```
peterjan: since we run `minShards` download workers in parallel we can't influence whether we try h1 here

ChrisSchinnerl: Can't we delete the hosts we don't want to try from the dialer? Or make them fail?

We can't randomly let this subtest pass without testing what needs to be tested.

slabs/downloads_test.go
```diff
@@ -165,18 +157,42 @@ func TestDownloadShards(t *testing.T) {
...
+		for _, shard := range sectors {
+			if shard == nil {
+				continue
+			}
+			nFetched++
+
+			// sanity check
+			isS1 := bytes.Equal(shard, sector1[:])
+			isS2 := bytes.Equal(shard, sector2[:])
+			isS3 := bytes.Equal(shard, sector3[:])
```
ChrisSchinnerl: This doesn't seem right. `downloadShards` should always return the sectors according to their oder in the slab. We rely on that in the migration code since the offset within the result affects the nonce we use for decrypting the data.

ChrisSchinnerl (changes requested):

slabs/downloads_test.go
peterjan: I rewrote it. I wanted to assert we have `minShard` sectors and they match up.

slabs/downloads_test.go
peterjan: We can't test the case where the download succeeds, but one of the sectors was marked as lost. I updated it to assert a lost sector is marked as lost, because that's what it's really about anyway. We can't influence what hosts the download code tries first, so if it we mock a lost sector on h1, it might try h2 and h3 and then the download succeeds.

We can rewrite this in a way we _can_ influence this but I don't really see the point. Fine with doing that though if you disagree and feel it's necessary.

slabs/downloads_test.go
```diff
@@ -69,19 +78,12 @@ func TestDownloadShards(t *testing.T) {
 	store := newMockStore()
 	am := newMockAccountManager(store)
 	hm := newMockHostManager()
-	account := types.GeneratePrivateKey()
-
-	settings := func(egress uint64) proto.HostSettings {
-		settings := goodSettings
-		settings.Prices.EgressPrice = types.NewCurrency64(egress)
-		return settings
```
ChrisSchinnerl, line 77: This is a good example of why we shouldn't refactor whenever we get the chance. Especially when in a rush to fix something.

By using `goodSettings` directly you end up 
1. breaking the assertions at the end of the test (you can put any number of sectors in and it will pass)
2. removing context about the pricing from a test that checks balances and moving it to a spot where it is easy to change without being aware of the consequences it might cause

ChrisSchinnerl (changes requested):

slabs/downloads_test.go
peterjan: We removed sorting so I still feel it's necessary to stop initialising hosts with different price settings. I added an assertion though that all hosts should have a non-zero read sector cost.

ChrisSchinnerl (approved):

PR #592, opened 2025-10-28
Add `HostsForFunding`

peterjan:
This PR adds `HostsForFunding` - making it more consistent but also avoids fetching full hosts since we'll need to scan the host anyway to get updated prices to calculate the fund target (see [this PR](https://github.com/SiaFoundation/indexd/pull/584/files#r2468791820) for context)

```
BenchmarkHostsForFunding         1660            771416 ns/op  
```

### DIFF
--- a/persist/postgres/hosts.go
+++ b/persist/postgres/hosts.go
@@ -843,8 +843,6 @@ func (s *Store) HostsForFunding(ctx context.Context) ([]types.PublicKey, error) 
 				SELECT 1
 				FROM contracts c
 				WHERE c.host_id = h.id AND c.state IN (0, 1) AND c.good AND c.renewed_to IS NULL
-			)  AND NOT EXISTS (
-				SELECT 1 FROM hosts_blocklist hb WHERE hb.public_key = h.public_key
 			)`)
 		if err != nil {
 			return fmt.Errorf("failed to query hosts for funding: %w", err)
--- a/persist/postgres/hosts.go
+++ b/persist/postgres/hosts.go
@@ -866,7 +866,7 @@ func (s *Store) HostsForFunding(ctx context.Context) ([]types.PublicKey, error) 
 }
 
 // HostsForPinning returns a list of host keys that can be used for sector
-// pinning. A host is eligble for pinning if it is not blocked, has unpinned
+// pinning. A host is eligible for pinning if it is not blocked, has unpinned
 // sectors and has an active contract.
 func (s *Store) HostsForPinning(ctx context.Context) ([]types.PublicKey, error) {
 	var hosts []types.PublicKey
--- a/persist/postgres/hosts.go
+++ b/persist/postgres/hosts.go
@@ -830,6 +830,41 @@ func (s *Store) HostsForIntegrityChecks(ctx context.Context, maxLastCheck time.T
 	return hosts, nil
 }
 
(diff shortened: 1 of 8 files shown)


peterjan:
Please note the base branch is `pj/fix-master` - probably want to hold off on merging this until the testsuite is 🟢 again on master.

chris124567 (dismissed):

persist/postgres/hosts.go
```diff
@@ -830,8 +830,43 @@ func (s *Store) HostsForIntegrityChecks(ctx context.Context, maxLastCheck time.T
...
+	var hosts []types.PublicKey
+	if err := s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		rows, err := tx.Query(ctx, `
+			SELECT h.public_key
+			FROM hosts h
+			WHERE EXISTS (
+				SELECT 1
+				FROM contracts c
+				WHERE c.host_id = h.id AND c.state IN (0, 1) AND c.good AND c.renewed_to IS NULL
+			)  AND NOT EXISTS (
```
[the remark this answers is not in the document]
peterjan: Nice try though, we want to check usability on a scanned host. Fetching usable hosts would imply duplicating the usability filters here while it's totally unnecessary

persist/postgres/hosts_test.go
```diff
@@ -1716,6 +1716,74 @@ func newTestHostSettings(pk types.PublicKey) proto4.HostSettings {
...
+	// reset
+	updateContractState(fcid1, contracts.ContractStateActive)
+	assertNumHostsForFunding(1)
+
+	// assert good is taken into account
+	if _, err := store.pool.Exec(context.Background(), `UPDATE contracts SET good = FALSE WHERE contract_id = $1`, sqlHash256(fcid1)); err != nil {
+		t.Fatal(err)
+	}
+	assertNumHostsForFunding(0)
+}
```
ChrisSchinnerl, line 1810: also check `renewed_to`

ChrisSchinnerl (changes requested):

persist/postgres/hosts.go
```diff
@@ -830,8 +830,43 @@ func (s *Store) HostsForIntegrityChecks(ctx context.Context, maxLastCheck time.T
...
+				SELECT 1 FROM hosts_blocklist hb WHERE hb.public_key = h.public_key
```
ChrisSchinnerl: In theory this is redundant. If a host is blocked, we should mark all of its contracts as bad. If that is not the case we should double check why.

ChrisSchinnerl (dismissed):

persist/postgres/hosts.go
peterjan: Yeah I thought about that actually but we have it in all of our `HostsForX`. I decided to keep it because it's the safest option. Perhaps though we should remove it in all places and rely on our `host.Blocked` checks to notify us of any issues through the logs?

ChrisSchinnerl: No need to introduce consistency just because the method names are similar. All those queries are specialized so it's fine for them to return exactly the hosts we need rather than extracting functionality into Go code.

ChrisSchinnerl (dismissed):

n8mgr (approved):

ChrisSchinnerl (approved):

Issue #350, opened 2025-08-15
Remove all Store mocks


n8mgr:
Replace all the store mocks with the Postgres store. Mocking should focus on triggering failure paths rather than *everything*.

Complete:
- `accounts`
- `pins`
- `hosts`
- `contracts`
- `slabs`
- `client`
- `client/v2`

PR #596, opened 2025-10-29
Remove store mocks from `hosts`

chris124567:
Re: #350

I also adjusted [TestScanTimeout](https://github.com/SiaFoundation/indexd/blob/master/hosts/manager_test.go#L427).  If the point of the test is to test that the timeout works, then we should check that the scan fails if we block for longer than the context deadline.  I noticed that it was succeeding when I was getting "context deadline exceeded" errors from the postgres store's `UpdateHost` being called.


cmd/indexd/run.go
```diff
@@ -121,7 +121,7 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
 		return fmt.Errorf("failed to create MaxMind locator: %w", err)
 	}
 
-	hm, err := hosts.NewManager(s, locator, store, hosts.WithLogger(log.Named("hosts")))
+	hm, err := hosts.NewManager(s, &hosts.HostScanner{}, locator, store, hosts.WithLogger(log.Named("hosts")))
```
peterjan: It should be an option. We don't want someone that's fine using the default scanner implementation to have to think about the scanner. In testing we can then pass `hosts.WithScanner`. I think we can probably treat the locator the same way. The syncer is different because it needs a listener so it has to be configured.

hosts/manager_test.go
```diff
@@ -185,226 +44,79 @@ func TestHostManager(t *testing.T) {
...
-	// assert host settings are returned
-	hs, err = fetchSettings(context.Background(), c, types.PublicKey{}, []chain.NetAddress{testMuxAddr("foo.bar")}, zap.NewNop())
-	if err != nil {
-		t.Fatal("unexpected", err)
-	} else if hs.Release != t.Name() {
-		t.Fatal("unexpected", hs)
-	}
-}
-
-func TestCalculateNextScanTime(t *testing.T) {
```
peterjan, line 290: Why did you move this test? `calculateNextScanTime` is still in `manager.go`. If you want to move large and helpers it's also good to consider whether it's really necessary for the PR, if not it's best to do it separately  because we have to now verify whether something has changed and it just blows up the diff.

hosts/manager_test.go
```diff
@@ -185,226 +44,79 @@ func TestHostManager(t *testing.T) {
...
-				if nextScan.Before(tt.expectedMinNextScan) || nextScan.After(tt.expectedMaxNextScan) {
-					t.Errorf("Expected next scan time between %v and %v, got %v", tt.expectedMinNextScan, tt.expectedMaxNextScan, nextScan)
-				}
-			}
 		})
-	}
-}
-
-// TestResolveHost unit tests the functionality of ResolveHost.
-func TestResolveHost(t *testing.T) {
```
peterjan, line 369: Same for this one, seems like `resolveHost` is still in `manager.go`

hosts/manager_test.go
```diff
@@ -425,38 +137,29 @@ func (bs *blockingScanner) ScanQuic(ctx context.Context, hk types.PublicKey, add
...
+		ctx, cancel := context.WithTimeout(t.Context(), 100*time.Millisecond)
 		defer cancel()
 
-		host, err := mgr.ScanHost(ctx, hostKey)
-		if err != nil {
-			t.Fatal(err)
-		} else if host.Settings.Release != release {
-			t.Fatal("unexpected", host.Settings)
+		_, err = mgr.ScanHost(ctx, hostKey)
+		if err == nil || !errors.Is(err, context.DeadlineExceeded) {
```
peterjan: This is testing the timeout now, great, but it should also test the happy case.

hosts/utils_test.go
```diff
@@ -42,3 +58,201 @@ func TestOnlineChecker(t *testing.T) {
...
+	}
+
+	hs, ok := c.settings[addr]
+	if ok {
+		return hs, nil
+	}
+	return proto4.HostSettings{}, errors.New("") // mock host being unavailable on unknown address
+}
+
+func TestFetchSettings(t *testing.T) {
```
peterjan: `fetchSettings` is still in `manager.go`

internal/testutils/mock/mock.go
```diff
@@ -0,0 +1,28 @@
+package mock
```
peterjan, line 1: It's a bit of an odd choice to add a `mock` package only to house the locator.  Do we have plans for this package or were you just looking for somewhere to put it. I'd put it in `locator.go` under `testutils` for the time being.

peterjan (changes requested):

hosts/manager_test.go
chris124567: Because `manager_test.go` is now in the `hosts_test` package and this function is unexported

hosts/manager_test.go
chris124567: Because `manager_test.go` is now in the `hosts_test` package and this function is unexported

hosts/utils_test.go
chris124567: Because `manager_test.go` is now in the `hosts_test` package and this function is unexported

internal/testutils/mock/mock.go
chris124567: If we have it in testutils then we get an import cycle

cmd/indexd/run.go
chris124567: I agree for the scanner but unlike the scanner, there isn't really a default locator, and even though there's no harm in it I don't think we should have multiple open Maxmind DB objects

peterjan (dismissed):
The import cycles are becoming a nuisance and I feel that more and more we're working around it by using the `_test` suffix, which is now causing us to relocate tests to avoid exporting  helpers and so on. That's probably fine for now since there's more important things to tackle at the moment, but we'll have to address it at some point.

I did a quick scan to see where most of the friction comes from and I think all unit tests that require the `TestStore` should never require a `_test` suffix. We could do that if we would move the store to a subpackage in `testutils`, like `testdb` for instance. Another cycle comes from the DB constructor requiring default contract and hosts settings.

internal/testutils/mock/mock.go
ChrisSchinnerl, line 1: I'd move this file to `internal/testutils/locator.go` to avoid starting a new generic `mock` package.

hosts/utils_test.go
ChrisSchinnerl: This is mostly scanning related code. Fetching settings, resolving a host etc.
So I'd move that to `scan_test.go` and maybe the corresponding methods you are testing to `scan.go`.
To make sure we stick to the pattern of testing functionality implemented in `foo.go` in a file called `foo_test.go`.

ChrisSchinnerl (changes requested):

internal/testutils/mock/mock.go
chris124567: Cannot do this due to import cycle

peterjan (approved):

ChrisSchinnerl (approved):

PR #597, opened 2025-10-30
Remove pagination from `/stats/hosts` 

peterjan:
Stats endpoints should return all values at all times. This is consistent with our other stats endpoints. 
The benchmark is not great but it's not slow either, this is 10k hosts and an M1 so should be good enough.

```
cpu: Apple M1 Max
BenchmarkHostStats-10                 15          70805567 ns/op        13205928 B/op     383090 allocs/op
PASS
```

### DIFF
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -73,7 +73,7 @@ type (
 		UsabilitySettings(ctx context.Context) (hosts.UsabilitySettings, error)
 		UpdateUsabilitySettings(ctx context.Context, us hosts.UsabilitySettings) error
 
-		Stats(ctx context.Context, offset, limit int) ([]hosts.HostStats, error)
+		Stats(ctx context.Context) ([]hosts.HostStats, error)
 	}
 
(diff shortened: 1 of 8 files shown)


n8mgr:
Why though? The other stats endpoints don't return multiple results. That's like saying the `/wallet/events` endpoint shouldn't paginate because `/wallet/balance` doesn't.

peterjan:
> Why though? The other stats endpoints don't return multiple results. That's like saying the `/wallet/events` endpoint shouldn't paginate because `/wallet/balance` doesn't.

It's annoying to scrape a stats endpoint that uses an offset and a limit. We either remove pagination or allow a sentinel value for unlimited or a very high limit, currently our max limit is 500, or update the prometheus template in the deploy repo to call that stats endpoint at different offsets but I feel all of those are worse than just removing it.

We currently only pull in 100 host stats in Grafana and we never even noticed.

n8mgr:
Create an issue so we can discuss and prioritize then. This is not it and there are significantly higher priority items to tackle.

PR #600, opened 2025-10-30
Add slow host logging

peterjan:
Our migrations are slow and it looks like it's mainly due to slow downloads. We don't have any exact numbers but it looks like it takes us more than 2 mins on average to download a single slab, so it's actually _very_ slow and it's likely just stalling on a couple of slow hosts. 

We do have some logs and it appears it's the same handful of hosts that are slow each and every time and are clogging up the works. This PR is meant to give us insights into how many times we're "stalled" in a slab download, what portion of the download time we're sitting idle and doing nothing, whether or not it's the same slow hosts that ruin the fun for everybody. Which is kind of interesting because we removed sorting and now randomize download candidates. Also I think data distribution is not even but also not the worst. 

Our download code doesn't have overdrive, if the shard was downloaded successfully we don't release the semaphore. 
```go
defer func() {
    if err != nil {
        <-sema // only release next candidate on failure
    }
    wg.Done()
}()
```

### DIFF
--- a/slabs/downloads.go
+++ b/slabs/downloads.go
@@ -67,7 +67,7 @@ func (dc *downloadCandidates) next() (hosts.Host, bool) {
 // downloadShards downloads at least the minimum number of shards required to
 // recover the slab.
 func (m *SlabManager) downloadShards(ctx context.Context, slab Slab, allHosts []hosts.Host, pool *connPool, log *zap.Logger) ([][]byte, error) {
-	start := time.Now()
+	dlStart := time.Now()
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
@@ -82,12 +82,10 @@ func (m *SlabManager) downloadShards(ctx context.Context, slab Slab, allHosts []
 	var wg sync.WaitGroup
 	sema := make(chan struct{}, slab.MinShards)
 
-	var stalledCnt uint8
-	var stalledDur time.Duration
-	const stallThreshold = 15 * time.Second
-
+	var waitDur time.Duration
+	const slowHostsThreshold = 10 * time.Second
 	var slowMu sync.Mutex
-	var slowHosts []types.PublicKey
+	slowHosts := make(map[types.PublicKey]time.Duration)
 
 outer:
 	for {
@@ -96,12 +94,10 @@ outer:
 		case <-ctx.Done():
 		case sema <- struct{}{}:
 		}
+		waitDur += time.Since(waiting)
 
-		if time.Since(waiting) > stallThreshold {
-			stalledCnt++
-			stalledDur += time.Since(waiting)
-		}
 		if ctx.Err() != nil {
+			// context cancelled, either due to timeout or enough shards downloaded
 			break outer
 		}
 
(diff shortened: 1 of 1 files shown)


slabs/downloads.go
```diff
@@ -85,22 +155,40 @@ func (m *SlabManager) downloadShards(ctx context.Context, slab Slab, allHosts []
...
 		case <-ctx.Done():
+			ctxDone = true
+		case sema <- struct{}{}:
+		}
+
+		if time.Since(waiting) > stallThreshold {
+			stalledCnt++
+			stalledDur += time.Since(waiting)
+		}
+
```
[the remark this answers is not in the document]
peterjan: Probably fine since we check `stallThreshold`

slabs/downloads.go
```diff
@@ -85,22 +155,40 @@ func (m *SlabManager) downloadShards(ctx context.Context, slab Slab, allHosts []
...
+		if ctxDone {
```
ChrisSchinnerl: I'd do `ctx.Err() != nil` here to get rid of the bool. Still a bit odd but I don't see a good way around it if you really want to log the time we spent waiting.

slabs/downloads.go
```diff
@@ -138,17 +229,38 @@ outer:
 	}
 
 	wg.Wait()
+
+	if time.Since(dlStart) > m.shardTimeout {
+		var slowHosts []types.PublicKey
+		for _, hk := range m.dlMetrics.topN(10, float64(stallThreshold.Milliseconds())) {
```
ChrisSchinnerl: `dlMetrics` seems to add a lot of complexity and pull in a new dependency to add 1 piece of information to this log line that is probably not even the most useful here. 

Why not collect all hosts that timed out during upload and add those as the `slowHosts`? You would then be able to tail the logs over any timeframe, grep and sum up the lines of each host to see if it's the same couple of hosts that block downloads all the time. 

Instead of adding 150 lines of stats code to end up with a list of hosts that might not even be relevant to this particular slab download.

peterjan: I thought it would be easier to track download performance across all hosts, and for every slow download fetch the top 10 slowest hosts and see whether it's the same couple of hosts that cause all downloads to slow down.

The 150 lines of stat code could prove useful should we need more info on a per host basis. I'll rewrite it though and grep through the logs.

slabs/downloads.go
peterjan: Ah good idea. Yeah I think the total time spent waiting would be interesting because we do have start off with `MinShards` of parallel requests.

slabs/downloads.go
```diff
@@ -85,15 +86,30 @@ func (m *SlabManager) downloadShards(ctx context.Context, slab Slab, allHosts []
...
+
 outer:
 	for {
+		waiting := time.Now()
 		select {
 		case <-ctx.Done():
+		case sema <- struct{}{}:
+		}
+
+		if time.Since(waiting) > stallThreshold {
```
ChrisSchinnerl: I'd always add to `stalledDur` to get a more accurate representation of the total time spent blocked here. The 15s threshold might make this quite odd since 10 seconds would also be very slow but contribute 0s to the total.

slabs/downloads.go
```diff
@@ -138,6 +161,17 @@ outer:
 	}
 
 	wg.Wait()
+
+	if len(slowHosts) > 0 {
+		m.log.Warn("slow download",
+			zap.Uint8("stallCount", stalledCnt),
+			zap.Duration("stallDuration", stalledDur),
+			zap.Duration("totalDuration", time.Since(start)),
+			zap.Stringers("slowestHosts", slowHosts),
```
ChrisSchinnerl: I'd also log the threshold here as well for reference.

slabs/downloads.go
```diff
@@ -119,8 +135,15 @@ outer:
 				m.markSectorLost(ctx, host, slab.Sectors[sectorIdx].Root, logger)
 				return
 			} else if err != nil {
+				elapsed := time.Since(start)
+				if elapsed >= m.shardTimeout {
```
ChrisSchinnerl: I'd rename `stalledThreshold` to `slowHostThreshold` and use that here instead of the `shardTimeout`. Then log both the slow hosts as well as how long it took them to finish.
Maybe also lower the threshold a bit to 10s or so.

ChrisSchinnerl (changes requested):

slabs/downloads.go
peterjan: Did that already 👍

peterjan:
Going to close this out since we have open issues to add host selection. The intention of this PR was to eventually improve download performance by either doing host selection and/or enabling overdrive.

PR #603, opened 2025-10-31
Only fund active accounts

### LINKED ISSUE #601: Only fund accounts that are active
Update `HostAccountsForFunding` to only return accounts we consider active.

chris124567:
Close #601

With changes:
```
goos: linux
goarch: amd64
BenchmarkHostAccountsForFunding/10000_accounts-4                  90      14735955 ns/op
BenchmarkHostAccountsForFunding/100000_accounts-4                 22      57609661 ns/op
BenchmarkHostAccountsForFunding/1000000_accounts-4               100      10689208 ns/op
PASS
ok      go.sia.tech/indexd/persist/postgres 430.014s
```

`master`:
```
goos: linux
goarch: amd64
BenchmarkHostAccountsForFunding/10000_accounts-4                 147       8004118 ns/op
BenchmarkHostAccountsForFunding/100000_accounts-4                 25      44758543 ns/op
BenchmarkHostAccountsForFunding/1000000_accounts-4                46      25817512 ns/op
PASS
ok      go.sia.tech/indexd/persist/postgres 368.942s
```

### DIFF
--- a/persist/postgres/accounts.go
+++ b/persist/postgres/accounts.go
@@ -141,8 +141,8 @@ func (s *Store) UpdateAccount(ctx context.Context, oldAK, newAK types.PublicKey)
 	})
 }
 
-// HostAccountsForFunding returns up to limit accounts for the given host key
-// that are due for funding.
+// HostAccountsForFunding returns up to `limit` active (after the `threshold`
+// time) accounts for the given host key that are due for funding.
 func (s *Store) HostAccountsForFunding(ctx context.Context, hk types.PublicKey, threshold time.Time, limit int) ([]accounts.HostAccount, error) {
 	if limit < 0 {
 		return nil, errors.New("limit can not be negative")
--- a/persist/postgres/accounts.go
+++ b/persist/postgres/accounts.go
@@ -143,7 +143,7 @@ func (s *Store) UpdateAccount(ctx context.Context, oldAK, newAK types.PublicKey)
 
 // HostAccountsForFunding returns up to limit accounts for the given host key
 // that are due for funding.
-func (s *Store) HostAccountsForFunding(ctx context.Context, hk types.PublicKey, limit int) ([]accounts.HostAccount, error) {
+func (s *Store) HostAccountsForFunding(ctx context.Context, hk types.PublicKey, threshold time.Time, limit int) ([]accounts.HostAccount, error) {
 	if limit < 0 {
 		return nil, errors.New("limit can not be negative")
 	} else if limit == 0 {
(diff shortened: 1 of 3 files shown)


persist/postgres/accounts.go
```diff
@@ -143,7 +143,7 @@ func (s *Store) UpdateAccount(ctx context.Context, oldAK, newAK types.PublicKey)
 
 // HostAccountsForFunding returns up to limit accounts for the given host key
 // that are due for funding.
-func (s *Store) HostAccountsForFunding(ctx context.Context, hk types.PublicKey, limit int) ([]accounts.HostAccount, error) {
+func (s *Store) HostAccountsForFunding(ctx context.Context, hk types.PublicKey, threshold time.Time, limit int) ([]accounts.HostAccount, error) {
```
peterjan, line 146: `nit` could update the docstring and mention it returns accounts that have been used since the `threshold` time

persist/postgres/accounts.go
```diff
@@ -325,15 +325,15 @@ func addAccount(ctx context.Context, tx *txn, account types.PublicKey, serviceAc
 	return nil
 }
 
-func newHostAccountsForFunding(ctx context.Context, tx *txn, hk types.PublicKey, hostID int64, limit int) ([]accounts.HostAccount, error) {
+func newHostAccountsForFunding(ctx context.Context, tx *txn, hk types.PublicKey, hostID int64, threshold time.Time, limit int) ([]accounts.HostAccount, error) {
```
peterjan, line 328: I would ignore `last_used` in `newHostAccountsForFunding` as it's just unnecessary complexity. It's fine to fund every account at least once and it's unlikely this condition would ever filter an account out anyway since the threshold is 1 week.

peterjan (changes requested):

persist/postgres/accounts.go
ChrisSchinnerl: This is responsible for creating/funding accounts with new hosts right? In that case we definitely want to take this into account here. The number of accounts on the indexer is only going to go up and the number of hosts too as the network grows. A single comparison here is a small price to pay for potentially tens of thousands of dollars leaking into inactive accounts.

ChrisSchinnerl (dismissed):

persist/postgres/accounts.go
peterjan: Both sub queries are responsible for catching new hosts, `newHostAccountsForFunding` is responsible for funding new accounts with all existing hosts. Since we trigger funding when we add an account, and we initialise `last_used` to `NOW()`, this condition should never filter out an account. I see the small price to pay argument, so fine with keeping it in but in practice I don't believe it will affect the resulting set of accounts.

ChrisSchinnerl: `last_used` being initialized to `NOW()` doesn't prevent a new host account from being returned by `newHostAccountsForFunding` weeks after the account was created. The host might not have existed back then and the account might be inactive by then.

peterjan (dismissed):

ChrisSchinnerl (approved):

n8mgr (approved):

PR #614, opened 2025-10-31
Fix migration logic

n8mgr:
The current `toMigrate` logic will cause migration on sectors that are unpinned, but otherwise healthy. `goodContract = false` whenever contract id is unset. This doesn't necessarily mean the sector needs migrated, it might just need pinned.

### DIFF
--- a/slabs/migrations.go
+++ b/slabs/migrations.go
@@ -197,17 +197,15 @@ func sectorsToMigrate(slab Slab, allHosts []hosts.Host, goodContracts []contract
 	// - the sector is stored on a bad contract
 	var toMigrate []int
 	for i, sector := range slab.Sectors {
-		if sector.HostKey != nil {
-			host, ok := hostsMap[*sector.HostKey]
-			if !ok {
-				// sector is on a bad host
-				toMigrate = append(toMigrate, i)
-				continue
-			}
-
-			set.Add(host.Info())
-		} else {
+		if sector.HostKey == nil {
 			// sector is lost
+			toMigrate = append(toMigrate, i)
+			continue
+		}
+
+		host, ok := hostsMap[*sector.HostKey]
+		if !ok {
+			// sector is on a bad host
 			toMigrate = append(toMigrate, i)
 			continue
 		}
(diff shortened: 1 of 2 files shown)


slabs/migrations.go
```diff
@@ -196,20 +196,25 @@ func sectorsToMigrate(slab Slab, allHosts []hosts.Host, goodContracts []contract
...
 
-		// add the host to the spaced set to ensure we don't use hosts that are
-		// too close to existing hosts
-		if host, ok := hostsMap[*sector.HostKey]; ok {
-			set.Add(host.Info())
+		if sector.HostKey != nil {
+			if host, ok := hostsMap[*sector.HostKey]; ok {
+				set.Add(host.Info())
+			}
 		}
```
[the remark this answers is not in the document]
n8mgr: > sectors with only a HostKey but no ContractID (i.e., sectors that are unpinned but stored on a host) will not be migrated

That should be the desired behavior.

peterjan (changes requested):
Nice catch, this was unit tested so I think we should add a case for this particular scenario in  `TestSectorsToMigrate`

peterjan (dismissed):

slabs/migrations.go
```diff
@@ -192,24 +192,33 @@ func sectorsToMigrate(slab Slab, allHosts []hosts.Host, goodContracts []contract
...
-		if isLost || !goodContract {
+		if sector.HostKey != nil {
+			host, ok := hostsMap[*sector.HostKey]
+			if !ok {
+				// sector is on a bad host
+				toMigrate = append(toMigrate, i)
+				continue
+			}
+
+			set.Add(host.Info())
```
ChrisSchinnerl: I think this is fine but I'm not 100% sure. At this point the host is good but the shard might still be in a bad contract. In that case the shard gets migrated to another host in which case we might not want to add the host to `set` here and instead do it at the end (after deleting from the good contract map) once we are sure the shard will not get migrated.

n8mgr: We should also delete from `hostsMap` and check whether the host is close enough to others and migrate if true, no?

ChrisSchinnerl (dismissed):

Issue #612, opened 2025-10-31
Investigate host queries


peterjan:
[This comment](https://github.com/SiaFoundation/indexd/pull/592#discussion_r2480878811) made me peek into our host queries and I found numerous inconsistencies and bugs. We recently merged [this PR](https://github.com/SiaFoundation/indexd/pull/518) that did something similar for all of our contracts queries.

Quick list of issues:
- [] redundant checks on blocked hosts
- [] redundant LEFT JOIN on blocklist
- [] redundant database setup in benchmarks taking into account the blocklist
- [] numerous active contracts checks that don't properly check for both `state`, `good` and `renewed_to`
- [] couple of missing indices (one for fetching hosts for pruning for instance)
- [x] main concern is `HostsWithUnpinnableSectors` which I believe is completely broken because of the state check

peterjan:
Note that fixing `numerous active contracts checks that don't properly check for both state, good and renewed_to` should get rid of the following error

```
{"level":"debug","ts":"2025-11-03T15:18:18Z","logger":"contracts.maintenance","caller":"contracts/maintenance.go:48","msg":"no contracts for funding","hostKey":"ed25519:ce1ad06650d66a7a7552d79ef812558d2881b15f75317eb78d9c296c9699c311"}
{"level":"debug","ts":"2025-11-03T15:18:18Z","logger":"contracts.maintenance","caller":"contracts/maintenance.go:48","msg":"no contracts for funding","hostKey":"ed25519:0eb6fb00c40cc94f72f8ee59a49665c3d4d202bd18f6feb3d41bb912da4e7bac"}
{"level":"debug","ts":"2025-11-03T15:18:18Z","logger":"contracts.maintenance","caller":"contracts/maintenance.go:48","msg":"no contracts for funding","hostKey":"ed25519:3f8ae80401e0e34d06d69c1f10899aa66fb2bcc0e8111fba127eb2b02c9fcd22"}
{"level":"debug","ts":"2025-11-03T15:18:18Z","logger":"contracts.maintenance","caller":"contracts/maintenance.go:48","msg":"no contracts for funding","hostKey":"ed25519:30fb5192af1a56dadc93715cfe120e94d26fa500c7bc195269b2f7141a2afae1"}
```

Issue #598, opened 2025-10-30
Figure out why blocked hosts keep getting blocked


ChrisSchinnerl:
@mike76-dev  reported that it seems like one of the hosts that got blocked keeps getting blocked due to uptime in every maintenance iteration.

```
2025-10-29T23:02:15+01:00	WARN	contracts.blockhosts	blocking unusable host	{"usability": ["Uptime"]}
```

However, that host should not even be returned for blocking since we filter out hosts that are already blocked. We should figure out if our filters are failing or if we aren't actually blocking the host properly.

peterjan:
I looked but it doesn't seem there's an issue for this bug report (?). It would be nice if we would get more of the log, the logging there was also broken. I checked `zeus` and it's not happening there. He could grep for `failed to block host`. In the mean time I'll run a node on testnet to try and reproduce it.

peterjan:
In the contractor we fetch all active contracts, but we do so by passing in `WithRevisable` true ,that does not take into account the `good` status which is what gets set to false when we block a host. We "evaluate" every host we have an active contract with so that would print that warning over and over again.

ChrisSchinnerl:
In that case we just need to fix the logging

PR #615, opened 2025-10-31
Ignore `bad` contracts when forming contracts

peterjan:
🧪  is very tricky since it's an option we forgot to pass, the host does get properly ignored the only side-effect here is that it messes up the logging, giving the impression we're re-evaluating blocked hosts to form contracts with. 

https://github.com/SiaFoundation/indexd/issues/598 

PS I already opened [an issue](https://github.com/SiaFoundation/indexd/issues/612) to discuss the matter, but this ties in with `HostsWithUnpinnableSectors`, which could potentially return a blocked host and cause us to forcefully form a contract with it. We should prioritise fixing that asap.

### DIFF
--- a/contracts/form.go
+++ b/contracts/form.go
@@ -96,7 +96,7 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period 
 	var activeContracts []Contract
 	const batchSize = 50
 	for offset := 0; ; offset += batchSize {
-		batch, err := cm.store.Contracts(ctx, offset, batchSize, WithRevisable(true))
+		batch, err := cm.store.Contracts(ctx, offset, batchSize, WithGood(true), WithRevisable(true))
 		if err != nil {
 			return fmt.Errorf("failed to fetch active contracts: %w", err)
 		}


chris124567 (approved):

ChrisSchinnerl (approved):

contracts/form.go
```diff
@@ -96,7 +96,7 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
 	var activeContracts []Contract
 	const batchSize = 50
 	for offset := 0; ; offset += batchSize {
-		batch, err := cm.store.Contracts(ctx, offset, batchSize, WithRevisable(true))
+		batch, err := cm.store.Contracts(ctx, offset, batchSize, WithGood(true), WithRevisable(true))
```
ChrisSchinnerl, line 99: Looking at this again we might not be able to do this. Seems like we called this `activeContracts` for a reason. e.g. we form contracts with hosts that have unpinnable sectors when they have no active contracts. Not considering bad contracts here means a host will also have a contract formed with it if all of its contracts are exclusively bad. It's a very subtle difference but definitely changes more behavior than one line of logging.

ChrisSchinnerl (changes requested):

contracts/form.go
peterjan: Wait hold on, we forcefully form contracts with hosts that have unpinnable sectors, but we fetch those separately using `HostsWithUnpinnableSectors` and then set `forceFormation` for all the hosts that are being returned. (This is broken by the way, see https://github.com/SiaFoundation/indexd/issues/612) 

How does not returning bbad contracts here mean a host will have a contract formed if all contracts are bad.  We only really use `activeContracts` for the `full` check and to update `wanted`, right?

peterjan: A host's contracts are only bad if the host is blocked and we never form a contract with a host that is not good, regardless of whether we `force` the contract formation. I think it's fine, might be missing the subtle difference but I don't see the change in behaviour since we perform a check on a scanned host right before forming anyways.

ChrisSchinnerl: You are right. But also that means (since we always check the host before forming a contract anyway) that there is no bug right? Even when we force a formation it won't happen with a bad host. I'd rather avoid changing core logic that we spent a lot of time reasoning about when we wrote it if we don't think we need to.

peterjan: It's relatively easy to reason about, it might influence `forceFormation`, it would not influence `wanted`. Since we check the host before forming a contract, it never influences actual contract formation. I don't really mind if you feel strongly, not sure if I follow though, it's an oversight and we can safely change this.

ChrisSchinnerl: Probably fine to close this now since Nate is "refactoring the thing"

peterjan: I would merge this and let him fix the conflict. 

edit:
Especially because I think this bug is actually still present in https://github.com/SiaFoundation/indexd/pull/619 and made worse because it now directly influences `needsFormation`. We only append to `needsFormation` if `hostsWithoutContracts` is true, a `bad` contract might've removed it from that list. 

We also no longer check the scanned host in `formContract` (nor in `refreshContract).

n8mgr (changes requested):
Superseded by #619

contracts/form.go
n8mgr: What would be the point of me fixing a merge conflict on logic that no longer exists?

Issue #611, opened 2025-10-31
Track the total pinned data per app key and limit the data per connect key


ChrisSchinnerl:
We already have a limit on the app key and on accounts. We enforce the per-account limit but we don't make sure that the sum of all account data is below the app connect key limit.

PR #616, opened 2025-10-31
Add a reference to the accounts table linking them to the `app_connect_keys` and return the key as part of the account API response

### LINKED ISSUE #605: Add a reference to the `accounts` table linking them to the `app_connect_keys` and return the key as part of the account API response

chris124567:
Close #605

Sadly we haven't been collecting this information so we cannot really do a proper migration here.

### DIFF
--- a/accounts/account.go
+++ b/accounts/account.go
@@ -66,7 +66,7 @@ type (
 	// Account represents an account in the indexer.
 	Account struct {
 		AccountKey     proto.Account `json:"accountKey"`
-		ConnectKey     *string       `json:"connectKey"`
+		ConnectKey     *string       `json:"connectKey,omitempty"`
 		ServiceAccount bool          `json:"serviceAccount"`
 		MaxPinnedData  uint64        `json:"maxPinnedData"`
 		PinnedData     uint64        `json:"pinnedData"`
--- a/accounts/account.go
+++ b/accounts/account.go
@@ -66,6 +66,7 @@ type (
 	// Account represents an account in the indexer.
 	Account struct {
 		AccountKey     proto.Account `json:"accountKey"`
+		ConnectKey     *string       `json:"connectKey"`
 		ServiceAccount bool          `json:"serviceAccount"`
 		MaxPinnedData  uint64        `json:"maxPinnedData"`
 		PinnedData     uint64        `json:"pinnedData"`
--- a/persist/postgres/accounts.go
+++ b/persist/postgres/accounts.go
@@ -34,8 +34,10 @@ func (s *Store) Accounts(ctx context.Context, offset, limit int, opts ...account
 
 	if err := s.transaction(ctx, func(ctx context.Context, tx *txn) (err error) {
 		rows, err := tx.Query(ctx, `
-			SELECT public_key, connect_key, service_account, max_pinned_data, pinned_data, description, logo_url, service_url, last_used
-			FROM accounts
+			SELECT a.public_key, ak.app_key, a.service_account, a.max_pinned_data, a.pinned_data, a.description, a.logo_url, a.service_url, a.last_used
+			FROM accounts a
+			LEFT JOIN app_connect_keys ak
+			ON ak.id = a.connect_key_id
 			WHERE ($1::boolean IS NULL OR service_account = $1::boolean)
 			LIMIT $2 OFFSET $3
 		`, queryOpts.ServiceAccount, limit, offset)
@@ -64,7 +66,11 @@ func (s *Store) Account(ctx context.Context, ak types.PublicKey) (accounts.Accou
 	var account accounts.Account
 	account.AccountKey = proto.Account(ak) // no need to fetch key
 	err := s.transaction(ctx, func(ctx context.Context, tx *txn) (err error) {
-		account, err = scanAccount(tx.QueryRow(ctx, `SELECT public_key, connect_key, service_account, max_pinned_data, pinned_data, description, logo_url, service_url, last_used FROM accounts WHERE public_key = $1`, sqlPublicKey(ak)))
+		account, err = scanAccount(tx.QueryRow(ctx, `SELECT a.public_key, ak.app_key, a.service_account, a.max_pinned_data, a.pinned_data, a.description, a.logo_url, a.service_url, a.last_used
+FROM accounts a
+LEFT JOIN app_connect_keys ak
+ON ak.id = a.connect_key_id
+WHERE public_key = $1`, sqlPublicKey(ak)))
 		return err
 	})
 	return account, err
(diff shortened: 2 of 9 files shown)


persist/postgres/migrations.go
```diff
@@ -451,4 +451,13 @@ FROM objects o;
 		_, err := tx.Exec(ctx, `ALTER TABLE global_settings ADD COLUMN wallet_hash BYTEA CHECK(wallet_hash IS NULL OR LENGTH(wallet_hash) = 32);`)
 		return err
 	},
+	// add connect_key to accounts
+	func(ctx context.Context, tx *txn, _ *zap.Logger) error {
+		_, err := tx.Exec(ctx, `
+-- add connect key column and create index
+ALTER TABLE accounts ADD COLUMN connect_key TEXT REFERENCES app_connect_keys(app_key);
```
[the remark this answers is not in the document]
chris124567: we check that key is not in use in `DeleteAppConnectKey`

persist/postgres/init.sql
```diff
@@ -1,6 +1,19 @@
...
+    created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    last_used TIMESTAMP WITH TIME ZONE,
+    max_pinned_data BIGINT NOT NULL CHECK (max_pinned_data >= 0)
+);
+
 CREATE TABLE accounts (
     id SERIAL PRIMARY KEY,
     public_key BYTEA UNIQUE NOT NULL CHECK (LENGTH(public_key) = 32),
+    connect_key TEXT REFERENCES app_connect_keys(app_key),
```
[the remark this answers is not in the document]
chris124567: we check that key is not in use by any accounts when deleting

persist/postgres/init.sql
```diff
@@ -1,6 +1,19 @@
+CREATE TABLE app_connect_keys (
+    app_key TEXT PRIMARY KEY,
```
n8mgr: Lets go ahead and add a numeric ID here to match the other FK.

persist/postgres/init.sql
n8mgr: Go ahead and make this `NOT NULL`

persist/postgres/migrations.go
```diff
@@ -451,4 +451,13 @@ FROM objects o;
 		_, err := tx.Exec(ctx, `ALTER TABLE global_settings ADD COLUMN wallet_hash BYTEA CHECK(wallet_hash IS NULL OR LENGTH(wallet_hash) = 32);`)
 		return err
 	},
+	// add connect_key to accounts
+	func(ctx context.Context, tx *txn, _ *zap.Logger) error {
+		_, err := tx.Exec(ctx, `
```
n8mgr, line 467: For migrations:

**Migrate app_connect_keys:**
1. Rename the existing table with `_tmp`
2. Create the new table with the new `id PRIMARY KEY`
3. Insert all records from the old table
4. Delete the old table

**Migrate accounts:**
1. `ALTER TABLE accounts ADD COLUMN connect_key_id INT REFERENCES app_connect_keys (id)`
2. pick a random one, who cares `UPDATE accounts SET connect_key_id = (SELECT id FROM app_connect_keys LIMIT 1);` 
3. Add not null: `ALTER TABLE accounts ALTER COLUMN connect_key_id SET NOT NULL;`

n8mgr (changes requested):

persist/postgres/migrations.go
chris124567: What connect key should a newly created service account be under?  Should there be a default one?

chris124567: I also didn't like the idea of picking a random key because we are eventually going to be imposing [limits](https://github.com/SiaFoundation/indexd/issues/611) per connect key so it could break things.  Though it's just us testing indexd at this point so it's not the end of the world

n8mgr: > What connect key should a newly created service account be under? Should there be a default one?

What a pain. That's a very good argument for service accounts not being stored in the same table as user accounts 😓

n8mgr: Guess we have to leave it as nullable and validate it in the API for now. Oh well.

persist/postgres/accounts.go
```diff
@@ -34,8 +34,10 @@ func (s *Store) Accounts(ctx context.Context, offset, limit int, opts ...account
 
 	if err := s.transaction(ctx, func(ctx context.Context, tx *txn) (err error) {
 		rows, err := tx.Query(ctx, `
-			SELECT public_key, service_account, max_pinned_data, pinned_data, description, logo_url, service_url, last_used
-			FROM accounts
+			SELECT a.public_key, ak.app_key, a.service_account, a.max_pinned_data, a.pinned_data, a.description, a.logo_url, a.service_url, a.last_used
+			FROM accounts a
+			LEFT JOIN app_connect_keys ak
+			ON ak.id = a.connect_key_id
```
peterjan: `nit` put the `ON` on the same line as `LEFT JOIN`

persist/postgres/accounts.go
```diff
@@ -306,14 +312,22 @@ func (s *Store) ServiceAccountBalance(ctx context.Context, hostKey types.PublicK
...
 		MaxPinnedData: math.MaxInt64, // no limit by default
 	}
 	for _, opt := range opts {
 		opt(&aao)
 	}
-	res, err := tx.Exec(ctx, `INSERT INTO accounts (public_key, service_account, max_pinned_data, description, logo_url, service_url) VALUES ($1, $2, $3, $4, $5, $6) ON CONFLICT DO NOTHING`, sqlPublicKey(account), serviceAccount, aao.MaxPinnedData, meta.Description, meta.LogoURL, meta.ServiceURL)
+
+	var connectKeyID sql.NullInt64
+	if connectKey != nil {
+		if err := tx.QueryRow(ctx, `SELECT id FROM app_connect_keys WHERE app_key = $1`, connectKey).Scan(&connectKeyID); err != nil {
```
peterjan: return `ErrKeyNotFound` on `sql.ErrNoRows`

persist/postgres/apps.go
```diff
@@ -123,6 +123,15 @@ func (s *Store) AppConnectKeys(ctx context.Context, offset, limit int) (keys []a
 // DeleteAppConnectKey deletes an application connection key from the database.
 func (s *Store) DeleteAppConnectKey(ctx context.Context, connectKey string) error {
 	return s.transaction(ctx, func(ctx context.Context, tx *txn) error {
+		var count int64
+		if err := tx.QueryRow(ctx, `SELECT COUNT(*) FROM accounts WHERE connect_key_id = (SELECT id FROM app_connect_keys WHERE app_key = $1)`, connectKey).Scan(&count); errors.Is(err, sql.ErrNoRows) {
```
peterjan: You'll never get `sql.ErrNoRows` on `SELECT COUNT()` and we're not checking `res.RowsAffected` in the subsequent delete so we would return `nil`. I would rewrite it to do `SELECT EXISTS` since we're not interested in the amount but whether one exists yes or no. I would also check rows affected on the delete for good measure.

peterjan: Probably want to cover this with a little test in `apps_test.go` since it got through unnoticed

peterjan (changes requested):

persist/postgres/accounts.go
```diff
@@ -383,6 +397,6 @@ LIMIT $2`, hostID, limit)
 }
 
 func scanAccount(s scanner) (account accounts.Account, err error) {
-	err = s.Scan((*sqlPublicKey)(&account.AccountKey), &account.ServiceAccount, &account.MaxPinnedData, &account.PinnedData, &account.Description, &account.LogoURL, &account.ServiceURL, &account.LastUsed)
+	err = s.Scan((*sqlPublicKey)(&account.AccountKey), &account.ConnectKey, &account.ServiceAccount, &account.MaxPinnedData, &account.PinnedData, &account.Description, &account.LogoURL, &account.ServiceURL, &account.LastUsed)
```
peterjan: Did you test this for both an account where connect key is set and unset? I think in the `NULL` case it might fail to scan because you're passing a pointer to `account.ConnectKey`. Probably want to use `sql.NullString` here in any case.

chris124567: In `TestAppConnectKeys` it is set and we check that it is set.  In `TestAccounts` it is unset in all those accounts and we are able to retrieve them.  But I will use sql.NullString anyways.

accounts/account.go
ChrisSchinnerl, line 1: Looking good overall but missing an update to the API docs for the new field

accounts/account.go
```diff
@@ -66,6 +66,7 @@ type (
 	// Account represents an account in the indexer.
 	Account struct {
 		AccountKey     proto.Account `json:"accountKey"`
+		ConnectKey     *string       `json:"connectKey"`
```
ChrisSchinnerl: (nit) could make it `omitempty` and mark it optional in the api docs mentioning it's only set for non-service accounts.

ChrisSchinnerl (changes requested):

n8mgr (approved):

ChrisSchinnerl (approved):

PR #619, opened 2025-10-31
Refactor the thing

### LINKED ISSUE #444: Contract renewal failed due to max collateral
I ran into the following on zeus. We should investigate why we don't respect the host's max collateral

```
indexd-1  | DEBUG	contracts.maintenance.refresh	failed to renew	{"hostKey": "ed25519:ec3331abf7d0a0bc4c0ad073644cdc5eca19c445e2939c2e7ab19ba82230c9db", "contractID": "6239050eec4933be166869ef565c8fd789ed9ce5da481a587c4983e340b04ddf", "remainingAllowance": "226.73475075661740982140928 SC", "initialAllowance": "280.3754650826326098837504 SC", "usedCollateral": "3.715902042836050319058665472 KS", "totalCollateral": "3.941027048622493626733166592 KS", "outOfFunds": false, "outOfCollateral": true, "error": "failed to refresh contract: failed to read host inputs response: required collateral 4.166152054408936934407667712 KS exceeds max collateral 4 KS (3) (3)"}
```

### LINKED ISSUE #506: Fix refreshes caused by contracts being out of collateral
`OutOfCollateral` is currently only looking at `TotalCollateral` and `UsedCollateral` to determine whether a contract is out of collateral.
The problem with that is that looking at the percentage of collateral remaining based on the `TotalCollateral` is not a good indicator for whether the contract is out of collateral or not.

Example:

A contract has 1000SC worth of risked collateral because at some point it held a lot of data that was pruned.
After pruning the contract is empty and `contractFunding` estimates that 10SC is plenty of collateral for that contract.
After refreshing the contract, the contract contains 1010SC `TotalCollateral`, 1000SC `UsedCollateral` and 10SC `remaining` as per `OutOfCollateral`.

That means at about 0.9% of remaining collateral, `OutOfCollateral` returns `true` repeatedly every time we refresh the contract and check again.

Solution:

- [ ] Update `OutOfCollateral` to only return `true` if `size == capacity` since we don't need more collateral if we have prepaid storage.
- [ ] Update `GoodForUpload` to ignore the allowance and collateral checks if `size < capacity` for the same reason
- [ ] Make sure that the pinning code will only pin a partial batch if capacity > size and the contract is otherwise out of funds

### LINKED ISSUE #589: Not getting requested number of contracts
I'm on Zen with the target contract count of 10. On the Hosts page, there are exactly 10 hosts that are usable. Yet I seem to be capped at 7 contracts, and nothing happens beyond that.

### LINKED ISSUE #598: Figure out why blocked hosts keep getting blocked

n8mgr:
Takes into account max collateral when renewing or refreshing contracts. Unifies the contract formation and refresh logic to reduce duplication and improve clarity. Changes the `GoodForUpload`, `OutOfCollateral`, and `NeedsRefresh` helpers to be more precise and removes any arbitrary values.

The goal is that each host has at least one contract that is both `GoodForUpload` and `GoodForAccountFunding` at all times. If the host does not, it will either refresh the smallest contract or form a new one if refreshing not possible (either hitting the max collateral or max storage size).

Fixes #506  - by making the contract helpers more accurate and making the logic consistent with the usage of the contracts instead of using arbitrary values.
Fixes #598 - contracts with unusable hosts are filtered out and not evaluated. 
Fixes #444 - the collateral parameters are now capped and consistent with the host logic
Fixes #589 - fixes edge-cases where new hosts may not be selected and adds additional debug logging for why a candidate was not considered.


peterjan (dismissed):

client/client.go
```diff
@@ -211,11 +211,20 @@ func (c *HostClient) ReadSector(ctx context.Context, hostPrices proto.HostPrices
...
 	if err := c.withRevision(ctx, params.ContractID, func(contract rhp.ContractRevision) (_ rhp.ContractRevision, _ proto.Usage, err error) {
-		if settings.ProtocolVersion.Cmp(rhp.ProtocolVersion500) >= 0 {
-			res, err = rhp.RPCRefreshContractPartialRollover(ctx, c.client, c.cm, c.signer, c.cm.TipState(), settings.Prices, contract.Revision, params)
-		} else {
-			res, err = rhp.RPCRefreshContractFullRollover(ctx, c.client, c.cm, c.signer, c.cm.TipState(), settings.Prices, contract.Revision, params)
+		if settings.ProtocolVersion.Cmp(rhp.ProtocolVersion500) < 0 {
+			return rhp.ContractRevision{}, proto.Usage{}, fmt.Errorf("host does not support contract refresh, protocol version %s < %s", settings.ProtocolVersion, rhp.ProtocolVersion500)
 		}
+
+		totalCollateral := contract.Revision.RiskedCollateral().Add(params.Collateral)
```
ChrisSchinnerl: In theory this is correct, but it seems like in `RPCRefreshContractRequest.Validate` we are doing `totalHostCollateral := existing.TotalCollateral.Add(req.Refresh.Collateral)` instead. 

So if we want to fix this issue right away we might want to apply the incorrect logic here for now until coreutils is fixed and then update this to the correct solution.

client/client.go
ChrisSchinnerl, line 1: Would be great to update `TestPerformContractRefreshes` and `TestPerformContractRenewals` to attempt a refresh/renewal beyond the `MaxCollateral` and verify that it got capped.

client/client.go
```diff
@@ -211,11 +211,20 @@ func (c *HostClient) ReadSector(ctx context.Context, hostPrices proto.HostPrices
...
+		if totalCollateral.Cmp(settings.MaxCollateral) > 0 {
+			capped, underflow := settings.MaxCollateral.SubWithUnderflow(contract.Revision.RiskedCollateral()) // cap to remaining collateral
+			if underflow {
+				capped = types.ZeroCurrency
+			}
+			params.Collateral = capped
```
ChrisSchinnerl, line 228: The problem with this is that unlike `contractFunding`, this doesn't take into account `minHostCollateral` which provided at least some sane lower limit.. So in theory `params.Collateral` might end up being `0` here. Sending us down a rapid refresh/renew loop since `OutOfCollateral` might return `true` again right after the refresh.

ChrisSchinnerl (changes requested):

client/client.go
n8mgr: In theory, we should detect that we can no longer add additional collateral and instead form a new contract instead of hitting this more than once.

contracts/form.go
```diff
@@ -162,162 +298,103 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
...
-		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
-			break
-		} else if errors.Is(err, hosts.ErrBadHost) {
-			continue // ignore bad host
-		} else if err != nil {
-			log.Error("failed to form contract", zap.Error(err))
-			continue
-		}
+			<-sema
+		}()
```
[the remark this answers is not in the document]
n8mgr: That is no longer true as of Go v1.23

contracts/form.go
```diff
@@ -162,162 +298,103 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
...
+		log := log.With(zap.Stringer("hostKey", contract.HostKey), zap.Stringer("contractID", contract.ID))
+		sema <- struct{}{}
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			if cm.refreshContract(ctx, contract, settings.Period, accountFundTarget, log) {
+				atomic.AddUint32(&refreshed, 1)
+			}
+			<-sema
+		}()
```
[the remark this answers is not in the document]
n8mgr: Ditto

contracts/form.go
```diff
@@ -162,162 +298,103 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
...
-
-			allowance, collateral := contractFunding(host.Settings, 0, target, minHostCollateral, period)
-			formationCtx, cancel := context.WithTimeout(ctx, 2*time.Minute) // note: broadcasting on the host-side can block for up to a minute by default
-			defer cancel()
-			hc, err := cm.dialer.DialHost(formationCtx, host.PublicKey, host.RHP4Addrs())
-			if err != nil {
-				return fmt.Errorf("failed to dial host: %w", err)
+			host, ok := hostMap[hostKey]
+			if !ok {
+				panic("host missing from host map") // should never happen
```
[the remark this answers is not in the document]
n8mgr: It cannot occur. The contract hosts are already checked to be part of the map.

api/app/app_test.go
```diff
@@ -99,7 +100,7 @@ func newAccount(t *testing.T, cluster *testutils.Cluster) (types.PrivateKey, acc
 func TestApplicationAPI(t *testing.T) {
 	ctx := t.Context()
 	// create cluster with three hosts
-	logger := testutils.NewLogger(false)
+	logger := zap.NewNop()
```
peterjan, line 103: Unrelated to the PR, and if you pass `zap.NewNop()` you might as well drop it from the options because the noop logger is the default.

contracts/contract.go
```diff
@@ -150,37 +150,57 @@ type (
...
-// OutOfCollateral indicates that a contract is running low on unallocated
-// collateral and should be refreshed.
-func (c Contract) OutOfCollateral() bool {
-	// TODO: change this to be based on host prices
-	if c.Capacity > c.Size {
-		return false
+// GoodForRefresh indicates whether a contract is likely to succeed refreshing.
+func (c Contract) GoodForRefresh(settings proto.HostSettings, fundTarget types.Currency, period uint64) error {
+	_, collateral := contractFunding(settings, c.Size, fundTarget, period)
+	totalCollateral := c.TotalCollateral.Add(collateral)
```
ChrisSchinnerl, line 196: You might want to mark this with a TODO as well saying it's incorrect but matches the current behavior. To make sure we don't forget updating it later.

contracts/form.go
```diff
@@ -87,70 +88,205 @@ func withinGougingLeeway(cost, limit types.Currency) bool {
...
+		batch, err := cm.store.Contracts(ctx, offset, batchSize, WithRevisable(true))
+		if err != nil {
+			return fmt.Errorf("failed to fetch active contracts: %w", err)
 		}
+		for _, contract := range batch {
+			host, ok := hostMap[contract.HostKey]
+			if !ok {
+				continue // host is not usable or is blocked
+			}
+			delete(hostsWithoutContracts, contract.HostKey) // host has at least one contract
```
peterjan, line 241: We might be deleting hosts from this list because of a `bad` contract in our batch. This influences `needsFormation`, e.g. we might not be forming contracts with hosts that need a contract formed because it has unpinnable sectors.

contracts/form.go
```diff
@@ -162,162 +298,103 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
...
-		return true
 	}
 
-	// determine how many contracts we need to form
-	for _, contract := range activeContracts {
-		log := log.With(zap.Stringer("contractID", contract.ID), zap.Stringer("hostKey", contract.HostKey))
-
-		// host checks
-		host, err := cm.hosts.Host(ctx, contract.HostKey)
+	if len(usableHostContracts) < int(settings.WantedContracts) && len(hostsWithoutContracts) > 0 {
```
peterjan: This condition feels off, did you mean to do an `OR` here? We delete from `hostsWithoutContracts` as soon as the host has a revisable contract, so we don't check all GFX'es or the `Good` status.

contracts/form.go
```diff
@@ -162,162 +298,103 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
...
+	}
 
-			contract := res.Contract
-			minerFee := res.FormationSet.Transactions[len(res.FormationSet.Transactions)-1].MinerFee
-			err = cm.store.AddFormedContract(ctx, hostKey, contract.ID, contract.Revision, host.Settings.Prices.ContractPrice, allowance, minerFee, res.Usage)
-			if err != nil {
-				return fmt.Errorf("failed to add formed contract: %w", err)
+	var formed, refreshed uint32
+	var wg sync.WaitGroup
+	sema := make(chan struct{}, 5) // concurrent formations and refreshes
```
peterjan: Concurrently forming and refreshing contracts is probably something we should tackle after we've taken care of the wallet maintenance issue.

slabs/migrations.go
```diff
@@ -68,14 +62,14 @@ func (m *SlabManager) migrateSlabs(ctx context.Context, slabIDs []SlabID, pool *
...
-func (m *SlabManager) migrateSlab(ctx context.Context, slabID SlabID, allHosts []hosts.Host, goodContracts []contracts.Contract, period uint64, pool *connPool, log *zap.Logger) error {
+func (m *SlabManager) migrateSlab(ctx context.Context, slabID SlabID, allHosts []hosts.Host, goodContracts []contracts.Contract, pool *connPool, log *zap.Logger) error {
 	start := time.Now()
 	slab, err := m.store.Slab(ctx, slabID)
 	if err != nil {
 		log.Error("failed to fetch slab", zap.Error(err))
 		return err
 	}
-	indices, uploadCandidates := sectorsToMigrate(slab, allHosts, goodContracts, period, m.minHostDistanceKm)
+	indices, uploadCandidates := sectorsToMigrate(slab, allHosts, goodContracts, m.chain.Tip().Height, m.minHostDistanceKm)
```
peterjan, line 72: It's a little unfortunate these checks have to happen in `sectorsToMigrate` if we already have a list of good contracts. E.g. we could have the store be aware of the chain manager and do something like `WithAppendable` instead.

contracts/maintenance.go
```diff
@@ -89,18 +89,13 @@ func (cm *ContractManager) performContractMaintenance(ctx context.Context, log *
...
-	}
-
 	// mark any contracts too close to their expiration height as bad
 	if err := cm.store.MarkUnrenewableContractsBad(ctx, blockHeight+settings.RenewWindow/2); err != nil {
 		return fmt.Errorf("failed to mark unrenewable contracts bad: %w", err)
 	}
 
 	// form new contracts until there are enough good contracts to use
-	if err := cm.performContractFormation(ctx, settings.Period, int64(settings.WantedContracts), log.Named("formation")); err != nil {
+	if err := cm.performContractFormation(ctx, settings, blockHeight, log.Named("maintenance")); err != nil {
```
peterjan, line 98: `nit` Reminder to update the name to `performContractFormationAndRefreshes` and perhaps name the logger `formation+refresh` or something

contracts/form.go
```diff
@@ -87,70 +88,205 @@ func withinGougingLeeway(cost, limit types.Currency) bool {
...
+	goodForAppend  error
+}
 
-	// fetch all revisable contracts
-	var activeContracts []Contract
-	const batchSize = 50
-	for offset := 0; ; offset += batchSize {
-		batch, err := cm.store.Contracts(ctx, offset, batchSize, WithRevisable(true))
+func (cm *ContractManager) formContract(ctx context.Context, host types.PublicKey, period uint64, fundTarget types.Currency, log *zap.Logger) bool {
+	err := cm.hosts.WithScannedHost(ctx, host, func(host hosts.Host) error {
```
peterjan, line 98: The purpose of `WithScannedHost` is to ensure we have the latest settings and the host is reachable, we should probably add a sanity check here to see if the host is good to form a contract with.

contracts/form.go
```diff
@@ -87,70 +88,205 @@ func withinGougingLeeway(cost, limit types.Currency) bool {
...
+	}
 
-		if full {
-			log.Debug("contract is full", zap.Bool("maxCollReached", maxCollReached),
-				zap.Bool("maxSizeReached", maxSizeReached),
-				zap.Uint64("size", contract.Size),
-				zap.Stringer("usedCollateral", contract.UsedCollateral),
-				zap.Stringer("totalCollateral", contract.TotalCollateral),
-				zap.Stringer("maxCollateral", s.MaxCollateral))
+	var needsFormation []types.PublicKey
```
ChrisSchinnerl: Shouldn't this be a map? Otherwise you are going to form a new contract for every bad contract a host already has rather than just one more.

contracts/form.go
```diff
@@ -162,162 +298,103 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
...
+		go func() {
+			defer wg.Done()
+			if cm.refreshContract(ctx, contract, settings.Period, accountFundTarget, log) {
+				atomic.AddUint32(&refreshed, 1)
+			}
+			<-sema
+		}()
+	}
+	wg.Wait()
+	log.Debug("formation finished", zap.Uint32("formedContracts", formed), zap.Uint32("refreshedContracts", refreshed))
```
peterjan, line 397: `nit` might be interesting to log the amount of contracts we're missing and the ones needing a refresh here

contracts/form.go
```diff
@@ -87,70 +88,205 @@ func withinGougingLeeway(cost, limit types.Currency) bool {
...
+					goodForFunding: fundingErr,
+					goodForAppend:  appendErr,
+				}
+				usableHostContracts[contract.HostKey] = cc
+			} else if cc.goodForAppend == nil && cc.goodForFunding == nil {
+				// host already has a contract good for both uploading and funding
+				continue
+			} else if contract.Size < cc.contract.Size {
+				// prefer smaller contracts since they are cheaper to refresh
+				cc = candidateContract{
```
peterjan: We should unit test the outcome of turning a batch of contracts into candidates.

contracts/form.go
```diff
@@ -87,70 +88,205 @@ func withinGougingLeeway(cost, limit types.Currency) bool {
...
-			log.Debug("contract is full", zap.Bool("maxCollReached", maxCollReached),
-				zap.Bool("maxSizeReached", maxSizeReached),
-				zap.Uint64("size", contract.Size),
-				zap.Stringer("usedCollateral", contract.UsedCollateral),
-				zap.Stringer("totalCollateral", contract.TotalCollateral),
-				zap.Stringer("maxCollateral", s.MaxCollateral))
+	var needsFormation []types.PublicKey
+	var needsRefresh []Contract
+	for hostKey, cc := range usableHostContracts {
+		switch {
```
peterjan: `nit` it would be more readable using an if statement for the good case and then form the reason right after, that would make it immediately clear what this code is doing, if it's not good, this is the reason, and then followed by either a refresh or forming a new contract with that host

contracts/form.go
```diff
@@ -162,162 +298,103 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
...
-	// we form contracts with all hosts in forceFormation and until we reach the
-	// wanted number of contracts
-	for i := range candidates {
-		if !forceFormation[candidates[i].PublicKey] && wanted <= 0 {
-			break
+		set := hosts.NewSpacedSet(cm.minHostDistanceKm)
+		for hostKey := range usableHostContracts {
+			host, ok := hostMap[hostKey]
+			if !ok {
+				panic("host missing from host map") // should never happen
```
peterjan: If we add the host to the `candidateContract` we get rid of these panics and it's immediately obvious why it's safe to do this because we're skipping hosts that aren't usable or blocked, see here:

```
host, ok := hostMap[contract.HostKey]
if !ok {
    continue // host is not usable or is blocked
}
```

we have the host at our disposal when we create the `candidateContract`s

contracts/form.go
```diff
@@ -87,70 +88,205 @@ func withinGougingLeeway(cost, limit types.Currency) bool {
...
+					goodForRefresh: refreshErr,
+					goodForFunding: fundingErr,
+					goodForAppend:  appendErr,
+				}
+				usableHostContracts[contract.HostKey] = cc
+			} else if cc.goodForAppend == nil && cc.goodForFunding == nil {
+				// host already has a contract good for both uploading and funding
+				continue
+			} else if contract.Size < cc.contract.Size {
+				// prefer smaller contracts since they are cheaper to refresh
```
ChrisSchinnerl: Shouldn't you check `contract.goodForAppend` and `contract.goodForFunding` first? If those are both `nil`, then `contract` should take precedence right?

contracts/form.go
```diff
@@ -87,70 +88,205 @@ func withinGougingLeeway(cost, limit types.Currency) bool {
...
+		return nil
+	})
+	if err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
+		log.Warn("failed to form contract", zap.Error(err))
 	}
+	return err == nil
+}
+
+func (cm *ContractManager) refreshContract(ctx context.Context, contract Contract, period uint64, fundTarget types.Currency, log *zap.Logger) bool {
+	err := cm.hosts.WithScannedHost(ctx, contract.HostKey, func(host hosts.Host) error {
```
peterjan, line 149: same here

contracts/form.go
```diff
@@ -87,70 +88,205 @@ func withinGougingLeeway(cost, limit types.Currency) bool {
...
+			continue // contract is good
+		case cc.goodForRefresh == nil:
+			needsRefresh = append(needsRefresh, cc.contract)
+			reason := cc.goodForAppend
+			if reason == nil {
+				reason = cc.goodForFunding
+			}
+			log.Debug("refreshing existing contract", zap.Stringer("hostKey", hostKey), zap.Stringer("contractID", cc.contract.ID), zap.NamedError("reason", reason))
+		default:
+			needsFormation = append(needsFormation, hostKey)
```
ChrisSchinnerl: If we don't check whether the host is good here, we'll end up failing later and possibly fail to do so repeatedly.

peterjan (changes requested):
I'm a little concerned we have less testing now, even though it's slightly more unit testable. This could've easily been separated into 3 PRs, and there's a couple of things we can leave out, for instance forming/refreshing in parallel. All in all I can see why it's a good refactor, I just feel it could've been a third of the size with 2 or 3 F/Us

contracts/form.go
```diff
@@ -162,162 +298,103 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
...
-		log := log.With(zap.Stringer("hostKey", hostKey), zap.Bool("force", forceFormation[hostKey]))
-
-		err := cm.hosts.WithScannedHost(ctx, hostKey, func(host hosts.Host) error {
-			// make sure host is still good
-			if !isGood(host, log) {
-				return fmt.Errorf("%w: %s", hosts.ErrBadHost, host.PublicKey)
+		additional := int(settings.WantedContracts) - len(usableHostContracts)
+		for hostKey := range hostsWithoutContracts {
+			if additional <= 0 {
+				break
```
ChrisSchinnerl, line 365: Doing it like this won't take into account planned formations that failed. So if a formation fails, we usually want to pick the next good candidate to make sure we reach the contract target.
In a worst case scenario the order we pick from is deterministic and we never end up with the number of contracts we want because we keep failing on the same few hosts.

contracts/form_test.go
```diff
@@ -187,19 +253,18 @@ func (s *hostManagerMock) UsabilitySettings(ctx context.Context) (hosts.Usabilit
 	return s.store.UsabilitySettings(ctx)
 }
 
-// TestPerformContractFormationWithoutContracts tests the
-// performContractFormation method assuming that we don't have any contracts
-// yet.
-func TestPerformContractFormationWithoutContracts(t *testing.T) {
+func TestPerformContractFormation(t *testing.T) {
```
ChrisSchinnerl, line 257: I think this is missing a test case for blocked hosts/contracts. To make sure we don't force any formations/refreshes when a host is blocked. 

Possibly also the edge case of forming contracts with two hosts at the same location.

contracts/contract.go
```diff
@@ -150,37 +150,57 @@ type (
...
+	case !c.Good:
+		return fmt.Errorf("contract is not good")
+	case c.Size >= maxContractSize:
+		return fmt.Errorf("contract has reached maximum size")
+	case c.ProofHeight <= height:
+		return fmt.Errorf("contract is not revisable")
+	case c.Size < c.Capacity:
+		return nil
+	}
+	// check if there is enough allowance and collateral to append at least one sector
```
peterjan, line 178: It's a little odd we go great lengths to get the `height` in here (instead of using the full period or something) to calculate whether we have enough money here to append a single sector. It wouldn't be the worst thing to return `false` here more often.

contracts/contract.go
```diff
@@ -150,37 +150,57 @@ type (
...
-		c.RemainingAllowance.Cmp(sectorAppendCost) > 0 &&
-		c.UsedCollateral.Add(sectorAppendCollateral).Cmp(c.TotalCollateral) < 0
-}
-
-// NeedsRefresh indicates that a contract should be refreshed.
-func (c Contract) NeedsRefresh() bool {
-	return c.Good && (c.OutOfFunds() || c.OutOfCollateral())
+// GoodForAccountFunding indicates whether a contract has enough remaining
+// allowance to spend the target amount.
+func (c Contract) GoodForAccountFunding(target types.Currency) error {
```
peterjan, line 155: This as kind of what I'm running into in my PR. This `target` is very likely the initial contract allowance. So immediately after forming it'll no longer be good for account funding.

contracts/form.go
```diff
@@ -87,70 +88,205 @@ func withinGougingLeeway(cost, limit types.Currency) bool {
...
-			forceFormation[contract.HostKey] = false
-		} else if _, hasContract := forceFormation[contract.HostKey]; !hasContract {
-			forceFormation[contract.HostKey] = true
+
+	// evaluate all existing contracts to see which hosts do not
+	// currently have a contract that is both good for uploading and
+	// funding
+	usableHostContracts := make(map[types.PublicKey]candidateContract)
+	for offset := 0; ; offset += batchSize {
+		batch, err := cm.store.Contracts(ctx, offset, batchSize, WithRevisable(true))
```
ChrisSchinnerl, line 232: You may or may not want to add `WithGood(true)` for pre-filtering here. However, if you do, you need to keep in mind that this changes `hostsWithoutContracts` for hosts that have exclusively bad contracts (e.g. blocked hosts).

ChrisSchinnerl (changes requested):

contracts/contract.go
n8mgr: Why would we use the full period? It’s wrong. It needs the duration based on the current height. Why skip a useful contract because it has enough funds for the remaining duration, but not enough for the whole period? Inconsistent logic is part of the reason this refactor is necessary in the first place.

contracts/contract.go
n8mgr: It *might* not be good for funding

api/app/app_test.go
n8mgr: Leaving it makes it easy to instrument later. Not logging in the test is pointless, but I resisted changing it here.

contracts/form.go
n8mgr: If every contract is bad, isn't it desirable to not continue to use that host? As far as I can tell, there are only two reasons a contract is marked as bad: 

1. The host refused to renew a contract
2. The host is blocked

contracts/form.go
n8mgr: We have a very thorough test in TestPerformContractFormation

contracts/form.go
n8mgr: That's impossible. It's built from a map `[host key]candidate`. It will only ever contain one entry per host. We determine the "best" contract above.

contracts/form.go
n8mgr: Unless the store is bugged, the filter criteria are exactly the same as `IsGood()`. Why check again when the same filtering already happened in the store? The only way it would change is if we grabbed it from the store again, which only happens in `WithScannedHost`.

contracts/form.go
n8mgr: `.IsGood()` only checks if the host is not blocked and the host usability is good, both of which are already filtered by the store. Why check again?

contracts/form.go
n8mgr: .IsGood() only checks if the host is not blocked and the host usability is good, both of which are already filtered by the store. Why check again?

contracts/form.go
n8mgr: We only want to try to form new contracts if "we have less than wanted contracts" AND "we have candidates".

contracts/form.go
peterjan: Because you're dealing with a `WithScannedHost` meaning up-to-date checks, you might as well sanity check the host didn't turn bad between the last scan and forming a contract.

contracts/form.go
n8mgr: Changed the logic to ensure that we can form a new contract if every contract with that host is bad for the edge case of renewals failing for random reasons.

n8mgr: Looking at this again, I don't think I needed to make this change. If a contract is !good, `GoodForRefresh`, `GoodForAppend` and `GoodForFunding` all return an error. A new contract would have been formed. This check prioritizes hosts we already have contracts with and are presumably happy with over forming a contract with a new host.

contracts/form.go
```diff
@@ -87,72 +86,248 @@ func withinGougingLeeway(cost, limit types.Currency) bool {
...
 
-	// fetch all revisable contracts
-	var activeContracts []Contract
-	const batchSize = 50
-	for offset := 0; ; offset += batchSize {
-		batch, err := cm.store.Contracts(ctx, offset, batchSize, WithRevisable(true))
+func (cm *ContractManager) formContract(ctx context.Context, host types.PublicKey, period uint64, fundTarget types.Currency, log *zap.Logger) bool {
+	err := cm.hosts.WithScannedHost(ctx, host, func(host hosts.Host) error {
+		if !host.IsGood() {
+			return fmt.Errorf("host is not good")
```
peterjan: `nit` no need for `Errorf` - you could also consider adding why it's not good

contracts/form_test.go
```diff
@@ -594,14 +530,15 @@ func newTestRevision(hk types.PublicKey) types.V2FileContract {
 	}
 }
 
+/*
```
peterjan: Oversight?

contracts/form.go
n8mgr: We don't know why it's not good. It returns a bool.

n8mgr:
> I'm a little concerned we have less testing now, even though it's slightly more unit testable. This could've easily been separated into 3 PRs, and there's a couple of things we can leave out, for instance forming/refreshing in parallel. All in all I can see why it's a good refactor, I just feel it could've been a third of the size with 2 or 3 F/Us

We may have one less test, but the new replacement is thorough and the logic is unified rather than split between three different decision trees each with its own inconsistent testing. 

> This could've easily been separated into 3 PRs, and there's a couple of things we can leave out, for instance forming/refreshing in parallel

Agreed on refreshing/forming in parallel. I've removed it. Hard disagree on anything else being irrelevant. Replacing the "out of collateral" and "needs refresh" checks meant reworking how formation decides between refreshing and forming new contracts. The new checks are not exactly the same as the old ones. Splitting those pieces across separate PRs would leave the intermediate state inconsistent (old decision tree calling the new helpers, or vice versa). It's safer to land them together.

contracts/form.go
```diff
@@ -162,162 +333,75 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, period
...
 			}
-
-			contract := res.Contract
-			minerFee := res.FormationSet.Transactions[len(res.FormationSet.Transactions)-1].MinerFee
-			err = cm.store.AddFormedContract(ctx, hostKey, contract.ID, contract.Revision, host.Settings.Prices.ContractPrice, allowance, minerFee, res.Usage)
-			if err != nil {
-				return fmt.Errorf("failed to add formed contract: %w", err)
+			log.Debug("forming contract with new host", zap.Int("remaining", additional))
+			if formContract(ctx, host.PublicKey, log) {
+				additional--
```
ChrisSchinnerl, line 389: Instead of `set.Add` you should call `set.CanAddHost` above and then call `set.Add` in here after the formation was successful. Otherwise, a failed formation blocks out a candidate for which it might succeed.

contracts/form.go
```diff
@@ -87,70 +86,242 @@ func withinGougingLeeway(cost, limit types.Currency) bool {
...
+			}
+		} else {
+			// contract is full or cannot be refreshed, form a new contract
+			reason := cc.goodForAppend
+			if reason == nil {
+				reason = cc.goodForFunding
+			}
+			log := log.With(zap.Stringer("hostKey", hostKey), zap.Stringer("existingContractID", cc.contract.ID))
+			log.Debug("forming new contract with existing host", zap.NamedError("reason", reason), zap.NamedError("refresh", cc.goodForRefresh))
+			if formContract(ctx, hostKey, log) {
```
ChrisSchinnerl, line 322: (nit) Since we no longer return the error, this loop can't skip anymore. So on shutdown or on other unrecoverable errors like an insufficient balance due to a lack of outputs we are going to loop through all hosts rather than skip the remaining loop cleanly.

contracts/form.go
ChrisSchinnerl: Reason number 3: https://github.com/SiaFoundation/indexd/issues/620#issuecomment-3487632411

ChrisSchinnerl (changes requested):
One more small change but otherwise I think it should be alright.

contracts/contract.go
```diff
@@ -150,37 +150,60 @@ type (
...
+		return nil
+	}
 }
 
-// OutOfFunds indicates that a contract is running low on funds and should be
-// refreshed.
-func (c Contract) OutOfFunds() bool {
-	return c.RemainingAllowance.Cmp(c.InitialAllowance.Div64(10)) < 0
+// GoodForAppend indicates whether a contract can be used to append sectors
+func (c Contract) GoodForAppend(prices proto.HostPrices, height uint64) error {
```
ChrisSchinnerl, line 167: This is the wrong spot to handle this but since I just ran into it here https://github.com/SiaFoundation/indexd/issues/632 I wanted to bring it up for discussion.

Do you think a full host would influence our contract formation code? In theory we can keep the contracts since the host is good but our active contracts with that host should probably not count towards the target of contracts we want to keep.

contracts/form.go
peterjan: I've tried poking a hole in this but the new logic is sound. I was tripping over this line `delete(hostsWithoutContracts, contract.HostKey)` mostly but !Good contracts are handled how they should be handled.

peterjan (dismissed):
This last version is pretty neat, very easy to follow

contracts/contract.go
n8mgr: I think that would only be a big issue if we had exactly 30 contracts rather than 50-100. We might cap the storage calculation in that case to limit spending and make the SDK storage aware instead of failing arbitrarily, but the complexity trade-off seems high for something that is probably a non-issue wrt contract management.

contracts/form.go
n8mgr: When the context is cancelled, formContract will return immediately without logging. If you run out of inputs, you deserve the log lines.

peterjan (approved):

ChrisSchinnerl (approved):

PR #623, opened 2025-11-04
Add `cluster.WaitForContracts` and `cluster.App`

peterjan:
This PR does two things:

1. gets rid of our two `WaitForX` helpers and moves it into a single method on the cluster itself that checks whether all hosts have an active contract. This is effectively a combination of the two but I think it's better located on the cluster and it's redundant to have two helpers. Waiting for active contracts instead of just the hosts will avoid races and NDFs.

2. It adds a helper called `App` that adds a test account and returns an app client for that account.

I'm pushing it separately because it's unrelated to account funding.

### DIFF
--- a/contracts/e2e_test.go
+++ b/contracts/e2e_test.go
@@ -20,24 +20,26 @@ func TestContractPruning(t *testing.T) {
 	logger := testutils.NewLogger(false)
 	cluster := testutils.NewCluster(t, testutils.WithLogger(logger), testutils.WithHosts(10))
 	indexer := cluster.Indexer
-	time.Sleep(time.Second)
 
-	// add an account
-	a1 := types.GeneratePrivateKey()
-	indexer.Store().AddTestAccount(t, a1.PublicKey())
+	// create an app
+	app := cluster.App(t)
 
-	time.Sleep(time.Second)
+	// fetch account
+	acc, err := app.Account(t.Context())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// wait for contracts to be formed
+	cluster.WaitForContracts(t)
+
+	// assert we have 10 usable hosts
 	hosts, err := indexer.Hosts().Hosts(context.Background(), 0, 10, hosts.WithUsable(true), hosts.WithActiveContracts(true))
 	if err != nil {
 		t.Fatal(err)
 	} else if len(hosts) != 10 {
 		t.Fatalf("expected 10 usable hosts, got %d", len(hosts))
 	}
-
-	// convenience variables
-	acc := proto.Account(a1.PublicKey())
-	client := indexer.App(a1)
-	store := indexer.Store()
 
 	// prepare pin params
 	params := slabs.SlabPinParams{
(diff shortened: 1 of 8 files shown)


contracts/e2e_test.go
```diff
@@ -116,14 +119,17 @@ func TestContractPruning(t *testing.T) {
 func TestSectorPinning(t *testing.T) {
 	// create cluster
 	logger := testutils.NewLogger(false)
-	cluster := testutils.NewCluster(t, testutils.WithLogger(logger), testutils.WithHosts(10))
+	cluster := testutils.NewCluster(t, testutils.WithLogger(logger), testutils.WithHosts(10), testutils.WithApps(1))
```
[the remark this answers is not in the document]
peterjan: man I saw it, I ignored it, I love that you nitpicked on it 👍 good stuff

internal/testutils/cluster.go
```diff
@@ -107,6 +119,17 @@ func NewCluster(t testing.TB, opts ...ClusterOpt) *Cluster {
 	return cluster
 }
 
+// AddApps adds n apps to the cluster.
+func (c *Cluster) AddApps(ctx context.Context, t testing.TB, n int) {
```
ChrisSchinnerl: There is really no need for apps to be tracked on the cluster or added using `WithApps`. 
We don't ever need more than one and it needs us to do this awful `app := cluster.Apps[0]` field access whenever we do.

Instead we could just have a method called `App` on the `cluster` which returns a new client with a registered account and then the test uses that. If you need more than one call it more than once.

slabs/e2e_test.go
```diff
@@ -16,31 +16,25 @@ import (
...
 	encryptionKey, shards, roots := slabs.NewTestShards(t, 1, 10)
 	for i := range shards {
-		client := indexer.HostClient(t, hosts[i].PublicKey)
+		client := indexer.HostClient(t, cluster.Hosts[i].PublicKey())
 		hs, err := client.Settings(context.Background())
 		if err != nil {
 			t.Fatal(err)
 		}
-		if _, err := client.WriteSector(context.Background(), hs.Prices, proto.NewAccountToken(a1, hosts[i].PublicKey), bytes.NewReader(shards[i]), proto.SectorSize); err != nil {
+		if _, err := client.WriteSector(context.Background(), hs.Prices, app.AccountToken(cluster.Hosts[i].PublicKey()), bytes.NewReader(shards[i]), proto.SectorSize); err != nil {
```
ChrisSchinnerl, line 39: This feels a bit clumsy. Shouldn't the `app` have a `WriteSector` method?

ChrisSchinnerl (changes requested):

slabs/e2e_test.go
peterjan: I agree but I'm not changing anything here. Looking at our codebase though `WriteSector` is defined on a `HostDialer` and a dialer is generally combined with an app client we call that an SDK client.

chris124567 (approved):

slabs/e2e_test.go
ChrisSchinnerl: Yeah it's just not very nice since you introduced the concept of an `App` which is an independent entity on the network that directly talks to a host but you didn't go all the way to make it independent and instead just use it to derive account tokens. Making it more of an `Account` rather than an `App`.

ChrisSchinnerl (approved):

PR #625, opened 2025-11-04
Update `store.MarkSlabRepaired`

peterjan:
I was looking at the repair code and I found out we don't set `next_repair_attempt` to the current timestamp when we've successfully repaired a slab. That's an oversight because it makes the slab jump the queue as soon as it becomes unhealthy. 

\- - 

Unrelated but I want to discuss it anyway: I think we should start prioritizing slabs based on their health. I ran some queries and, on average, the position in our queue is about 100k positions off of its true position in the queue if the unhealthy slabs were sorted on their health... with our worst slabs having 16 bad sectors, this is becoming an issue.


peterjan:
I was mistaken, we update `next_repair_attempt` when we select a slab, so all is good.

PR #628, opened 2025-11-04
Fix missing pruning offset

### LINKED ISSUE #627: Investigate Contract Pruning
We see a dramatic increase in lost sectors whenever data is removed from contracts. There may be an issue with how we are determining the indices to prune on the indexer side or the host side might be deleting the incorrect indices.

n8mgr:
Pruning currently deletes arbitrary sectors from the start of the contract since it does not account for the current offset when calling `FreeSectors` with a batch.

Fixes #627

### DIFF
--- a/contracts/pruning.go
+++ b/contracts/pruning.go
@@ -161,7 +161,7 @@ func (cm *ContractManager) pruneContract(ctx context.Context, client HostClient,
 		var indices []uint64
 		for i, root := range res.Roots {
 			if _, found := prunable[root]; found {
-				indices = append(indices, uint64(i))
+				indices = append(indices, offset+uint64(i))
 			}
 		}
 		if len(indices) == 0 {


chris124567 (approved):

contracts/pruning.go
```diff
@@ -161,7 +161,7 @@ func (cm *ContractManager) pruneContract(ctx context.Context, client HostClient,
 		var indices []uint64
 		for i, root := range res.Roots {
 			if _, found := prunable[root]; found {
-				indices = append(indices, uint64(i))
+				indices = append(indices, offset+uint64(i))
```
peterjan, line 164: WOW! This needs a regression test. I'm looking into why our tests didn't catch this :/.

peterjan (changes requested):

peterjan (approved):
I'll F/U with a PR that extends `TestContractPruning` - testing 1TB contracts in testing might be tricky 😅 
(maybe this warrants a test option in the contracts manager though)

ChrisSchinnerl (approved):

Issue #607, opened 2025-10-31
Add support for filtering accounts returned by `/accounts` by the connect key and make use of that in the UI


ChrisSchinnerl:
- [x] Implemented in backend
- [ ] Implemented in frontend

chris124567:
Unassigning myself because backend part is done

Edit: actually there can be multiple assignees so might as well just leave it

ChrisSchinnerl:
@chris124567 assigned you to make sure the issue shows up in `TODO` again.

PR #629, opened 2025-11-05
Add support for filtering accounts returned by `/accounts` by the connect key 

chris124567:
#607

### DIFF
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -430,7 +430,8 @@ func (a *admin) handleGETAccounts(jc jape.Context) {
 
 	accs, err := a.accounts.Accounts(jc.Request.Context(), offset, limit, opts...)
 	if errors.Is(err, accounts.ErrKeyNotFound) {
-		jc.Error(err, http.StatusNotFound)
+		// cast needed or empty response will be written
+		jc.Encode([]accounts.Account(nil))
 		return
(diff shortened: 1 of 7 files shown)


api/admin/admin.go
```diff
@@ -424,12 +424,18 @@ func (a *admin) handleGETAccounts(jc jape.Context) {
...
 	}
+	if connectKey := jc.Request.FormValue("connectkey"); connectKey != "" {
+		opts = append(opts, accounts.WithConnectKey(connectKey))
+	}
 
-	accounts, err := a.accounts.Accounts(jc.Request.Context(), offset, limit, opts...)
-	if jc.Check("failed to get accounts", err) != nil {
+	accs, err := a.accounts.Accounts(jc.Request.Context(), offset, limit, opts...)
+	if errors.Is(err, accounts.ErrKeyNotFound) {
+		jc.Error(err, http.StatusNotFound)
```
peterjan: Returning a 404 here is a bit odd. You return a 404 when a path-identified resource is not found. You could maybe argue a 400 is in place here but you return a 400 if a parameter is illegal or absent, which is not the case. This should return a 200 and an empty list of accounts.

peterjan (changes requested):

peterjan (approved):

ChrisSchinnerl (approved):

PR #631, opened 2025-11-05
Update `TestContractPruning`

peterjan:
This PR turns `TestContractPruning` into a more solid and robust test that would've caught the `offset` bug fixed by Nate in #628 . `contracts.WithSectorRootsBatchSize` is a little out there but I feel it's warranted in this particular case.


contracts/manager.go
```diff
@@ -237,6 +239,14 @@ func WithMinHostDistance(km float64) ContractManagerOpt {
 	}
 }
 
+// WithSectorRootsBatchSize sets the batch size for fetching sector roots.
+// The default is 1TB worth of sectors.
+func WithSectorRootsBatchSize(batchSize uint64) ContractManagerOpt {
+	return func(cm *ContractManager) {
+		cm.sectorRootsBatchSize = batchSize
```
ChrisSchinnerl, line 247: Might want to sanity check that this is >0 to avoid some unexpected surprises.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (dismissed):

ChrisSchinnerl (approved):

Issue #630, opened 2025-11-05
Mark contracts bad when we have a newer revision than the host


ChrisSchinnerl:
We have 5 hosts for which we get the error `failed to sync revision: local revision is newer than host revision`. This means we are kind of stuck on these contracts since we can't recover from it.

So when we identify this issue, we should mark that contract as bad and log a warning when doing so. That way, any data on that contract will be migrated away from it.

PR #634, opened 2025-11-05
Add store.MarkContractBad

peterjan:
Related to https://github.com/SiaFoundation/indexd/issues/630

REVIEWER NOTE: I've based it off of https://github.com/SiaFoundation/indexd/pull/631 to help with the NDF in that test


ChrisSchinnerl (dismissed):
I appreciate the benchmark but the query is to simple that if it turned out slow there wouldn't really be a lot we could do about it ^^

peterjan:
> I appreciate the benchmark but the query is to simple that if it turned out slow there wouldn't really be a lot we could do about it ^^

😂 - it was just in the fingers to be honest

n8mgr (dismissed):

n8mgr (approved):

PR #635, opened 2025-11-05
Fix BenchmarkUnhealthySlabs

peterjan:
I found out it's broken when I was trying out some health queries.

### DIFF
--- a/persist/postgres/sectors_test.go
+++ b/persist/postgres/sectors_test.go
@@ -1869,7 +1869,7 @@ func BenchmarkUnhealthySlabs(b *testing.B) {
 
 	// prepare base db
 	for range dbBaseSize / slabSize {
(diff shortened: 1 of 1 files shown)


chris124567 (approved):
As long as using `PinSectors` instead of setting `contract_sectors_map_id` directly doesn't increase the setup time by a ton

peterjan:
> As long as using `PinSectors` instead of setting `contract_sectors_map_id` directly doesn't increase the setup time by a ton

Should be fine. It's also a little annoying now to do it manually since we no longer have a single contract ID. I didn't time it but on my machine I didn't notice any meaningful slowdown in running the benchmark, that is doing the setup I mean, the benchmark results are the same.

ChrisSchinnerl (approved):

Issue #632, opened 2025-11-05
Handle hosts that are out of storage


ChrisSchinnerl:
One of our hosts, `Atlas` is out of storage. It keeps spamming the following error:

`hostd-mainnet-1  | 2025-11-05T10:46:34Z	ERROR	rhp4.siamux	RPC failed	{"peerAddress": "<redacted>", "streamID": "2fa76ca9", "rpc": "WriteSector", "error": "failed to store sector: failed to store sector: transaction failed (attempt 1): failed to get empty location: not enough storage", "elapsed": "88.294178ms"}`

Since this is the `WriteSector` RPC coming from Zeus' IP address, this probably means that:

1. Our junk uploader receives Atlas' public key and address from the indexer even though it is full
2. `Indexd` doesn't ignore full hosts during migration

chris124567:
I think we can close this?  All the sub issues are done.

PR #640, opened 2025-11-06
Ignore full hosts during migration

### LINKED ISSUE #642: Migrations out of storage
Check whether we are actively trying to write sectors to completely full hosts during migrations. If we are, it needs to be detected and fail early.

chris124567:
Fixes #642

Check that host has at least `MinRemainingStorage` bytes available to be an upload candidate for migrations.

### DIFF
--- a/slabs/dialer.go
+++ b/slabs/dialer.go
@@ -176,6 +176,9 @@ func (c *connPool) retry(ctx context.Context, hostKey types.PublicKey, addrs []c
 		// is similar, indicating that the stream hit a timeout which was set
 		// using SetDeadline. In both cases, the mux is still healthy.
 		return err
+	} else if errors.Is(err, errNotEnoughStorage) {
+		// No sense in retrying if the host does not have enough storage left
+		return err
 	}
 
 	// clear connection if we got transport error
@@ -219,6 +222,9 @@ func (c *connPool) uploadShard(ctx context.Context, h hosts.Host, migrationToken
 		settings, err := client.Settings(ctx)
 		if err != nil {
 			return fmt.Errorf("failed to fetch host settings: %w", err)
+		}
+		if settings.RemainingStorage < proto.SectorSize {
+			return errNotEnoughStorage
 		}
 
 		result, err = client.WriteSector(ctx, settings.Prices, migrationToken, shard, proto.SectorSize)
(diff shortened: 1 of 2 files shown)


n8mgr (changes requested):
We want to stop uploading to hosts that have no storage available, not set a leeway if they have remaining storage. An almost full host is still a good host. Probably an indicator it's a very good host.

On the migration end (it may not be broken), since we always scan their settings, I'd check the settings at the scan time. If `remaining storage < SectorSize`, fail early.

n8mgr:
I've split #632 into two issues #642 (fixed by this PR) and #641 for the SDK side of things.

ChrisSchinnerl (commented):
To give some additional context to Nate's response. `connPool.uploadShard` would probably be a good spot since that's the closest to the `WriteSector` call. You could return an `ErrNotEnoughStorage` there.

slabs/dialer.go
```diff
@@ -220,6 +223,9 @@ func (c *connPool) uploadShard(ctx context.Context, h hosts.Host, migrationToken
 		if err != nil {
 			return fmt.Errorf("failed to fetch host settings: %w", err)
 		}
+		if settings.RemainingStorage < proto.SectorSize {
+			return errNotEnoughStorage
```
ChrisSchinnerl: If you reuse the `photo.ErrNotEnoughStorage` error you can avoid the change to `retry` since it should be caught by `proto.ErrorCode(err) != proto.ErrorCodeTransport` then.

e.g. `return fmt.Errorf("%w: remaining storage less than sector size", proto.ErrNotEnoughStorage)`

ChrisSchinnerl (changes requested):

slabs/dialer.go
```diff
@@ -220,6 +223,9 @@ func (c *connPool) uploadShard(ctx context.Context, h hosts.Host, migrationToken
 		if err != nil {
 			return fmt.Errorf("failed to fetch host settings: %w", err)
 		}
+		if settings.RemainingStorage < proto.SectorSize {
```
n8mgr: The doc string for `RemainingStorage` says that its in sectors, so this should be `< 1` or `<= 0`

n8mgr (changes requested):

n8mgr (approved):

chris124567:
Failing cause of TestContractPruning NDF

ChrisSchinnerl (approved):

PR #644, opened 2025-11-06
Mark unsynced contracts bad

### LINKED ISSUE #630: Mark contracts bad when we have a newer revision than the host

peterjan:
Fixes #630

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -57,6 +57,7 @@ type (
 	RevisionStore interface {
 		ContractRevision(ctx context.Context, contractID types.FileContractID) (rhp.ContractRevision, bool, error)
 		UpdateContractRevision(ctx context.Context, contract rhp.ContractRevision, usage proto.Usage) error
+		MarkContractBad(ctx context.Context, contractID types.FileContractID) error
 	}
 )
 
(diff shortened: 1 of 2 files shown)


client/client.go
```diff
@@ -57,6 +57,7 @@ type (
 	RevisionStore interface {
 		ContractRevision(ctx context.Context, contractID types.FileContractID) (rhp.ContractRevision, bool, error)
 		UpdateContractRevision(ctx context.Context, contract rhp.ContractRevision, usage proto.Usage) error
+		MarkContractBad(ctx context.Context, contractID types.FileContractID) error
```
peterjan, line 60: It took everything I had to not refactor this. I don't think the client should have a `RevisionStore`, it should have a `ContractManager`. This sends you down quite the rabbit hole though, so I opted for a minimum LoC PR instead.

chris124567 (approved):

n8mgr (approved):

PR #646, opened 2025-11-06
Move slabs and sectors methods/tests

peterjan:
Been a long time coming, no changes. Would appreciate fast merge, I don't think anyone has open PRs with files in either of these packages so it's a good time.

### DIFF
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -7,10 +7,7 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/jackc/pgx/v5"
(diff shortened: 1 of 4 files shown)


n8mgr (changes requested):
There's no real point to this right now. We don't have a consistent style guide available and files shouldn't be alphabetized anyway. There are plenty of high priority bugs, issues, and features in `indexd` and our other repos spending time on a massive change for the sake of reordering is a waste.


My preferred structure of a Go file is this:
- imports
- consts
- vars
- types
- private methods
- public methods
- private funcs
- public funcs

At the very least, it is better to group methods and funcs logically over alphabetically.

PR #654, opened 2025-11-07
Track the total pinned data per app key and limit the data per key

### LINKED ISSUE #611: Track the total pinned data per app key and limit the data per connect key

chris124567:
Close #611

Almost thinking maybe we should just remove the limit at the account level and only limit at the connect key level


accounts/account.go
```diff
@@ -23,7 +23,7 @@ var (
 
 	// ErrStorageLimitExceeded is returned when an operation fails due to the
 	// account exceeding its storage limit.
-	ErrStorageLimitExceeded = errors.New("storage limit exceeded")
+	ErrStorageLimitExceeded = errors.New("account storage limit exceeded")
```
n8mgr: lets do "app storage limit" here and "account storage limit" for the other

chris124567:
Failing cause of TestContractPruning NDF

accounts/account.go
ChrisSchinnerl: It's a bit counterintuitive but Nate was right. `ErrStorageLimitExceeded` should be `app storage limit exceeded` and `ErrKeyStorageLimitExceeded` should be `account storage limit exceeded`.

Because from a user's perspective, they will have 1 account and let's say 5 apps they can link to it. So the app connect key limits the total storage of their account but the account limit limits the storage of their app.

I'd also rename these errors to `ErrAccountStorageLimitExceeded` and `ErrAppKeyStorageLimitExceeded`. It's more verbose but more obvious. 

I'd maybe also explain in the docstring why the app key says "account" and the account key says "app" in the actual error.

ChrisSchinnerl (changes requested):

n8mgr (dismissed):

persist/postgres/sectors.go
```diff
@@ -353,6 +366,19 @@ func (s *Store) PinSlabs(ctx context.Context, account proto.Account, nextIntegri
 				}
 			}
 		}
+
+		_, err = tx.Exec(ctx, `UPDATE accounts SET last_used=NOW(), pinned_data = $1 WHERE id = $2`, pinnedData, accountID)
```
peterjan: I'm a little concerned about atomicity of these updates. I think we're missing a row lock when we `SELECT` on line 219. So it should be `SELECT ... FOR UPDATE` to then update the app connect key (and the account for that matter) after pinning all slabs in a for loop.

I think this is worth verifying in a test. If a bug in accounting can be reproduced by pinning/unpinning slabs from multiple threads, we should consider hardening these updates. For example by moving the `max` calculation over the postgres, or have both... we can keep track of pinned data, return errors from Go, and when we update do `UPDATE accounts SET x = y WHERE id = z AND pinned_data + $1 < max` and see if `res.RowsAffected` is 1.

Before you take action though I'm roping in @ChrisSchinnerl here because it's a reasonably big ask to refactor this and try and reproduce it in a little unit test.

persist/postgres/sectors.go
```diff
@@ -393,18 +419,26 @@ RETURNING a.slab_id;`, accountID, args)
...
+	var delta uint64
+	if err := tx.QueryRow(ctx, `SELECT COUNT(*) * $1
+	FROM slab_sectors
+	WHERE slab_id = ANY($2)`, proto.SectorSize, sIDs).Scan(&delta); err != nil {
+		return fmt.Errorf("failed to get storage delta: %w", err)
+	}
+
+	var connectKeyID sql.NullInt64
+	err = tx.QueryRow(ctx, `UPDATE accounts
+SET pinned_data = pinned_data - $1
```
peterjan, line 432: This is another spot where I think an underflow can happen.

peterjan (changes requested):

persist/postgres/sectors.go
ChrisSchinnerl: In theory this is correct but I don't think it applies here since we start the transaction with an update to `accounts`. Effecting causing the whole pinning to be serial on a per-account basis anyway.

That said, I think we can improve parallel pinning under the same account by optimistically pinning slabs like so:

```
index 349526c6..b374daab 100644
@@ -206,22 +206,13 @@ func (s *Store) PinSlabs(ctx context.Context, account proto.Account, nextIntegri
 		var accountID int64
 		var connectKeyID sql.NullInt64
-		var pinnedData, maxPinnedData uint64
-		err := tx.QueryRow(ctx, "UPDATE accounts SET last_used = NOW() WHERE public_key = $1 RETURNING id, connect_key_id, pinned_data, max_pinned_data", sqlPublicKey(account)).Scan(&accountID, &connectKeyID, &pinnedData, &maxPinnedData)
+		err := tx.QueryRow(ctx, "SELECT id, connect_key_id FROM accounts WHERE public_key = $1", sqlPublicKey(account)).Scan(&accountID, &connectKeyID)
 			return accounts.ErrNotFound
 		} else if err != nil {
 			return err
 		}

-		var keyPinnedData, keyMaxPinnedData uint64
-		if connectKeyID.Valid {
-			err = tx.QueryRow(ctx, `SELECT pinned_data, max_pinned_data FROM app_connect_keys WHERE id = $1`, connectKeyID).Scan(&keyPinnedData, &keyMaxPinnedData)
-			if err != nil {
-				return fmt.Errorf("failed to get pinned data for connect key: %w", err)
-			}
-		}
-
 		hostRows, err := tx.Query(ctx, `SELECT h.public_key FROM contracts c INNER JOIN hosts h ON c.host_id = h.id WHERE state IN (0,1) AND renewed_to IS NULL AND good`)
 		if err != nil {
 			return fmt.Errorf("failed to get good hosts: %w", err)
@@ -240,6 +231,7 @@ func (s *Store) PinSlabs(ctx context.Context, account proto.Account, nextIntegri
 			return fmt.Errorf("failed to get good host rows: %w", err)
 		}

+		newPinnedData := uint64(0)
 		for _, slab := range toPin {
 			if slab.MinShards <= 0 || uint(len(slab.Sectors)) < slab.MinShards {
 				return slabs.ErrMinShards
@@ -278,19 +270,7 @@ func (s *Store) PinSlabs(ctx context.Context, account proto.Account, nextIntegri
 			// check here since we need to know if the slab is a new slab or whether
 			// it was just repinned.
 			if res.RowsAffected() > 0 {
-				newPinnedData := pinnedData + slab.Size()
-				if newPinnedData > maxPinnedData {
-					return accounts.ErrAccountStorageLimitExceeded
-				}
-				pinnedData = newPinnedData
-
-				if connectKeyID.Valid {
-					newKeyPinnedData := keyPinnedData + slab.Size()
-					if newKeyPinnedData > keyMaxPinnedData {
-						return accounts.ErrAppKeyStorageLimitExceeded
-					}
-					keyPinnedData = newKeyPinnedData
-				}
+				newPinnedData += slab.Size()
 			}

 			// if the slab already existed, we don't need to insert the sectors
@@ -367,15 +347,23 @@ func (s *Store) PinSlabs(ctx context.Context, account proto.Account, nextIntegri
 			}
 		}

-		_, err = tx.Exec(ctx, `UPDATE accounts SET last_used=NOW(), pinned_data = $1 WHERE id = $2`, pinnedData, accountID)
+		var pinnedData, maxPinnedData uint64
+		err = tx.QueryRow(ctx, `UPDATE accounts SET last_used=NOW(), pinned_data = pinned_data + $1 WHERE id = $2 RETURNING pinned_data, max_pinned_data`, newPinnedData, accountID).
+			Scan(&pinnedData, &maxPinnedData)
 		if err != nil {
 			return fmt.Errorf("failed to update account's pinned data: %w", err)
+		} else if pinnedData > maxPinnedData {
+			return accounts.ErrAccountStorageLimitExceeded
 		}

 		if connectKeyID.Valid {
-			_, err = tx.Exec(ctx, `UPDATE app_connect_keys SET pinned_data = $1 WHERE id = $2`, keyPinnedData, connectKeyID)
+			var pinnedData, maxPinnedData uint64
+			err = tx.QueryRow(ctx, `UPDATE app_connect_keys SET pinned_data = pinned_data + $1 WHERE id = $2 RETURNING pinned_data, max_pinned_data`, newPinnedData, connectKeyID).
+				Scan(&pinnedData, &maxPinnedData)
 			if err != nil {
 				return fmt.Errorf("failed to update connect key's pinned data: %w", err)
+			} else if pinnedData > maxPinnedData {
+				return accounts.ErrAppKeyStorageLimitExceeded
 			}
 		}
```

That way, we only rely on the `SELECT` to give us the immutable account's id and the bottleneck of acquiring an exclusive lock on the account happens at the very end.

@chris124567 can you try this in a benchmark with 16 goroutines pinning in parallel vs what we have now? The actual impact would be interesting to see since pinning in parallel is definitely going to be a common theme. e.g. gompa is uploading 16 slabs in parallel.

persist/postgres/sectors.go
ChrisSchinnerl: This one might be fine as well. Since we only update the pinned data for slabs we previously deleted. Which can only happen once.

ChrisSchinnerl (commented):
Commenting to unassign myself since we got some open discussions here.

persist/postgres/sectors.go
chris124567: This is with GOMAXPROCS=4 but the results were slightly worse for the parallel case when it was 16, possibly because I only have 4 cores.

master:
```
goos: linux
goarch: amd64
BenchmarkSlabs/PinSlab-4                 100      10207714 ns/op    4108.96 MB/s
BenchmarkSlabs/PinSlab_parallel-4                100      11877187 ns/op    3531.39 MB/s
BenchmarkSlabs/Slab-4                           2053        508922 ns/op    82415.46 MB/s
BenchmarkSlabs/PinnedSlab-4                      427       3147900 ns/op    13324.13 MB/s
BenchmarkSlabs/Slabs-40MiB-4                    1347        875387 ns/op    47913.73 MB/s
BenchmarkSlabs/Slabs-400MiB-4                    376       3168535 ns/op    13237.36 MB/s
BenchmarkSlabs/Slabs-4GiB-4                       49      24974959 ns/op    1679.40 MB/s
BenchmarkSlabs/SlabIDs-4                          39      31937105 ns/op
PASS
ok      go.sia.tech/indexd/persist/postgres 382.266s
```


With changes you suggested:
```
goos: linux
goarch: amd64
BenchmarkSlabs/PinSlab-4                 100      10030922 ns/op    4181.37 MB/s
BenchmarkSlabs/PinSlab_parallel-4                122       9707322 ns/op    4320.76 MB/s
BenchmarkSlabs/Slab-4                           2169        510182 ns/op    82211.99 MB/s
BenchmarkSlabs/PinnedSlab-4                      426       2795604 ns/op    15003.21 MB/s
BenchmarkSlabs/Slabs-40MiB-4                    1354        837565 ns/op    50077.37 MB/s
BenchmarkSlabs/Slabs-400MiB-4                    380       3109697 ns/op    13487.82 MB/s
BenchmarkSlabs/Slabs-4GiB-4                       50      22996722 ns/op    1823.87 MB/s
BenchmarkSlabs/SlabIDs-4                          38      30032095 ns/op
PASS
ok      go.sia.tech/indexd/persist/postgres 357.095s
```

So your suggestions do make it parallelize better.

persist/postgres/sectors.go
chris124567: The only difference is that we get the amount to subtract in a separate query.  The logic for calculating the amount is the same: https://github.com/SiaFoundation/indexd/blob/master/persist/postgres/sectors.go#L411

So I think it should be safe

persist/postgres/sectors.go
ChrisSchinnerl: Ok then I'd keep them and then this should be good to go.

peterjan (dismissed):

ChrisSchinnerl (dismissed):

persist/postgres/sectors.go
```diff
@@ -282,15 +283,7 @@ func (s *Store) PinSlabs(ctx context.Context, account proto.Account, nextIntegri
...
-				newPinnedData := pinnedData + slab.Size()
-				if newPinnedData > maxPinnedData {
-					return accounts.ErrStorageLimitExceeded
-				}
-
-				_, err := tx.Exec(ctx, `UPDATE accounts SET last_used=NOW(), pinned_data = $1 WHERE id = $2`, newPinnedData, accountID)
-				if err != nil {
-					return fmt.Errorf("failed to update account's pinned data: %w", err)
-				}
+				newPinnedData += slab.Size()
```
peterjan, line 284: `nit` outdated comment

ChrisSchinnerl (approved):

PR #659, opened 2025-11-10
Fix TestMigrations

### LINKED ISSUE #651: Fix TestMigrations NDF
This issue is simply to track the NDF, it should be low priority as it feels locally for me about 1/50 times. I've seen it once on our CI.

```
=== RUN   TestMigrations
    e2e_test.go:113: expected sector 45fa0a7af8fa2caf24787d1a775ba6af7a280bcd6ccd520104c30099bfaa8838 on host ed25519:dfaf6b6b1bd11a4d11d029e851362fe354dd6099a3a572e99883dbbe162c0ea3, got 45fa0a7af8fa2caf24787d1a775ba6af7a280bcd6ccd520104c30099bfaa8838 on host ed25519:31eac5dec990bca0501b04d8fc7006c957fe958e6cb9639f84da81152623e66f
--- FAIL: TestMigrations (41.11s)
```

peterjan:
This PR tweaks the configuration of our test cluster to avoid NDFs, in particular `TestMigrations`. 

To fix `TestMigrations`, all that is necessary is to increase the maintenance frequency to a higher, more reasonable interval. The reason it fails is because there's a race between fetching a `Slab` and fetching good hosts and contracts. A hydrated slab goes through the contract sectors map, by the time the slab is being repaired, that contract might no longer be revisable, because it was renewed for instance -> causing it to be selected for migration.

I decided to move some of our configuration into the default cluster config because they are sane values.
I ran `TestMigrations` 100 times successfully both on changing only the maintenance freq _and_ all other config options.

Closes https://github.com/SiaFoundation/indexd/issues/651

### DIFF
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -57,4 +57,4 @@ jobs:
       - name: Test
         uses: n8maninger/action-golang-test@v2
         with:
-          args: -race;-failfast
+          args: -race;-failfast;-timeout=1h;-count=25;-run=^TestMigrations$
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -57,4 +57,4 @@ jobs:
       - name: Test
         uses: n8maninger/action-golang-test@v2
         with:
-          args: -race;-failfast;-timeout=1h;-count=100;-run=^TestMigrations$
+          args: -race;-failfast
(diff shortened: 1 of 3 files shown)


ChrisSchinnerl (approved):
Might want to rebase this. For some reason the diff is quite polluted with stuff already in master but Github won't say that the branch is outdated.

ChrisSchinnerl (approved):

PR #661, opened 2025-11-11
Add integrity check stats

### LINKED ISSUE #653: Add integrity check stats
Add stats for integrity checks. I don't think we need to make this per host since we have lost sectors for failed integrity checks, but I'm open to it.

- total number of sectors checked
- number of sectors lost
- number of checks failed

Having a count of the number of sectors queued would be nice, but I'm not sure a `COUNT(*) WHERE next_check < NOW()` query on our largest table by far is worth it. Might be a good candidate for a materialized view. Would give us insights on how integrity checks are working as node size grows.

chris124567:
Close #653

### DIFF
--- a/api/admin/types.go
+++ b/api/admin/types.go
@@ -41,11 +41,14 @@ type (
 
 	// SectorsStatsResponse is the response body for the [GET] /stats/sectors
 	SectorsStatsResponse struct {
-		Slabs      int64 `json:"slabs"`
-		Migrated   int64 `json:"migrated"`
-		Pinned     int64 `json:"pinned"`
-		Unpinnable int64 `json:"unpinnable"`
-		Unpinned   int64 `json:"unpinned"`
+		Slabs       int64 `json:"slabs"`
+		Migrated    int64 `json:"migrated"`
+		Pinned      int64 `json:"pinned"`
+		Unpinnable  int64 `json:"unpinnable"`
+		Unpinned    int64 `json:"unpinned"`
+		Lost        int64 `json:"lost"`
+		Checked     int64 `json:"checked"`
+		CheckFailed int64 `json:"checkFailed"`
 	}
 
 	// AccountStatsResponse is the response body for the [GET] /stats/accounts.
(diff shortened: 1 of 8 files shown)


persist/postgres/sectors.go
```diff
@@ -70,6 +70,10 @@ func (s *Store) MarkSectorsLost(ctx context.Context, hostKey types.PublicKey, ro
 		if err := updateSectorStats(ctx, tx, -pinned, -unpinned, totalLost); err != nil {
 			return fmt.Errorf("failed to update sector stats: %w", err)
 		}
+		if err := incrementNumSectorsLost(ctx, tx, totalLost); err != nil {
```
peterjan: we've combined these updates in `updateSectorStats`, seems like we already pass `totalLost` so I'd definitely move this there

persist/postgres/stats.go
```diff
@@ -39,6 +39,21 @@ func incrementNumUnpinnedSectors(ctx context.Context, tx *txn, delta int64) erro
 	return err
 }
 
+func incrementNumSectorsLost(ctx context.Context, tx *txn, delta int64) error {
```
peterjan: we only ever add to this field, I would make that clear by changing `delta int64` to `n uint64` 
(same goes for the other two of course)

persist/postgres/stats_test.go
```diff
@@ -225,6 +225,104 @@ func TestSectorStats(t *testing.T) {
 	}
 }
 
+func TestIntegrityCheckStats(t *testing.T) {
```
peterjan, line 228: This covers the `checked` stats, I would also extend `TestSectorStats` to cover `num_sectors_lost`.
I know you cover it here as well but it would be a shame for `TestSectorStats` to not cover everything. I probably would look into merging this test into `TestSectorStats` so we have one test that covers all sector stats.

peterjan (changes requested):

api/admin/types.go
ChrisSchinnerl, line 1: Seems like we are missing an update to the openapi spec

persist/postgres/stats_test.go
ChrisSchinnerl: I think this test is fine. `TestSectorStats` is pretty hard to read at this point and is probably ripe for being split up into logical groups of sector stats. But not in this PR.

ChrisSchinnerl (changes requested):

n8mgr (approved):

peterjan (approved):

PR #663, opened 2025-11-12
Remove `ctx` parameter from store methods

### LINKED ISSUE #662: Remove `ctx` parameter from store methods
We keep running into issues with errors like `failed to pin sectors	{"hostKey": "ed25519:f291371aedfd74fd8a7f2a42b56ff46ecda298bc72bfd2dd7618408d75a8fc26", "error": "failed to scan host, failed to begin transaction: context deadline exceeded"` 
Where the slow part of an operation is done (e.g. pinning the sectors) but persisting the change failed for no good reason.
Even on shutdown we want any database operations to finish first.

The solution is to update our database interfaces to not accept a `ctx` so that we can never pass in a context that either times out or gets accidentally cancelled on shutdown.

chris124567:
Close #662

### DIFF
--- a/accounts/connect.go
+++ b/accounts/connect.go
@@ -60,35 +60,35 @@ type (
 
 // AddAppConnectKey adds a new app connect key.
 func (m *AccountManager) AddAppConnectKey(ctx context.Context, key UpdateAppConnectKey) (ConnectKey, error) {
-	return m.store.AddAppConnectKey(ctx, key)
+	return m.store.AddAppConnectKey(key)
 }
 
 // UpdateAppConnectKey updates an existing app connect key.
 // If the key does not exist, it returns [ErrKeyNotFound].
 func (m *AccountManager) UpdateAppConnectKey(ctx context.Context, key UpdateAppConnectKey) (ConnectKey, error) {
-	return m.store.UpdateAppConnectKey(ctx, key)
+	return m.store.UpdateAppConnectKey(key)
 }
 
 // DeleteAppConnectKey deletes an existing app connect key.
 // If the key does not exist, it returns [ErrKeyNotFound].
 func (m *AccountManager) DeleteAppConnectKey(ctx context.Context, key string) error {
-	return m.store.DeleteAppConnectKey(ctx, key)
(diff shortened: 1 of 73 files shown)


n8mgr (dismissed):

accounts/connect.go
```diff
@@ -60,35 +60,35 @@ type (
...
 func (m *AccountManager) AppConnectKeys(ctx context.Context, offset, limit int) ([]ConnectKey, error) {
-	return m.store.AppConnectKeys(ctx, offset, limit)
+	return m.store.AppConnectKeys(offset, limit)
 }
 
 // UseAppConnectKey uses an existing app connect key to add an account. If the key is exhausted, it
 // returns [ErrKeyExhausted]. If the key is not found, it returns [ErrKeyNotFound].
 func (m *AccountManager) UseAppConnectKey(ctx context.Context, key string, pk types.PublicKey, meta AccountMeta) error {
-	if err := m.store.UseAppConnectKey(ctx, key, pk, meta); err != nil {
+	if err := m.store.UseAppConnectKey(key, pk, meta); err != nil {
```
peterjan, line 91: Are we going to F/U with removing the `ctx` parameter from the methods that wrap the store in the managers?

persist/postgres/store.go
```diff
@@ -53,6 +53,11 @@ func (s *Store) transaction(ctx context.Context, fn func(context.Context, *txn)
 	return nil
 }
 
+// ctx returns the default context used by the store.
```
peterjan: I would get rid of this method, it's obfuscating the fact that we're passing in a background context. 
If you get rid of the `ctx` parameter in `s.transaction` we only have to use `context.Background()` there and then it's just informative and only in that spot. Perhaps we can even add a comment there why we get rid of all contexts in our store methods.

peterjan (changes requested):

accounts/connect.go
chris124567: Yeah

peterjan (approved):

ChrisSchinnerl (approved):

PR #664, opened 2025-11-13
Add uniFFIed client

n8mgr:
Adds a unified RPC client and dialer and switches the SDK to use it. Next will be switching pinning.


internal/testutils/cluster.go
```diff
@@ -201,7 +201,8 @@ func (c *Cluster) WaitForContracts(t *testing.T) {
 		required[h.PublicKey()] = struct{}{}
 	}
 
-	for range 100 {
+	var formed int
+	for {
```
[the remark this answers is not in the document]
n8mgr: The test will timeout itself

client/v2/client.go
```diff
@@ -0,0 +1,329 @@
...
+		go func(addr chain.NetAddress) {
+			defer func() {
+				<-sema
+				wg.Done()
+			}()
+			var transport rhp.TransportClient
+			var err error
+			switch addr.Protocol {
+			case siamux.Protocol:
+				transport, err = siamux.Dial(ctx, addr.Address, hostKey)
```
peterjan: pass the `dialCtx`

client/v2/client.go
```diff
@@ -0,0 +1,329 @@
...
+			case quic.Protocol:
+				transport, err = quic.Dial(ctx, addr.Address, hostKey)
```
peterjan: pass the `dialCtx`

client/v2/client.go
```diff
@@ -0,0 +1,329 @@
...
+
+	// reset the transport and try again.
+	transport, err = c.hostTransport(ctx, hostKey, true)
+	if err != nil {
+		return err
+	}
+	return fn(ctx, transport)
+}
+
+func (c *Client) prices(hostKey types.PublicKey, forceUpdate bool) (proto.HostPrices, error) {
```
peterjan: I see why you didn't add one but I think `prices` needs to respect the context of the RPC that called it. Fetching the prices here might block and it will obfuscate what's going on in upload/download algorithms. We'll be hanging on cancelled writes without knowing what's going only to find out later we hit a refresh and we're hanging on fetching the settings. It should be fine to fail here if the context was cancelled and leave the refresh up to the next RPC.

client/v2/client_test.go
```diff
@@ -0,0 +1,156 @@
...
+	var sector [proto.SectorSize]byte
+	copy(sector[:], data)
+	expectedRoot := proto.SectorRoot(&sector)
+
+	hostKey, ok := candidates.Next()
+	if !ok {
+		t.Fatal("expected candidate")
+	}
+
+	root, err := client.WriteSector(context.Background(), accountKey, hostKey, data)
```
peterjan: `nit` use `t.Context()`

internal/testutils/indexer.go
```diff
@@ -290,6 +299,16 @@ func NewIndexer(t testing.TB, c *ConsensusNode, log *zap.Logger, opts ...Indexer
 	}
 }
 
+// AddTestAccount adds a test account to the indexer and funds it.
+func (idx *Indexer) AddTestAccount(t *testing.T, pk types.PublicKey) {
```
peterjan, line 306: We have a method on the cluster called `App` that does something similar, but it feels off. I do think it's silly to have two helpers. Perhaps we should have `App` return the `SDK` and expose the app client on it?

sdk/client.go
```diff
@@ -403,30 +406,17 @@ func stripedJoin(dst io.Writer, dataShards [][]byte, writeLen int) error {
...
-	go func() {
-		root := proto4.SectorRoot(sector)
-		rootCh <- root
-	}()
-
-	uploaded, err := dialer.WriteSector(ctx, hostKey, sector)
-	if err != nil {
-		return types.Hash256{}, fmt.Errorf("failed to upload shard to host %s: %w", hostKey.String(), err)
-	} else if root := <-rootCh; uploaded != root {
-		return types.Hash256{}, fmt.Errorf("uploaded shard root %s does not match expected root %s", uploaded.String(), root.String())
```
peterjan, line 427: This was removed, I'm fine with that but I always thought we wanted that sanity check.

peterjan (changes requested):
Nice refactor

client/v2/candidates.go
```diff
@@ -0,0 +1,256 @@
...
+func (p *Provider) AddWriteSample(hostKey types.PublicKey, latency time.Duration) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	metric, exists := p.metrics[hostKey]
+	if !exists {
+		metric = &hostMetric{}
+		p.metrics[hostKey] = metric
+	}
+	metric.rpcWriteAverage.AddSample(latency)
+	metric.rpcFailRate.AddSample(true)
```
ChrisSchinnerl, line 199: I think this is something that might bite us. We sort hosts by the `rpcFailRate` so as a host, as soon as you failed once, you end up at the end of the queue and will potentially never be used again.
However, to improve/decay the rate, you need to be used. 

This should work fine if you create a new provider for each upload or download but if you intend to keep it around forever it might not be optimal. Maybe your fastest host had a connection issue for 5 minutes and now you never use it again.

So I think there also needs to be a time component here. e.g. if you call `Value` on a `failureRate` and it hasn't been updated within the last 5 minutes, you give it a free success sample.

client/v2/candidates.go
```diff
@@ -0,0 +1,256 @@
...
+		p.metrics[hostKey] = metric
+	}
+	metric.rpcFailRate.AddSample(false)
+}
+
+// Addresses returns the network addresses for the specified host.
+func (p *Provider) Addresses(hostKey types.PublicKey) ([]chain.NetAddress, error) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	return p.store.Addresses(hostKey)
```
ChrisSchinnerl, line 216: Why serialize access to the store in all methods? The provider only ever needs read access to the store which Postgres could nicely parallelize.

client/v2/client.go
```diff
@@ -0,0 +1,329 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to get transport: %w", err)
+	}
+
+	if err := fn(ctx, transport); err == nil {
+		return nil
+	} else if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) || proto.ErrorCode(err) != proto.ErrorCodeTransport {
+		// do not reset on context errors or non-transport errors
+		return err
+	}
```
ChrisSchinnerl, line 181: This seems like it would reintroduce the bug where `ErrClosedStream` on a single stream nukes the whole mux. Because `ErrClosedStream` is returned when the context is closed while reading from a stream instead of `context.Canceled` which is interpreted as `proto.ErrCodeTransport`.

client/v2/client.go
```diff
@@ -0,0 +1,329 @@
...
+// AppendSectors appends sectors to the specified contract on the host.
+func (c *Client) AppendSectors(ctx context.Context, signer rhp.ContractSigner, cs consensus.State, revision rhp.ContractRevision, sectors []types.Hash256) (res rhp.RPCAppendSectorsResult, err error) {
+	done, err := c.tg.Add()
+	if err != nil {
+		return rhp.RPCAppendSectorsResult{}, err
+	}
+	defer done()
+
+	err = c.rpcFn(ctx, revision.Revision.HostPublicKey, func(ctx context.Context, transport rhp.TransportClient) error {
+		prices, err := c.prices(revision.Revision.HostPublicKey, false)
```
ChrisSchinnerl: Not necessarily bad but I would like to highlight the implications here.

Before, we accepted that a suddenly gouging host might just empty our account but we wouldn't revise that contract anymore
If we start to use cached prices for `AppendSectors`, we effectively also accept that we might spend all the funds within the contract on pinning without checking that the prices are within our limits.

hosts/provider.go
```diff
@@ -0,0 +1,44 @@
...
+	host, err := ps.store.Host(context.Background(), hostKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get host: %w", err)
+	}
+	return host.Addresses, nil
+}
+
+// Usable implements the [HostStore] interface.
+func (ps *ProviderStore) Usable(hostKey types.PublicKey) (bool, error) {
+	host, err := ps.store.Host(context.Background(), hostKey)
```
ChrisSchinnerl: Due to us now caching prices instead of going through `WithScannedHost`, the only thing that updates the prices in the database will be the actual scanning code. Which happens once a day or so. So `Usable` would be more effective if the `Client`, upon refreshing a price table, also persisted it in the `ProviderStore` so that `Usable` actually operated on fresh prices rather than potentially 1 day old ones. 

NOTE: For the cached store this could be a no-op since it refreshes directly from the indexer anyway.

Also another though on why updating the database makes sense is that it will also cause hosts to become unusable faster meaning that clients will stop using them when they update their list of usable hosts. Which happens every 10 minutes. Otherwise they might continue to use a bad host for 24 hours or however long it takes for us to scan it again.

client/v2/client.go
```diff
@@ -0,0 +1,329 @@
...
+	// reset the transport and try again.
+	transport, err = c.hostTransport(ctx, hostKey, true)
+	if err != nil {
+		return err
+	}
+	return fn(ctx, transport)
+}
+
+func (c *Client) prices(hostKey types.PublicKey, forceUpdate bool) (proto.HostPrices, error) {
+	if !forceUpdate {
```
ChrisSchinnerl: `forceUpdate` is never `true` in any of the calls. 

Also right now there is also no time buffer in the validation and the error returned by `Validate` will also bypass the retry mechanism since it's not a `TransportError`. So if we start an RPC a few ms before the prices expire, we might punish the host for no good reason. Forcing a refresh if the prices expire less than 5 seconds from now might be an easy way around this. Don't really need much since the prices are checked first anyway.

sdk/client.go
ChrisSchinnerl: Isn't it redundant? The RPC handler itself should perform that check already.
Computing a sector root is too expensive of an operation to do for a redundant sanity check.

hosts/provider.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"context"
+	"fmt"
+
+	"go.sia.tech/core/types"
+	"go.sia.tech/coreutils/chain"
+)
+
+// A ProviderStore implements the HostStore interface using a Store.
+type ProviderStore struct {
```
ChrisSchinnerl: (nit) can we call this `HostStore`? Also the docstring mentions a `HostStore` interface which seems to be called `Store` in `candidates.go`.

ChrisSchinnerl (changes requested):

internal/testutils/indexer.go
n8mgr: We should not expose the raw app client on the SDK. The SDK is a currated experience for external developers. We do not want to expose a bunch of methods that are effectively useless to them to get rid of one extra test helper.

client/v2/client.go
n8mgr: We're getting into weird territory where you're saying it's also the client's problem to validate gouging settings and updating the database, something the SDK knows nothing about and can't do. 

I'd argue that the client is not the appropriate place for that logic. It would be more appropriate for that logic to live in the caller. A dialer should be dumb with minimal logic, like a store. It makes it easier to test and keeps the logic where it is actually useful.

hosts/provider.go
n8mgr: This isn't meant to be a thorough check. It's meant to be "do we know about this host, yes/no. We should not be polluting the client with a bunch of extra unrelated logic. Make that the caller's responsibility, like it is today with `WithScannedHost`. Having logic in multiple levels makes it hard to maintain and update. Very easy for conflicts where one callsite gets updated and the other doesn't.

client/v2/candidates.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+// Addresses returns the network addresses for the specified host.
+func (p *Provider) Addresses(hostKey types.PublicKey) ([]chain.NetAddress, error) {
+	return p.store.Addresses(hostKey)
+}
+
+// Candidates returns all host candidates ordered by their
+// historical performance.
+func (p *Provider) Candidates() (*Candidates, error) {
```
peterjan, line 221: This method is accessing metrics outside of the lock

client/v2/candidates.go
```diff
@@ -0,0 +1,256 @@
...
+func (p *Provider) sortHosts(hosts []types.PublicKey) {
+	sort.SliceStable(hosts, func(i, j int) bool {
+		return p.cmpMetrics(hosts[i], hosts[j]) < 0
+	})
+}
+
+// Prioritize reorders the given slice of hosts in place based
+// on their historical performance. The reordered slice is returned with
+// unusable hosts removed.
+func (p *Provider) Prioritize(hosts []types.PublicKey) []types.PublicKey {
```
peterjan, line 239: This method is accessing metrics outside of the lock

client/v2/client.go
```diff
@@ -0,0 +1,326 @@
...
+	err = c.rpcFn(ctx, hostKey, func(ctx context.Context, transport rhp.TransportClient) error {
+		settings, err = rhp.RPCSettings(ctx, transport)
+		return err
+	})
+	if err != nil {
+		return proto.HostSettings{}, err
+	}
+	if settings.Prices.Validate(hostKey) != nil {
+		c.mu.Lock()
+		c.cachedPrices[hostKey] = settings.Prices
```
peterjan, line 227: We're caching these prices on a failed validation

client/v2/candidates_test.go
```diff
@@ -0,0 +1,281 @@
...
+		var fr failureRate
+
+		fr.AddSample(false)
+		expected := 1.0
+		if v := fr.Value(); v != expected {
+			t.Fatalf("expected %f, got %f", expected, v)
+		}
+
+		time.Sleep(totalDecayMinutes * time.Minute)
+		synctest.Wait()
```
peterjan, line 74: Ok this is awesome. I thought this would require a dependency that mocks the time.

peterjan (changes requested):

client/v2/client.go
ChrisSchinnerl: Yeah definitely not the client's problem. But for the caller to be able to do this it needs to be able to pass in the validated prices for the client to use.

hosts/provider.go
ChrisSchinnerl: I get that but from your PR looks like we are moving away from `WithScannedHost` for more and more RPCs until we eventually just use this client and its cached, unvalidated prices.
The only thing that would remain that still checks prices is the daily scan which is the only set of prices we actually don't care about.

So before approving a PR like this I'd like to come up with a plan for where price validations eventually end up.

n8mgr: `WithScannedHost` isn't in the existing client, it's in the `HostManager`. This PR does not touch that? There is nothing preventing `WithScannedHost` from using the cached client in the `HostManager`?

peterjan (dismissed):

ChrisSchinnerl (approved):

PR #667, opened 2025-11-14
Fix verify sector cost computation

ChrisSchinnerl:
This potentially slows down our integrity checks. Since we end up subtracting 1000x the amount we actually pay from the service account.

### DIFF
--- a/slabs/sectorverifier.go
+++ b/slabs/sectorverifier.go
@@ -11,6 +11,7 @@ import (
 	proto "go.sia.tech/core/rhp/v4"
 	"go.sia.tech/core/types"
 	"go.sia.tech/indexd/hosts"
+	"go.sia.tech/mux/v2"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


slabs/sectorverifier.go
```diff
@@ -125,7 +126,7 @@ func (v *SectorVerifier) VerifySectors(ctx context.Context, host hosts.Host, roo
 
 		// verify the sector
 		res, err := hc.VerifySector(ctx, host.Settings.Prices, v.token(host.PublicKey), root)
-		if errors.Is(err, context.Canceled) {
+		if errors.Is(err, context.Canceled) || errors.Is(err, mux.ErrClosedStream) {
```
ChrisSchinnerl, line 129: Added some logs for debugging and saw a bunch of these pop up.

n8mgr (approved):

PR #668, opened 2025-11-14
Log failed sector verifications

ChrisSchinnerl:


slabs/sectorverifier.go
```diff
@@ -141,6 +145,10 @@ func (v *SectorVerifier) VerifySectors(ctx context.Context, host hosts.Host, roo
 		} else if strings.Contains(err.Error(), proto.ErrSectorNotFound.Error()) {
 			results = append(results, SectorLost)
 		} else {
+			v.log.Debug("failed to verify sector",
```
ChrisSchinnerl: Alright this might seem like a horrible idea but it shouldn't be that bad. It might highlight issues with checks that we consider failed for no good reason. Most checks shouldn't fail and instead just be lost. 

I've deployed this for an hour or so already and the only line I have seen so far is a batch of 25 `peer closed underlying connection` errors for host `a8544ee30760baa192a0dd32a34be3cba62f73d9b6562904f4d681491c8b1606`

n8mgr: My first thought was, you want to log an error per sector and we're storing 100TiB? My second thought was, well it's only if there's a problem should be okay.

slabs/sectorverifier.go
```diff
@@ -92,6 +94,8 @@ func (v *SectorVerifier) ResetBalance(ctx context.Context, host hosts.Host) erro
 // should handle any remaining results and then interrupt the integrity checks
 // for the host.
 func (v *SectorVerifier) VerifySectors(ctx context.Context, host hosts.Host, roots []types.Hash256) ([]CheckSectorsResult, error) {
+	logger := v.log.With(zap.Stringer("hostKey", host.PublicKey))
```
n8mgr: `log` for consistency

n8mgr (changes requested):

slabs/sectorverifier.go
```diff
@@ -141,6 +145,10 @@ func (v *SectorVerifier) VerifySectors(ctx context.Context, host hosts.Host, roo
 		} else if strings.Contains(err.Error(), proto.ErrSectorNotFound.Error()) {
 			results = append(results, SectorLost)
 		} else {
+			log.Debug("failed to verify sector",
```
peterjan, line 148: you could wrap this in a `sync.Once`, you'd very likely get the same information and you'd be somewhat protected from excessive log spam if things go wrong

peterjan (approved):

Issue #604, opened 2025-10-31
Remove pagination from `/stats/hosts`


peterjan:
I would like to propose we remove pagination from `/stats/hosts` for two reasons:

1) In our prometheus config we collect all of our node stats and we are not taking into account possible pagination. That's because all other stats endpoints return a single object. To fetch all hosts stats we would have to either temporarily pass a `limit` of 500 which would get all stats but if our host network grows we would probably not think to update the config to paginate through the result set.

2) It is inconsistent with our other stats endpoints because originally we returned a list of host stats, where it made sense to add an `offset` and ` limit`. On the PR there was a request to wrap the hosts slice into a response object, making pagination parameters a little quirky because they would be applied to the hosts slice on the response object. It would argue it was an oversight on my part not to remove it then and there.

This is not the highest priority but it is linked to migrations, which is our top priority item. 
If we decide to go ahead with this issue, there is a closed PR here https://github.com/SiaFoundation/indexd/pull/597

ChrisSchinnerl:
I'd set it to 500 in the prometheus config for now and keep this issue around as a reminder to come up with a proper solution.

peterjan:
Seems like Nate already took care of that https://github.com/SiaFoundation/deploy/commit/3fd7cd02d1ade7d01fd755b48e09174228e09cee

ChrisSchinnerl:
Considering that we are getting closer and closer to stabilizing the API (and also 500 hosts) we should revisit this.
Prometheus doesn't leave us a lot of options considering that it doesn't support pagination nor arrays. We also don't want to return aggregates since we like our per-host insights.

That might. leve us with only one option. We keep pagination for the common case and drop the limit argument when `prometheus` is supplied to perform the pagination internally and then return all stats.

PR #669, opened 2025-11-15
Add global host scan metrics

### LINKED ISSUE #666: Host scanning metrics
Similar to data integrity checks it would be nice to have the following metrics

- Total number of performed host scans 
- Total number of failed host scans
- Total number of host scans per host
- Total number of failed host scans per host

This should be quite easy to track in `UpdateHost` and will help us visualize why certain hosts have low uptime all of a sudden.

chris124567:
#666 

Add
- Total number of performed host scans
- Total number of failed host scans

### DIFF
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -953,11 +953,11 @@ func (a *admin) handleGETStatsHosts(jc jape.Context) {
 		return
 	}
 
-	var res HostStatsResponse
+	var res []HostStats
 	for _, s := range stats {
-		res.Hosts = append(res.Hosts, HostStats(s))
+		res = append(res, HostStats(s))
 	}
-	writeResponse(jc, res)
+	writeResponse(jc, HostStatsResponse(res))
 }
 
 func (a *admin) handleGETStatsHostsScans(jc jape.Context) {
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -105,7 +105,7 @@ type (
 	Store interface {
 		AccountStats() (AccountStatsResponse, error)
 		ContractsStats() (ContractsStatsResponse, error)
-		HostScanStats() (HostScanStatsResponse, error)
+		ScanStats() (ScansStatsResponse, error)
 		HostStats(offset, limit int) ([]hosts.HostStats, error)
 		SectorStats() (SectorsStatsResponse, error)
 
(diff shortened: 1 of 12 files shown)


api/admin/admin.go
```diff
@@ -244,10 +245,11 @@ func NewAPI(chain ChainManager, accounts Accounts, contracts ContractManager, ho
...
 
 		// stats endpoints
-		"GET /stats/accounts":  a.handleGETStatsAccounts,
-		"GET /stats/contracts": a.handleGETStatsContracts,
-		"GET /stats/hosts":     a.handleGETStatsHosts,
-		"GET /stats/sectors":   a.handleGETStatsSectors,
+		"GET /stats/accounts":    a.handleGETStatsAccounts,
+		"GET /stats/contracts":   a.handleGETStatsContracts,
+		"GET /stats/hosts":       a.handleGETStatsHosts,
+		"GET /stats/hosts/scans": a.handleGETStatsHostsScans,
```
peterjan: We wrapped the host stats in `HostStatsResponse` to allow extending it with other host statistics. I think instead of adding a new route, we should extend  `/stats/hosts`. This PR is also missing additions to our OpenAPI docs.

persist/postgres/stats.go
```diff
@@ -54,6 +54,16 @@ func incrementNumSectorsFailed(ctx context.Context, tx *txn, delta uint64) error
 	return err
 }
 
+func incrementNumScans(ctx context.Context, tx *txn, delta uint64) error {
+	_, err := tx.Exec(ctx, "UPDATE stats SET num_scans = num_scans + $1", delta)
+	return err
+}
+
+func incrementNumFailedScans(ctx context.Context, tx *txn, delta uint64) error {
```
peterjan: It would be slightly cleaner to have `incrementNumScans(ctx context.Context, tx *txn, success bool)` that increments `num_scans` at all times and updates `num_scans_failed` in case of !success.

peterjan (changes requested):

api/admin/admin.go
chris124567: That seemed inappropriate because /stats/hosts takes an offset and limit.  So we would always be returning the same values for the totals regardless of the offset.

peterjan: cc @ChrisSchinnerl are you fine having two routes? I don't really mind it's just that we wrapped it for this very reason. edit: personally I would probably unwrap the `/stats/hosts` to return an array and keep the `/scans` route.

ChrisSchinnerl: I don't mind a second route but we might as well just call it `/scans` rather than `/hosts/scans`. 
We can figure out what to do about `/stats/hosts` as we tackle https://github.com/SiaFoundation/indexd/issues/604

api/admin/types.go
```diff
@@ -51,16 +51,20 @@ type (
 		CheckFailed int64 `json:"checkFailed"`
 	}
 
+	// HostScanStatsResponse is the response body for [GET] /stats/hosts/scans.
+	HostScanStatsResponse struct {
+		Scans       int64 `json:"scans"`
+		ScansFailed int64 `json:"scansFailed"`
```
ChrisSchinnerl: Since we are already within the scan response struct here I'd rename these to `Total` and `Failed`. Similar to how we use `Pinned`, `Migrated` etc and not `SectorsPinned`, `SectorsMigrated`  in `SectorsStatsResponse`.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

PR #671, opened 2025-11-17
Add SDK helpers

ChrisSchinnerl:
Splits uploading into `NewEmptyObject` and `SaveObject` rather than doing it all in one. 

Fixes https://github.com/SiaFoundation/sia-sdk-rs/issues/230


sdk/object.go
```diff
@@ -84,13 +92,22 @@ func (o *Object) UpdateMetadata(meta json.RawMessage) {
...
 }
 
-// Object retrieves the object with the given key.
-func (s *SDK) Object(ctx context.Context, objectKey types.Hash256) (Object, error) {
-	lo, err := s.client.Object(ctx, objectKey)
-	if err != nil {
-		return Object{}, fmt.Errorf("failed to get locked object: %w", err)
+// NewObject creates a new Object with the given master key, slabs, and
+// metadata.
+func NewObject(slabs []slabs.SlabSlice, meta json.RawMessage) Object {
```
n8mgr: This constructor makes no sense now. You can't have slabs without already having an encryption key. Realistically it should just be unexported.

sdk/client.go
```diff
@@ -320,6 +326,24 @@ func (s *SDK) Close() error {
...
+// PinObject pins the given object to the indexer. This usually only needs to be
+// caleld if the upload was done with the WithSkipSave option.
+func (s *SDK) PinObject(ctx context.Context, obj Object) error {
+	return s.client.SaveObject(ctx, obj.Seal(s.appKey))
+}
+
+// SealObject seals the object for storage in the indexer. A sealed object can
+// be safely stored and shared without granting access to the underlying
+// metadata or object data.
+func (s *SDK) SealObject(obj Object) slabs.SealedObject {
```
n8mgr: These will need to be changed either here or in the Rust SDK after #682 introduces `AppKey` for consistencies sake.

Either `Object.Seal` here or `SDK.SealObject` in Rust.

sdk/client.go
```diff
@@ -320,6 +326,24 @@ func (s *SDK) Close() error {
...
+	return obj.Seal(s.appKey)
+}
+
+// OpenSealedObject opens a sealed object from the indexer.
+func (s *SDK) OpenSealedObject(so slabs.SealedObject) (Object, error) {
```
n8mgr: ditto

sdk/object.go
```diff
@@ -121,42 +138,19 @@ func (s *SDK) ListObjects(ctx context.Context, cursor slabs.Cursor, limit int) (
...
-		return nil, fmt.Errorf("failed to unlock metadata: %w", err)
+		return "", fmt.Errorf("failed to open object: %w", err)
 	}
-	return metadata, nil
+	return s.client.CreateSharedObjectURL(ctx, obj.ID(), obj.masterKey, validUntil)
 }
 
-func objectFromSealedObject(so slabs.SealedObject, appKey types.PrivateKey) (Object, error) {
+// ObjectFromSealedObject unlocks a SealedObject using the given app key.
+func ObjectFromSealedObject(so slabs.SealedObject, appKey types.PrivateKey) (Object, error) {
```
n8mgr: Since this one already takes an app key, why not make it a method on the SealedObject? i.e. `SealedObject.open`

Consistency in developer experience is key. We've got some attached to the SDK and some as exported funcs.

sdk/object.go
n8mgr: If it's really important for testing somewhere else, turn it into `NewEmptyObject` with encryption key and metadata set but no slabs.

n8mgr: That could also be the basis for resumable uploads. Change `Upload` to take an object pointer and the uploader appends slabs to it rather than building the object from scratch. The encryption offset could be calculated based on the number of slabs.

When you're done call `SaveObject`.

Gets rid of the extra option on Upload too.

ChrisSchinnerl: Problem with `NewEmptyObject` is that all empty objects have the same ID. So you can't easily mock the `sdk` and do something like the following:

```go
obj := sdk.Upload(r) // this can only return empty objects

if obj.Size() != expected { // obj.Size() is always 0
  return errors.New("mismatch") 
}

sdk.Pin(obj) // pin always gets the same `ID`

```

Potential alternatives could be:
1. providing an in-memory SDK as a reference implementation for testing in the SDK itself. Similar to the in-memory components for RHP4 we have in `coreutils`.
2. Exposing a way to create the `sdk` with custom implementations of `AppClient` and `HostClient`

sdk/client.go
ChrisSchinnerl: Yeah this is not super urgent. I'll wait for our breaking changes to be merged first and then update this.

ChrisSchinnerl:
I removed a lot of the stuff I added since it didn't really end up being useful.

sdk/client.go
```diff
@@ -168,8 +167,10 @@ func (s *SDK) AppKey() types.PrivateKey {
 
 // Upload uploads the data to hosts and pins it to the indexer.
 //
-// Returns the metadata of the slabs that were pinned
-func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) (Object, error) {
+// Appends the metadata of the slabs that were uploaded to the given object.
+// After uploading the object, the caller must call SaveObject to save the
```
peterjan, line 224: This is not reflected in the tests. We should check all call sites of `sdk.Upload` and ensure we follow it up with a call to `SaveObject`.

sdk/client.go
```diff
@@ -168,8 +167,10 @@ func (s *SDK) AppKey() types.PrivateKey {
...
+// object metadata to the indexer.
+func (s *SDK) Upload(ctx context.Context, obj *Object, r io.Reader, opts ...UploadOption) error {
```
peterjan, line 226: Mutating the object is a little nasty in my opinion, I think it's fine to keep it though but we should either document or perhaps even assert the provided object doesn't have slabs yet. Something like `if len(obj.slabs) > 0 { return ErrObjectNotEmpty` wouldn't be a bad sanity check

sdk/client.go
```diff
@@ -185,16 +186,15 @@ func (s *SDK) Upload(ctx context.Context, r io.Reader, opts ...UploadOption) (Ob
 
 	totalShards := int(uo.dataShards) + int(uo.parityShards)
 	if err := slabs.ValidateECParams(int(uo.dataShards), totalShards); err != nil {
-		return Object{}, err
+		return err
 	}
 
-	obj := Object{dataKey: frand.Bytes(32)}
```
peterjan, line 243: Passing in a pointer might lead to a panic on the following line when we dereference the object to get at the data key

peterjan (changes requested):

sdk/client.go
ChrisSchinnerl: I updated `TestE2E` to make sure the object isn't pinned implicitly and to then save it and check it was saved. Don't think we need every test to do that if they don't end up using the pinned object anyway.

sdk/client.go
ChrisSchinnerl: It's what enables resumable uploads.

sdk/client.go
ChrisSchinnerl: It will definitely lead to that but I'd say it's fair to call that a developer error

peterjan (dismissed):

sdk/client.go
peterjan: Ooooh 👀

sdk/client.go
```diff
@@ -206,14 +206,14 @@ top:
...
 			err := slab.err
 			if errors.Is(err, io.EOF) {
 				// all slabs complete
 				break top
 			} else if slab.err != nil {
-				return Object{}, slab.err
+				return slab.err
 			}
 
 			slabIndex := len(obj.slabs)
```
n8mgr: It's unlikely, but can't slabs arrive down the chan out of order depending on the concurrency settings?

peterjan (dismissed):

peterjan:
re-approved but Nate is technically right about the out of order slabs, other than that looks fine to me

sdk/object.go
```diff
@@ -96,6 +96,16 @@ func (o *Object) UpdateMetadata(meta json.RawMessage) {
 	o.metadata = slices.Clone(meta)
 }
 
+// NewEmptyObject creates a new Object to use in [Upload].
+func NewEmptyObject() Object {
```
peterjan, line 100: `nit` we could allow optional metadata, right now you'd have to create `NewEmptyObject` and then call `UpdateMetadata`

peterjan (dismissed):
LGTM 👍

peterjan (approved):

n8mgr (approved):

PR #676, opened 2025-11-18
Add download range support

### LINKED ISSUE #670: SDK: Add support for specifying an offset + length in `Download` method
Users might not always want to download a full object and we need this for S3 compatibility.

peterjan:
Fixes https://github.com/SiaFoundation/indexd/issues/670

### DIFF
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -104,7 +104,7 @@ func (s *SDK) downloadSlab(ctx context.Context, slab slabs.PinnedSlabSlice, maxI
 
 	// calculate offset and length that's required from each sector to recover
 	// the data referenced by the slab slice
-	offset, length := slab.SectorRegion()
+	offset, length := sectorRegion(slab)
 
 	// prioritize hosts
 	slabHosts = s.hosts.Prioritize(slabHosts)
@@ -268,34 +268,44 @@ func (s *SDK) Download(ctx context.Context, w io.Writer, obj Object, opts ...Dow
 		return errors.New("requested range exceeds object size")
 	}
 
-	// determine which slabs are required for the requested range
-	required := obj.SlabsForRange(do.offset, do.length)
-	if len(required) == 0 {
-		return errors.New("no slabs that cover the requested range")
-	}
-
 	// decrypt stream using the object's master key
 	if len(obj.masterKey) != 32 {
 		return fmt.Errorf("invalid master key length: %d", len(obj.masterKey))
 	}
 	w = decrypt((*[32]byte)(obj.masterKey), w, uint64(do.offset))
 
-	var curr int
+	var i int
+	offset := do.offset
+	length := do.length
+	for i = range obj.slabs {
+		slabLength := uint64(obj.slabs[i].Length)
+		if offset < slabLength {
+			break
+		}
+		offset -= slabLength
+	}
+
 	return s.downloadSlabs(ctx, w, do.maxInflight, do.hostTimeout, func() (slabs.PinnedSlabSlice, error) {
-		if curr >= len(required) {
+		if i >= len(obj.slabs) || length == 0 {
 			return slabs.PinnedSlabSlice{}, nil
 		}
-		slab := required[curr]
-		curr++
 
+		slab := obj.slabs[i]
 		pinned, err := s.client.Slab(ctx, slab.SlabID)
 		if err != nil {
-			return slabs.PinnedSlabSlice{}, fmt.Errorf("failed to get slab %d metadata: %w", curr, err)
+			return slabs.PinnedSlabSlice{}, fmt.Errorf("failed to get slab %d metadata: %w", i, err)
 		}
+		i++
+
+		slabOffset := slab.Offset + uint32(offset) // cannot overflow, offset < slabLength
+		slabLength := min(uint64(slab.Length)-offset, length)
+		offset = 0
+		length -= slabLength
+
 		return slabs.PinnedSlabSlice{
 			PinnedSlab: pinned,
-			Offset:     slab.Offset,
-			Length:     slab.Length,
+			Offset:     slabOffset,
+			Length:     uint32(slabLength),
 		}, nil
 	})
 }
@@ -324,23 +334,37 @@ func (s *SDK) DownloadSharedObject(ctx context.Context, w io.Writer, sharedURL s
 		return errors.New("requested range exceeds object size")
 	}
 
-	// determine which slabs are required for the requested range
-	required := obj.SlabsForRange(do.offset, do.length)
-	if len(required) == 0 {
-		return errors.New("no slabs that cover the requested range")
-	}
-
 	// decrypt stream using the object's master key
 	w = decrypt((*[32]byte)(encryptionKey), w, uint64(do.offset))
 
-	var curr int
+	var i int
+	offset := do.offset
+	length := do.length
+	for i = range obj.Slabs {
+		slabLength := uint64(obj.Slabs[i].Length)
+		if offset < slabLength {
+			break
+		}
+		offset -= slabLength
+	}
+
 	return s.downloadSlabs(ctx, w, do.maxInflight, do.hostTimeout, func() (slabs.PinnedSlabSlice, error) {
-		if curr >= len(required) {
+		if i >= len(obj.Slabs) || length == 0 {
 			return slabs.PinnedSlabSlice{}, nil
 		}
-		slab := required[curr]
-		curr++
-		return slab, nil
+		slab := obj.Slabs[i]
+		i++
+
+		slabOffset := slab.Offset + uint32(offset) // cannot overflow, offset < slabLength
+		slabLength := min(uint64(slab.Length)-offset, length)
+		offset = 0
+		length -= slabLength
+
+		return slabs.PinnedSlabSlice{
+			PinnedSlab: slab.PinnedSlab,
+			Offset:     slabOffset,
+			Length:     uint32(slabLength),
+		}, nil
 	})
 }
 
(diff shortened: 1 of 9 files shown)


slabs/objects.go
```diff
@@ -66,14 +78,121 @@ type (
...
+// Decrypt xors shards with the keystream derived from s.Key (starting at the
+// slice offset), using a different nonce for each shard.
+func (ss SharedSlabSlice) Decrypt(shards [][]byte) {
+	offset := ss.Offset / (proto.LeafSize * uint32(ss.MinShards))
+	var wg sync.WaitGroup
+	for i := range shards {
+		wg.Add(1)
+		go func(i int) {
+			nonce := make([]byte, 24)
+			nonce[0] = byte(i)
```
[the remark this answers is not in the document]
peterjan: We currently do not support that many shards.

slabs/objects.go
```diff
@@ -66,14 +78,121 @@ type (
...
+			c, _ := chacha20.NewUnauthenticatedCipher(ss.EncryptionKey[:], nonce)
```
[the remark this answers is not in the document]
peterjan: We ignore this everywhere

sdk/client.go
```diff
@@ -245,39 +250,48 @@ top:
...
+	if len(obj.masterKey) != 32 {
+		return fmt.Errorf("invalid master key length: %d", len(obj.masterKey))
+	}
+	w = decrypt((*[32]byte)(obj.masterKey), w, uint64(do.offset))
+
 	var curr int
-	return s.downloadSlabs(ctx, w, do.maxInflight, do.hostTimeout, func() (slabs.SharedSlab, error) {
-		if curr >= len(obj.slabs) {
-			return slabs.SharedSlab{}, nil
+	return s.downloadSlabs(ctx, w, do.maxInflight, do.hostTimeout, func() (slabs.PinnedSlabSlice, error) {
```
peterjan, line 294: I briefly considered hydrating the object data with `PinnedSlab` data. This currently happens on the fly in the iterator, where we call `s.client.Slab`. However, if we were to pay the price upfront, maybe even by adding a method to the client that fetches slab metadata in bulk, we would be able to deduplicate the download code between `sdk.Download` and `sdk.DownloadSharedObject` and get rid of the duplicated `SlabsForRange` and `Size` helpers.

sdk/object.go
```diff
@@ -69,11 +70,62 @@ func (o *Object) Seal(appKey types.PrivateKey) slabs.SealedObject {
...
 // Slabs returns a copy of the object's slabs.
 func (o *Object) Slabs() []slabs.SlabSlice {
 	return slices.Clone(o.slabs)
 }
 
+// SlabsForRange returns the slabs that cover the given range of data within the
+// object.
+func (o *Object) SlabsForRange(offset, length uint64) []slabs.SlabSlice {
+	// declare a helper to cast a uint64 to uint32 with overflow detection. This
+	// could should never produce an overflow.
```
chris124567: nit: remove "could"

sdk/client_test.go
```diff
@@ -93,17 +94,61 @@ func TestDownload(t *testing.T) {
...
+
+		buf.Reset()
+		if err := s.DownloadSharedObject(t.Context(), buf, sharedURL); err != nil {
+			t.Fatalf("failed to download shared object: %v", err)
+		} else if !bytes.Equal(buf.Bytes(), data) {
+			t.Fatal("data mismatch")
+		}
+	})
+
+	t.Run("ranges", func(t *testing.T) {
```
chris124567, line 132: Maybe add cases where length=0 or offset=dataSize etc

chris124567 (changes requested):

sdk/client_test.go
peterjan: Good idea

sdk/client.go
```diff
@@ -245,39 +251,48 @@ top:
...
 		opt(&do)
 	}
 
+	// validate range
+	if do.offset+do.length > obj.Size() {
+		return errors.New("requested range exceeds object size")
+	}
+
+	// determine which slabs are required for the requested range
+	required := obj.SlabsForRange(do.offset, do.length)
```
n8mgr: Why add an exported helper? It's called once. It can be inlined neatly and gets rid of the downcast helper, cloning the slice, and iterating over the slabs effectively four times. Ditto for the shared version.

```go
// Download downloads object metadata
func (s *SDK) Download(ctx context.Context, w io.Writer, obj Object, opts ...DownloadOption) error {
	// parse options
	do := downloadOption{
		hostTimeout: 4 * time.Second, // ~10 Mbps
		maxInflight: 10,
		offset:      0,
		length:      obj.Size(),
	}
	for _, opt := range opts {
		opt(&do)
	}

	// validate range
	if do.offset+do.length > obj.Size() {
		return errors.New("requested range exceeds object size")
	}

	// decrypt stream using the object's master key
	if len(obj.masterKey) != 32 {
		return fmt.Errorf("invalid master key length: %d", len(obj.masterKey))
	}
	w = decrypt((*[32]byte)(obj.masterKey), w, uint64(do.offset))

	var i int
	offset := do.offset
	length := do.length
	for i = range obj.slabs {
		slabLength := uint64(obj.slabs[i].Length)
		if offset < slabLength {
			break
		}
		offset -= slabLength
	}
	return s.downloadSlabs(ctx, w, do.maxInflight, do.hostTimeout, func() (slabs.PinnedSlabSlice, error) {
		if i >= len(obj.slabs) || length == 0 {
			return slabs.PinnedSlabSlice{}, nil
		}

		slab := obj.slabs[i]
		pinned, err := s.client.Slab(ctx, slab.SlabID)
		if err != nil {
			return slabs.PinnedSlabSlice{}, fmt.Errorf("failed to get slab %d metadata: %w", i, err)
		}
		i++

		slabOffset := slab.Offset + uint32(offset) // cannot overflow, offset < slabLength
		slabLength := min(uint64(slab.Length)-offset, length)
		offset = 0
		length -= slabLength
		return slabs.PinnedSlabSlice{
			PinnedSlab: pinned,
			Offset:     slabOffset,
			Length:     uint32(slabLength),
		}, nil
	})
}
```

slabs/objects.go
```diff
@@ -26,19 +28,26 @@ type (
...
-		Slabs             []SharedSlab `json:"slabs"`
-		EncryptedMetadata []byte       `json:"encryptedMetadata"`
+		Slabs             []PinnedSlabSlice `json:"slabs"`
+		EncryptedMetadata []byte            `json:"encryptedMetadata"`
+	}
+
+	// A PinnedSlabSlice represents a slice of a slab that is part of an object.
+	// It contains all the metadata needed to retrieve a slab.
+	PinnedSlabSlice struct {
+		PinnedSlab
```
n8mgr: I know it was here before, but the embedding is bugging me. What happens if you call `PinnedSlabSlice.EncodeTo`?

slabs/objects.go
```diff
@@ -66,14 +75,69 @@ type (
...
+		}
+		lastLength -= uint64(ss.Length)
+	}
+	slabs[len(slabs)-1].Length = cast32(lastLength)
+	return slabs
+}
+
+// SectorRegion returns the offset and length of the sector region that must be
+// downloaded in order to recover the data referenced by the slice.
+func (ss PinnedSlabSlice) SectorRegion() (offset, length uint64) {
```
n8mgr: Same with this. Why here? Why export it? It's called once in the SDK. It can be inlined or a generic helper `sectorRegion(dataShards, offset, length uint64) (uint64, uint64)`

n8mgr (changes requested):

sdk/client.go
n8mgr: Looks like it was copied over from `renterd`. That has bit us in the ass more than once. This is pretty minor, but a good reminder: make sure we're building the best systems we can rather than dragging in legacy code to tick the box on a completed issue. Design for `indexd`. Don't assume the same logic is the best fit just because you've done it before.

slabs/objects.go
peterjan: I removed the embedding, it was missing its custom impl. of EncodeTo and DecodeFrom to take into account the offset and length. This bothered me too, didn't spot this encoding issue though. Definitely unfortunate we can't easily go from a `SlabSlice` to a `PinnedSlabSlice`, I really wanted to deduplicate the logic between `Download` and `DownloadSharedObject`, which is already largely deduplicated but the part that's not is still very much duplicate code.

n8mgr (dismissed):

sdk/client.go
```diff
@@ -245,73 +251,130 @@ top:
...
 		opt(&do)
 	}
 
-	var curr int
-	return s.downloadSlabs(ctx, w, do.maxInflight, do.hostTimeout, func() (slabs.SharedSlab, error) {
-		if curr >= len(obj.slabs) {
-			return slabs.SharedSlab{}, nil
+	// validate range
+	if do.offset+do.length > obj.Size() {
+		return errors.New("requested range exceeds object size")
```
ChrisSchinnerl: Can we expose an `ErrInvalidRange` for this? In some circumstances like building an http API the caller might want to check for that error to then return the right status code.

sdk/client.go
```diff
@@ -245,73 +251,130 @@ top:
...
 		opt(&do)
 	}
 
-	var curr int
-	return s.downloadSlabs(ctx, w, do.maxInflight, do.hostTimeout, func() (slabs.SharedSlab, error) {
-		if curr >= len(obj.Slabs) {
-			return slabs.SharedSlab{}, nil
+	// validate range
+	if do.offset+do.length > obj.Size() {
+		return errors.New("requested range exceeds object size")
```
ChrisSchinnerl: ditto

sdk/client_test.go
```diff
@@ -93,17 +94,83 @@ func TestDownload(t *testing.T) {
...
+			return [2]uint64{offset, length}
+		}
+
+		cases := [][2]uint64{
+			{0, proto.SectorSize},
+			{proto.SectorSize, proto.SectorSize},
+			{proto.LeafSize, proto.LeafSize},
+			{dataSize - proto.SectorSize, proto.SectorSize},
+			{dataSize - proto.LeafSize, proto.LeafSize},
+			{dataSize, 0},
```
ChrisSchinnerl, line 148: Would be nice to have the following edge cases tested as well:
1. read within the boundaries of a single leaf (e.g. offset 130 with length 20)
2. read from the middle of a leaf until the middle of another leaf (e.g. offset 96 and length 64)
3. extend the test data to 3 slabs and start a read in the middle of slab 1 and finish in the middle of slab 3 

to force a bit of testing around the not so perfectly aligned ranges

ChrisSchinnerl (changes requested):

sdk/client_test.go
peterjan: I had 10 random offset/length reads to cover that but good idea to document them explicitly.

ChrisSchinnerl (approved):

PR #679, opened 2025-11-19
Improve perf. of UsableHosts

### LINKED ISSUE #612: Investigate host queries

peterjan:
This PR improves the benchmark performance of a bunch of our queries, most notably though it speeds up `UsableHosts` 5x in some cases, which you could see as a performance improvement for the SDK.

Fixes https://github.com/SiaFoundation/indexd/issues/612

### DIFF
--- a/persist/postgres/hosts.go
+++ b/persist/postgres/hosts.go
@@ -207,7 +207,7 @@ WHERE
 	-- blocked host filter
 	AND (($4::boolean IS NULL) OR ($4::boolean = hosts.blocked))
 	-- active contracts filter
-	AND (($5::boolean IS NULL) OR ($5::boolean = EXISTS (SELECT 1 FROM contracts WHERE host_id = hosts.id AND state IN (0,1) AND renewed_to IS NULL AND good)))
+	AND (($5::boolean IS NULL) OR ($5::boolean = EXISTS (SELECT 1 FROM contracts WHERE host_id = hosts.id AND state IN (0,1))))
 	-- public key filter
 	AND ((CARDINALITY($6::bytea[]) = 0) OR (public_key = ANY($6)))
 	%s -- orderClause
@@ -669,7 +669,7 @@ WHERE
 	settings_free_sector_price <= globals.one_sc / globals.sectors_per_tb AND
 	-- country filter
 	($3::text IS NULL OR country_code = $3::text) AND
-	-- active contracts
+	-- active and good contracts
 	EXISTS (SELECT 1 FROM contracts WHERE host_id = hosts.id AND state IN (0,1) AND renewed_to IS NULL AND good) AND
 	-- protocol filter
 	($4::smallint IS NULL OR EXISTS (SELECT 1 FROM host_addresses WHERE host_id = hosts.id AND protocol = $4::smallint)) `
--- a/persist/postgres/hosts.go
+++ b/persist/postgres/hosts.go
@@ -14,7 +14,6 @@ import (
 	proto4 "go.sia.tech/core/rhp/v4"
 	"go.sia.tech/core/types"
 	"go.sia.tech/coreutils/chain"
-	"go.sia.tech/indexd/contracts"
 	"go.sia.tech/indexd/geoip"
 	"go.sia.tech/indexd/hosts"
 )
@@ -208,11 +207,11 @@ WHERE
 	-- blocked host filter
 	AND (($4::boolean IS NULL) OR ($4::boolean = hosts.blocked))
 	-- active contracts filter
-	AND (($5::boolean IS NULL) OR ($5::boolean = EXISTS (SELECT 1 FROM contracts WHERE host_id = hosts.id AND state >= $6 AND state <= $7)))
+	AND (($5::boolean IS NULL) OR ($5::boolean = EXISTS (SELECT 1 FROM contracts WHERE host_id = hosts.id AND state IN (0,1) AND renewed_to IS NULL AND good)))
 	-- public key filter
-	AND ((CARDINALITY($8::bytea[]) = 0) OR (public_key = ANY($8)))
+	AND ((CARDINALITY($6::bytea[]) = 0) OR (public_key = ANY($6)))
 	%s -- orderClause
-	LIMIT $1 OFFSET $2`, orderClause), limit, offset, opts.Usable, opts.Blocked, opts.ActiveContracts, contracts.ContractStatePending, contracts.ContractStateActive, hks)
+	LIMIT $1 OFFSET $2`, orderClause), limit, offset, opts.Usable, opts.Blocked, opts.ActiveContracts, hks)
 		if err != nil {
 			return fmt.Errorf("failed to query hosts: %w", err)
 		}
@@ -341,7 +340,7 @@ func (s *Store) HostsWithUnpinnableSectors() ([]types.PublicKey, error) {
 			) AND NOT EXISTS (
 				SELECT 1
 				FROM contracts
-				WHERE host_id = hosts.id AND renewed_to IS NULL AND good AND state IN (0, 1)
+				WHERE host_id = hosts.id AND state IN (0,1) AND renewed_to IS NULL AND good
 			)
 		`)
 		if err != nil {
@@ -644,8 +643,7 @@ WITH globals AS (
 		settings_signature,
 		(get_byte(settings_protocol_version, 0) << 16) + (get_byte(settings_protocol_version, 1) << 8) + (get_byte(settings_protocol_version, 2)) as settings_version
 	FROM hosts
-	LEFT JOIN hosts_blocklist hb ON hosts.public_key = hb.public_key
-	WHERE hb.public_key IS NULL AND last_successful_scan IS NOT NULL -- not blocked and has settings
+	WHERE last_successful_scan IS NOT NULL -- has settings
 )
 SELECT
 	hosts.id,
@@ -672,7 +670,7 @@ WHERE
 	-- country filter
 	($3::text IS NULL OR country_code = $3::text) AND
 	-- active contracts
-	EXISTS (SELECT 1 FROM contracts WHERE host_id = hosts.id AND state <= 1) AND
+	EXISTS (SELECT 1 FROM contracts WHERE host_id = hosts.id AND state IN (0,1) AND renewed_to IS NULL AND good) AND
 	-- protocol filter
 	($4::smallint IS NULL OR EXISTS (SELECT 1 FROM host_addresses WHERE host_id = hosts.id AND protocol = $4::smallint)) `
 		args := []any{limit, offset, queryOpts.CountryCode, (*sqlNetworkProtocol)(queryOpts.Protocol)}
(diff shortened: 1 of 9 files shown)


persist/postgres/hosts.go
```diff
@@ -632,7 +630,7 @@ WHERE
 	-- country filter
 	($3::text IS NULL OR country_code = $3::text) AND
 	-- active contracts
-	EXISTS (SELECT 1 FROM contracts WHERE host_id = hosts.id AND state <= 1) AND
+	EXISTS (SELECT 1 FROM contracts WHERE host_id = hosts.id AND state IN (0,1) AND renewed_to IS NULL AND good) AND
 	-- protocol filter
 	($4::smallint IS NULL OR EXISTS (SELECT 1 FROM host_addresses WHERE host_id = hosts.id AND protocol = $4::smallint)) `
```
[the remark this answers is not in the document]
peterjan: When we block hosts we mark all contracts as bad. We can remove the join on blocklist for any query that asserts the host has an active contract.

chris124567 (dismissed):

persist/postgres/hosts.go
```diff
@@ -604,8 +603,7 @@ WITH globals AS (
 		settings_signature,
 		(get_byte(settings_protocol_version, 0) << 16) + (get_byte(settings_protocol_version, 1) << 8) + (get_byte(settings_protocol_version, 2)) as settings_version
 	FROM hosts
-	LEFT JOIN hosts_blocklist hb ON hosts.public_key = hb.public_key
-	WHERE hb.public_key IS NULL AND last_successful_scan IS NOT NULL -- not blocked and has settings
```
ChrisSchinnerl, line 648: I'd not get rid of this until there is a significant improvement in performance. This makes it very easy to verify that we don't return bad hosts. 
If we ever break the assumption that a blocked host will have all of its contracts marked as bad either on purpose or by introducing a bug, this will no longer work.

persist/postgres/hosts.go
```diff
@@ -870,21 +868,18 @@ func (s *Store) HostsForPinning() ([]types.PublicKey, error) {
...
-					SELECT 1
-					FROM sectors
-					WHERE sectors.host_id = h.id AND contract_sectors_map_id IS NULL
-				) AND
-				EXISTS (
-					SELECT 1
-					FROM contracts
-					WHERE contracts.host_id = h.id AND contracts.state <= $1 AND contracts.good = TRUE
-				) AND
-				NOT EXISTS (SELECT 1 FROM hosts_blocklist hb WHERE hb.public_key = h.public_key)
```
ChrisSchinnerl, line 926: Same thing here

persist/postgres/hosts.go
```diff
@@ -910,19 +905,13 @@ func (s *Store) HostsForPruning() ([]types.PublicKey, error) {
...
 		rows, err := tx.Query(ctx, `
-			SELECT h.public_key
-			FROM hosts h
-			WHERE
-				EXISTS (
-					SELECT 1
-					FROM contracts
-					WHERE contracts.host_id = h.id AND contracts.state <= $1 AND contracts.good = TRUE AND contracts.next_prune < NOW()
-				) AND
-				NOT EXISTS (
```
ChrisSchinnerl, line 961: and here

persist/postgres/hosts.go
```diff
@@ -208,11 +207,11 @@ WHERE
 	-- blocked host filter
 	AND (($4::boolean IS NULL) OR ($4::boolean = hosts.blocked))
 	-- active contracts filter
-	AND (($5::boolean IS NULL) OR ($5::boolean = EXISTS (SELECT 1 FROM contracts WHERE host_id = hosts.id AND state >= $6 AND state <= $7)))
+	AND (($5::boolean IS NULL) OR ($5::boolean = EXISTS (SELECT 1 FROM contracts WHERE host_id = hosts.id AND state IN (0,1) AND renewed_to IS NULL AND good)))
```
ChrisSchinnerl: Quite sure adding `good` here introduces a bug where we no longer form a new contract with a host that only has a bad contract for some reason.

ChrisSchinnerl (changes requested):

persist/postgres/hosts.go
peterjan: Hm I'll look into that. I mean it's a bug not to add `good` here so if this does introduce a bug we'll have to fix it on the call site. This is part of the reason why I want to close that issue, we're testing the indexer on broken queries so I'd rather fix it and see what the side-effects are than postponing it

persist/postgres/hosts.go
peterjan: That argument can be made for a number of things though. I don't mind re-adding it but this does significantly improve the benchmarks. In production we don't have that many blocked hosts  of course. 

edit: I see why it's safer to keep it around of course, but it's a nasty LEFT JOIN. I think Nate proposed to remove them once (can't recall when and where that was though so I apologise if I'm mistaken about that).

edit2: We also have numerous `h.Blocked` checks as a safety to warn us should we ever break the assumption without thinking to update these queries.

persist/postgres/hosts.go
ChrisSchinnerl: I'm not sure where you get the idea from that this query is broken. A contract is revisable when it hasn't been resolved yet. A revisable contract can be both good or bad.

peterjan: I'll look into it first and make sure I haven't made a mistake here, but I think you might be confusing this query with the contracts query? When we fetch contracts we have a `revisable` filter and that does indeed only check the state and whether it's renewed or not. This query fetches hosts, there we have an active contracts filter. An active contract should be revisable _and_ good.

peterjan: I checked and it's safe. This will influence wallet maintenance (make it more correct) and migrations (no effect). It fixes the filter so it'll also fix the output of the admin API.

persist/postgres/hosts.go
peterjan: The main argument for me is that this speeds up `UsableHosts`:

| Benchmark                          | Before (ms) | After (ms) | Improvement |
|-----------------------------------|-------------|------------|-------------|
| DefaultLimit                      | 1.605       | 0.934      | **41.8%**   |
| WithProtocol_DefaultLimit         | 4.165       | 1.272      | **69.5%**   |
| WithCountry_DefaultLimit          | 2.157       | 1.401      | **35.0%**   |
| WithProximity_DefaultLimit        | 8.530       | 3.074      | **64.0%**   |
| MaxLimit                          | 2.369       | 1.987      | **16.1%**   |
| WithProtocol_MaxLimit             | 3.266       | 2.790      | **14.6%**   |
| WithCountry_MaxLimit              | 1.679       | 1.406      | **16.3%**   |
| WithProximity_MaxLimit            | 4.901       | 3.787      | **22.7%**   |

Currently we don't have a lot of blocked hosts, but I can imagine we have to block a large number of hosts at one point. At which point we'll have a hidden `LEFT JOIN` that we left in there because at the time it wasn't posing a real issue.

persist/postgres/hosts.go
ChrisSchinnerl: I don't agree. A host without good contracts is still a host we form a new contract with if it isn't bad itself. Considering that, wallet maintenance should take that host into account. Same goes for `blockBadHosts`. A host that has only bad contracts should still be blocked if it becomes unusable to avoid forming a new contract with it.

So let's get rid of this to make sure we can merge the rest of the PR.

ChrisSchinnerl (changes requested):

persist/postgres/hosts.go
peterjan: We should indeed remove it from `blockBadHosts`. I am having a really hard time getting rid of this line that fixes the filter. The filter is incorrect the way it is. Our definition of an active contract is that it's not renewed, active or pending _and_ good. We have that assumption in multiple places, I don't see why we would not fix this filter and fix whatever call site that needs fixing. 

Can I remove `hosts.WithActiveContracts(true)` in blockBadHosts instead?
I rarely strongly disagree but I just don't get it, the filter is incorrect, why would we not change it.

peterjan: Double checking I actually noticed we are missing this is missing in the `ContractStats` too. I once fixed that but forgot to cherry pick it over to this PR.

edit: I guess the question is: "Do we want to include a bad contract in the total capacity and size?"
edit2: I guess the answer might actually be yes to that, in my head it was only active contracts

ChrisSchinnerl: > Our definition of an active contract is that it's not renewed, active or pending and good

I think this is where the confusion comes from. This is not specified anywhere and might be based on your own assumption based off `renterd` terminology. Active has always related to the consensus state of the contract being either pending or active in `indexd` and I don't see any usage of it in `indexd` where we would want to explicitly retrieve only good contracts.

peterjan: Ah, well thanks for that. It's not an assumption based off `renterd` though. It comes from `UsableHosts` where we check for "active contracts" and we _do_ check `good` there. But there it makes sense of course. Other than that I thought we added that filter for migrations, but we build the upload candidates based off of the `goodContracts`.

I ended up updating `blockBadHosts` and wallet maintenance. I think both are good right? I can open a separate PR for those if you want.

> We have that assumption in multiple places

Here I was referencing all of the `HostForX` methods by the way. But again, it makes sense to also check for `good` in all of those instances.

ChrisSchinnerl (dismissed):

persist/postgres/hosts.go
```diff
@@ -208,11 +207,11 @@ WHERE
 	-- blocked host filter
 	AND (($4::boolean IS NULL) OR ($4::boolean = hosts.blocked))
 	-- active contracts filter
-	AND (($5::boolean IS NULL) OR ($5::boolean = EXISTS (SELECT 1 FROM contracts WHERE host_id = hosts.id AND state >= $6 AND state <= $7)))
+	AND (($5::boolean IS NULL) OR ($5::boolean = EXISTS (SELECT 1 FROM contracts WHERE host_id = hosts.id AND state IN (0,1))))
```
peterjan, line 210: @ChrisSchinnerl our codebase is now _almost_ perfectly consistent aside from not checking `renewed_to` here. 

Not sure if it's worth it at this point but we could consider renaming `WithActiveContracts` to `WithRevisableContracts`, update the query here and also the `HostStats`. If we do that the definition of an active contract is a revisable contract that is good. It's probably good enough as it is but these small things tend to bother me sometimes and I feel it's important we're all perfectly aligned on these terms because they cause confusion when we talk things though.

ChrisSchinnerl: We can rename this to `WithRevisableContracts` to be consistent with the contracts query. That should be fine.
Checking `renewed_to` here should also be fine.

peterjan: I'll do it separately, it's only a couple of lines anyway. I'm happy though it'll be consistent and this confusion won't have been for nothing.

PR #682, opened 2025-11-20
Update app auth

### LINKED ISSUE #680: Improve App Auth Security
The goal with auth is that the user can deterministically generate a private key unique to an "app." In the event a user deletes the app or loses their device, they should be able to recover on a new device using just their mnemonic phrase and `indexd` "account" information.

Currently, we derive an ed25519 key pair using two pieces of information: a BIP39 mnemonic phrase supplied by the user and a constant [32]byte app ID supplied by the app the user wants to access. This works fine, but has a major problem, the app IDs are hard coded and can be spoofed trivially. If a malicious app were to gain access to a user's mnemonic they could spoof access to any app without user approval or knowledge. Obviously leaking your mnemonic in a malicious app is bad, but it doesn't have to be a complete death sentence for the whole account.

Here is our current app auth flow:

1. User creates a mnemonic phrase, this is private to the user
2. App creates a constant unique App ID. It is 32 bytes. This is public and can be discovered by anyone 
3. User inputs mnemonic into the app 
4. App derives `entropy` from that phrase 
5. App performs `pk, sk = KDF(entropy, app id)` to derive an ed25519 key pair
6.  App sends a signed auth request for `pk` to the server
7. User is shown a dialog that asks for their connect password and to approve/deny the connection
9. If the user approves, the server registers `pk` and starts funding it and the app saves `sk` for later use

I propose we add a third bit of information into the derivation known only to the `indexd` server. Apps that the user has explicitly granted approval from would be able to store the key as they do today, but knowing just the mnemonic and app ID would not be enough information to access anything. The app would have to go back through the authorization flow to get explicit approval from the user for each app ID. Entering your seed somewhere malicious is still a major no-no, but there's at least some limit to overall exposure in the event it does happen.

1. User creates a mnemonic phrase (private)
2. App creates a constant unique `app id` (public)
3. App generates a random 32 byte `request ID`
4. App sends the `request ID` and app metadata to the `indexd` server to start an authorization request
5. User is shown a dialog that asks for their connect password and to approve/deny the connection to the app
6. If the user approves, the server derives `server token = KDF(app id, server secret)` and sends it to the app. `server secret` must be known only to the server and unique per account.
7. The user enters their mnemonic phrase into the app 
8. The app performs `pk, sk = KDF(mnemonic, app id, server token)` to derive an ed25519 key pair
9. The app sends a signed request with the `pk` and `auth token` to the `indexd` server. The server registers `pk` and starts funding it. The app saves `sk` for later use.

This changes the auth to require three pieces of information instead of two. There is nothing we can do to prevent app ID spoofing (there are legitimate scenarios two separate apps might use the same ID), but we can at least make it require explicit user approval per app ID rather than an attack that compromises the whole account.

n8mgr:
Fixes #680

### DIFF
--- a/accounts/connect.go
+++ b/accounts/connect.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"go.sia.tech/core/types"
+	"go.sia.tech/indexd/keys"
 )
 
 var (
@@ -59,11 +60,12 @@ type (
 		RemainingUses int    `json:"remainingUses"`
 	}
 
-	// AccountMeta contains additional metadata associated with an account.
-	AccountMeta struct {
-		Description string
-		LogoURL     string
-		ServiceURL  string
+	// AppMeta contains additional metadata associated with an account.
+	AppMeta struct {
+		ID          types.Hash256 `json:"id"`
+		Description string        `json:"description"`
+		LogoURL     string        `json:"logoURL"`
+		ServiceURL  string        `json:"serviceURL"`
 	}
 )
 
@@ -94,11 +96,11 @@ func (m *AccountManager) AppConnectKeys(ctx context.Context, offset, limit int) 
 	return m.store.AppConnectKeys(offset, limit)
 }
 
-// UseAppConnectKey uses an existing app connect key to add an account. If the key is exhausted, it
+// RegisterAppKey uses an existing app connect key to add an account. If the key is exhausted, it
 // returns [ErrKeyExhausted]. If the key is not found, it returns [ErrKeyNotFound].
-func (m *AccountManager) UseAppConnectKey(ctx context.Context, key string, pk types.PublicKey, meta AccountMeta) error {
-	if err := m.store.UseAppConnectKey(key, pk, meta); err != nil {
-		return fmt.Errorf("failed to use app connect key: %w", err)
+func (m *AccountManager) RegisterAppKey(key string, pk types.PublicKey, meta AppMeta) error {
+	if err := m.store.RegisterAppKey(key, pk, meta); err != nil {
+		return fmt.Errorf("failed to register app connect key: %w", err)
 	}
 	return nil
 }
@@ -107,4 +109,14 @@ func (m *AccountManager) UseAppConnectKey(ctx context.Context, key string, pk ty
 // returns [ErrKeyNotFound].
 func (m *AccountManager) ValidAppConnectKey(ctx context.Context, key string) (bool, error) {
 	return m.store.ValidAppConnectKey(key)
+}
+
+// AppSecret derives a unique application secret using a stored user secret
+// associated with the given connect key and the provided app ID.
+func (m *AccountManager) AppSecret(connectKey string, appID types.Hash256) (types.Hash256, error) {
+	secret, err := m.store.AppConnectKeyUserSecret(connectKey)
+	if err != nil {
+		return types.Hash256{}, err
+	}
+	return types.Hash256(keys.Derive(secret[:], appID[:], []byte("server app secret"), 32)), nil
 }
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -56,7 +56,8 @@ type (
 	// Accounts defines the account management interface for the application API.
 	Accounts interface {
 		ValidAppConnectKey(context.Context, string) (bool, error)
-		UseAppConnectKey(context.Context, string, types.PublicKey, accounts.AccountMeta) error
+		RegisterAppKey(string, types.PublicKey, accounts.AppMeta) error
+		AppSecret(connectKey string, appID types.Hash256) (types.Hash256, error)
 
 		HasAccount(context.Context, types.PublicKey) (bool, error)
 		Account(context.Context, types.PublicKey) (accounts.Account, error)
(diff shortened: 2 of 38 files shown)


api/app/app.go
```diff
@@ -520,34 +483,126 @@ func (a *app) handlePOSTAuthConnect(jc jape.Context) {
 		return
 	}
```
ChrisSchinnerl, line 485: One issue I'm seeing here after going through all the trouble of deriving the key from another factor is a client app spoofing the web ui.

It could open the browser on `malicious.app.storage` and display the same web ui we display with the expected app id. Then the user would send their secret to the third party server which then proxies the request to our URL.

I'm not sure there is a way around this before SiaHub when we start using OAuth on the web ui and users will be able to tell who they are authorizing with.

api/app/app.go
```diff
@@ -505,10 +469,9 @@ func (a *app) handleGETAuthConnectUI(jc jape.Context) {
...
 }
 
+// 3. web UI sends approval/rejection
```
ChrisSchinnerl, line 472: I think you are missing an update to `auth.html` to display the app id and some warning string. Unless you want to leave that to Alex.

api/app/app.go
n8mgr: That’s not a unique issue to us or something we can resolve.

api/app/app.go
```diff
@@ -68,25 +69,30 @@ type (
...
+		LogoURL     string        `json:"logoURL"`
+		ServiceURL  string        `json:"serviceURL"`
+		CallbackURL string        `json:"callbackURL"`
 	}
 
 	// AuthConnectStatusResponse is the response body for checking the status of an
 	// application connection request.
 	AuthConnectStatusResponse struct {
-		Approved bool `json:"approved"`
+		Approved   bool          `json:"approved"`
```
peterjan, line 94: `nit` could argue that `Approved` is redundant since a lack of user secret indicates it was not approved - squashing this field would get rid of `panic("user secret is empty for approved request") // should never happen`

api/app/app.go
```diff
@@ -520,34 +483,126 @@ func (a *app) handlePOSTAuthConnect(jc jape.Context) {
...
+	} else if time.Now().After(authReq.Expiration) {
+		jc.Error(fmt.Errorf("request expired"), http.StatusGone)
+		return
+	} else if authReq.Approved {
+		return
 	}
 
-	var approveReq ApproveAppRequest
-	if err := jc.Decode(&approveReq); err != nil {
+	if !approveReq.Approve {
```
peterjan, line 498: If a user rejects the request, shouldn't it be deleted from `authRequests`? In this case maybe we should perform this check before fetching the auth request and potentially returning errors depending on its value.

api/app/app.go
```diff
@@ -520,34 +483,126 @@ func (a *app) handlePOSTAuthConnect(jc jape.Context) {
...
+			UserSecret: authReq.UserSecret,
+		})
+	}
+}
+
+// 5. app registers the public key with indexd using a request signed
+// with the derived private key
+func (a *app) handleAuthRegister(jc jape.Context) {
+	var requestID string
+	if err := jc.DecodeParam("requestID", &requestID); err != nil {
```
peterjan, line 556: this is inconsistent with parameters like `contractid` and `slabid`

accounts/connect.go
```diff
@@ -59,8 +60,9 @@ type (
 		RemainingUses int    `json:"remainingUses"`
 	}
 
-	// AccountMeta contains additional metadata associated with an account.
-	AccountMeta struct {
+	// AppMeta contains additional metadata associated with an account.
+	AppMeta struct {
```
peterjan, line 64: should we add JSON tags here? This is being returned along with accounts

peterjan (commented):
Left a few small remarks but overall definitely LGTM

api/app/app.go
n8mgr: Harmless to be explicit here. Rather than requiring len == 0.

api/app/app.go
n8mgr: It's a path param, who cares? Not following our normal casing is more incorrect, imo.

contracts/fund.go
```diff
@@ -0,0 +1,100 @@
+//go:build ignore
```
ChrisSchinnerl: Why is this file in here?

ChrisSchinnerl (changes requested):

ChrisSchinnerl (dismissed):

peterjan (changes requested):
Missing update to OpenAPI spec. Fine to F/U with that though if you want to get this in.

api/app/app.go
```diff
@@ -520,34 +483,126 @@ func (a *app) handlePOSTAuthConnect(jc jape.Context) {
...
 	}
 
-	var approveReq ApproveAppRequest
-	if err := jc.Decode(&approveReq); err != nil {
+	// derive shared secret
+	sharedSecret, err := a.accounts.AppSecret(connectKey, authReq.Request.AppID)
+	if err != nil {
+		jc.Error(fmt.Errorf("failed to derive app secret: %w", err), http.StatusInternalServerError)
+		return
+	}
```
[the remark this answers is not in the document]
n8mgr: auth is handled by the `wrapBasicAuth` handler.

peterjan (dismissed):

ChrisSchinnerl (dismissed):

n8mgr:
Let’s hold off on merging until the rust sdk is updated so we don’t break everything right before break

Issue #608, opened 2025-10-31
Add support for deleting an account to both backend and UI via soft deletion + background pruning


ChrisSchinnerl:
Needs to be done via soft deletion since accounts are potentially associated with a lot of objects, slabs and sectors in our database. So we should add a `deleted` field which we use for pruning as well as for excluding deleted accounts from existing queries.

- [x] Implemented in backend
- [x] Implemented in frontend

chris124567:
Does this sound OK for the pruning part?  The rest I think is straightforward.
 
- DeleteAccount sets `deleted` = `true`, `pruned` = `false`
- Slab manager in maintenance loop calls new store method `AccountsForPruning`, and for each of those accounts, we:
    - Delete all of the objects associated with that account (or maybe 50-100)
    - Call `PruneSlabs` to remove all of the slabs on that account that are now not attached to any object
    - Set `pruned` = `true` on the account once all objects have been removed and all slabs have been pruned

My only worry is that `PruneSlabs` can be very expensive.

ChrisSchinnerl:
You shouldn't need `pruned`. As soon as `deleted` is `true` we won't allow adding anything to that account. 

Apart from that it should be as you described. Delete 100 or so objects per batch and once you run out of objects, call `PruneSlabs` which will delete all slabs of that account in batches.
Once that is done, delete the account for good and increment the remaining uses of the app key by 1 again.

`PruneSlabs` itself performs batching already so it should be fine. If it's not, we will need to update it. But considering that a user should be able to call it too, we need to make sure it's sane anyway.

alexfreska:
@chris124567 I see a change was merged, does this change the existing API or require changes to the existing account delete functionality?

<img width="730" height="202" alt="Image" src="https://github.com/user-attachments/assets/4c7114ae-63ac-41a3-aec5-4e0b24bd5b59" />

chris124567:
@alexfreska yeah nothing should be required.  There are no new endpoints.  Once the account is marked as deleted, it won't show up in the response to /accounts and the auto deletion of all the objects etc will be done in the background by the accounts manager.  Before deletion would fail if the account still had objects/slabs attached to it but now it should always succeed (unless it's a service account)

alexfreska:
@chris124567 ok great, sounds like this issue is complete.

PR #694, opened 2025-11-21
Add support for deleting an account to backend via soft deletion + background pruning

chris124567:
#608


accounts/manager.go
```diff
@@ -190,9 +190,11 @@ func (m *AccountManager) Accounts(ctx context.Context, offset, limit int, opts .
 	return m.store.Accounts(offset, limit, opts...)
 }
 
-// DeleteAccount deletes the account for the given public key.
+// DeleteAccount soft deletes the account with the given public key.
+// Objects/slabs/sectors associated with the account will be cleaned up by the
+// slabs manager.
 func (m *AccountManager) DeleteAccount(ctx context.Context, ak types.PublicKey) error {
-	return m.store.DeleteAccount(ak)
+	return m.store.DeleteAccount(ak, false)
```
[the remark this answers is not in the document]
chris124567: fixed

persist/postgres/init.sql
```diff
@@ -22,9 +22,11 @@ CREATE TABLE accounts (
     description TEXT NOT NULL DEFAULT '',
     logo_url TEXT NOT NULL DEFAULT '',
     service_url TEXT NOT NULL DEFAULT '',
-    last_used TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
+    last_used TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    deleted BOOLEAN NOT NULL DEFAULT FALSE
```
peterjan: Generally soft deletion uses a `deleted_at` timestamp. Same thing but it gives us a little more info/context when debugging.

persist/postgres/accounts.go
```diff
@@ -121,18 +122,39 @@ func (s *Store) ActiveAccounts(threshold time.Time) (count uint64, err error) {
 }
 
 // DeleteAccount deletes the account in the database with given account key.
-func (s *Store) DeleteAccount(ak types.PublicKey) error {
+func (s *Store) DeleteAccount(ak types.PublicKey, soft bool) error {
```
peterjan: `Delete` should probably always soft-delete the account. I think you want to use this method for hard deleting an account right? That's not immediately obvious to me when I pass `soft=false`, but also that should happen in a single transaction that gets rids of everything that's attached to that account.

persist/postgres/objects.go
```diff
@@ -207,47 +207,91 @@ func (s *Store) ListObjects(account proto.Account, cursor slabs.Cursor, limit in
...
+	if err != nil {
+		return fmt.Errorf("failed to delete object slabs: %w", err)
+	}
+	_, err = tx.Exec(ctx, `DELETE FROM objects WHERE id = ANY($1)`, objectIDs)
+	if err != nil {
+		return fmt.Errorf("failed to delete object: %w", err)
+	}
+
+	_, err = tx.Exec(ctx, `
+            UPDATE object_events SET was_deleted = TRUE, updated_at = NOW()
```
peterjan: unrelated to the PR but `object_events` should probably have a `create_at`, `updated_at` and `deleted_at` timestamp if we want to capture all that information. Updated_at turning into deleted_at if was_deleted is true is a little odd

persist/postgres/objects.go
```diff
@@ -310,13 +354,16 @@ AND slabs.digest = ANY($2)`, accountID, args).Scan(&count); err != nil {
 	})
 }
 
-func accountID(ctx context.Context, tx *txn, account proto.Account) (int64, error) {
+func accountID(ctx context.Context, tx *txn, account proto.Account, deletedOK bool) (int64, error) {
```
peterjan: Perhaps we should return both the ID and a boolean that indicates whether the account was deleted. `deletedOK` is not the best name either, and in general passing booleans around makes things a lot less readable. I'd prefer the following over `accountID (ctx, tx, account, true)` anyday:

```go
id, deleted := accountID(ctx, tx, account)
if deleted {
    ... handle
}
```

slabs/manager.go
```diff
@@ -409,3 +413,70 @@ func (m *SlabManager) performSlabMigrations(ctx context.Context) error {
...
+	var accExhausted bool
+	for !accExhausted && ctx.Err() == nil {
+		batch, err := m.store.AccountsForPruning(accBatchSize)
+		if err != nil {
+			return fmt.Errorf("failed to get accounts for pruning: %w", err)
+		} else if len(batch) < accBatchSize {
+			accExhausted = true
+		}
+
+		for _, acc := range batch {
```
peterjan: Haven't looked into it too closely but this feels like it should have a method for fetching all accounts that can be pruned, along with a method on the store called `PruneAccount` that takes an account ID and gets rid of everything that's attached to it in a single transaction.

peterjan: I guess maybe you wanted to reuse existing methods that delete everything associated to one account? In any case I personally feel it should be done in a single transaction. `ListObjects` is quite a wasteful query to get at the object keys we want to delete. Also usually when deleting a large number of rows it's more optimal to delete it with a query as opposed to fetching a list of keys or identifiers in batches to then delete in batches.

peterjan: Feel free to check in with Chris or Nate though if you disagree or feel this would needlessly overcomplicate things.

peterjan (changes requested):

accounts/manager.go
```diff
@@ -70,7 +70,7 @@ type (
 		Account(types.PublicKey) (Account, error)
 		Accounts(offset, limit int, opts ...QueryAccountsOpt) ([]Account, error)
 		HasAccount(types.PublicKey) (bool, error)
-		DeleteAccount(ak types.PublicKey) error
+		DeleteAccount(ak types.PublicKey, soft bool) error
```
ChrisSchinnerl: This shouldn't be necessary. The `Store` should always soft delete.

persist/postgres/accounts.go
```diff
@@ -156,6 +178,40 @@ func (s *Store) UpdateAccount(oldAK, newAK types.PublicKey) error {
 	})
 }
 
+// AccountsForPruning returns up to `limit` soft deleted accounts that have not
+// been entirely deleted yet.
+func (s *Store) AccountsForPruning(limit int) ([]proto.Account, error) {
```
ChrisSchinnerl: I'd restructure this into a `PruneAccounts(limit)` which performs:

1. fetch an account that was deleted 
2. delete up to `limit` objects on it 
3. if there are no more objects and you are still below the limit start deleting slabs
4. if there are no more slabs delete the account

Then every batch by itself is ACID and it's all within a single `store` method that can be called within the `AccountManager` rather than `SlabManager`.

ChrisSchinnerl (changes requested):

persist/postgres/accounts.go
chris124567: OK, I did this except I prune the slabs after we delete the objects each time rather than all at the end.  This way the final call to PruneAccounts won't be very slow compared to the prior calls and as long as we are deleting reasonable numbers of objects at once we shouldn't need to use multiple transactions.  If you think this is worse I can do it the way you proposed.

persist/postgres/accounts.go
```diff
@@ -156,6 +155,77 @@ func (s *Store) UpdateAccount(oldAK, newAK types.PublicKey) error {
 	})
 }
 
+// PruneAccount deletes up to `limit` objects from an account that has been
+// marked as soft deleted.  It will then prune the deleted objects' associated
+// slabs and sectors.  If there are no objects left on the account to delete,
+// it will be hard deleted.  If there are no pending soft deleted accounts,
+// accounts.ErrNotFound is returned.
+func (s *Store) PruneAccount(limit int) error {
```
peterjan: This does the job but it feels a bit sloppy in my opinion. The method is called `PruneAccount` but we don't pass an identifier. Instead we pass a `limit` that doesn't impose itself on accounts per se but rather on objects attached to accounts. There's also no order when we select a deleted account, so we might delete `limit` objects from account `a` and on a consecutive call delete `limit` objects from account `b` instead of continuing to delete the remainder of objects attached to `a`.


I would reintroduce `AccountsForPruning` or `DeletedAccounts` that returns account keys in the order they were deleted, which is possible now that deleted_at is a timestamp. For every account in that loop I would call `store.PruneAccount(accountID)` which would in turn delete associated objects in batches of 100, but I'd loop through the batches until it's done. 

That way if `PruneAccount` returns without error we know it's been pruned and the account was deleted from the store, and we can also neatly log this event perhaps. This will also make it very obvious when we're dealing with a backlog, because the deleted accounts are being pruned in the order they were deleted, which is a nice property imo.


edit: I only just now saw Chris's comment on your PR, I'm fine with keeping it but I would rename it to `PruneAccounts` and I would personally order by `deleted_at` so consecutive calls to this method work on the same account until it's gone.

peterjan (changes requested):

persist/postgres/accounts.go
peterjan: This would also be a good candidate for a benchmark. Perhaps one that plays around with the batch count on objects.

accounts/manager.go
```diff
@@ -246,15 +263,74 @@ func HostFundTarget(host hosts.Host) types.Currency {
...
+				}
+			}
+		}()
+	}
+
+	launch("prune accounts", m.performPruneAccounts)
+	wg.Wait()
+}
+
+func (m *AccountManager) performPruneAccounts(ctx context.Context) error {
```
ChrisSchinnerl: We can drop the `ctx` here since we don't use it anyway.

persist/postgres/accounts.go
ChrisSchinnerl: There is no need to split it up but we should call it `PruneAccounts`. I also agree that this needs a benchmark.

persist/postgres/accounts.go
```diff
@@ -156,6 +155,77 @@ func (s *Store) UpdateAccount(oldAK, newAK types.PublicKey) error {
...
+	if limit < 0 {
+		return errors.New("limit can not be negative")
+	}
+
+	return s.transaction(func(ctx context.Context, tx *txn) error {
+		var accountID int64
+		err := tx.QueryRow(ctx, `SELECT id FROM accounts WHERE deleted_at IS NOT NULL`).Scan(&accountID)
```
ChrisSchinnerl: Order this by timestamp to make the pruning deterministic and add a `LIMIT 1` to make the query more efficient.

persist/postgres/accounts.go
```diff
@@ -156,6 +155,77 @@ func (s *Store) UpdateAccount(oldAK, newAK types.PublicKey) error {
...
+			var objKey types.Hash256
+			if err := rows.Scan((*sqlHash256)(&objKey)); err != nil {
+				return fmt.Errorf("failed to scan object ID: %w", err)
+			}
+			objKeys = append(objKeys, objKey)
+		}
+		if err := rows.Err(); err != nil {
+			return fmt.Errorf("failed to get rows: %w", err)
+		}
+
```
ChrisSchinnerl, line 203: Note that we don't want to delete a full object and all of its slabs at once if it is too big. That is what the limit is for. The limit is a combined limit for the number of objects + slabs we want to delete.

So after finding the account to delete the first thing we do is delete all of the account's objects.
We don't need (or want) `deleteObjects` for that because that creates deletion events which will hang around forever.

Instead we should be able to get away with a simple `DELETE from objects WHERE account_id = $1 LIMIT $2 RETURNING object_key` followed by deleting all those keys from `object_events` as well.

Then you do `limit -= rowsAffected` and if `limit == 0`, you `return nil`.

If `limit > 0` you delete up to `limit` slabs. Only then, if you run out of slabs to delete, delete the account.

----------

The reason for this is that `1 object != 1 object` when it comes to how long it takes to delete it. Since an object can have an arbitrary number of slabs. So we delete the objects first without deleting the slabs and then we delete the slabs.

The only thing that definitely needs to be atomic is deleting from `objects`, `object_slabs` and `object_events`. If an object gets deleted it needs to be deleted without a trace.

Slabs are independent from objects and can temporarily stick around without being pruned alongside the object.

persist/postgres/accounts.go
```diff
@@ -156,6 +155,77 @@ func (s *Store) UpdateAccount(oldAK, newAK types.PublicKey) error {
...
+		// prune slabs and delete the user if we have no objects left
+		if len(objKeys) > 0 {
+			// only delete up to limit objects
+			err := deleteObjects(ctx, tx, accountID, objKeys[:min(len(objKeys), limit)])
+			if err != nil {
+				return fmt.Errorf("failed to delete objects: %w", err)
+			}
+
+			slabIDs, err := getPrunableSlabs(ctx, tx, accountID, math.MaxInt64)
```
ChrisSchinnerl: This also gets easier when you prune all objects first. Because if no objects exist, all slabs for an account are prunable. So this turns into `SELECT id FROM slabs WHERE account_id = $1 LIMIT $2, account, limit)`

persist/postgres/objects.go
ChrisSchinnerl: Definitely unrelated and something we can revisit when we introduce migration events.

ChrisSchinnerl (changes requested):

chris124567:
```
goos: linux
goarch: amd64
BenchmarkPruneAccounts/10-4                   61      32868899 ns/op
BenchmarkPruneAccounts/50-4                   14      74925543 ns/op
BenchmarkPruneAccounts/100-4                   8     129222802 ns/op
BenchmarkPruneAccounts/1000-4                 12     102000705 ns/op
PASS
ok      go.sia.tech/indexd/persist/postgres 869.491s
```

persist/postgres/accounts.go
```diff
@@ -156,6 +155,102 @@ func (s *Store) UpdateAccount(oldAK, newAK types.PublicKey) error {
...
+			if err := rows.Scan(&objKey); err != nil {
+				return fmt.Errorf("failed to scan object ID: %w", err)
+			}
+			objKeys = append(objKeys, objKey)
+		}
+		if err := rows.Err(); err != nil {
+			return fmt.Errorf("failed to get rows: %w", err)
+		}
+
+		_, err = tx.Exec(ctx, `DELETE FROM object_events WHERE account_id = $1 AND object_key = ANY($2)`, accountID, objKeys)
```
peterjan: We should be able to get rid of this step entirely because of the `CASCADE DELETE` on the `account_id`, right? I would definitely take advantage of that, it will simplify the code and should be a noticeable speedup in the benchmarks as well.

persist/postgres/accounts_test.go
```diff
@@ -932,3 +1091,90 @@ func BenchmarkActiveAccounts(b *testing.B) {
...
 }
+
+func BenchmarkPruneAccounts(b *testing.B) {
+	store := initPostgres(b, zap.NewNop())
+
+	hostKeys := make([]types.PublicKey, 30)
+	for i := range hostKeys {
+		hostKeys[i] = types.GeneratePrivateKey().PublicKey()
+		if err := store.UpdateChainState(func(tx subscriber.UpdateTx) error {
+			return tx.AddHostAnnouncement(hostKeys[i], chain.V2HostAnnouncement{{Protocol: quic.Protocol, Address: "[::]:4848"}}, time.Now())
```
peterjan: `nit`  use the helper

```
hk := store.addTestHost(b)
store.addTestContract(b, hk)
```

persist/postgres/accounts_test.go
```diff
@@ -932,3 +1091,90 @@ func BenchmarkActiveAccounts(b *testing.B) {
...
+			pk := types.GeneratePrivateKey().PublicKey()
+			store.addTestAccount(b, pk)
+
+			for range frand.Intn(10) {
+				pinObject(proto.Account(pk))
+			}
+
+			// delete 1/10 accounts
+			if i%10 == 0 {
+				if err := store.DeleteAccount(pk); err != nil {
```
peterjan: I would update `store.DeleteAccount` to take a `proto.Account` key and bubble that up all the way to the admin API. It's the same thing but it's a little annoying in the code because all other store methods that deal with an account expect a `proto.Account`.

persist/postgres/accounts.go
```diff
@@ -156,6 +155,102 @@ func (s *Store) UpdateAccount(oldAK, newAK types.PublicKey) error {
...
+// account to delete, it will prune the associated slabs and sectors.  If there
+// are no slabs left it will hard delete the account.  If there are no pending
+// soft deleted accounts, accounts.ErrNotFound is returned
+func (s *Store) PruneAccounts(limit int) error {
+	if limit < 0 {
+		return errors.New("limit can not be negative")
+	}
+
+	return s.transaction(func(ctx context.Context, tx *txn) error {
+		getAccountSlabs := func(accountID int64, limit int) ([]int64, error) {
```
peterjan: This helper is only called once, I would inline it because it's confusing to have it here at the top

persist/postgres/accounts.go
```diff
@@ -156,6 +155,102 @@ func (s *Store) UpdateAccount(oldAK, newAK types.PublicKey) error {
...
+		if err != nil {
+			return fmt.Errorf("failed to delete object events: %w", err)
+		}
+
+		limit -= len(objKeys)
+		if limit == 0 {
+			return nil
+		}
+
+		slabIDs, err := getAccountSlabs(accountID, limit)
```
peterjan: I don't think this is the way we want to handle the `limit`. We want to batch remove associated objects in an ACID fashion. Meaning that after every batch the resulting state should have intact objects, just less of them. 

Our schema is working against you here, unfortunately. For every `object` you delete, you have to get at the `slabID`, you can access that through `object_slabs` but that'll only get you the slab digest, from there you need to get the `slabID` to pass it to `unpinSlabs`.

The `object_slabs` should sort themselves out because of the `CASCADE DELETE`, same with `object_events`.

persist/postgres/accounts_test.go
```diff
@@ -932,3 +1091,90 @@ func BenchmarkActiveAccounts(b *testing.B) {
 		}
 	}
 }
+
+func BenchmarkPruneAccounts(b *testing.B) {
```
peterjan, line 1075: While reviewing I played around with some parameters, seeing as I have it anyway I thought I would share it here

```
func BenchmarkPruneAccounts(b *testing.B) {
	store := initPostgres(b, zap.NewNop())

	const (
		numHosts    = 30
		numAccounts = 100
		numObjects  = 1000
		numSlabs    = 2
	)

	var hks []types.PublicKey
	for range numHosts {
		hks = append(hks, store.addTestHost(b))
		store.addTestContract(b, hks[len(hks)-1])
	}

	var accs []types.PublicKey
	for range numAccounts {
		sk := types.GeneratePrivateKey()
		store.addTestAccount(b, sk.PublicKey())
		accs = append(accs, sk.PublicKey())
	}

	for i, acc := range accs {
		for range numObjects {
			obj := slabs.SealedObject{
				EncryptedMasterKey: frand.Bytes(72),
				EncryptedMetadata:  []byte("hello world"),
				Signature:          (types.Signature)(frand.Bytes(64)),
			}

			var slabParams []slabs.SlabPinParams
			for range numSlabs {
				s := slabs.SlabPinParams{
					MinShards:     3,
					EncryptionKey: frand.Entropy256(),
					Sectors:       make([]slabs.PinnedSector, 30),
				}
				for i := range s.Sectors {
					s.Sectors[i].HostKey = hks[i]
					s.Sectors[i].Root = frand.Entropy256()
				}
				slabParams = append(slabParams, s)
			}

			slabIDs, err := store.PinSlabs(proto.Account(acc), time.Time{}, slabParams...)
			if err != nil {
				b.Fatal(err)
			}

			for _, slabID := range slabIDs {
				obj.Slabs = append(obj.Slabs, slabs.SlabSlice{
					SlabID: slabID,
					Offset: 0,
					Length: 256,
				})
			}

			err = store.SaveObject(proto.Account(acc), obj)
			if err != nil {
				b.Fatal(err)
			}
		}
		if err := store.DeleteAccount(acc); err != nil {
			b.Fatal(err)
		}
		if i > 0 && i%10 == 0 {
			b.Logf("prepared %d/%d accounts", i, len(accs))
		}
	}

	for _, limit := range []int{100, 250, 500} {
		b.Run(fmt.Sprint(limit), func(b *testing.B) {
			for b.Loop() {
				if err := store.PruneAccounts(limit); errors.Is(err, accounts.ErrNotFound) {
					b.Logf("pruning error: %v", err)
				} else if err != nil {
					b.Fatal(err)
				}
			}
		})
	}
}
```

Please note though this is not good to copy paste or merge. The setup takes too long and I'm not quite happy with the parameters. The results however do indicate we are not quite there yet with the pruning code:

```
BenchmarkPruneAccounts/100-10                  5         206893717 ns/op           10115 B/op        223 allocs/op
BenchmarkPruneAccounts/250-10                  3         382462556 ns/op           11154 B/op        260 allocs/op
BenchmarkPruneAccounts/500-10                  4         307264917 ns/op           10292 B/op        242 allocs/op
```

and that's an on M5

peterjan (changes requested):

persist/postgres/accounts_test.go
```diff
@@ -932,3 +1091,90 @@ func BenchmarkActiveAccounts(b *testing.B) {
...
+		}); err != nil {
+			b.Fatal(err)
+		}
+		store.addTestContract(b, hostKeys[i])
+	}
+
+	pinObject := func(acc proto.Account) {
+		b.Helper()
+
+		obj := slabs.SealedObject{
```
peterjan: `obj` is unused here, I think you are missing a call to `store.SaveObject` so objects aren't even making it into the database

persist/postgres/accounts.go
ChrisSchinnerl: This is fine. The objects, including their object events and object slabs are deleted atomically. 
The slabs and associated tables can be deleted separately. 

This is probably also the only way to cleanly do it considering that 2 objects mights reference the same slab. So you can't delete that slab before deleting both objects anyway.

persist/postgres/accounts.go
ChrisSchinnerl: Right we have a cascade there. That makes things simpler.

persist/postgres/accounts_test.go
chris124567: The first query in `unpinSlabs` had an unnecessary `USING` that I think was slowing things down a lot.  100 seems reasonable for a limit.  There seems to be more variability with higher limits.

With 1000 accounts, 500 objects per account, 5 slabs per object:

```
goos: linux
goarch: amd64
BenchmarkPruneAccounts
BenchmarkPruneAccounts/100
BenchmarkPruneAccounts/100-4                 135       8820641 ns/op
BenchmarkPruneAccounts/250
BenchmarkPruneAccounts/250-4                   9     289581152 ns/op
BenchmarkPruneAccounts/500
BenchmarkPruneAccounts/500-4                   1    1513515212 ns/op
PASS
ok      go.sia.tech/indexd/persist/postgres 3511.300s
```

```
goos: linux
goarch: amd64
BenchmarkPruneAccounts
BenchmarkPruneAccounts/100
BenchmarkPruneAccounts/100-4                  56      22028707 ns/op
BenchmarkPruneAccounts/250
BenchmarkPruneAccounts/250-4                  60      19723378 ns/op
^C
```

accounts/manager.go
```diff
@@ -246,15 +263,74 @@ func HostFundTarget(host hosts.Host) types.Currency {
...
+	}()
+
+	return m, nil
+}
+
+// maintenanceLoop performs any background tasks that the accounts manager
+// needs to perform on accounts
+func (m *AccountManager) maintenanceLoop(ctx context.Context) {
+	var wg sync.WaitGroup
+	launch := func(descr string, task func() error) {
```
peterjan: defining a `launch` function is not necessary in this loop because there's only one job to perform

accounts/manager.go
```diff
@@ -246,15 +263,74 @@ func HostFundTarget(host hosts.Host) types.Currency {
...
+	start := time.Now()
+	log := m.log.Named("prune")
+	log.Debug("starting account pruning")
+
+	const objectBatchSize = 100
+	for {
+		if err := m.store.PruneAccounts(objectBatchSize); errors.Is(err, ErrNotFound) {
+			break
+		} else if err != nil {
+			return fmt.Errorf("failed to prune account: %w", err)
```
peterjan:
```suggestion
			return fmt.Errorf("failed to prune accounts: %w", err)
```

persist/postgres/accounts.go
```diff
@@ -156,6 +155,96 @@ func (s *Store) UpdateAccount(oldAK, newAK types.PublicKey) error {
...
+RETURNING o.object_key;`, accountID, limit)
+		if err != nil {
+			return fmt.Errorf("failed to delete objects: %w", err)
+		}
+
+		var objKeys []sqlHash256
+		for rows.Next() {
+			var objKey sqlHash256
+			if err := rows.Scan(&objKey); err != nil {
+				return fmt.Errorf("failed to scan object ID: %w", err)
```
peterjan:
```suggestion
				return fmt.Errorf("failed to scan object key: %w", err)
```

persist/postgres/accounts_test.go
```diff
@@ -793,6 +795,156 @@ func TestServiceAccounts(t *testing.T) {
...
+			ss = append(ss, slabs.SlabSlice{
+				SlabID: ids[0],
+				Offset: 10,
+				Length: 120,
+			})
+		}
+		return ss
+	}
+
+	randomObject := func(s []slabs.SlabSlice) slabs.SealedObject {
```
peterjan: These 40 lines are copied directly from `object_test`, we could use `store.newTestObject` that returns a `SealedObject`

persist/postgres/accounts_test.go
```diff
@@ -932,3 +1086,83 @@ func BenchmarkActiveAccounts(b *testing.B) {
...
+
+					batch.Queue(`INSERT INTO slabs(digest, encryption_key, min_shards) VALUES ($1, $2, 1);`, slabDigest, sqlHash256(encryptionKey))
+					slabID++
+
+					batch.Queue(`INSERT INTO account_slabs(account_id, slab_id) VALUES ($1, $2)`, accountID, slabID)
+					batch.Queue(`INSERT INTO object_slabs(object_id, slab_digest, slab_index, slab_offset, slab_length) VALUES ($1, $2, $3, 0, 0)`, objectID, slabDigest, k)
+				}
+			}
+
+			// delete 1/10 accounts
```
peterjan: I don't really mind this line but it doesn't really contribute much other than make 90% of the test setup useless because it's not linked to a deleted account. This shouldn't have any effect on performance right?

persist/postgres/accounts_test.go
peterjan: `ok      go.sia.tech/indexd/persist/postgres 3511.300s` does this benchmark take 1h to complete on your laptop? That's a little excessive 😅  we run all benchmarks in a job on master. Why do you have to `reset` your database setup between runs? With those parameters you should be able to run all batches without running out of objects to delete. You can safely delete 100% of the accounts imo, it doesn't make any difference to our query whether everything is "marked to be deleted" yes or no.

persist/postgres/accounts_test.go
```diff
@@ -932,3 +1086,83 @@ func BenchmarkActiveAccounts(b *testing.B) {
...
+			if accountID%10 == 0 {
+				b.Log("Deleting:", accountID)
```
peterjan: This log line is meaningless because you batch all queries

peterjan (changes requested):
I'm only requesting changes because `BenchmarkPruneAccounts` takes way too long to finish. I left some small comments, perhaps we can take those along too. Overall looks good to me though, nice!

persist/postgres/accounts_test.go
peterjan: Hm ok I ran it after reviewing your PR and it's a lot faster on my machine

```
BenchmarkPruneAccounts/100-10                421           2545879 ns/op           38635 B/op        771 allocs/op
... bunch of delete logs
BenchmarkPruneAccounts/250-10                319           3730167 ns/op           80438 B/op       1674 allocs/op
... bunch of delete logs
BenchmarkPruneAccounts/500-10                193           7894731 ns/op          176019 B/op       3035 allocs/op
PASS
ok      go.sia.tech/indexd/persist/postgres     418.444s
```

I'll leave it up to Chris to decide. I would do the setup once and delete all accounts personally. 
If I do that, it runs perfectly fine on my machine (without pruning err logs) and finishes in 90s. 

```
cpu: Apple M5
BenchmarkPruneAccounts
BenchmarkPruneAccounts/100
BenchmarkPruneAccounts/100-10                502           2425885 ns/op           39195 B/op        778 allocs/op
BenchmarkPruneAccounts/250
BenchmarkPruneAccounts/250-10                285           3854061 ns/op           81584 B/op       1687 allocs/op
BenchmarkPruneAccounts/500
BenchmarkPruneAccounts/500-10                242           5013548 ns/op          175598 B/op       3001 allocs/op
PASS
ok      go.sia.tech/indexd/persist/postgres     96.830s
```

persist/postgres/accounts_test.go
chris124567: The accounts need to be soft deleted for them to be considered by `PruneAccounts`.  Alternatively, we could set deleted_at=NOW() in the `INSERT INTO accounts` query.

persist/postgres/accounts_test.go
chris124567: Alright I removed the resetting.  I think my CPU and or postgres verison might be slower than yours but if it's that fast on your machine I guess it's OK.
 
```
goos: linux
goarch: amd64
BenchmarkPruneAccounts
BenchmarkPruneAccounts/100
BenchmarkPruneAccounts/100-4                 154       7637516 ns/op
BenchmarkPruneAccounts/250
BenchmarkPruneAccounts/250-4                  75      13692975 ns/op
BenchmarkPruneAccounts/500
BenchmarkPruneAccounts/500-4                  92     372282401 ns/op
PASS
ok      go.sia.tech/indexd/persist/postgres 782.860s
```

persist/postgres/objects_test.go
```diff
@@ -19,6 +19,15 @@ import (
 	"lukechampine.com/frand"
 )
 
+func (s *Store) testRandomObject(ss []slabs.SlabSlice) slabs.SealedObject {
```
peterjan: Why is this defined on the store? I was suggesting to combine `pinSlabs` and `randomSlabs` in a method called `newTestObject` or something that generates a random object with a requested number of slabs for a certain account, pins it, saves the sealed object and returns it.

chris124567: I misread your message initially.  There was no reason for it to be on the store.  There is a reason now that I just updated it to save the object.  Some of the tests use the slabs in a different enough way that I think pinning the slabs in the helper is more trouble than it's worth.

peterjan: Hm ok, that's unfortunate but thanks for checking it out!

persist/postgres/accounts_test.go
peterjan: Cool. Yeah an hour is just a really long time :D

peterjan (approved):

ChrisSchinnerl (approved):

PR #703, opened 2025-11-30
Race migrations

n8mgr:
Prevents slow hosts from slowing down migrations by periodically spawning additional shard downloads when there are open semaphore slots.

### DIFF
--- a/slabs/downloads.go
+++ b/slabs/downloads.go
@@ -94,12 +94,14 @@ func (m *SlabManager) downloadShards(ctx context.Context, slab Slab, log *zap.Lo
 		return nil
 	}
 
+	if len(candidates) < int(slab.MinShards) {
+		return nil, fmt.Errorf("only %d available sectors, minimum required: %d: %w", len(candidates), slab.MinShards, errNotEnoughShards)
+	}
+
 	// start initial shards
 	failedCh := make(chan struct{}, slab.MinShards)
-	initial := min(len(candidates), int(slab.MinShards))
-	log.Debug("starting initial shard downloads", zap.Int("count", initial), zap.Int("required", int(slab.MinShards)), zap.Int("totalCandidates", len(candidates)))
 initialLoop:
-	for _, hostKey := range candidates[:initial] {
+	for _, hostKey := range candidates[:int(slab.MinShards)] {
 		select {
 		case <-ctx.Done():
 			break initialLoop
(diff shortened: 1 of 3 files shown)


slabs/downloads.go
```diff
@@ -50,64 +50,105 @@ func (m *SlabManager) downloadShards(ctx context.Context, slab Slab, log *zap.Lo
...
-			// debit the service account for the read since the host may charge for it
-			// even if it is cancelled quickly. This is best effort, it's fine to
-			// log the error and continue on failure.
-			cost := prices.RPCReadSectorCost(proto.SectorSize).RenterCost()
-			if err = m.am.DebitServiceAccount(context.Background(), hostKey, m.migrationAccount, cost); err != nil {
-				log.Debug("failed to debit service account for sector read", zap.Error(err))
-			}
+	// start initial shards
+	failedCh := make(chan struct{}, slab.MinShards)
+	initial := min(len(candidates), int(slab.MinShards))
```
peterjan: You should be able to return early if there aren't enough candidates

slabs/downloads.go
```diff
@@ -50,64 +50,105 @@ func (m *SlabManager) downloadShards(ctx context.Context, slab Slab, log *zap.Lo
...
-		}()
+	t := time.NewTicker(m.shardTimeout / 2)
+	defer t.Stop()
+raceLoop:
+	for i := initial; downloaded.Load() < uint32(slab.MinShards) && i < len(candidates); i++ {
+		hostKey := candidates[i]
+		select {
+		case <-ctx.Done():
+			break raceLoop
+		case <-failedCh:
```
peterjan, line 133: Could argue we need to reset the ticker here to avoid racing a slow shard immediately after launching a new candidate upon failure

slabs/downloads.go
```diff
@@ -50,64 +50,105 @@ func (m *SlabManager) downloadShards(ctx context.Context, slab Slab, log *zap.Lo
...
 			}
+		})
+	}
 
-			shards[sector.index] = buf.Bytes()
-			if n := downloaded.Add(1); n >= uint32(slab.MinShards) {
-				cancel()
-			}
-		}()
+	t := time.NewTicker(m.shardTimeout / 2)
```
peterjan: Shard timeout is 2m so this is still quite slow to kick in?

peterjan (dismissed):
Left some small comments but overall def. looks good to go 👍

ChrisSchinnerl (dismissed):

PR #708, opened 2025-12-02
Fix TestApplicationAPI NDF

chris124567:
Bumped into this a few times recently


peterjan (approved):
Did you run it in a loop on the CI? That's something I usually do on these Fix NDF PRs. More often than not I end up tweaking my fix until it passes 250 times with ease. Approved it though

ChrisSchinnerl (approved):

PR #714, opened 2025-12-04
Use fallback resolver if OS fails when getting host IP addresses

### LINKED ISSUE #709: Use nomad's fallback when resolving hostnames
nomad falls back to Cloudflare if the local resolver fails. We should use that in indexd to make lookups more robust.

chris124567:
Close #709

### DIFF
--- a/hosts/manager.go
+++ b/hosts/manager.go
@@ -200,13 +200,21 @@ func (hm *HostManager) UnblockHost(ctx context.Context, hk types.PublicKey) erro
 
 // NewManager creates a new host manager.
 func NewManager(syncer Syncer, locator Locator, client HostClient, store Store, opts ...Option) (*HostManager, error) {
+	// uses Cloudflare 1.1.1.1 instead of OS default
+	fallbackResolver := &net.Resolver{
+		PreferGo: true,
+		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+			return (&net.Dialer{}).DialContext(ctx, network, "1.1.1.1:53")
+		},
+	}
+
 	m := &HostManager{
 		announcementMaxAge: time.Hour * 24 * 365,
 		scanFrequency:      time.Hour,
 		scanInterval:       time.Hour * 24,
 
 		onlineChecker: &onlineChecker{addresses: fallbackSites, syncer: syncer},
-		resolver:      &net.Resolver{},
+		resolver:      &resolver{main: &net.Resolver{}, fallback: fallbackResolver},
 		scanner:       &scanner{},
 		locator:       locator,
 		store:         store,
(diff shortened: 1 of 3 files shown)


hosts/manager.go
```diff
@@ -200,13 +200,21 @@ func (hm *HostManager) UnblockHost(ctx context.Context, hk types.PublicKey) erro
 
 // NewManager creates a new host manager.
 func NewManager(syncer Syncer, locator Locator, client HostClient, store Store, opts ...Option) (*HostManager, error) {
+	// uses Cloudflare 1.1.1.1 for when OS resolver fails
+	fallbackResolver := &net.Resolver{
+		PreferGo: true,
+		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+			return (&net.Dialer{}).DialContext(ctx, network, "1.1.1.1:53")
```
peterjan, line 209: `nit` might want to reuse the dialer here instead of allocating a new one on each dial

hosts/manager.go
```diff
@@ -200,13 +200,21 @@ func (hm *HostManager) UnblockHost(ctx context.Context, hk types.PublicKey) erro
 
 // NewManager creates a new host manager.
 func NewManager(syncer Syncer, locator Locator, client HostClient, store Store, opts ...Option) (*HostManager, error) {
+	// uses Cloudflare 1.1.1.1 for when OS resolver fails
+	fallbackResolver := &net.Resolver{
+		PreferGo: true,
```
peterjan, line 207: Maybe drop a small comment indicating why this has to be `true`. It's because it forces usage of the `Dial` function we specify here right? Otherwise it might use the systems resolver? Docs are a little unclear

hosts/manager.go
ChrisSchinnerl: This should be fine. It's a common pattern that we use ourselves in other places. `net.Dial` itself creates a new one every time as well.

hosts/utils_test.go
```diff
@@ -42,3 +46,67 @@ func TestOnlineChecker(t *testing.T) {
...
+type mockResolverFallback struct {
+	addrs []net.IPAddr
+	err   error
+}
+
+func (m *mockResolverFallback) LookupIPAddr(ctx context.Context, host string) ([]net.IPAddr, error) {
+	return m.addrs, m.err
+}
+
+func TestResolver_LookupIPAddr(t *testing.T) {
```
ChrisSchinnerl:
```suggestion
func TestResolver(t *testing.T) {
```

That should be fine. No need to introduce a new naming convention for tests.

hosts/utils_test.go
```diff
@@ -42,3 +46,67 @@ func TestOnlineChecker(t *testing.T) {
...
+	ctx := context.Background()
```
ChrisSchinnerl: Make use of `t.Context()` instead and inline it. It's only used in one spot.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

n8mgr (approved):

PR #715, opened 2025-12-05
Fixed incorrect version number in `go.mod`

skunk-ink:
Running `go build ./cmd/indexd` will produce:

```
go: download go1.25 for linux/amd64: toolchain not available
```

This is due to the incorrect version format in `go.mod`. The version number was `1.25` but should be `1.25.0`.

This PR corrects the mistake.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module go.sia.tech/indexd
 
-go 1.25.0
+go 1.25
+
+toolchain go1.25.5
 
 require (
 	github.com/jackc/pgerrcode v0.0.0-20240316143900-6e2875d9b438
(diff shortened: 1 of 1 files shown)


ChrisSchinnerl:
What version of Go did you have installed? If you look at the `go.mod` [reference](https://go.dev/doc/modules/gomod-ref). the version should be fine. In fact we usually don't set it ourselves but it is automatically set whenever an underlying dependency requires an upgrade.

skunk-ink:
```
user@ubuntu:~/docker/indexd$ go version
go version go1.25.0 linux/amd64
```

n8mgr:
@ChrisSchinnerl we should probably just bite the bullet and finally add the silly toolchain line...

```
go 1.25
toolchain go1.25.5
```

PR #720, opened 2025-12-06
Remove service accounts from database and store them only in memory

### LINKED ISSUE #624: Remove service accounts from database and store them only in memory
Service accounts seem to turn out to be just different enough to warrant special handling. Considering that there are only 2-3 of them in total, we should be able to track them in ram only and add them to the front of the list returned by the database when fetching accounts for funding as needed. This will also allow us to enforce a `NOT NULL` constraint between accounts and app connect keys.

chris124567:
Close #624

### DIFF
--- a/accounts/account.go
+++ b/accounts/account.go
@@ -58,14 +58,6 @@ type (
 	QueryAccountsOpt func(o *QueryAccountsOptions)
 )
 
-// WithServiceAccount sets the service account filter for querying accounts.
-// Defaults to all accounts.
-func WithServiceAccount(serviceAccount bool) QueryAccountsOpt {
-	return func(opt *QueryAccountsOptions) {
-		opt.ServiceAccount = &serviceAccount
-	}
-}
-
 // WithConnectKey filters the accounts by the connect key they are associated
 // with.
 func WithConnectKey(connectKey string) QueryAccountsOpt {
--- a/accounts/account.go
+++ b/accounts/account.go
@@ -17,10 +17,6 @@ var (
 	// account not being found.
 	ErrNotFound = errors.New("account not found")
 
-	// ErrServiceAccount is returned by operations that fail due to an account
-	// being a service account.
-	ErrServiceAccount = errors.New("account is a service account")
-
 	// ErrAccountStorageLimitExceeded is returned when an operation fails due
 	// to the account exceeding its storage limit.  We use the term "app
 	// storage limit" here because from the user's perspective, they will have
@@ -50,8 +46,7 @@ func WithMaxPinnedData(maxPinnedData uint64) AddAccountOption {
 type (
 	// QueryAccountsOptions holds options for querying accounts.
 	QueryAccountsOptions struct {
-		ServiceAccount *bool
-		ConnectKey     *string
+		ConnectKey *string
 	}
 
 	// QueryAccountsOpt is a functional option for querying accounts.
@@ -69,13 +64,12 @@ func WithConnectKey(connectKey string) QueryAccountsOpt {
 type (
 	// Account represents an account in the indexer.
 	Account struct {
-		AccountKey     proto.Account `json:"accountKey"`
-		ConnectKey     *string       `json:"connectKey,omitempty"`
-		ServiceAccount bool          `json:"serviceAccount"`
-		MaxPinnedData  uint64        `json:"maxPinnedData"`
-		PinnedData     uint64        `json:"pinnedData"`
-		App            AppMeta       `json:"app,omitempty"`
-		LastUsed       time.Time     `json:"lastUsed"`
+		AccountKey    proto.Account `json:"accountKey"`
+		ConnectKey    *string       `json:"connectKey,omitempty"`
+		MaxPinnedData uint64        `json:"maxPinnedData"`
+		PinnedData    uint64        `json:"pinnedData"`
+		App           AppMeta       `json:"app,omitempty"`
+		LastUsed      time.Time     `json:"lastUsed"`
 	}
 
 	// HostAccount represents an ephemeral account on a host.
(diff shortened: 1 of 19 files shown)


accounts/manager.go
```diff
@@ -188,6 +188,23 @@ func (m *AccountManager) FundAccounts(ctx context.Context, host hosts.Host, cont
...
+			HostKey:    host.PublicKey,
+		})
+	}
+
+	if len(serviceAccounts) > 0 {
+		// fund them
+		_, _, err := m.funder.FundAccounts(ctx, host, contractIDs, serviceAccounts, fundTarget, log)
+		if err != nil {
+			return fmt.Errorf("failed to fund service accounts: %w", err)
+		}
```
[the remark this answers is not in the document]
chris124567: fixed

accounts/serviceaccounts.go
```diff
@@ -53,6 +55,12 @@ func (m *AccountManager) serviceAccountExists(account proto.Account) bool {
 	return exists
 }
 
+func (m *AccountManager) serviceAccs() []proto.Account {
```
peterjan: You range over this return value to build a list of HostAccounts so you might as well make this a function that receives a host key and returns the host accounts directly.

accounts/account.go
```diff
@@ -58,14 +58,6 @@ type (
 	QueryAccountsOpt func(o *QueryAccountsOptions)
 )
 
-// WithServiceAccount sets the service account filter for querying accounts.
-// Defaults to all accounts.
-func WithServiceAccount(serviceAccount bool) QueryAccountsOpt {
-	return func(opt *QueryAccountsOptions) {
-		opt.ServiceAccount = &serviceAccount
```
peterjan, line 65: Need to remove this option on `QueryAccountsOptions`

api/admin/admin.go
```diff
@@ -420,13 +420,6 @@ func (a *admin) handleGETAccounts(jc jape.Context) {
 		return
 	}
 	var opts []accounts.QueryAccountsOpt
-	if jc.Request.FormValue("serviceaccount") != "" {
-		var serviceAccount bool
-		if jc.DecodeForm("serviceaccount", &serviceAccount) != nil {
```
peterjan, line 425: Also remove from docs

accounts/manager.go
```diff
@@ -176,18 +176,35 @@ func (m *AccountManager) FundAccounts(ctx context.Context, host hosts.Host, cont
...
+	for _, serviceAccount := range m.serviceAccs() {
+		serviceAccounts = append(serviceAccounts, HostAccount{
+			AccountKey: serviceAccount,
+			HostKey:    host.PublicKey,
+		})
+	}
+
+	if len(serviceAccounts) > 0 {
+		// fund them
+		_, _, err := m.funder.FundAccounts(ctx, host, contractIDs, serviceAccounts, fundTarget, log)
```
peterjan: Why discard `funded`? We have to take that into account when updating the service account balances

peterjan (changes requested):

persist/postgres/init.sql
```diff
@@ -91,10 +90,10 @@ CREATE TABLE account_hosts (
 CREATE INDEX account_hosts_host_id_next_fund_idx ON account_hosts (host_id, next_fund);
 
 CREATE TABLE service_accounts (
```
ChrisSchinnerl: Not sure if @n8mgr feels the same but I'd move the `service_accounts` table into memory. 
At 80 or so bytes per account we are looking at 240kib for 3 accounts on 1000 hosts. 

After a reboot, we can assume that the balance is 0SC until the first time we have called Replenish on the account which we will do soon enough.

n8mgr: That’s fine with me

peterjan (dismissed):

accounts/serviceaccounts.go
```diff
@@ -26,15 +26,16 @@ func (m *AccountManager) ResetAccountBalance(ctx context.Context, hostKey types.
...
 		return ErrNotFound
 	}
-	return m.store.UpdateServiceAccountBalance(hostKey, account, types.ZeroCurrency)
+	m.serviceAccounts[account][hostKey] = types.ZeroCurrency
+	return nil
 }
 
 // ServiceAccountBalance returns the balance of a locked service account.
 func (m *AccountManager) ServiceAccountBalance(ctx context.Context, hostKey types.PublicKey, account proto.Account) (types.Currency, error) {
 	if !m.serviceAccountExists(account) {
```
ChrisSchinnerl: All the methods we call `m.serviceAccountExists` in have broken locking now.

We should delete `m.serviceAccountExists` since it no longer saves us any lines and instead lock + defer unlock `m.serviceAccountsMu` in all the exposed methods directly. 

Then things are nice and atomic.

persist/postgres/migrations.go
```diff
@@ -574,4 +574,21 @@ ALTER TABLE app_connect_keys ADD CONSTRAINT app_connect_keys_user_secret_key UNI
...
+		_, err := tx.Exec(ctx, `
+ALTER TABLE service_accounts ADD COLUMN public_key BYTEA;
+
+ALTER TABLE service_accounts ADD CONSTRAINT service_accounts_public_key_length CHECK (LENGTH(public_key) = 32);
+UPDATE service_accounts sa
+SET public_key = a.public_key
+FROM accounts a
+WHERE sa.account_id = a.id;
+
+ALTER TABLE accounts DROP COLUMN service_account;
```
ChrisSchinnerl, line 580: Unnecessary migration code above this line.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

n8mgr (approved):

peterjan (approved):

Issue #683, opened 2025-11-20
Split Data and Metadata encryption


n8mgr:
Currently, user metadata is also shared when sharing files. This may not be ideal as there may be private information specific to the user in the metadata. After looking at removing it, I realized that if we share files we would be leaking the metadata encryption key even if we weren't leaking the metadata. 

I am suggest updating the sealed object struct to encrypt the metadata separately to avoid always leaking the key when sharing:

```go
// A SealedObject is an object that has been locked with an app key.
// It can be safely serialized and shared, but cannot be used to access
// the underlying data until it has been unlocked with the app key.
SealedObject struct {
	EncryptedDataKey []byte      `json:"encryptedDataKey"`
	Slabs            []SlabSlice `json:"slabs"`
	// DataSignature is a signature of the blake2b(object ID, encrypted_data_key)
	// to attest that the object data has not been tampered with. This attestation
	// is separate from the metadata attestation to allow for validating the
	// encryption key when sharing the object which does not include the metadata.
	DataSignature types.Signature `json:"dataSignature"`
	
	EncryptedMetadataKey []byte `json:"encryptedMetadataKey"`
	EncryptedMetadata    []byte `json:"encryptedMetadata"`
	// MetadataSignature is a signature of the blake2b(object ID, metadata key and encrypted metadata)
	// to attest that the metadata has not been tampered with.
	MetadataSignature types.Signature `json:"metadataSignature"`

	CreatedAt time.Time `json:"createdAt"`
	UpdatedAt time.Time `json:"updatedAt"`
}
```

ChrisSchinnerl:
I got a few questions:

1. Should this be optional when sharing? So depending on whether you want to give access to the metadata you'd also include the key in the link? I could see reasons why you'd want to share the metadata as well. e.g. you share a picture and want to also share the associated location data or timestamp.
2. What was the reason for including a signature rather than using ChaCha20 with poly1305 to implicitly get authentication again? Seems like the only reason would be separating the used keys but maybe that's obsolete now?
3. I assume we want to do both now? Drop the metadata in the server response + have a separate key?

n8mgr:
We have a signature so the encryption key can be authenticated. An HMAC validates that the encryption key is encrypted correctly. A signature attestation validates that the user set it not the server.

alexfreska:
[...]

PR #723, opened 2025-12-09
Split object encryption

ChrisSchinnerl:
Fixes https://github.com/SiaFoundation/indexd/issues/683

### DIFF
--- a/sdk/object.go
+++ b/sdk/object.go
@@ -94,8 +94,7 @@ func (o *Object) Metadata() json.RawMessage {
 	return slices.Clone(o.metadata)
 }
 
-// UpdateMetadata updates the object's metadata and generates a new metadata key
-// if necessary.
+// UpdateMetadata updates the object's metadata.
 func (o *Object) UpdateMetadata(meta json.RawMessage) {
 	o.metadata = slices.Clone(meta)
 }
--- a/sdk/object.go
+++ b/sdk/object.go
@@ -54,7 +54,7 @@ func (o *Object) Seal(appKey types.PrivateKey) slabs.SealedObject {
 	}
 	encryptedDataKey := seal(dataKeyCipher(appKey, objectID), o.dataKey)
 	encryptedMetaKey := seal(metadataKeyCipher(appKey, objectID), o.metaDataKey)
-	encryptedMetadata := seal(metadataCipher(o.metaDataKey, objectID), o.metadata)
+	encryptedMetadata := seal(metadataCipher(o.metaDataKey), o.metadata)
 
 	so := slabs.SealedObject{
 		EncryptedDataKey:     encryptedDataKey,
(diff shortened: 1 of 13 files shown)


slabs/objects.go
```diff
@@ -116,16 +130,50 @@ func ObjectID(slabs []SlabSlice) types.Hash256 {
...
+}
+
+// MetaSigHash returns the hash used for signing/verifying the metadata
+// signature. It covers the slabs through the object ID, the encrypted metadata
+// key, and the encrypted metadata.
+func (so *SealedObject) MetaSigHash() types.Hash256 {
 	h := types.NewHasher()
 	so.ID().EncodeTo(h.E)
-	h.E.Write(so.EncryptedMasterKey)
+	h.E.Write(so.EncryptedMetadataKey)
```
ChrisSchinnerl, line 148: I'd like to highlight this. In the issue it says `metadata key` but I wasn't sure if that was a typo.
If we want to use the plain key here we need to first decrypt the metadata key before we can verify that the metadata hasn't been tempered with.

Similar consideration is true for encrypted metadata vs metadata.

slabs/objects.go
```diff
@@ -19,12 +19,19 @@ type (
...
-		// Signature is a signature of the blake2b(object ID, encrypted_master_key, and encrypted_metadata)
+		EncryptedDataKey []byte      `json:"encryptedDataKey"`
+		Slabs            []SlabSlice `json:"slabs"`
+		// DataSignature is a signature of the blake2b(object ID, encrypted_data_key)
+		// to attest that the object data has not been tampered with. This attestation
+		// is separate from the metadata attestation to allow for validating the
+		// encryption key when sharing the object which does not include the metadata.
+		DataSignature types.Signature `json:"dataSignature"`
+
+		EncryptedMetadataKey []byte `json:"encryptedMetadataKey,omitempty"`
```
peterjan, line 30: Should these be `omitempty`? I understand the metadata is optional but if you look at `Object.VerifySignatures` it verifies both signatures, meaning both are required or it would be impossible to save an object. Looking at `Object.Seal` both are set regardless of whether the metadata is empty or not.

peterjan: Updating this would also have repercussion on the database level, we can make those fields `NOT NULL` in the case we want to update it. In the case we want it to be optional we probably want to update both `Seal` and `VerifySignature` plus add an app test to check we can save a sealed object both with and without metadata.

peterjan (changes requested):

slabs/objects.go
ChrisSchinnerl: Yeah we definitely want the signature to always exist. I updated that. That way we can attest that the metadata is in fact supposed to be empty. 

The metadata key only really needs to exist if there is metadata.

slabs/objects.go
n8mgr: I think you're right, we should use the encrypted key not the plain key

sdk/object.go
```diff
@@ -84,7 +94,8 @@ func (o *Object) Metadata() json.RawMessage {
 	return slices.Clone(o.metadata)
 }
 
-// UpdateMetadata updates the object's metadata.
+// UpdateMetadata updates the object's metadata and generates a new metadata key
+// if necessary.
```
n8mgr: nit: it does not

n8mgr (approved):

peterjan (approved):

n8mgr (approved):

PR #724, opened 2025-12-10
Use unified host client in account funding

### LINKED ISSUE #718: Use unified host client in account funding

chris124567:
Close #718

### DIFF
--- a/accounts/manager_test.go
+++ b/accounts/manager_test.go
@@ -36,35 +36,6 @@ type testStore struct {
 	testutils.TestStore
 }
 
-func (s testStore) resetNextFund(t testing.TB) {
-	t.Helper()
-	if _, err := s.Exec(t.Context(), `UPDATE account_hosts SET next_fund = NOW()`); err != nil {
-		t.Fatal(err)
-	}
-}
-
-func (s testStore) hostAccounts(t testing.TB) (result []accounts.HostAccount) {
-	t.Helper()
-
-	rows, err := s.Query(t.Context(), `SELECT next_fund FROM account_hosts`)
-	if err != nil {
-		t.Fatal(err)
-	}
-	defer rows.Close()
-
-	for rows.Next() {
-		var account accounts.HostAccount
-		if err := rows.Scan(&account.NextFund); err != nil {
-			t.Fatal(err)
-		}
-		result = append(result, account)
-	}
-	if err := rows.Err(); err != nil {
-		t.Fatal(err)
-	}
-	return
-}
-
 func newTestStore(t testing.TB) testStore {
 	s := testutils.NewDB(t, contracts.DefaultMaintenanceSettings, zaptest.NewLogger(t))
 	t.Cleanup(func() {
--- a/accounts/manager_test.go
+++ b/accounts/manager_test.go
@@ -8,28 +8,8 @@ import (
 
 	"go.sia.tech/indexd/accounts"
 	"go.sia.tech/indexd/contracts"
-	"go.sia.tech/indexd/hosts"
 	"go.sia.tech/indexd/internal/testutils"
 	"go.uber.org/zap/zaptest"
-)
-
-var (
-	// goodUsability is the usability of a host that passes all checks
-	goodUsability = hosts.Usability{
-		Uptime:              true,
-		MaxContractDuration: true,
-		MaxCollateral:       true,
-		ProtocolVersion:     true,
-		PriceValidity:       true,
-		AcceptingContracts:  true,
-
-		ContractPrice:   true,
-		Collateral:      true,
-		StoragePrice:    true,
-		IngressPrice:    true,
-		EgressPrice:     true,
-		FreeSectorPrice: true,
-	}
 )
 
 type testStore struct {
--- a/accounts/manager_test.go
+++ b/accounts/manager_test.go
@@ -1,23 +1,15 @@
 package accounts_test
 
 import (
-	"context"
-	"errors"
 	"math"
-	"reflect"
 	"slices"
 	"testing"
 	"time"
 
-	proto "go.sia.tech/core/rhp/v4"
-	"go.sia.tech/core/types"
-	"go.sia.tech/coreutils/chain"
-	"go.sia.tech/coreutils/rhp/v4/siamux"
 	"go.sia.tech/indexd/accounts"
 	"go.sia.tech/indexd/contracts"
 	"go.sia.tech/indexd/hosts"
 	"go.sia.tech/indexd/internal/testutils"
-	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest"
 )
 
@@ -39,31 +31,6 @@ var (
 		FreeSectorPrice: true,
 	}
 )
-
-type fundAccountCall struct {
-	host        hosts.Host
-	accounts    []accounts.HostAccount
-	contractIDs []types.FileContractID
-	target      types.Currency
-}
-
-type mockFunder struct {
-	calls []fundAccountCall
-	fail  bool
-}
-
-func (f *mockFunder) FundAccounts(ctx context.Context, host hosts.Host, contractIDs []types.FileContractID, accounts []accounts.HostAccount, target types.Currency, log *zap.Logger) (funded int, drained int, _ error) {
-	f.calls = append(f.calls, fundAccountCall{
-		host:        host,
-		accounts:    accounts,
-		contractIDs: contractIDs,
-		target:      target,
-	})
-	if f.fail {
-		return 0, 0, nil
-	}
-	return len(accounts), 1, nil
-}
 
 type testStore struct {
 	testutils.TestStore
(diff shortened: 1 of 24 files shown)


accounts/funder_test.go
```diff
@@ -81,7 +63,9 @@ func TestFunder(t *testing.T) {
 		res:    rhp.RPCReplenishAccountsResult{Revision: types.V2FileContract{RenterOutput: types.SiacoinOutput{Value: target}}},
 		funded: len(accounts) - 1,
 	}
-	dialer.clients[host.PublicKey] = hc
+
+	// prepare funder
+	f := &Funder{client: hc}
```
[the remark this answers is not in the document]
chris124567: no reason to mock it if we aren't going to use it

client/v2/client.go
```diff
@@ -346,12 +359,34 @@ func shouldResetTransport(err error) bool {
...
+}
+
 // New creates a new Client.
-func New(hosts *Provider) *Client {
-	return &Client{
-		tg:           threadgroup.New(),
-		hosts:        hosts,
-		cachedPrices: make(map[types.PublicKey]proto.HostPrices),
-		transports:   make(map[types.PublicKey]*transport),
+func New(hosts *Provider, cm ChainManager, store RevisionStore, log *zap.Logger, opts ...Option) *Client {
```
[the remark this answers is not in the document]
chris124567: we don't do this anywhere else and this makes it harder to write tests

client/v2/contracts.go
```diff
@@ -202,16 +225,141 @@ func (c *Client) AppendSectors(ctx context.Context, signer rhp.ContractSigner, c
...
 	}
 	defer done()
 
-	err = c.rpcFn(ctx, params.Contract.Revision.HostPublicKey, func(ctx context.Context, transport rhp.TransportClient) error {
-		result, err = rhp.RPCReplenishAccounts(ctx, transport, params, chain.TipState(), signer)
-		return err
+	err = c.withRevision(ctx, params.Contract.ID, func(contract rhp.ContractRevision) (rhp.ContractRevision, proto.Usage, error) {
+		if contract.Revision.RenterOutput.Value.Cmp(params.Target) < 0 {
+			return rhp.ContractRevision{}, proto.Usage{}, ErrContractInsufficientFunds
+		}
```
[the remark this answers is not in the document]
chris124567: target will never be 0

client/v2/client.go
```diff
@@ -145,10 +146,22 @@ top:
...
+	MarkContractBad(contractID types.FileContractID) error
+}
+
 // A Client is used to interact with Sia hosts over RHP4.
 type Client struct {
-	tg    *threadgroup.ThreadGroup
-	hosts *Provider
+	tg                       *threadgroup.ThreadGroup
+	hosts                    *Provider
+	store                    RevisionStore
```
peterjan: All these new fields on the `Client` feel a little off. 

I know you didn't introduce this interface, but I'm not sure whether we should be copying it into the new client implementation either. The client should not be interacting with the store directly, but rather go through the contract manager. A field like `revisionSubmissionBuffer` should be an option for the manager and not the client.

Another thing that tripped me up for instance is that you removed the `chain` manager argument from the interface but kept the `signer`. You could argue that a `Client` could be instantiated with its own signer and that would allow you to remove it from all interfaces as well as the `Funder` that now takes a signer only to pass it on to the client and nothing more.

I understand you copied `withRevision` and `syncRevision` from the old client to the new, but there's numerous RPCs on the client that need `withRevision` to be called that currently don't call it. More importantly though I'm not sure whether that should even be something the client manages, this too should be handled by the contract manager that uses the client only for the RPC.

I feel these are worth discussing before proceeding with this PR. Personally I would prefer we keep the client purely for interacting with the hosts over RHP4, we move the revision management to the contract manager.

peterjan (changes requested):

client/v2/client.go
ChrisSchinnerl: PJ is right. The client is used by both indexer as well as the SDK and the reason it has a `Provider` is because both indexer and SDK need to get hosts from somewhere. The indexer from its store and the SDK from the indexer's app api. We can't expect users of the SDK to require a revision store to use a client. 

So instead of moving all this logic in the client we should fetch the revision out side the client, pass it to `ReplenishAccounts` and if the operation fails due to the revision not matching, the caller will need to handle that. 

Upon success `ReplenishAccounts` will return the latest revision through the `RPCReplenishAccountsResult` and the caller is responsible for persisting it.

ChrisSchinnerl (changes requested):

chris124567:
Sadly we can't use the contracts manager methods for storing/retrieving the revision as the contracts manager depends on the account manager so we need to use the store directly.  I also wasn't sure if we wanted a generic withRevision method or where we wanted to put that if so I didn't abstract anything but I think this is a good starting point.

client/v2/contracts.go
```diff
@@ -202,16 +210,38 @@ func (c *Client) AppendSectors(ctx context.Context, signer rhp.ContractSigner, c
 // The balance of each account will be topped up to the target balance and the
 // amount of funds added to each account will be returned in the result in the
 // same order as the provided accounts.
-func (c *Client) ReplenishAccounts(ctx context.Context, signer rhp.ContractSigner, chain ChainManager, params rhp.RPCReplenishAccountsParams) (result rhp.RPCReplenishAccountsResult, err error) {
+func (c *Client) ReplenishAccounts(ctx context.Context, signer rhp.ContractSigner, chain ChainManager, revisionSubmissionBuffer uint64, params rhp.RPCReplenishAccountsParams) (result rhp.RPCReplenishAccountsResult, funded int, err error) {
```
n8mgr: Let's leave the `revisionSubmissionBuffer` check in the caller.

n8mgr: Same with the `funded` logic. Attempt to fund the exact accounts the caller passes in. If the caller didn't calculate it correctly, an error is fine.

Logic spanning multiple levels is something we want to avoid in future code. Just makes it harder to track and reason about.

n8mgr: Realistically you shouldn't need to make any changes to the client.

peterjan (changes requested):
This is already a big improvement over the first version, but it still doesn't feel right. The funder becomes much harder to read and we're deduplicating revision management. 

With the old client, every flow that updates a revision had to go through `withRevision`. That's a good property to have and we should try and protect it. The new client is an improvement in that it focuses on RHP4 communication with the host, great, so we have to do our own revision management on the caller side.

I think account funding is in the wrong place. The account manager should be restricted to fetching and updating account information, nothing more. The contract manager should be responsible for funding accounts.

The CM needs the AM for
- funding accounts
- calculating the contract fund target (how weird is that - I added that 😢 )

Instead it should ask the AM for
- accounts that need funding (should expose a cursor though so we can keep the accounts batch size)
- number of active accounts, so it can calculate the target itself

The CM should be the driving force, that way we can house all of the revision management over to the CM. We definitely should try and make sure a contract's revision is only ever update from one location.

As a F/U step I think it would be good to split off some responsibilities from the CM. It should only concern itself with contract lifecycle management. So anything that is not forming/refreshing/renew should be pulled out. I guess you could argue that revising can be part of that list, but funding an account and sector pinning doesn't belong there imo. Pruning can go either way but I wouldn't consider that part of a contract's lifecycle either.

n8mgr:
> As a F/U step I think it would be good to split off some responsibilities from the CM. It should only concern itself with contract lifecycle management. So anything that is not forming/refreshing/renew should be pulled out. I guess you could argue that revising can be part of that list, but funding an account and sector pinning doesn't belong there imo. Pruning can go either way but I wouldn't consider that part of a contract's lifecycle either.

Disagree with this being a good goal. The contract manager should be the sole location of anything that revises a contract, otherwise we need to re-add locking. There's no reason to have a dozen different managers that all revise contracts differently with a manager on top to synchronize them. The only thing that needs to be protected is the serial access to contracts, which gets harder the more intermediaries you add. Keep logic in one place, don't spread it out.

The only place this currently isn't the case is account funding for some reason. Probably best to fold that into the contract manager. There's very little logic there anyway.

peterjan:
> > As a F/U step I think it would be good to split off some responsibilities from the CM. It should only concern itself with contract lifecycle management. So anything that is not forming/refreshing/renew should be pulled out. I guess you could argue that revising can be part of that list, but funding an account and sector pinning doesn't belong there imo. Pruning can go either way but I wouldn't consider that part of a contract's lifecycle either.
> 
> Disagree with this being a good goal. The contract manager should be the sole location of anything that revises a contract, otherwise we need to re-add locking. There's no reason to have a dozen different managers that all revise contracts differently with a manager on top to synchronize them. The only thing that needs to be protected is the serial access to contracts, which gets harder the more intermediaries you add. Keep logic in one place, don't spread it out.
> 
> The only place this currently isn't the case is account funding for some reason. Probably best to fold that into the contract manager. There's very little logic there anyway.

Hm didn't consider the locking aspect, that was dumb. I guess I was looking at the contract manager and felt it was doing too much, without realising we have to because we avoid locking that way - not sure why that didn't came to mind. As long as we move funding and revising into the contract manager and update the contract revision in a single location I'm fine.

chris124567:
- Do we want to remove the idea of a Funder entirely and just put it on the contracts manager itself?
- I can replace all the uses of the v1 client on the contracts manager if we want but this PR was already getting pretty large so I didn't

n8mgr:
> - Do we want to remove the idea of a Funder entirely and just put it on the contracts manager itself?
> - I can replace all the uses of the v1 client on the contracts manager if we want but this PR was already getting pretty large so I didn't

Maybe in a follow up. Lets keep it tightly scoped. This PR is getting large.

contracts/funder.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+	for _, opt := range opts {
+		opt(f)
+	}
+	return f
+}
+
+// FundAccounts tops up the provided accounts to the target balance using the
+// specified contracts in order. The given accounts should not exceed the batch
+// size used in the replenish RPf. This method returns two numbers, the first
```
peterjan:
```suggestion
// size used in the replenish RPC. This method returns two numbers, the first
```

contracts/fund_test.go
```diff
@@ -21,25 +22,50 @@ type fundAccountsCall struct {
...
 	am.mu.Lock()
 	defer am.mu.Unlock()
-	am.calls = append(am.calls, fundAccountsCall{
+	cpy := make([]accounts.HostAccount, len(am.accountsToFund))
+	copy(cpy, am.accountsToFund)
+	return cpy, nil
+}
+
+func (am *accountsManagerMock) ActiveAccounts(threshold time.Time) (uint64, error) {
+	return am.activeAccounts, nil
```
peterjan, line 39: Acquire the mutex lock

accounts/manager_test.go
```diff
@@ -107,155 +45,6 @@ func newTestStore(t testing.TB) testStore {
...
-	f.fail = true
-	for range 3 {
-		s.resetNextFund(t)
-		err = am.FundAccounts(context.Background(), host, contractIDs, false, zap.NewNop())
-		if err != nil {
-			t.Fatal(err)
-		}
-	}
-
-	// assert the exponential backoff was applied
```
peterjan, line 190: This test was removed and I don't think we added equivalent test coverage in the contracts package

peterjan (changes requested):

peterjan:
I think the tests would be good to port over, other than that this PR looks good to me!

ChrisSchinnerl (commented):
Got nothing to add to the open comments.

ChrisSchinnerl (approved):

peterjan (approved):

PR #728, opened 2025-12-15
Add endpoint to reset lost sectors

### LINKED ISSUE #727: Endpoint to reset lost sector count for a host
Add an endpoint to reset the lost sectors for a host.

chris124567:
Close #727

### DIFF
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -209,9 +209,9 @@ func NewAPI(chain ChainManager, accounts Accounts, contracts ContractManager, ho
 		"GET /explorer/exchange-rate/siacoin/:currency": a.handleGETExplorerSiacoinExchangeRate,
 
 		// host endpoints
-		"GET    /host/:hostkey":             a.handleGETHost,
-		"POST   /host/:hostkey/scan":        a.handlePOSTHostScan,
-		"DELETE /host/:hostkey/lostsectors": a.handleDELETEHostLostSectors,
+		"GET    /host/:hostkey":                   a.handleGETHost,
+		"POST   /host/:hostkey/scan":              a.handlePOSTHostScan,
+		"POST   /host/:hostkey/lostsectors/reset": a.handlePOSTHostLostSectorsReset,
 
 		// hosts endpoints
 		"GET    /hosts":                    a.handleGETHosts,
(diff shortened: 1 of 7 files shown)


api/admin/admin.go
```diff
@@ -208,8 +209,9 @@ func NewAPI(chain ChainManager, accounts Accounts, contracts ContractManager, ho
 		"GET /explorer/exchange-rate/siacoin/:currency": a.handleGETExplorerSiacoinExchangeRate,
 
 		// host endpoints
-		"GET    /host/:hostkey":      a.handleGETHost,
-		"POST   /host/:hostkey/scan": a.handlePOSTHostScan,
+		"GET    /host/:hostkey":             a.handleGETHost,
+		"POST   /host/:hostkey/scan":        a.handlePOSTHostScan,
+		"DELETE /host/:hostkey/lostsectors": a.handleDELETEHostLostSectors,
```
peterjan: I would go for `PUT /hosts/:hostkey/lostsectors/count` - `DELETE` feels off because we're not deleting any resouces, `POST` is an option but you'd expect for a resource to be created in that case and it's not idempotent, `PUT` is probably our best option here.

peterjan (changes requested):

api/admin/admin.go
ChrisSchinnerl: Considering it's not cleanly fitting into CRUD I'd be explicit about it and make it `POST /host/:hostkey/lostsectors/reset`

Neither `DELETE` nor `PUT` seem right since `lostsectors` is not a resource and `PUT` should ideally replace a resource with a new version of it. In this case though we are issuing more of a command rather than updating a resource.

Imo trying to make `PUT` work by adding a `count` is 1. trying a bit too hard to make something that's not a resource a resource and 2. is less intuitive than just calling it by what it does which is `reset`.

ChrisSchinnerl (changes requested):
Missing schema update

api/admin/admin.go
peterjan: Yeah fine with `POST`, still think `PUT` is a little bit more correct but I don't really care. It technically is a count, the difference here with what you'd usually do is here we don't allow sending a request body and implicitly `PUT` the count to 0. As long as it's not `DELETE` I'm fine 👍

peterjan: edit: I def. see the command argument, truth be told I would've probably made it a `POST` if I had to write it. It's just that thinking about it made me lean towards a PUT

peterjan (approved):

ChrisSchinnerl (approved):

PR #729, opened 2025-12-16
Update downloadSlab to match rust sdk

### LINKED ISSUE #205: SDK: Make slab download more efficient
Right now for every shard that completes we free the semaphore again and launch a new shard. Which isn't horribly efficient since downloading 9 shards and waiting for the 10th to finish will launch 9 more.

A potential solution to this is to only free the semaphore if the shard failed to download and add 10% or so extra shards to the semaphore channel's buffer to begin with.

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/indexd/issues/205

### DIFF
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"io"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/klauspost/reedsolomon"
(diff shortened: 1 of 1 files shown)


sdk/client.go
```diff
@@ -115,46 +114,88 @@ func (s *SDK) downloadSlab(ctx context.Context, slab slabs.SlabSlice, maxInfligh
...
+	}
+
+	// launch more downloads as results come in or periodically if there is
+	// still capacity in the semaphore
+	var successful int
+	shards := make([][]byte, len(slab.Sectors))
+	timer := time.NewTimer(time.Second)
+	defer timer.Stop()
+	for {
+		select {
```
peterjan, line 164: This select is missing a `ctx.Done()` check, right?

peterjan (changes requested):
I think this deadlocks.

If all initial shards get launched and the context gets cancelled, and for some reason we never get to min shard successful downloads... Then we have to get responses to get out of the loop. The timer will start trying sectors, but they might not send responses down the channel, and eventually we might get stuck in the timer branch which just continues the loop because there's no more slab hosts.

peterjan (approved):

peterjan (commented):
LGTM 🔥

PR #730, opened 2025-12-16
Update updated_at column of object when sector is migrated

### LINKED ISSUE #684: Object Migration Event
I think we should change the way clients are expected to store metadata. The SDK currently exposes only the slab ID and offsets to the user:

```
// SlabSlice represents a slice of a slab that is part of an object.
SlabSlice struct {
	SlabID SlabID `json:"slabID"`
	Offset uint32 `json:"offset"`
	Length uint32 `json:"length"`
}
```

The SDK then lazily fetches each slab from the indexer before downloading. What if the indexer is offline for a short period? The user will not be able to download their data.

There are two potential changes here. The first is to expose all of the slab metadata as part of the object and sealed object. The second is to make migrations register an object event so the user can sync down the updated locations.

Exposing all the slab metadata on the object in the SDK will also let us unify the download logic and get rid of the split between a SlabSlab vs SharedSlab.

This is a pretty large refactor with broad implications. I'm open to not changing it, but if we want to make such a large breaking change we should do so now.

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/indexd/issues/684

### DIFF
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -795,10 +795,11 @@ func (s *Store) MigrateSector(root types.Hash256, hostKey types.PublicKey) (migr
 	err = s.transaction(func(ctx context.Context, tx *txn) error {
 		var hostID sql.NullInt64
 		var contractMapID sql.NullInt64
+		var sectorID int64
 		err := tx.QueryRow(ctx, `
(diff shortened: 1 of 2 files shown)


peterjan (approved):
nice

PR #733, opened 2025-12-18
Better notifications for stuck hosts

### LINKED ISSUE #692: Better notifications for stuck hosts
I spent some more time staring at logs to debug the reason for our 300GB of unpinned sectors. As it turns out, all that data was stored on 2 hosts which have been returning host funding errors for a while. After blocking them the amount of unpinned sectors dropped to 13GB.

The underlying issue was that the contracts were not good for pinning data but refreshing them failed due to the hosts not having funded wallets. Of course there might be other reasons for getting stuck as well. Such as hosts just returning "internal error" all the time.

So I'd like to propose the following to discover these "stuck" hosts sooner in the future.

- add a `stuck_since` column to the hosts table which is a timestamp
- if we fail to refresh, renew or form a contract with a host, we set `stuck_since` if it is `NULL`
- if we succeed any of the above we set `stuck_since` to `NULL`
- we consider a host "stuck" if the time since `stuck_since` is 24+ hours and the host has any active contracts (good or bad)
- we add the number of stuck hosts to `/stats/hosts`
- we register an alert for every stuck host (if #721 is done at that point we can include the unpinned data for each stuck host)

chris124567:
Close #692

```
goos: linux
goarch: amd64
BenchmarkStuckHosts
BenchmarkStuckHosts-4   	     373	   3228131 ns/op
PASS
ok  	go.sia.tech/indexd/persist/postgres	12.314s
```


openapi/admin.yml
```diff
@@ -1397,19 +1397,18 @@ paths:
...
+  /stats/hosts/all:
     get:
       tags:
         - stats
-      summary: Get stats about host scans
-      operationId: getStatsHostsScans
+      summary: Get total stats about hosts and their scans
+      operationId: getStatsHostsAll
       responses:
         "200":
```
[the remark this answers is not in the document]
chris124567: It won't convey anything new

persist/postgres/init.sql
```diff
@@ -43,6 +43,7 @@ CREATE TABLE hosts (
     country_code TEXT NOT NULL DEFAULT '',
     location POINT NOT NULL DEFAULT POINT(0.0, 0.0),
     lost_sectors INTEGER NOT NULL DEFAULT 0,
+    stuck_since TIMESTAMP WITH TIME ZONE,
```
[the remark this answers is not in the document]
chris124567: We would need orders of magnitude more hosts for it to be useful

api/admin/admin.go
```diff
@@ -251,7 +251,7 @@ func NewAPI(chain ChainManager, accounts Accounts, contracts ContractManager, ho
 		"GET /stats/accounts":  a.handleGETStatsAccounts,
 		"GET /stats/contracts": a.handleGETStatsContracts,
 		"GET /stats/hosts":     a.handleGETStatsHosts,
-		"GET /stats/scans":     a.handleGETStatsScans,
+		"GET /stats/hosts/all": a.handleGETStatsHostsAll,
```
peterjan: I'll rope in @ChrisSchinnerl  but `stuck` should be a new field on `hosts.Host` (similar to `blocked`) and exposed through `/stats/hosts` instead of updating this route.

contracts/manager.go
```diff
@@ -132,6 +132,9 @@ type (
 		UpdateContractRevision(contract rhp.ContractRevision, usage proto.Usage) error
 		UpdateNextPrune(contractID types.FileContractID, nextPrune time.Time) error
 
+		SetHostStuck(hk types.PublicKey) error
```
peterjan: These methods are a little clunky imo. I would rename it to `UpdateHostStuckSince` and allow passing `nil` to unstuck the host. Or `UpdateStuckSince`

hosts/manager.go
```diff
@@ -560,6 +580,35 @@ loop:
 	m.log.Debug("host scans finished", zap.Int("hosts", len(hosts)), zap.Duration("duration", time.Since(start)))
 }
 
+func newStuckHostsAlert(hks []types.PublicKey) alerts.Alert {
+	return alerts.Alert{
+		ID:       alertStuckHostsID,
+		Severity: alerts.SeverityWarning,
+		Message:  "Host(s) are stuck",
+		Data: map[string]any{
+			"hostKeys": hks,
```
peterjan: I would've probably made this alert unique to a host. Yes, that does make it more complicated, especially unregistering the alerts, but in a F/U we can add more information on a per host basis. E.g. the amount of data that is stuck. 

I'm on the fence myself though, theoretically we can add all the information in the alert's metadata and have the frontend figure it out, historically though I think we would've made this a host alert. I can also imagine eventually passing on the stuck reason, which might also be useful information. E.g. the contract that's failing to renew/refresh etc.

peterjan (changes requested):

hosts/manager.go
peterjan: I would probably also ask Nate or Chris's input on this one. Alerts are super important to us and it would be nice to get them right. I can see why they would be against an alert on a per host basis for instance, because in `renterd` we never really succeeded at having good alerting. 

There might be an issue with the node which causes all hosts to get stuck for instance, in that case we should try and avoid registering an alert for every host.

peterjan (commented):
It's also missing an index on `stuck_since` btw.

api/admin/admin.go
ChrisSchinnerl: I'd start with `hosts.HostStats` since that's what we use for the dashboard already. We can later add it to `hosts.Host` if we feel the need.

hosts/manager.go
ChrisSchinnerl: I think we can use a single one since that's what we do for lost sectors as well. That way we don't get spammed by 100 alerts if 100 hosts are stuck. If we then feel like we need more information we can still turn `hostKeys` into a `hosts` field which contains an object for every stuck host rather than just the key.

api/admin/admin.go
peterjan: Oh yeah sorry that's actually totally what I meant, I was addressing the fact that we changed the `scans` route whereas we should be updating the `/stats/hosts` route instead.

api/admin/client.go
```diff
@@ -309,9 +309,9 @@ func (c *Client) StatsHosts(ctx context.Context, offset, limit int) (resp HostSt
 	return
 }
 
-// StatsScans returns statistics about host scans for all hosts.
-func (c *Client) StatsScans(ctx context.Context) (resp ScansStatsResponse, err error) {
-	err = c.c.GET(ctx, "/stats/scans", &resp)
+// StatsAllHosts returns statistics about all hosts.
+func (c *Client) StatsAllHosts(ctx context.Context) (resp AllHostsStatsResponse, err error) {
+	err = c.c.GET(ctx, "/stats/hosts/all", &resp)
```
ChrisSchinnerl: Reminder that this will also require a PR to the `deploy` repo for `prometheus.yml.template`.

api/admin/types.go
```diff
@@ -51,10 +51,11 @@ type (
 		CheckFailed int64 `json:"checkFailed"`
 	}
 
-	// ScansStatsResponse is the response body for [GET] /stats/hosts/scans.
-	ScansStatsResponse struct {
+	// AllHostsStatsResponse is the response body for [GET] /stats/hosts/all.
+	AllHostsStatsResponse struct {
```
ChrisSchinnerl: How about `HostsStats`? Then we got `HostsStats` for all hosts and `HostStats` for a single one.

contracts/manager.go
ChrisSchinnerl: Agreed. I left another comment on the database code on how I think we can simplify this a bit.

hosts/manager.go
peterjan: Hm, yeah well I figured. I'm not too sure though, for now of course it's definitely ok but we have alerts because we want the user/admin to eventually take action. I'm wondering whether it doesn't make more sense to have these alerts on a per host basis, and then house everything under that one alert. Like lost sectors, and like being stuck, that way the user has the entire view of that host and he can make his decision to block that host yes or no. From the context of blocking a host it makes more sense to have all of the info in one alert.

contracts/renew.go
```diff
@@ -95,4 +95,14 @@ func (cm *ContractManager) renewContract(ctx context.Context, contract Contract,
...
 	})
+	if err != nil {
+		if setErr := cm.store.SetHostStuck(contract.HostKey); setErr != nil {
+			log.Error("failed to mark host stuck", zap.Error(setErr))
+		}
+		return err
+	}
+	if err := cm.store.ClearHostStuck(contract.HostKey); err != nil {
+		log.Error("failed to clear host stuck", zap.Error(err))
+	}
```
ChrisSchinnerl: I think we need to move this one layer up. The goal is to only set a host stuck if 
1. it has contracts
2. all of them are bad in some way (e.g. need refresh)
3. we can't refresh any of them or form a new one

As soon as we were able to refresh a contract or form a new one we are golden. Honestly we should be able to completely ignore the `renew` case even. 

All of this should be able to go into `PerformContractFormation`.

hosts/manager.go
```diff
@@ -560,6 +580,35 @@ loop:
...
+			"hint":     "Contract operations (form/renew/refresh) have been failing for these hosts for more than 24 hours. Consider blocking these hosts through the blocklist feature.",
+		},
+		Timestamp: time.Now(),
+	}
+}
+
+func (m *HostManager) registerStuckHostsAlert() {
+	hks, err := m.store.StuckHosts()
```
ChrisSchinnerl: If you also return `stuck_since` here we can register that for each host as well. Might be useful.

persist/postgres/hosts.go
```diff
@@ -1012,6 +1012,66 @@ func (s *Store) HostsWithLostSectors() ([]types.PublicKey, error) {
 	return hks, nil
 }
 
+// SetHostStuck sets the stuck_since timestamp to now if it is currently NULL.
+// This is called when a contract operation (form, renew, refresh) fails for a host.
+func (s *Store) SetHostStuck(hk types.PublicKey) error {
```
ChrisSchinnerl: While typing my other comment (about a host only being stuck if all contracts are bad in `performContractFormation` and we can't form a new one or refresh an existing one to make it good) I had an idea.

I think you can do the following:

1. In `performContractFormation` collect all hosts that are stuck
2. Remove `SetHostStuck` and `ClearHostStuck` here and change them to `UpdateStuckHosts(...)`
3. Call `UpdateStuckHosts` with the stuck hosts at the end of `performContractFormation`

`UpdateStuckHosts` would do the following:
1. clear `stuck_since` for all hosts that are currently stuck but no longer stuck according to the input
2. mark any host that is not stuck as stuck
3. keep the timestamp the same for already stuck hosts

That will then also allow us to simplify `StuckHosts` to only look at `stuck_since` and ignore the contract state. Why? Because a stuck host that has no contracts would end up being unstuck at the end of `performContractFormation` anyway.

persist/postgres/stats.go
```diff
@@ -99,9 +99,27 @@ func (s *Store) AccountStats() (admin.AccountStatsResponse, error) {
...
+		if err := tx.QueryRow(ctx, "SELECT num_scans, num_scans_failed FROM stats").Scan(&stats.Total, &stats.Failed); err != nil {
+			return fmt.Errorf("failed to get scan counts: %w", err)
+		}
+		// count stuck hosts
+		err := tx.QueryRow(ctx, `
+			SELECT COUNT(DISTINCT h.id)
+			FROM hosts h
+			WHERE h.stuck_since IS NOT NULL
+				AND h.stuck_since < NOW() - INTERVAL '24 hours'
+				AND EXISTS (
```
ChrisSchinnerl: This will also get simpler if you implement the change I suggested and ignore the state here.

ChrisSchinnerl (changes requested):

api/admin/client.go
chris124567: https://github.com/SiaFoundation/deploy/pull/35

chris124567:
```
goos: linux
goarch: amd64
BenchmarkStuckHosts
BenchmarkStuckHosts-4   	     667	   1653807 ns/op
PASS
ok  	go.sia.tech/indexd/persist/postgres	11.210s
```

hosts/manager.go
```diff
@@ -84,6 +92,7 @@ type (
 		ScansFailed         int64                  `json:"scansFailed"`
 		Blocked             bool                   `json:"blocked"`
 		BlockedReasons      []string               `json:"blockedReasons"`
+		Stuck               bool                   `json:"stuck"`
```
peterjan, line 95: We don't need to add `Stuck` to the `Host` yet. In my comment earlier I meant to say `HostStats`, where you also added it. The only place to add this `bool` is in `HostStats`.

api/admin/types.go
```diff
@@ -51,10 +51,11 @@ type (
 		CheckFailed int64 `json:"checkFailed"`
 	}
 
-	// ScansStatsResponse is the response body for [GET] /stats/hosts/scans.
-	ScansStatsResponse struct {
+	// HostsStatsResponse is the response body for [GET] /stats/hosts/all.
+	HostsStatsResponse struct {
```
peterjan: Might be me but I don't understand why this PR updates the `ScansStatsResponse`. You added a `stuck` bool to the `HostStats`, but you don't export it it's `PrometheusMetric` method. The only place where we need it should be there. We can easily get the total stuck hosts by aggregating over the host stat metrics in grafana.

If we do want to update the `/stats/hosts/scans` route and turn it into an all host stats endpoint, we should still export the stuck bool in the prometheus metric of host stats too.

peterjan (changes requested):

api/admin/types.go
ChrisSchinnerl: That's a good point. We could leave that endpoint be for now and save the update to `deploy`.

peterjan (changes requested):
Not sure if this was ready for another 👀 , so I apologise if you were still working on it. The latest version however adds `indexd_total_hosts_stuck` and updates `HostsStatsResponse`, that should not  be necessary to export the stuck host stats. We should keep everything as is, and add the `stuck` bool to `HostStats` and make sure it gets exported.

persist/postgres/hosts_test.go
```diff
@@ -2719,6 +2884,58 @@ func BenchmarkHostStats(b *testing.B) {
 	}
 }
 
+func BenchmarkStuckHosts(b *testing.B) {
```
peterjan, line 2875: This is good to have but looking at the queries I'd be more concerned with `UpdateStuckHosts`, if we have a benchmark anyway we might as well extend it to cover both get and set.

openapi/components.yml
```diff
@@ -862,7 +862,7 @@ components:
               type: string
             description: Reasons why the host was blocked, if any
 
-    HostScanStatsResponse:
+    ScansStatsResponse:
```
peterjan, line 865: Updating `HostScanStatsResponse` is a little out of scope but if you want to take it along here we should probably also correct the properties while we're at it.

persist/postgres/hosts_test.go
```diff
@@ -1088,6 +1088,171 @@ func TestResetLostSectors(t *testing.T) {
...
+					break
+				}
+			}
+			if !found {
+				t.Fatalf("expected host %v to be stuck", exp)
+			}
+		}
+	}
+
+	assertNumStuckHosts := func(expected int64) {
```
peterjan: `nit` would merge this into `assertStuckHosts` - you're always asserting two things where the second count is the length of the stuck hosts you're asserting so you might as well check the stats in `assertStuckHosts`

peterjan (changes requested):
Left some final comments, LGTM otherwise 👍 

edit: actually I just though of something, I'll leave it in a [comment](https://github.com/SiaFoundation/indexd/pull/733#discussion_r2642792119)

contracts/form.go
```diff
@@ -435,6 +442,15 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, setting
 		}
 	}
 
+	// update stuck hosts status
+	stuckHostsList := make([]types.PublicKey, 0, len(stuckHosts))
+	for hk := range stuckHosts {
+		stuckHostsList = append(stuckHostsList, hk)
+	}
+	if err := cm.store.UpdateStuckHosts(stuckHostsList); err != nil {
```
peterjan: Two things came to mind here:

1. We only call `UpdateStuckHosts` in `performContractFormation`, which handles formations and refreshes (might want to update the name), so that means we're missing renewals? Both the issue and some comments in this PR indicate we want to mark a host stuck if we fail to renew a contract.

2. `UpdateStuckHosts` will clear `stuck_since` for all hosts not in the provided set, but not being part of the provided set doesn't mean they got unstuck. Perhaps a solution could be to update `UpdateStuckHosts` to have two parameters `stuck` and `unstuck`, so we can clear the hosts in the unstuck slice and set the hosts and the stuck slice.

persist/postgres/hosts_test.go
chris124567:
```
goos: linux
goarch: amd64
BenchmarkStuckHosts
BenchmarkStuckHosts/StuckHosts
BenchmarkStuckHosts/StuckHosts-4         	     723	   2598967 ns/op
BenchmarkStuckHosts/UpdateStuckHosts
BenchmarkStuckHosts/UpdateStuckHosts-4   	      18	  66632287 ns/op
PASS
ok  	go.sia.tech/indexd/persist/postgres	10.648s
```

with 10k hosts and using 5% of hosts as stuck and 5% as unstuck

persist/postgres/hosts.go
```diff
@@ -1012,6 +1014,73 @@ func (s *Store) HostsWithLostSectors() ([]types.PublicKey, error) {
...
+			_, err := tx.Exec(ctx, `
+				UPDATE hosts
+				SET stuck_since = COALESCE(stuck_since, NOW())
+				WHERE public_key = ANY($1)`, sqlHks)
+			if err != nil {
+				return fmt.Errorf("failed to set stuck hosts: %w", err)
+			}
+		}
+
+		// clear stuck_since for unstuck hosts
```
peterjan: `nit` probably want to have stuck take precedence for hosts that are part of both slices - that should never happen but better to be safe than sorry

persist/postgres/init.sql
```diff
@@ -70,6 +71,9 @@ CREATE TABLE hosts (
 );
 CREATE INDEX hosts_next_scan_idx ON hosts(next_scan);
 
+-- speed up querying for stuck hosts
+CREATE INDEX hosts_stuck_since_idx ON hosts(stuck_since);
```
peterjan: `nit` turn this into a partial index (`WHERE stuck_since IS NOT NULL`) to make it smaller

peterjan (dismissed):
Left two small remarks, looks good to go though!

peterjan (dismissed):

persist/postgres/hosts.go
```diff
@@ -1012,6 +1014,73 @@ func (s *Store) HostsWithLostSectors() ([]types.PublicKey, error) {
 	return hks, nil
 }
 
+// UpdateStuckHosts updates the stuck_since timestamp for hosts. Hosts in the
+// stuck slice will be marked as stuck (preserving existing timestamps), and
+// hosts in the unstuck slice will have their stuck status cleared.
+func (s *Store) UpdateStuckHosts(stuck, unstuck []types.PublicKey) error {
```
ChrisSchinnerl: Unless I'm missing an edge case here you don't need to pass `unstuck`. Every host that is not `stuck` should implicitly be `unstuck`. That should simplify the update here to `SET stuck_since = NULL` only if it's 1. not already `NULL` and 2. the key is not in the list of stuck keys. 

Otherwise if we rely on `unstuck` too, we might run into edge cases where hosts without contracts never get unstuck.

contracts/renew.go
```diff
@@ -40,6 +50,10 @@ func (cm *ContractManager) performContractRenewals(ctx context.Context, period,
 		}
 	}
 
+	if err := cm.store.UpdateStuckHosts(slices.Collect(maps.Keys(stuckHosts)), slices.Collect(maps.Keys(unstuckHosts))); err != nil {
```
ChrisSchinnerl: Drop this from `performContractRenewals` for now. e.g. if 1 contract is up for renewal and that fails, the host is considered stuck even if it has more good contracts. Which is not exactly what we want.

contracts/form.go
```diff
@@ -279,6 +281,10 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, setting
 	}
 
 	var formed, refreshed uint32
+	// track hosts that fail contract operations
+	stuckHosts := make(map[types.PublicKey]struct{})
+	unstuckHosts := make(map[types.PublicKey]struct{})
```
ChrisSchinnerl: (nit) I'd rename these to `hasStuckContract` and `hasGoodContract`. A stuck host would then be a host that has a stuck contract but no good contract.

contracts/form.go
ChrisSchinnerl, line 350: You missed this edge case here. Where a good contract already exists. If a host has a good contract already and one that failed to refresh, we don't consider the host stuck. 

So here we also need to add the host to the map of hosts that aren't stuck. Basically every time we do `goodContracts++` we also want to make sure we added the host to the map of unstuck hosts. Since we neither call `formContract` nor `refreshContract` here that's not the case.

ChrisSchinnerl (changes requested):

persist/postgres/hosts.go
peterjan: That's on me I guess. I commented saying that it was safer to explicitly "unstuck" hosts if we are sure they are effectively unstuck. I felt it was incorrect to unstuck a host because he's not in the stuck slice, because the host may have been skipped for instance, or not returned by the query that fetches hosts for formation/refresh for whatever reason.

Two questions:
1) Why is it bad to explicitly unstuck? Theoretically this should result in a correct `stuck` status for all hosts at all times
2) You end with "we might run into edge cases where hosts without contracts never get unstuck." - isn't `stuck` the proper status for a host without contracts (edit: if it got stuck in the first place)

ChrisSchinnerl: Because we know that a host can only be stuck or not. So that is all the information the database code needs. Meaning the database will always correctly represent the results of whatever we determined in `performContractFormation`. If you pass 4 you open yourself up to hosts that are neither or both. Neither meaning the host will keep its previous state (nondeterminism) and both means we ran into a developer error.

A host without contracts is only stuck if we attempted to form a contract with it and failed. Which is only the case if a host has unpinned sectors. If a host has no contracts and no unpinned sectors it's just an unused host and not stuck.

peterjan: The problem I was trying to address is that `performContractFormation` only fetches usable and unblocked hosts. That means that all hosts _not_ in that list, are being `unstuck` (since they can never end up in the `stuck` list). I was under the impression we would not want that to be the case. I figured a `stuck` host should only be `unstuck` if we were explicitly able to form/refresh its contract. Apologies @chris124567 🤷🏻‍♂️

ChrisSchinnerl: It does. But if the host is unusable or blocked then they usually remain blocked forever and never unstuck themselves. So we would just end up with alerts that we can't act on which is something we want to avoid.

chris124567:
```
goos: linux
goarch: amd64
BenchmarkStuckHosts
BenchmarkStuckHosts/StuckHosts
BenchmarkStuckHosts/StuckHosts-4                 858       1422461 ns/op
BenchmarkStuckHosts/UpdateStuckHosts
BenchmarkStuckHosts/UpdateStuckHosts-4            62      22319105 ns/op
PASS
ok      go.sia.tech/indexd/persist/postgres 6.211s
```

ChrisSchinnerl (approved):

peterjan (dismissed):

contracts/form.go
```diff
@@ -435,6 +446,28 @@ func (cm *ContractManager) performContractFormation(ctx context.Context, setting
...
+	// but no good contract
+	var stuckHosts []types.PublicKey
+	for _, hk := range candidates {
+		if _, ok := hasGoodContract[hk]; !ok {
+			stuckHosts = append(stuckHosts, hk)
+		}
+	}
+	if err := cm.store.UpdateStuckHosts(stuckHosts); err != nil {
+		log.Error("failed to update stuck hosts", zap.Error(err))
+	}
```
ChrisSchinnerl: This should be all that's needed now right?

```suggestion
	// update stuck hosts status - a host is stuck if it has a stuck contract
	// but no good contract
	var stuckHosts []types.PublicKey
	for hk := range hasStuckContract {
		if _, ok := hasGoodContract[hk]; !ok {
			stuckHosts = append(stuckHosts, hk)
		}
	}
	if err := cm.store.UpdateStuckHosts(stuckHosts); err != nil {
		log.Error("failed to update stuck hosts", zap.Error(err))
	}
```

ChrisSchinnerl (changes requested):

contracts/form.go
chris124567: Yeah that should be right performContractFormation considers all the hosts

peterjan (approved):

ChrisSchinnerl (approved):

PR #748, opened 2026-01-08
Track unpinned sectors

### LINKED ISSUE #721: Track unpinned sectors per host
We ran into another couple of hosts that accumulated unpinned sectors over the weekend. One of them more than 500GB of sectors. Usually the way this starts is by checking the unpinned sector graph which keeps going up, then running a slow query on the database directly to find the host causing it and then an ssh and grep on zeus to see if it's reasonable to block that host yet.

If we tracked unpinned sectors per host, we could provide a pie chart which makes this easier by telling us directly what host is responsible for 99% of all unpinned sectors. 
We should also register a warning alert for a host with more than 100GB of unpinned data.

peterjan:
Fixes https://github.com/SiaFoundation/indexd/issues/721


chris124567 (dismissed):

persist/postgres/hosts.go
```diff
@@ -1078,6 +1082,22 @@ func (s *Store) StuckHosts() ([]hosts.StuckHost, error) {
 	return result, nil
 }
 
+func incrementHostUnpinnedSectors(ctx context.Context, tx *txn, hostID int64, delta int64) error {
```
ChrisSchinnerl: These should be in `stats.go` alongside the other stat related `increment*` functions.

hosts/host.go
```diff
@@ -123,6 +123,7 @@ type (
 		Blocked                bool                `json:"blocked"`
 		BlockedReasons         []string            `json:"blockedReasons"`
 		LostSectors            uint64              `json:"lostSectors"`
+		UnpinnedSectors        uint64              `json:"unpinnedSectors"`
```
ChrisSchinnerl, line 126: Missing update to `prometheus.go`.

hosts/host.go
```diff
@@ -141,8 +142,9 @@ type (
 	// StuckHost contains the public key of a stuck host and the time it
 	// became stuck.
 	StuckHost struct {
-		PublicKey  types.PublicKey `json:"publicKey"`
-		StuckSince time.Time       `json:"stuckSince"`
+		PublicKey       types.PublicKey `json:"publicKey"`
+		StuckSince      time.Time       `json:"stuckSince"`
+		UnpinnedSectors uint64          `json:"unpinnedSectors"`
```
ChrisSchinnerl, line 147: Same. 

While you are at it you could also expose `StuckSince` as  a `indexd_host_stuck_since` metric which is 0 when the host is not stuck. 
The convention in prometheus is returning a `float64` for time afaik and we already do that for `indexd_runtime`.

ChrisSchinnerl (changes requested):

persist/postgres/hosts.go
peterjan: I put them here on purpose. The other stats methods are for the stats table. This and lost sector is "host stats" so I figured it belongs in hosts.go.

hosts/host.go
peterjan: Not sure what you meant with "same" here. Stuck hosts aren't exported.

ChrisSchinnerl: ha interesting. I guess I assumed they were in some way due to the json tags but I guess it's just the alert

n8mgr (dismissed):

n8mgr:
@peterjan what's the status on this one?

peterjan:
> @peterjan what's the status on this one?

I forgot this was still in review, thought it got merged.

ChrisSchinnerl (approved):

PR #749, opened 2026-01-08
Update shouldReplaceContract

### LINKED ISSUE #743: Excessive contract formation?
Within the last week we have formed 92 new contracts while only increasing the total contract size by 76 TB. 
We should double check that there is no bug there and that those contracts were formed for good reason.

Valid reasons would be reaching a host's max collateral limit or the 10TB contract size.

ChrisSchinnerl:
Fixes https://github.com/SiaFoundation/indexd/issues/743

We were running into an edge case where we would pick the wrong candidate when a contract is up for renewal. Because the `current` contract ends up being not good for append or refresh but good for funding. While the candidate (which was drained) ended up being not good for append, not good for funding but good for refresh. So we always kept the current contract which wasn't good for refreshing, realized it's bad and formed a new contract instead of trying to refresh the refreshable candidate.

This PR fixes that and adds a test.

### DIFF
--- a/contracts/form.go
+++ b/contracts/form.go
@@ -469,27 +469,18 @@ func shouldReplaceContract(current, candidate candidateContract) bool {
 	case current.goodForAppend == nil && current.goodForFunding == nil:
 		// current contract is already good for both uploading and funding
 		return false
-	case candidate.goodForAppend == nil && candidate.goodForFunding == nil:
-		// candidate contract is good for both uploading and funding
-		return true
-	case current.goodForRefresh == nil && candidate.goodForRefresh != nil:
-		// current contract can be refreshed to become good, but candidate cannot
+	case current.goodForRefresh == nil && candidate.goodForRefresh != nil && (candidate.goodForAppend != nil || candidate.goodForFunding != nil):
+		// current contract can be refreshed to become good, candidate cannot
+		// and is not good for both upload and funding.
 		return false
-	case current.goodForRefresh != nil && candidate.goodForRefresh == nil:
-		// candidate contract can be refreshed to become good, but current cannot
-		return true
-	case current.goodForAppend == nil && candidate.goodForAppend != nil:
-		// current contract is good for uploading, but candidate is not
+	case current.goodForAppend == nil && candidate.goodForAppend != nil && candidate.goodForRefresh != nil:
+		// current contract is only good for uploading, candidate is not and
+		// can't be refreshed.
 		return false
-	case current.goodForAppend != nil && candidate.goodForAppend == nil:
-		// candidate contract is good for uploading, but current is not
-		return true
-	case current.goodForFunding == nil && candidate.goodForFunding != nil:
-		// current contract is good for funding, but candidate is not
+	case current.goodForFunding == nil && candidate.goodForFunding != nil && candidate.goodForRefresh != nil:
+		// current contract is only good for funding, candidate is not and can't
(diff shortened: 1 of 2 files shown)


contracts/form.go
```diff
@@ -482,3 +458,38 @@ func contractFunding(settings proto.HostSettings, existingData uint64, minAllowa
...
+		// candidate contract is good for uploading, but current is not
+		return true
+	case current.goodForFunding == nil && candidate.goodForFunding != nil:
+		// current contract is good for funding, but candidate is not
+		return false
+	case current.goodForFunding != nil && candidate.goodForFunding == nil:
+		// candidate contract is good for funding, but current is not
+		return true
+	}
+	return false
```
n8mgr:
```suggestion
	switch {
	case current.goodForAppend == nil && current.goodForFunding == nil:
		// current contract is already good for both uploading and funding
		return false
	case current.goodForRefresh == nil && candidate.goodForRefresh != nil && (candidate.goodForAppend != nil || candidate.goodForFunding != nil):
		// current contract can be refreshed to become good, candidate cannot and is not good for both upload and funding.
		return false
	case current.goodForAppend == nil && candidate.goodForAppend != nil && candidate.goodForRefresh != nil:
		// current contract is only good for uploading, candidate is not and can't be
		// refreshed.
		return false
	case current.goodForFunding == nil && candidate.goodForFunding != nil && candidate.goodForRefresh != nil:
		// current contract is only good for funding, candidate is not and can't be
		// refreshed.
		return false
	}
	return true
```

Would prefer keeping the match arms consistent. Early return here makes the logic less clear imo. (equally bad contracts will return `true`, but all other cases pass).

contracts/form_test.go
```diff
@@ -652,3 +652,89 @@ func TestContractFunding(t *testing.T) {
...
+		if !gff {
+			cc.goodForFunding = fmt.Errorf("not good for funding")
+		}
+		if !gfr {
+			cc.goodForRefresh = fmt.Errorf("not good for refresh")
+		}
+		return cc
+	}
+
+	tests := []struct {
```
n8mgr, line 671: missing a few cases, but one that stood out:


```
// contract refreshable, candidate good for upload and funding
current: (false, false, true)
candidate: (true, true, false)
```

n8mgr (changes requested):

contracts/form_test.go
```diff
@@ -652,3 +652,134 @@ func TestContractFunding(t *testing.T) {
...
+		}
+	}
+
+	// a good contract should never be replaced
+	goodContracts := enumerate([]bool{true}, []bool{true}, []bool{true, false})
+	anyContracts := enumerate([]bool{true, false}, []bool{true, false}, []bool{true, false})
+	assertShouldReplace(goodContracts, anyContracts, false)
+
+	// a contract that is not good should be replaced by any good contract
+	badContracts := slices.DeleteFunc(slices.Clone(goodContracts), func(c candidateContract) bool {
```
peterjan: This leaves `badContracts` empty and isn't really testing anything.  Did you mean to clone anyContracts?

peterjan (changes requested):

contracts/form_test.go
ChrisSchinnerl: Indeed I did

peterjan (approved):

n8mgr (approved):

PR #750, opened 2026-01-08
Add "GoodForUpload" field to usable host response

### LINKED ISSUE #740: Add "GoodForUpload" field to usable host response
Should usually be `true` except when

- host is out of storage
- host is stuck

This list is probably incomplete but should be a good start.

chris124567:
Close https://github.com/SiaFoundation/indexd/issues/740

### DIFF
--- a/hosts/distance.go
+++ b/hosts/distance.go
@@ -22,7 +22,7 @@ func NewSpacedSet(minDistanceKm float64) *SpacedSet {
 
 // Add adds the host to the set if it is sufficiently spaced apart from the
 // existing hosts. It returns true if the host was added, false otherwise.
-func (s *SpacedSet) Add(h HostInfo) bool {
+func (s *SpacedSet) Add(h Host) bool {
 	if s.CanAddHost(h) {
 		s.selected[h.PublicKey] = h.Location()
 		return true
@@ -32,7 +32,7 @@ func (s *SpacedSet) Add(h HostInfo) bool {
 
 // CanAddHost returns true if the host is sufficiently far removed from the
 // existing hosts in the set. Uniqueness is always enforced.
-func (s *SpacedSet) CanAddHost(h HostInfo) bool {
+func (s *SpacedSet) CanAddHost(h Host) bool {
 	if _, exists := s.selected[h.PublicKey]; exists {
 		return false
 	}
--- a/hosts/host.go
+++ b/hosts/host.go
@@ -125,6 +125,7 @@ type (
 		LostSectors            uint64              `json:"lostSectors"`
 		AccountFunding         types.Currency      `json:"accountFunding"`
 		TotalSpent             types.Currency      `json:"totalSpent"`
+		GoodForUpload          bool                `json:"goodForUpload"`
 	}
 
 	// HostInfo is a subset of the Host struct that contains only the public
@@ -197,11 +198,12 @@ type (
 // Info returns the HostInfo of the host.
 func (h *Host) Info() HostInfo {
 	return HostInfo{
-		PublicKey:   h.PublicKey,
-		Addresses:   h.Addresses,
-		CountryCode: h.CountryCode,
-		Latitude:    h.Latitude,
-		Longitude:   h.Longitude,
+		PublicKey:     h.PublicKey,
+		Addresses:     h.Addresses,
+		CountryCode:   h.CountryCode,
+		Latitude:      h.Latitude,
+		Longitude:     h.Longitude,
+		GoodForUpload: h.GoodForUpload,
 	}
 }
 
--- a/hosts/host.go
+++ b/hosts/host.go
@@ -125,7 +125,6 @@ type (
 		LostSectors            uint64              `json:"lostSectors"`
 		AccountFunding         types.Currency      `json:"accountFunding"`
 		TotalSpent             types.Currency      `json:"totalSpent"`
-		GoodForUpload          bool                `json:"goodForUpload"`
 	}
 
 	// HostInfo is a subset of the Host struct that contains only the public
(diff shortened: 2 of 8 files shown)


persist/postgres/hosts.go
```diff
@@ -651,7 +652,8 @@ SELECT
 	hosts.id,
 	hosts.public_key,
 	hosts.country_code,
-	hosts.location
+	hosts.location,
+	(stuck_since IS NULL OR stuck_since >= NOW() - INTERVAL '24 hours') AND settings_remaining_storage > 0
```
peterjan: `nit` we should probably check whether it can store a sector

peterjan (changes requested):
The `host.Info()` method doesn't populate the GFU setting

persist/postgres/hosts_test.go
```diff
@@ -856,6 +858,34 @@ func TestUsableHosts(t *testing.T) {
...
+	// test GoodForUpload field
+	// set uh1 to have no remaining storage - should have GoodForUpload=false
+	settingsNoStorage := newTestHostSettings(uh1)
+	settingsNoStorage.RemainingStorage = 0
+	if err := db.UpdateHostScan(uh1, settingsNoStorage, locationUS, true, time.Now().Add(time.Hour)); err != nil {
+		t.Fatal(err)
+	}
+
+	// set uh2 to be stuck - should have GoodForUpload=false
+	if _, err := db.pool.Exec(t.Context(), `UPDATE hosts SET stuck_since = NOW() - INTERVAL '25 hours' WHERE public_key = $1`, sqlPublicKey(uh2)); err != nil {
```
peterjan: `nit` small improvement to the test would be to check a host that's been stuck since 23 hours to see whether that one's considered not-stuck.

hosts/host.go
```diff
@@ -131,11 +131,12 @@ type (
...
-		Addresses   []chain.NetAddress `json:"addresses"`
-		CountryCode string             `json:"countryCode"`
-		Latitude    float64            `json:"latitude"`
-		Longitude   float64            `json:"longitude"`
+		PublicKey     types.PublicKey    `json:"publicKey"`
+		Addresses     []chain.NetAddress `json:"addresses"`
+		CountryCode   string             `json:"countryCode"`
+		Latitude      float64            `json:"latitude"`
+		Longitude     float64            `json:"longitude"`
+		GoodForUpload bool               `json:"goodForUpload"`
```
peterjan, line 139: I think it's probably fine to expose this on `hosts.Host` too? cc @ChrisSchinnerl wdyt? The HostInfo should probably be a subset of the default host info.

persist/postgres/hosts.go
```diff
@@ -643,15 +646,17 @@ WITH globals AS (
 		settings_tip_height,
 		settings_valid_until,
 		settings_signature,
-		(get_byte(settings_protocol_version, 0) << 16) + (get_byte(settings_protocol_version, 1) << 8) + (get_byte(settings_protocol_version, 2)) as settings_version
+		(get_byte(settings_protocol_version, 0) << 16) + (get_byte(settings_protocol_version, 1) << 8) + (get_byte(settings_protocol_version, 2)) as settings_version,
+		(stuck_since IS NULL AND settings_remaining_storage >= $5) AS good_for_upload
```
ChrisSchinnerl: Might want to double check this but I think `remaining_storage` is already in sectors. So this could be `> 0`.

hosts/host.go
```diff
@@ -125,17 +125,19 @@ type (
 		LostSectors            uint64              `json:"lostSectors"`
 		AccountFunding         types.Currency      `json:"accountFunding"`
 		TotalSpent             types.Currency      `json:"totalSpent"`
+		GoodForUpload          bool                `json:"goodForUpload"`
```
ChrisSchinnerl: Let's not add that to the `Host` for now. and get rid of the `Info` method on the `Host` type. The methods on the spaced set can then accept a `Host` instead of `HostInfo`.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

persist/postgres/hosts.go
```diff
@@ -643,15 +645,17 @@ WITH globals AS (
 		settings_tip_height,
 		settings_valid_until,
 		settings_signature,
-		(get_byte(settings_protocol_version, 0) << 16) + (get_byte(settings_protocol_version, 1) << 8) + (get_byte(settings_protocol_version, 2)) as settings_version
+		(get_byte(settings_protocol_version, 0) << 16) + (get_byte(settings_protocol_version, 1) << 8) + (get_byte(settings_protocol_version, 2)) as settings_version,
+		(stuck_since IS NULL AND settings_remaining_storage > 0) AS good_for_upload
```
n8mgr, line 649: should you add that they have at least one good contract? Maintenance should ensure that is true, but we have run into it a number of times.

ChrisSchinnerl: That might be a good idea. `stuck_since` will guarantee that a host has at least one good contract but only if the host has >0 active contracts in the first place.
So we might want to drop in an `EXISTS` that checks for at least 1 active contract.

chris124567: I think we already filter for this in UsableHosts: https://github.com/SiaFoundation/indexd/blob/master/persist/postgres/hosts.go#L675

ChrisSchinnerl: You are right 👍

hosts/host.go
peterjan: This was resolved with a thumbs up, but was it eventually added? I'm asking because if it was added we don't have to update `dbHost` since it should already be there.

hosts/distance.go
```diff
@@ -22,7 +22,7 @@ func NewSpacedSet(minDistanceKm float64) *SpacedSet {
 
 // Add adds the host to the set if it is sufficiently spaced apart from the
 // existing hosts. It returns true if the host was added, false otherwise.
-func (s *SpacedSet) Add(h HostInfo) bool {
+func (s *SpacedSet) Add(h Host) bool {
```
peterjan, line 25: Fine with this change because it doesn't really matter I guess but generally it's a little sad to require more information than necessary. We only need the host's location for the spaced set so I'm not sure why we would all of a sudden have it required a full host. Fine to keep though, just wondering what the reason was changing it, seems unrelated to the PR

peterjan (changes requested):
Looks good to me, just checking to see if we should expose it on `hosts.Host` too. I think we should.

hosts/distance.go
chris124567: https://github.com/SiaFoundation/indexd/pull/750#discussion_r2704601303

hosts/host.go
chris124567: it was added then removed in response to https://github.com/SiaFoundation/indexd/pull/750#discussion_r2704601303

hosts/distance.go
peterjan: Missed that

PR #757, opened 2026-01-14
Provide real-time sector events

n8mgr:
Fun fun

### DIFF
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -12,6 +12,7 @@ import (
 	"time"
 	"unicode/utf8"
 
+	"github.com/gorilla/websocket"
 	"go.sia.tech/core/consensus"
 	proto "go.sia.tech/core/rhp/v4"
 	"go.sia.tech/core/types"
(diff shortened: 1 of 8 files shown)


ChrisSchinnerl:
Copilot made a few valid remarks but more importantly what do we do about the authentication? Make it public?

api/admin/admin.go
```diff
@@ -256,6 +268,9 @@ func NewAPI(chain ChainManager, accounts Accounts, contracts ContractManager, ho
 		"GET /stats/hosts":     a.handleGETStatsHosts,
 		"GET /stats/scans":     a.handleGETStatsScans,
 		"GET /stats/sectors":   a.handleGETStatsSectors,
+
+		// fun
+		"GET /events": a.handleWebsocket,
```
peterjan, line 273: Missing docs update

peterjan (changes requested):
Most (if not all) of the copilot remarks are valid and should be handled.

PR #770, opened 2026-01-21
Include renew period in collateral calculation logic

n8mgr:
Fixes contract renewal and refreshes to use the host's height for duration calculations instead of an abstract `period` to match RHP4 calculations.


n8mgr:
A note on contract durations: we form a contract for `period` blocks, but we renew it for `period + renew window` blocks. With the default settings we are forming a contract for 4w then renewing it for 6w, with 2w left on the original contract. That keeps the period constant which is probably the behavior we want, but it does make reasoning about the collateral logic during renewals more difficult. I'm assuming this is why our renewal estimates are failing. The host has enough max collateral for the data at 4w, but not at 6w.

ChrisSchinnerl (approved):

PR #772, opened 2026-01-23
Add `DELETE /contracts/:contractid` endpoint

### LINKED ISSUE #768: Add `DELETE /contracts/:contractid` endpoint
The endpoint should do the following for the provided contract

1. unpin all sectors from that contract by `NULL`ing the corresponding column
2. update the unpinned sector stats
3. mark the contract as !good

Reminder to also update the openapi docs

chris124567:
Close #768

### DIFF
--- a/persist/postgres/contracts.go
+++ b/persist/postgres/contracts.go
@@ -706,6 +706,12 @@ func (s *Store) DeleteContract(contractID types.FileContractID) error {
 			return fmt.Errorf("failed to unpin sectors: %w", err)
 		}
 
+		// delete the entry in contract_sectors_map_id
+		_, err = tx.Exec(ctx, `DELETE FROM contract_sectors_map WHERE id = $1`, contractMapID)
+		if err != nil {
+			return fmt.Errorf("failed to delete from contract sectors map: %w", err)
+		}
+
 		// update stats
 		if pinned > 0 {
 			if err := incrementNumPinnedSectors(ctx, tx, -pinned); err != nil {
--- a/persist/postgres/contracts.go
+++ b/persist/postgres/contracts.go
@@ -668,16 +668,20 @@ func (s *Store) MarkUnrenewableContractsBad(minProofHeight uint64) error {
 	})
 }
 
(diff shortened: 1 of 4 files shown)


persist/postgres/contracts.go
```diff
@@ -681,6 +681,51 @@ func (s *Store) MarkContractBad(contractID types.FileContractID) error {
...
+			if err := incrementNumPinnedSectors(ctx, tx, -pinned); err != nil {
+				return fmt.Errorf("failed to update pinned sectors stat: %w", err)
+			}
+			if err := incrementNumUnpinnedSectors(ctx, tx, pinned); err != nil {
+				return fmt.Errorf("failed to update unpinned sectors stat: %w", err)
+			}
+		}
+
+		// mark the contract as bad
+		_, err = tx.Exec(ctx, `UPDATE contracts SET good = FALSE WHERE contract_id = $1`, sqlHash256(contractID))
```
peterjan: we can be more strict here and check rows affected to ensure it's exactly 1 (`MarkContractBad` does this), we could also extract it and call it in these two spots to ensure the impl. is identical

persist/postgres/contracts.go
```diff
@@ -681,6 +681,51 @@ func (s *Store) MarkContractBad(contractID types.FileContractID) error {
 	})
 }
 
+// DeleteContract unpins all sectors from a contract, updates stats, and marks
+// the contract as bad.
+func (s *Store) DeleteContract(contractID types.FileContractID) error {
+	return s.transaction(func(ctx context.Context, tx *txn) error {
+		var contractMapID int64
+		err := tx.QueryRow(ctx, `SELECT id FROM contract_sectors_map WHERE contract_id = $1`, sqlHash256(contractID)).Scan(&contractMapID)
```
peterjan, line 693: I think we can cleanup the row when we're done right? I think it'll get pruned eventually (by `PruneContractSectorsMap`) but it's cleaner to get rid of it here.

peterjan (changes requested):

n8mgr (approved):

peterjan (approved):

Issue #636, opened 2025-11-05
Deduplicate networking code between the connection pool in the slab manager and the SDK


ChrisSchinnerl:
They share a lot of logic such as the retry logic and should be deduplicated.

chris124567:
This was done in in #696 and #664 right?

n8mgr:
Still need to replace it in ~~migrations, sector verification, account funding and pinning~~.

PR #775, opened 2026-01-28
Replace store mocks with postgres in `client` and `client/v2` tests

### LINKED ISSUE #350: Remove all Store mocks

chris124567:
Close #350

### DIFF
--- a/client/export_test.go
+++ b/client/export_test.go
@@ -1,0 +1,30 @@
+package client
+
+import (
+	"context"
+
+	proto "go.sia.tech/core/rhp/v4"
+	"go.sia.tech/core/types"
+	"go.sia.tech/coreutils/rhp/v4"
+	"go.uber.org/zap"
+)
+
+type RPCLatestRevisionFn = rpcLatestRevisionFn
+
+func NewHostClient(hk types.PublicKey, cm ChainManager, client rhp.TransportClient, signer rhp.FormContractSigner, store RevisionStore, revisionSubmissionBuffer uint64, log *zap.Logger) *HostClient {
+	return newHostClient(hk, cm, client, signer, store, revisionSubmissionBuffer, log)
+}
+
+func (c *HostClient) SetLatestRevisionFn(fn RPCLatestRevisionFn) {
+	c.latestRevisionFn = fn
+}
+
+func (c *HostClient) WithRevision(ctx context.Context, contractID types.FileContractID, reviseFn func(contract rhp.ContractRevision) (rhp.ContractRevision, proto.Usage, error)) error {
+	return c.withRevision(ctx, contractID, reviseFn)
+}
+
+const DefaultRevisionSubmissionBuffer = defaultRevisionSubmissionBuffer
+
+func IsBeyondMaxRevisionHeight(proofHeight, revisionSubmissionBuffer, blockHeight uint64) bool {
+	return isBeyondMaxRevisionHeight(proofHeight, revisionSubmissionBuffer, blockHeight)
+}
--- a/client/export_test.go
+++ b/client/export_test.go
@@ -6,14 +6,16 @@ import (
 	proto "go.sia.tech/core/rhp/v4"
 	"go.sia.tech/core/types"
 	"go.sia.tech/coreutils/rhp/v4"
-	"go.uber.org/zap"
 )
 
 type RPCLatestRevisionFn = rpcLatestRevisionFn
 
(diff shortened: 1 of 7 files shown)


client/v2/candidates_test.go
```diff
@@ -73,59 +77,84 @@ func TestFailureRateTimeDecay(t *testing.T) {
...
+		MaxCollateral:       types.Siacoins(10000),
+		MaxContractDuration: 10000,
+		RemainingStorage:    oneTB,
+		Prices: proto.HostPrices{
+			TipHeight:    0,
+			ValidUntil:   time.Now().Add(time.Hour),
+			Collateral:   types.Siacoins(1000).Div64(oneTB).Div64(blocksPerMonth),
+			StoragePrice: types.Siacoins(100).Div64(oneTB).Div64(blocksPerMonth),
+		},
+	}
```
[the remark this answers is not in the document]
chris124567: It didn't break anything but I fixed this anyways

client/export_test.go
```diff
@@ -0,0 +1,30 @@
...
+	c.latestRevisionFn = fn
+}
+
+func (c *HostClient) WithRevision(ctx context.Context, contractID types.FileContractID, reviseFn func(contract rhp.ContractRevision) (rhp.ContractRevision, proto.Usage, error)) error {
+	return c.withRevision(ctx, contractID, reviseFn)
+}
+
+const DefaultRevisionSubmissionBuffer = defaultRevisionSubmissionBuffer
+
+func IsBeyondMaxRevisionHeight(proofHeight, revisionSubmissionBuffer, blockHeight uint64) bool {
```
peterjan: This is kind of unfortunate and it looks like `isBeyondMaxRevisionHeight` is duplicated in two places. Maybe we should move this to the contracts package and expose it as a function there. If we get rid of that we can keep these test exports focussed on the revision functions.

peterjan (changes requested):
These export files are a little unfortunate but it's quite a nasty import cycle indeed. Not sure whether it's that much of an issue, apparently this is the idiomatic approach, although it feels like we could/should try and get rid of the import cycle that causes it. That's quite the refactor though so not sure if it's worth tackling at this point.

client/v2/export_test.go
ChrisSchinnerl, line 1: If you rename this file to `stats_test.go` and move `TestFailureRate`, `TestFailureRateTimeDecay` and `TestRPCAverage` in here, you don't need to export anything. 

Better to move tests around than re-exporting things.

client/export_test.go
ChrisSchinnerl, line 1: This isn't great but I don't think it's worth looking into ways to resolve the import cycle between the v1 client and the contracts package since we are [planning](https://github.com/SiaFoundation/indexd/issues/636) on getting rid of the v1 client and as a result of these tests anyway.

ChrisSchinnerl (changes requested):

client/export_test.go
chris124567: So do we want to only do the changes for `client/v2` then?

client/v2/candidates_test.go
```diff
@@ -1,131 +1,96 @@
...
-		fr.AddSample(false)
-		expected := 1.0
-		if v := fr.Value(); v != expected {
-			t.Fatalf("expected %f, got %f", expected, v)
-		}
-
-		time.Sleep(totalDecayMinutes * time.Minute)
-		synctest.Wait()
+// addUsableHost adds a host considered usable by UsableHosts
+func (s *testStore) addUsableHost(t testing.TB, pk types.PublicKey, goodForUpload bool) {
```
peterjan: Perhaps the test store could keep a counter and drop the GFU bool. We only ever pass `false` once anyway. That way we get something like:

```
h1 := s.addUsableHost(t)
h2 := s.addUsableHost(t)
h3 := s.addUsableHost(t)

// mark h3 as !GFU
badSettings := s.HostSettings(h3)
badSettings.RemainingStore = 0
s.UpdateHostScan(h3, badSettings, ...)
```

peterjan (dismissed):

ChrisSchinnerl (commented):
Lgtm overall but tests are failing due to db issues

peterjan (approved):

ChrisSchinnerl (approved):

PR #784, opened 2026-02-03
Add packing support

### LINKED ISSUE #776: SDK: Upload packing
Same as in the Rust SDK but for our Go SDK.

peterjan:
Closes https://github.com/SiaFoundation/indexd/issues/776

### DIFF
--- a/sdk/README.md
+++ b/sdk/README.md
@@ -93,7 +93,7 @@ into shared slabs, reducing overhead and improving upload efficiency.
 
 ```go
 // create a packed upload
-packed, err := client.UploadPacked(ctx)
+packed, err := client.UploadPacked()
 if err != nil {
 	log.Fatal("failed to create packed upload")
 }
@@ -133,7 +133,7 @@ const paddingTarget = 0.05
 var uploads [][]byte
 for len(uploads) > 0 {
 	err := func() error {
-		packed, err := client.UploadPacked(ctx)
+		packed, err := client.UploadPacked()
 		if err != nil {
 			return err
 		}
--- a/sdk/README.md
+++ b/sdk/README.md
@@ -97,6 +97,7 @@ packed, err := client.UploadPacked(ctx)
 if err != nil {
 	log.Fatal("failed to create packed upload")
 }
+defer packed.Close()
 
 // add multiple objects
 for _, data := range smallObjects {
@@ -115,32 +116,59 @@ if err != nil {
 for i, obj := range objects {
 	var buf bytes.Buffer
 	if err := client.Download(ctx, &buf, obj); err != nil {
-		log.Fatal("failed to download object", i)
+		log.Fatalf("failed to download object %d", i)
 	}
 }
 ```
 
 ### Optimizing Packed Uploads
 
-To minimize padding and maximize efficiency, use `Remaining()` to check how
-much space is left in the current slab before it spills over to a new one:
+Use `Remaining()` and `Length()` to monitor efficiency and decide when to
+finalize. This is useful when you have a stream of objects and want to avoid
+excessive padding:
 
 ```go
-packed, _ := client.UploadPacked(ctx)
+const paddingTarget = 0.05
 
-// prioritize objects that fit in the remaining space
-for _, obj := range sortBySize(objects) {
-	if int64(len(obj)) <= packed.Remaining() {
-		packed.Add(ctx, bytes.NewReader(obj))
+var uploads [][]byte
+for len(uploads) > 0 {
+	err := func() error {
+		packed, err := client.UploadPacked(ctx)
+		if err != nil {
+			return err
+		}
+		defer packed.Close()
+
+		for len(uploads) > 0 {
+			_, err := packed.Add(ctx, bytes.NewReader(uploads[0]))
+			if err != nil {
+				return err
+			}
+			uploads = uploads[1:]
+
+			if packed.Length() == 0 {
+				continue
+			}
+
+			padding := float64(packed.Remaining()) / float64(packed.Length())
+			if padding <= paddingTarget {
+				break
+			}
+		}
+
+		objects, err := packed.Finalize(ctx)
+		if err != nil {
+			return err
+		}
+
+		// pin objects
+
+		return nil
+	}()
+	if err != nil {
+		log.Fatal(err)
 	}
 }
-
-// add remaining objects
-for _, obj := range remainingObjects {
-	packed.Add(ctx, bytes.NewReader(obj))
-}
-
-objects, _ := packed.Finalize(ctx)
 ```
 
 ### Limitations
--- a/sdk/README.md
+++ b/sdk/README.md
@@ -84,3 +84,67 @@ if err != nil {
 The `metadata` returned from the `Upload` function contains information about
 the uploaded file required to download it later. Applications need to store
 this information in a database or some other persistent storage.
+
+## Packed Uploads
+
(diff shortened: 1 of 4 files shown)


sdk/README.md
```diff
@@ -84,3 +84,67 @@ if err != nil {
...
+		packed.Add(ctx, bytes.NewReader(obj))
+	}
+}
+
+// add remaining objects
+for _, obj := range remainingObjects {
+	packed.Add(ctx, bytes.NewReader(obj))
+}
+
+objects, _ := packed.Finalize(ctx)
```
ChrisSchinnerl: If you push all objects anyway before finalizing then the order you push them in shouldn't matter. So in that regard this example might be a bit confusing. It also won't upload objects greater than a slab.

I think it's tricky to provide a good but also simple example here but maybe we can demonstrate how to use `Remaining()` and `Length()` here to determine when to call `finalize` on an arbitrary number of objects to upload. 

e.g.
```go
paddingTarget := 0.05 // 95% efficiency

for {
  padding := float64(packed.Remaining()) / float64(packed.Length())
  if padding <= paddingTarget && packed.Length() > 0 {
    // we are efficient! finalize and start a new packed upload!
    packed.Finalize()
    return
  }

  // add more stuff
}
```

sdk/packing_test.go
```diff
@@ -0,0 +1,214 @@
...
+		n, err := u.Add(t.Context(), bytes.NewReader(data))
+		if err != nil {
+			t.Fatal(err)
+		} else if n != int64(len(data)) {
+			t.Fatalf("unexpected number of bytes returned, %d != %d", len(data), n)
+		}
+		total += len(data)
+	}
+
+	// assert total - and remaining length
```
ChrisSchinnerl:
```suggestion
	// assert total and remaining length
```

sdk/packing.go
```diff
@@ -0,0 +1,293 @@
...
+// slabSize returns the size of a slab based on the number of data shards.
+func (u *PackedUpload) slabSize() int64 {
+	return int64(u.dataShards) * proto4.SectorSize
+}
+
+// UploadPacked creates a new packed upload. This allows multiple objects to be
+// packed together for more efficient uploads. The returned PackedUpload can be
+// used to add objects and then finalized to get the resulting objects. A packed
+// upload is not thread-safe.
+func (s *SDK) UploadPacked(ctx context.Context, opts ...UploadOption) (*PackedUpload, error) {
```
ChrisSchinnerl: Overall the code looks good but the synchronization is a bit odd. `UploadPacked` doesn't block on anything yet it is passed a `ctx`.
On the other hand `Add` is the method that blocks but the `ctx` that is passed to it is useless since it is only check in the beginning and then the copy can block forever. Of course only until the `ctx` is closed that is passed to `UploadPacked` but that is quite sneaky. 

So I would try and get rid of the `ctx` here and instead of storing `cancel` on the `PackedUpload`, store the `reader` you get from `io.Pipe`. In `Add`, you can then spin up a goroutine which calls `reader.CloseWithError(ctx.Err())` if the ctx is closed before `Add` finishes. In `Finalize` after closing the `writer` you could defer close the `reader` as well for completeness. 

That will give the `ctx` in `Add` the ability to actually interrupt the upload.

ChrisSchinnerl (changes requested):
Left a few nits and one concern regarding the synchronization

sdk/packing.go
peterjan: Aah nice, yeah I was aware that io.Copy wasn't respecting the given context but didn't consider closing the reader!

sdk/README.md
peterjan: good idea

sdk/packing.go
```diff
@@ -0,0 +1,304 @@
...
+			u.reader.CloseWithError(context.Cause(ctx))
+		}
+	}()
+
+	// pipe data into writer
+	n, err := io.Copy(u.writer, r)
+	if err != nil {
+		// add was cancelled, overwrite the error
+		if ctx.Err() != nil {
+			err = context.Cause(ctx)
```
ChrisSchinnerl: Is this necessary? Doesn't `io.Copy` return the error you passed to `CloseWithError`?

sdk/packing.go
```diff
@@ -0,0 +1,304 @@
...
+// slabSize returns the size of a slab based on the number of data shards.
+func (u *PackedUpload) slabSize() int64 {
+	return int64(u.dataShards) * proto4.SectorSize
+}
+
+// UploadPacked creates a new packed upload. This allows multiple objects to be
+// packed together for more efficient uploads. The returned PackedUpload can be
+// used to add objects and then finalized to get the resulting objects. A packed
+// upload is not thread-safe.
+func (s *SDK) UploadPacked(ctx context.Context, opts ...UploadOption) (*PackedUpload, error) {
```
ChrisSchinnerl: I think we might need a `Close` too maybe. I can't think of a good way to clean up the background goroutine without it.
We can't really rely on `ctx` because it might be the `Background` context and if `Finalize` isn't called the goroutine will be leaked. It also feels odd to have a `ctx` passed in a non-blocking method.

If we combine a `Close` with an internal context/channel/threadgroup it would be clear that `Close` needs to be called to clean up resources.

peterjan: I considered that but I didn't want to move away too much from the Rust impl since we have an issue to ensure consistent UX across all of our SDKs. I do agree a `Close` is necessary though so I'll add that, I figured we might also leave it up to the developer to call `Finalize` but having a close is def. better.

edit: it's kind of annoying in practice though, because you have to defer the close so if you're looping over uploads and creating packed uploads you have to defer from inside a loop which is awful

sdk/packing.go
peterjan: It does but there's a race condition where if you close both ends of the pipe it returns `io: read/write on closed pipe`. I'll double back on this to see if it's really necessary though, I thought it was pretty harmless. I was on the fence on adding a cause anyway but I think we should be doing that more often, especially in our SDKs since there might be developers that consider that a good practice.

sdk/packing.go
n8mgr: +1 for `defer upload.Close()`

At least from the rust side it's common to have `Drop` abort things implicitly. The shape of the API will be slightly different in each language. Idiomatic for the language > 1:1 match.

sdk/packing.go
```diff
@@ -0,0 +1,337 @@
...
+
+		reader: reader,
+		writer: writer,
+
+		resultAvailCh: make(chan struct{}),
+		tg:            threadgroup.New(),
+	}
+
+	// register with threadgroup to ensure proper cleanup of goroutines on Close
+	bgCtx, bgCancel, err := u.tg.AddContext(ctx)
```
ChrisSchinnerl: Use the background context here and don't pass `ctx` to UploadPacked. `Close` will handle the cleanup now and `UploadPacked` itself doesn't require unblocking. 

If we wanted to be extra brave we could panic on error here since `AddContext` only returns an error on a closed ThreadGroup. Then `UploadPacked` doesn't even require returning an error anymore.

ChrisSchinnerl (changes requested):

sdk/packing.go
peterjan: We can't get rid of the error because we have to validate the options and they might not be valid EC params. I didn't consider getting rid of the context though, I'll update that, thanks!

n8mgr (approved):

ChrisSchinnerl (approved):

PR #789, opened 2026-02-05
Introduce `maxRenewableContractSize`

ChrisSchinnerl:
This PR does 2 things to reduce our issues with contracts not getting renewed due to too much host collateral being requested.

1. updates renewals to use the period rather than period + renew window to make sure both formations and renewals use the same duration
2. introduce a maximum contract size and enforce it

The latter is a computed maximum contract size over a full contract period. This is really just a less reliable version of a max contract size field on the host settings but since we still don't know how to best implement that this is what we can do right now.

The reason we need this is the fact that right now collateral for a sector gets cheaper as time passes. Let's assume 1 sector costs 1SC per block in collateral. If the sector is uploaded right when forming a contract with a period of 100 blocks this means 100SC of collateral for that sector. If you upload the sector right at the end it costs 1SC. 
So imagine a host has a max collateral of 50SC and we upload 50 sectors at the end of the contract. To the host this is fine but when we renew that contract we need to pay for the next 100 blocks of collateral which would be 5000SC which is above the max collateral and therefore fails.

This PR solves this by making sure we consider a contract as not good as soon as we approach that max size and form a new one and by not going above the limit.

It's not foolproof since a host might update its settings which is why I gave it a 20% safety margin until the host itself enforces a max contract size. Once we have that, `maxRenewableContractSize` basically becomes `return min(hostSettings.MaxContractSize, maxContractSize)`

### DIFF
--- a/contracts/renew.go
+++ b/contracts/renew.go
@@ -13,7 +13,7 @@ import (
 func (cm *ContractManager) performContractRenewals(ctx context.Context, period, renewWindow uint64, log *zap.Logger) error {
 	bh := cm.chain.TipState().Index.Height
 	minProofHeight := bh + renewWindow
-	newProofHeight := bh + period + renewWindow
+	newProofHeight := bh + period
 
 	batchSize := 50
 	for offset := 0; ; offset += batchSize {
--- a/client/client.go
+++ b/client/client.go
@@ -17,8 +17,6 @@ import (
 )
 
 const (
-	maxContractSize = 10 * 1 << 40 // 10TB
-
 	// defaultRevisionSubmissionBuffer is a buffer that mainnet hosts apply on
 	// the contract's proof height before they consider a contract revisable, so
 	// if the current block height plus the buffer exceed the proof height, the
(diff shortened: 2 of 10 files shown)


contracts/renew.go
```diff
@@ -13,7 +13,7 @@ import (
 func (cm *ContractManager) performContractRenewals(ctx context.Context, period, renewWindow uint64, log *zap.Logger) error {
 	bh := cm.chain.TipState().Index.Height
 	minProofHeight := bh + renewWindow
-	newProofHeight := bh + period + renewWindow
+	newProofHeight := bh + period
```
peterjan, line 16: `minProofHeight := bh + renewWindow` shouldn't we be updating that too?

peterjan (changes requested):
Makes sense! left one small remark

contracts/renew.go
ChrisSchinnerl: To what should that be updated?

peterjan: Hm yeah nevermind, it looked like you forgot to apply the change to the `minProofHeight` also. We never validate the period is always strictly higher than the renew window, might make sense to add that. We don't expose that but we do have `WithMaintenanceSettings` theoretically.

peterjan (approved):

n8mgr (approved):

PR #793, opened 2026-02-05
Add host stat for number of hosts with active contracts

### LINKED ISSUE #777: Add host stat for number of hosts with active contracts
Meaning that the host has good and not stuck contracts

chris124567:
Close #777

### DIFF
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -106,8 +106,8 @@ type (
 	// A Store is a persistent store for the indexer.
 	Store interface {
 		AccountStats() (AccountStatsResponse, error)
+		AllHostsStats() (AllHostsStatsResponse, error)
 		ContractsStats() (ContractsStatsResponse, error)
-		ScanStats() (ScansStatsResponse, error)
 		HostStats(offset, limit int) ([]hosts.HostStats, error)
 		SectorStats() (SectorsStatsResponse, error)
 
@@ -256,7 +256,7 @@ func NewAPI(chain ChainManager, accounts Accounts, contracts ContractManager, ho
 		"GET /stats/accounts":  a.handleGETStatsAccounts,
 		"GET /stats/contracts": a.handleGETStatsContracts,
 		"GET /stats/hosts":     a.handleGETStatsHosts,
-		"GET /stats/scans":     a.handleGETStatsScans,
+		"GET /stats/hosts/all": a.handleGETStatsHostsAll,
 		"GET /stats/sectors":   a.handleGETStatsSectors,
 	}
 
@@ -1011,9 +1011,9 @@ func (a *admin) handleGETStatsHosts(jc jape.Context) {
 	writeResponse(jc, HostStatsResponse(res))
 }
 
-func (a *admin) handleGETStatsScans(jc jape.Context) {
-	stats, err := a.store.ScanStats()
-	if jc.Check("failed to retrieve host scan stats", err) != nil {
+func (a *admin) handleGETStatsHostsAll(jc jape.Context) {
+	stats, err := a.store.AllHostsStats()
+	if jc.Check("failed to retrieve all hosts stats", err) != nil {
 		return
 	}
 	writeResponse(jc, stats)
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -106,7 +106,7 @@ type (
 	// A Store is a persistent store for the indexer.
 	Store interface {
 		AccountStats() (AccountStatsResponse, error)
-		AllHostsStats() (AllHostsStatsResponse, error)
+		AggregatedHostStats() (AggregatedHostStatsResponse, error)
 		ContractsStats() (ContractsStatsResponse, error)
 		HostStats(offset, limit int) ([]hosts.HostStats, error)
 		SectorStats() (SectorsStatsResponse, error)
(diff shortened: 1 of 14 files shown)


persist/postgres/contracts.go
```diff
@@ -78,9 +78,30 @@ func (s *Store) ContractsStats() (resp admin.ContractsStatsResponse, _ error) {
...
+			SELECT COUNT(DISTINCT c.host_id)
+			FROM contracts c
+			INNER JOIN hosts h ON c.host_id = h.id
+			CROSS JOIN globals
+			WHERE
+				c.good = TRUE AND
+				c.state IN (0,1) AND
+				c.renewed_to IS NULL AND
+				c.proof_height > globals.scanned_height AND
+				(h.stuck_since IS NULL OR h.stuck_since > NOW() - INTERVAL '24 hours')
```
ChrisSchinnerl: Since we have this check in multiple places now how about we get rid of the `INTERVAL` part everywhere?
From staring at the dashboard for the last couple of weeks I came to the conclusion that it actually feels detrimental to have this 24 hours lag between what the data already shows us (sectors getting stuck) and not seeing why.

ChrisSchinnerl (changes requested):

peterjan (changes requested):
This PR extends the `ContractsStatsResponse`, while this should be tied to hosts more than contracts. We have a `HostStats` endpoint but that one's a little annoying since it returns a stats object on a per-host basis. That should be enough to get us all the stats we want though. An alternative would be to extend the host stat with a `Usable` field maybe? That feels like a cleaner approach because it already has `blocked` and `stuck` status, would be nice to extend that with `usable` and we can aggregate it in the dashboard to get the total count. We apply this pattern a couple of times by now so it would be more consistent as well. CC'ing @ChrisSchinnerl  to confirm this request before making any changes

ChrisSchinnerl:
> This PR extends the `ContractsStatsResponse`, while this should be tied to hosts more than contracts. We have a `HostStats` endpoint but that one's a little annoying since it returns a stats object on a per-host basis. That should be enough to get us all the stats we want though. An alternative would be to extend the host stat with a `Usable` field maybe? That feels like a cleaner approach because it already has `blocked` and `stuck` status, would be nice to extend that with `usable` and we can aggregate it in the dashboard to get the total count. We apply this pattern a couple of times by now so it would be more consistent as well. CC'ing @ChrisSchinnerl to confirm this request before making any changes

The reason for this change is the setup dialog in the UI which currently shows wrong numbers and doesn't have an easy way to get the number of hosts we have good contracts with. Ideally we provide a single number they can use rather than adding more roundabout ways to get it. 

I agree that adding it to the contracts stats is not ideal. The other option would be an aggregated host stats endpoint. 
But then we need to revisit the whole `/hosts` issue too because I don't want to touch that again knowing that we have to change it yet again later when we reach more than 500 active hosts.

openapi/components.yml
```diff
@@ -820,6 +820,10 @@ components:
           type: integer
           format: int64
           description: Total number of active, bad contracts managed by this instance
+        activeHosts:
```
ChrisSchinnerl, line 823: Outdated now

ChrisSchinnerl (changes requested):

persist/postgres/hosts.go
```diff
@@ -1065,8 +1065,7 @@ func (s *Store) StuckHosts() ([]hosts.StuckHost, error) {
 		rows, err := tx.Query(ctx, `
 			SELECT public_key, stuck_since, unpinned_sectors
 			FROM hosts
-			WHERE stuck_since IS NOT NULL
-				AND stuck_since < NOW() - INTERVAL '24 hours'`)
+			WHERE stuck_since IS NOT NULL`)
```
peterjan, line 1068: This change is unrelated to the PR but now that we have it we might as well keep it. Do a search though for '24 hours', there's various locations where we mention that and I don't think this PR gets rid of all of them

persist/postgres/stats.go
```diff
@@ -215,7 +238,7 @@ func (s *Store) HostStats(offset, limit int) ([]hosts.HostStats, error) {
 			); err != nil {
 				return err
 			}
-			if stuckSince.Valid && time.Since(stuckSince.Time) >= 24*time.Hour {
+			if stuckSince.Valid {
 				hs.StuckSince = &stuckSince.Time
```
peterjan, line 242: `nit` do we need both `Stuck` and `StuckSince`? It was already a bit weird but now that `NOT NULL` means `stuck` it might make sense to get rid of the two fields that essentially mean exactly the same thing. Leaving that up to @ChrisSchinnerl  whether or not to tackle that. Definitely a `nit`

api/admin/admin.go
```diff
@@ -256,7 +256,7 @@ func NewAPI(chain ChainManager, accounts Accounts, contracts ContractManager, ho
 		"GET /stats/accounts":  a.handleGETStatsAccounts,
 		"GET /stats/contracts": a.handleGETStatsContracts,
 		"GET /stats/hosts":     a.handleGETStatsHosts,
-		"GET /stats/scans":     a.handleGETStatsScans,
+		"GET /stats/hosts/all": a.handleGETStatsHostsAll,
```
peterjan: I wasn't going to say anything at first but `all` really feels weird. Wouldn't `aggregated` be a better choice here? All feels like you get what you would get from `/stats/hosts` but without a limit applied...

peterjan (changes requested):

persist/postgres/stats.go
ChrisSchinnerl: We could drop Stuck indeed and make StuckSince omit empty if it isn't already. The dashboard doesn't rely on that field anymore and instead uses stuck since already.

api/admin/admin.go
ChrisSchinnerl: I'd maybe flip it around. `/contracts` and `/accounts` are all aggregated so `/hosts` could be too and then we find a good term for the more detailed response. `individual`, `detailed` etc.

openapi/components.yml
```diff
@@ -885,26 +882,31 @@ components:
...
               type: string
             description: Reasons why the host was blocked, if any
-          stuck:
-            type: boolean
-            description: Whether the host is currently stuck. A host is stuck when contract operations have been failing for more than 24 hours.
           stuckSince:
             type: string
             format: date-time
             nullable: true
             description: Timestamp when the host became stuck. Only present when stuck is true.
```
peterjan: remove `Only present when stuck is true.`

persist/postgres/hosts.go
```diff
@@ -1058,15 +1058,14 @@ func (s *Store) UpdateStuckHosts(stuck []types.PublicKey) error {
 }
 
 // StuckHosts returns a list of stuck hosts with the timestamp they first
-// became stuck. A host is stuck if stuck_since is more than 24 hours ago.
+// became stuck. A host is considered stuck if stuck_since is non null.
```
peterjan:
```suggestion
// became stuck.
```
That's more or less implied now

peterjan (dismissed):

ChrisSchinnerl (approved):

PR #796, opened 2026-02-06
Add retry mechanism in store.transaction

### LINKED ISSUE #792: Add retry mechanism to database transaction
Similar to `hostd` and `renterd` which already do it for their SQLite and MySQL transactions we should attempt retrying a transaction if a deadlock was detected.

This also means going over all transactions we open and making sure they are safe to be retried. They might not be if they are not atomic or capture/modify variables from outside the transactions scope. 

We should also make sure to log an error if a transaction fails all attempts.

peterjan:
Closes https://github.com/SiaFoundation/indexd/issues/792

### DIFF
--- a/persist/postgres/accounts.go
+++ b/persist/postgres/accounts.go
@@ -18,7 +18,7 @@ import (
 )
 
 // Accounts returns a list of account keys.
-func (s *Store) Accounts(offset, limit int, opts ...accounts.QueryAccountsOpt) ([]accounts.Account, error) {
+func (s *Store) Accounts(offset, limit int, opts ...accounts.QueryAccountsOpt) (accs []accounts.Account, err error) {
 	if err := validateOffsetLimit(offset, limit); err != nil {
 		return nil, err
 	} else if limit == 0 {
@@ -30,7 +30,6 @@ func (s *Store) Accounts(offset, limit int, opts ...accounts.QueryAccountsOpt) (
 		opt(&queryOpts)
 	}
 
-	accs := make([]accounts.Account, 0, limit)
 	if err := s.transaction(func(ctx context.Context, tx *txn) (err error) {
 		accs = accs[:0] // reuse same slice if transaction retries
 
@@ -163,7 +162,7 @@ func (s *Store) PruneAccounts(limit int) error {
 	}
 
 	return s.transaction(func(ctx context.Context, tx *txn) error {
-		txLimit := limit // reset per transaction attempt
+		remaining := limit // reset per transaction attempt
 
 		var accountID int64
 		err := tx.QueryRow(ctx, `SELECT id FROM accounts WHERE deleted_at IS NOT NULL ORDER by deleted_at LIMIT 1`).Scan(&accountID)
@@ -182,7 +181,7 @@ USING (
 	LIMIT $2
 ) d
 WHERE o.id = d.id
-RETURNING o.object_key;`, accountID, txLimit)
+RETURNING o.object_key;`, accountID, remaining)
 		if err != nil {
 			return fmt.Errorf("failed to delete objects: %w", err)
 		}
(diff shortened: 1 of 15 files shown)


persist/postgres/store.go
```diff
@@ -37,18 +48,66 @@ func (ci ConnectionInfo) String() string {
...
 	}
-	defer tx.Rollback(ctx)
+	start := time.Now()
+	defer func() {
+		err := tx.Rollback(ctx)
+		if err != nil {
+			log.Error("failed to rollback transaction", zap.Error(err))
+		}
+		if time.Since(start) > longTxnDuration {
+			log.Debug("long transaction", zap.Duration("elapsed", time.Since(start)), zap.Stack("stack"), zap.Bool("failed", err != nil))
```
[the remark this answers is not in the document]
peterjan: Wait what? This should be marked outdated

ChrisSchinnerl (changes requested):
Don't forget the second part of the issue. Checking whether calling `transaction` with an implicit retry is safe. It's seems like a majority of methods we have at the very least appends to a slice that is returned or does things like subtracting from a limit.

peterjan:
> Don't forget the second part of the issue. Checking whether calling `transaction` with an implicit retry is safe. It's seems like a majority of methods we have at the very least appends to a slice that is returned or does things like subtracting from a limit.

Oh yeah, man there's quite a few 😓 . I've updated them. I think the approach I took is probably fine, but I was considering alternative approaches that forces the developer to think about transaction retries. I added a comment to nearly every case to help copilot spot issues when opening new PRs. 

edit: there's also two in `renterd` that went by unnoticed - I'll open a PR to fix those

persist/postgres/accounts.go
```diff
@@ -158,8 +163,9 @@ func (s *Store) PruneAccounts(limit int) error {
 	}
 
 	return s.transaction(func(ctx context.Context, tx *txn) error {
-		var accountID int64
+		txLimit := limit // reset per transaction attempt
```
ChrisSchinnerl: (nit) `remaining := limit`

persist/postgres/hosts.go
```diff
@@ -143,8 +143,10 @@ func (s *Store) Hosts(offset, limit int, queryOpts ...hosts.HostQueryOpt) ([]hos
 		return nil, err
 	}
 
-	var hosts []hosts.Host
+	hosts := make([]hosts.Host, 0, min(limit, 1000))
```
ChrisSchinnerl: I'd keep all of those the way they were since it's not related to the retry mechanism.
It's really a hit and miss anyway and in some cases like further down passing a high limit might actually cause massive allocations.

persist/postgres/accounts.go
```diff
@@ -248,6 +254,9 @@ func (s *Store) HostAccountsForFunding(hk types.PublicKey, threshold time.Time,
 
 	accs := make([]accounts.HostAccount, 0, limit)
 	if err := s.transaction(func(ctx context.Context, tx *txn) error {
+		accs = accs[:0]  // reuse same slice if transaction retries
+		txLimit := limit // reset per transaction attempt
```
ChrisSchinnerl: (nit) `remaining` is probably a better name than `txLimit`. Not just here but other spots as well.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (dismissed):

persist/postgres/slabs.go
```diff
@@ -204,12 +208,11 @@ LIMIT $2
...
 				return fmt.Errorf("failed to get slabs to unpin: %w", err)
-			} else if len(slabIDs) < batchSize {
-				exhausted = true
-			}
-			if err := s.unpinSlabs(ctx, tx, id, slabIDs); err != nil {
+			} else if err := s.unpinSlabs(ctx, tx, id, slabIDs); err != nil {
 				return fmt.Errorf("failed to unpin slabs: %w", err)
 			}
+
+			exhausted = true
```
chris124567: Shouldn't we do this?

```
            exhausted = len(slabIDs) < batchSize
```

peterjan: heh :/ thanks for that, not sure how that got in there

ChrisSchinnerl (approved):

chris124567 (approved):

PR #797, opened 2026-02-06
Add quotas for app connect keys

ChrisSchinnerl:
This PR adds a `quotas` table which contains custom settings we might want to associate with app connect keys and their associated accounts. 
Initially only the "default" quota exists which is also the quota assigned by the migration in this PR. 

What this PR doesn't do is add handling for when a user ends up exceeding their limit due to lowering their quota. We will cross that bridge when we come to it.

This is also going to be useful for suspending accounts without deleting them. A "suspended" quota would be all zero limits and lead to accounts not being funded at all. While we still migrate the data.

### DIFF
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -385,9 +385,16 @@ func (a *admin) handlePUTAppConnectKeys(jc jape.Context) {
 	if jc.Decode(&key) != nil {
 		return
 	}
+	if key.Quota == "" {
+		jc.Error(errors.New("quota is required"), http.StatusBadRequest)
+		return
+	}
 
 	_, err := a.accounts.UpdateAppConnectKey(jc.Request.Context(), key)
 	if errors.Is(err, accounts.ErrKeyNotFound) {
+		jc.Error(err, http.StatusNotFound)
+		return
+	} else if errors.Is(err, accounts.ErrQuotaNotFound) {
 		jc.Error(err, http.StatusNotFound)
 		return
 	} else if jc.Check("failed to update app connect key", err) != nil {
@@ -478,6 +485,10 @@ func (a *admin) handlePUTQuota(jc jape.Context) {
 
 	var req accounts.PutQuotaRequest
 	if jc.Decode(&req) != nil {
+		return
+	}
+	if req.TotalUses < 0 {
+		jc.Error(errors.New("totalUses must be non-negative"), http.StatusBadRequest)
 		return
 	}
 
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/hex"
 	"net/http"
+	"strings"
 
 	"errors"
 	"fmt"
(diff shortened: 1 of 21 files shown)


api/admin/admin.go
```diff
@@ -414,6 +429,83 @@ func (a *admin) handleDELETEAppConnectKeys(jc jape.Context) {
...
+	// validate key
+	if key == "" {
+		jc.Error(errors.New("key cannot be empty"), http.StatusBadRequest)
+		return
+	} else if len(key) > 32 {
+		jc.Error(errors.New("key cannot be longer than 32 characters"), http.StatusBadRequest)
+		return
+	} else if strings.Contains(key, " ") {
+		jc.Error(errors.New("key cannot contain spaces"), http.StatusBadRequest)
+		return
```
ChrisSchinnerl, line 480: This might seem a bit arbitrary but I think restricting a quota key to a sane character set and size makes sense.

n8mgr (dismissed):
Did you want to add max account funding in here too?

ChrisSchinnerl:
> Did you want to add max account funding in here too?

Already got it queued up locally for another PR

persist/postgres/quotas.go
```diff
@@ -0,0 +1,99 @@
...
+			return accounts.ErrQuotaNotFound
+		}
+		return err
+	})
+	return
+}
+
+// Quotas retrieves a list of quotas from the database.
+func (s *Store) Quotas(offset, limit int) (quotas []accounts.Quota, err error) {
+	err = s.transaction(func(ctx context.Context, tx *txn) error {
```
peterjan, line 77: we'll have to make sure to update this to be able to handle a transaction retry

persist/postgres/quotas_test.go
```diff
@@ -0,0 +1,128 @@
+package postgres
+
+import (
+	"testing"
+
+	"go.sia.tech/indexd/accounts"
+	"go.uber.org/zap/zaptest"
+)
+
+func TestQuotas(t *testing.T) {
```
peterjan, line 10: `nit` could add a small case that asserts offset and limit

api/admin/admin.go
```diff
@@ -1047,3 +1148,9 @@ func writeResponse(jc jape.Context, resp prometheus.Marshaller) {
 		jc.Encode(resp)
 	}
 }
+
+var validQuotaKey = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)
```
peterjan: since we have the regex, should we add `{1,32}` to check the length here?

persist/postgres/quotas.go
```diff
@@ -0,0 +1,99 @@
...
+// If the quota does not exist, it returns [accounts.ErrQuotaNotFound].
+// If the quota is in use by connect keys, it returns [accounts.ErrQuotaInUse].
+func (s *Store) DeleteQuota(key string) error {
+	return s.transaction(func(ctx context.Context, tx *txn) error {
+		// check if quota exists
+		var exists bool
+		if err := tx.QueryRow(ctx, `SELECT EXISTS(SELECT 1 FROM quotas WHERE name = $1)`, key).Scan(&exists); err != nil {
+			return fmt.Errorf("failed to check quota: %w", err)
+		} else if !exists {
+			return accounts.ErrQuotaNotFound
```
peterjan, line 42: is this necessary?

peterjan (changes requested):

peterjan:
left some minor remarks and questions, should be good to go though

persist/postgres/quotas.go
ChrisSchinnerl: It makes for a nice error.

peterjan (approved):

PR #800, opened 2026-02-11
Add contract fund target to quota

### LINKED ISSUE #787: Discussion: Add Quotas
Right now we have 1 table for app connect keys and 1 for accounts which are created from connect keys. Each one of them has a limit on how much data can be pinned. One limit is the global limit for a user on the app connect key table and the other one the local limit for a single account.

With SiaStorage in mind and the fact that we might want to change the quota for a lot of accounts and connect keys at once I've been thinking whether it makes sense to drop the `max_pinned_data` columns on these tables and instead add a new `quotas` table. The app connect key table would link each connect key to a quota using a foreign key.

A quota contains 
1. global max pinned data for the app connect key
2. per account max funding target
3. max usages

The reasoning is that it will become quite easy to have SiaStorage update a quota if necessary. e.g. if we ever decide a trial tier needs to go from 100GB to 1TB or so. Otherwise we need to update the limits for every account and app key individually when a tier changes.

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/indexd/issues/787


n8mgr (dismissed):

contracts/funding.go
```diff
@@ -35,47 +46,55 @@ func (cm *ContractManager) FundAccounts(ctx context.Context, host hosts.Host, co
...
+
+			contractIDs = contractIDs[drained:]
+			if len(contractIDs) == 0 {
+				log.Debug("not all accounts could be funded, no more contracts available", zap.String("quota", quota.Key))
+				break OUTER
+			}
 		}
 	}
 
 	serviceAccounts := cm.accounts.ServiceAccounts(host.PublicKey)
```
peterjan: Should we consider funding service accounts first? Integrity checks are important 😅

contracts/funding.go
```diff
@@ -35,47 +46,55 @@ func (cm *ContractManager) FundAccounts(ctx context.Context, host hosts.Host, co
...
+			accounts.UpdateFundedAccounts(accs, funded)
+			err = cm.accounts.UpdateHostAccounts(accs)
+			if err != nil {
+				return fmt.Errorf("failed to update accounts: %w", err)
+			}
+
+			contractIDs = contractIDs[drained:]
+			if len(contractIDs) == 0 {
+				log.Debug("not all accounts could be funded, no more contracts available", zap.String("quota", quota.Key))
+				break OUTER
```
peterjan, line 105: If we `break OUTER` here because we are out of contracts, we should handle it gracefully when we try and fund the service accounts, currently we would pass it to `accountFunder.FundAccounts` which would error out. Not the worst but we can probably improve the logs there.

persist/postgres/quotas.go
```diff
@@ -10,21 +10,22 @@ import (
...
+			INSERT INTO quotas (name, description, max_pinned_data, total_uses, fund_target_bytes)
+			VALUES ($1, $2, $3, $4, $5)
 			ON CONFLICT (name) DO UPDATE SET
 				description = EXCLUDED.description,
 				max_pinned_data = EXCLUDED.max_pinned_data,
-				total_uses = EXCLUDED.total_uses
-		`, key, req.Description, req.MaxPinnedData, req.TotalUses)
+				total_uses = EXCLUDED.total_uses,
+				fund_target_bytes = EXCLUDED.fund_target_bytes
+		`, key, req.Description, req.MaxPinnedData, req.TotalUses, req.FundTargetBytes)
```
peterjan, line 28: `nit` FundTargetBytes is a pointer here, we validate it higher up but should we explicitly deref it here to make it obvious or maybe even check for nil as a sanity check?

api/admin/admin.go
```diff
@@ -483,10 +483,12 @@ func (a *admin) handlePUTQuota(jc jape.Context) {
 	var req accounts.PutQuotaRequest
 	if jc.Decode(&req) != nil {
 		return
-	}
-	if req.TotalUses < 0 {
+	} else if req.TotalUses < 0 {
 		jc.Error(errors.New("totalUses must be non-negative"), http.StatusBadRequest)
 		return
+	} else if req.FundTargetBytes == nil {
+		jc.Error(errors.New("fundTargetBytes is required"), http.StatusBadRequest)
```
peterjan, line 490: why make it a pointer?

peterjan (changes requested):

contracts/funding.go
ChrisSchinnerl: I don't think it matters too much since funding is very fast but considering it's just 2 service accounts I moved it up.

api/admin/admin.go
ChrisSchinnerl: To be able to explicitly handle the case where a client doesn't provide the field rather than silently succeed and end up with a quota that never funds accounts.

persist/postgres/quotas.go
ChrisSchinnerl: I added a test for the validation but if this is called with `nil` it will just error out on the NOT NULL constraint which should be better than a panic.

peterjan (dismissed):

api/admin/admin.go
peterjan: shouldn't we get rid of the pointer then and not allow a target of zero?

ChrisSchinnerl: A zero target can be useful to create quotas for idle accounts. e.g. if a user stops paying you could move them over to an "suspended" quota where accounts don't get funded.

n8mgr (dismissed):

n8mgr:
@ChrisSchinnerl merge conflict

PR #803, opened 2026-02-11
Add `Usable` and `GoodForUpload` fields to individual host stats

### LINKED ISSUE #798: Add `Usable` and `GoodForUpload` fields to individual host stats

chris124567:
Close #798


persist/postgres/stats.go
```diff
@@ -180,9 +191,28 @@ func (s *Store) HostStats(offset, limit int) ([]hosts.HostStats, error) {
...
+						h.settings_valid_until >= h.last_successful_scan + INTERVAL '15 minutes' AND
+						h.settings_accepting_contracts AND
+						h.settings_contract_price <= g.one_sc AND
+						h.settings_collateral >= g.hosts_min_collateral AND
+						h.settings_collateral >= 2 * h.settings_storage_price AND
+						h.settings_storage_price <= g.hosts_max_storage_price AND
+						h.settings_ingress_price <= g.hosts_max_ingress_price AND
+						h.settings_egress_price <= g.hosts_max_egress_price AND
+						h.settings_free_sector_price <= g.one_sc / g.sectors_per_tb AND
+						EXISTS (SELECT 1 FROM contracts WHERE host_id = h.id AND state IN (0,1) AND renewed_to IS NULL AND good),
```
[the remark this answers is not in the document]
chris124567: fixed

persist/postgres/stats_test.go
```diff
@@ -499,6 +499,64 @@ func TestHostStats(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	} else if stats[0].Usable || stats[1].Usable {
+		t.Fatal("expected neither host to be usable")
+	} else if stats[0].GoodForUpload || stats[1].GoodForUpload {
+		t.Fatal("expected neither host to be good for upload")
+	}
+
+	// lower global settings so test host settings pass usability checks, reset
+	// scanned height so contracts are active, scan hk2, and boost its uptime
```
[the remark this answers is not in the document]
chris124567: fixed

persist/postgres/stats.go
```diff
@@ -180,9 +191,36 @@ func (s *Store) HostStats(offset, limit int) ([]hosts.HostStats, error) {
...
+						h.settings_egress_price <= g.hosts_max_egress_price AND
+						h.settings_free_sector_price <= g.one_sc / g.sectors_per_tb AND
+						EXISTS (
+							SELECT 1
+							FROM contracts
+							WHERE host_id = h.id
+								AND state IN (0,1)
+								AND renewed_to IS NULL
+								AND good
+								AND proof_height > g.scanned_height
```
peterjan, line 221: Hm so it seems we consider contracts that are in the proof window as not usable for everything related to stats, however the `UsableHosts` query as well as `HostsForX` don't take that into account. I remember we ran into this before at one point.

I did some searching and it was introduced in this PR https://github.com/SiaFoundation/indexd/pull/467 .
I think we should probably fix that throughout the code base and be consistent with our definition of an active contract.

I think, in as a follow-up, we should consider deduplicating this logic. We have it 3 times now, maybe even 4 times. At a minimum we should probably add a comment in every location referencing the other locations so copilot picks up these inconsistencies that sneak in.

peterjan: I know this would add some complexity and we want to avoid it, but an active_contracts view alongside maybe even a host_usability view would be pretty nice.

peterjan (changes requested):

persist/postgres/stats.go
chris124567: I changed the logic to be more consistent across the other store methods

I also made https://github.com/SiaFoundation/indexd/issues/815 as a follow up

peterjan (approved):

ChrisSchinnerl (approved):

PR #806, opened 2026-02-12
Fix OpenAPI spec

peterjan:
I found a tab with an open review on https://github.com/SiaFoundation/indexd/pull/793. It got merged so decided to quickly open a PR for it myself.

### DIFF
--- a/hosts/manager.go
+++ b/hosts/manager.go
@@ -589,7 +589,7 @@ func newStuckHostsAlert(hosts []StuckHost) alerts.Alert {
 		Message:  "Host(s) are stuck",
 		Data: map[string]any{
 			"hosts": hosts,
-			"hint":  "Contract operations (form/renew/refresh) have been failing for these hosts for more than 24 hours. Consider blocking these hosts through the blocklist feature.",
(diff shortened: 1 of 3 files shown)


persist/postgres/stats.go
```diff
@@ -142,13 +142,15 @@ func (s *Store) AggregatedHostStats() (stats admin.AggregatedHostStatsResponse,
...
 			FROM contracts c
 			INNER JOIN hosts h ON c.host_id = h.id
+			LEFT JOIN hosts_blocklist hb ON hb.public_key = h.public_key
 			CROSS JOIN globals
 			WHERE
 				c.good = TRUE AND
 				c.state IN (0,1) AND
 				c.renewed_to IS NULL AND
 				c.proof_height > globals.scanned_height AND
-				h.stuck_since IS NULL
```
peterjan, line 151: We probably want to take into account blocked hosts right? We do it in `HostStats`... I used a separate commit so we could easily revert it.

ChrisSchinnerl: I don't think we do. After all we use that endpoint to report how many hosts are blocked in the dashboard and what for.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

PR #809, opened 2026-02-13
Handle violations in UpdateAppConnectKey

peterjan:
This PR adds handling of unique constraint violation and foreign key constraint violation. I've updated `AddAppConnectKey` to handle it in a similar fashion so the implementations are consistent.

### DIFF
--- a/accounts/connect.go
+++ b/accounts/connect.go
@@ -11,8 +11,8 @@ import (
 )
 
 var (
-	//  ErrKeyAlreadyExists is returned when trying to add an app connect key
-	//  with an app key that already exists.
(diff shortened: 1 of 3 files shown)


ChrisSchinnerl (changes requested):
Copilot has a few good remarks.

ChrisSchinnerl (changes requested):
The comments were all resolved but it seems like no commit was pushed.

peterjan:
> The comments were all resolved but it seems like no commit was pushed.

Heh, I remember fixing them, but I lost the commit it seems. Thanks for noticing.

ChrisSchinnerl (approved):

PR #811, opened 2026-02-13
Fix ndf in TestApplicationAPI

ChrisSchinnerl:
Also removes an unused parameter from `App()`.

The issue is that sectors sometimes get marked as lost from hosts because we didn't actually upload them. Same as `TestSharedObjects` before.

### DIFF
--- a/api/app/app_test.go
+++ b/api/app/app_test.go
@@ -60,7 +60,7 @@ func newAccount(t *testing.T, cluster *testutils.Cluster) (types.PrivateKey, acc
 	indexer := cluster.Indexer
 
 	sk := types.GeneratePrivateKey()
-	client := indexer.App()
+	client := indexer.App
 
 	key, err := indexer.Admin.AddAppConnectKey(ctx, accounts.AddConnectKeyRequest{
 		Quota: "default",
@@ -162,7 +162,7 @@ func TestApplicationAPI(t *testing.T) {
 	sk, key := newAccount(t, cluster)
 	sk2, key2 := newAccount(t, cluster)
 
-	client := indexer.App()
+	client := indexer.App
 
 	// check that the key has been used
 	keys, err := adminClient.AppConnectKeys(ctx, 0, 2)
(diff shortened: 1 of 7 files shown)


chris124567 (dismissed):

testutils/indexer.go
```diff
@@ -55,7 +55,7 @@ type (
 		appApiAddr string
 
 		Admin *admin.Client
-		App   func(types.PrivateKey) *app.Client
+		App   func() *app.Client
```
peterjan: The function can be refactored out now that we no longer pass a private key.

api/app/app_test.go
```diff
@@ -124,44 +158,34 @@ func TestApplicationAPI(t *testing.T) {
...
 
-	// prepare account
+	// prepare accounts for the test
 	sk, key := newAccount(t, cluster)
-	client := indexer.App(sk)
+	sk2, key2 := newAccount(t, cluster)
+
+	client := indexer.App()
 
 	// check that the key has been used
```
peterjan, line 167: have they been used?

api/app/app_test.go
```diff
@@ -723,16 +689,16 @@ func TestSharedObjects(t *testing.T) {
 		t.Fatal("shared object mismatch")
 	}
 
-	// make sure client2 has no objects
-	if objs, err := client2.ListObjects(ctx, sk2, slabs.Cursor{}, 100); err != nil {
+	// make sure appClient has no objects
```
peterjan: would update the comments here

testutils/cluster.go
```diff
@@ -114,7 +114,7 @@ func (c *Cluster) App(t testing.TB) (*app.Client, types.PrivateKey) {
 	t.Helper()
 	sk := types.GeneratePrivateKey()
 	c.Indexer.Store().AddTestAccount(t, sk.PublicKey())
-	return c.Indexer.App(sk), sk
+	return c.Indexer.App(), sk
```
peterjan: The `App` method on the cluster no longer makes sense now that we don't pass the private key to the client. I would rename it to `AppClient maybe (or just have callers fetch it from the indexer) and add a new method `AddTestAccount` or whatever on the cluster that adds an account to the store and returns the private key. I checked and there's not too many call sites.

peterjan (changes requested):

peterjan (approved):

PR #818, opened 2026-02-19
Fix stats race

ChrisSchinnerl:
I ran into a few of these on Zeus:

```
{"level":"debug","ts":"2026-02-18T09:39:20Z","logger":"slabs.integrity","caller":"slabs/dataintegrity.go:77","msg":"performed integrity checks","hostKey":"ed25519:4489f1744948a981035f901217976bb407bbdc7cf8b913b7003708a49d055699","lost":0,"failed":0,"successful":1000}
```

The problem was probably caused by a host being blocked, triggering a huge update of unpinnable sectors. Looking at PostgreSQLs default transaction read isolation level it probably stems from selecting a lot of rows we want to update matching certain conditions (e.g. host_id IT NOT NULL) and then updating these rows without them being locked.

There are 2 options to fix this which this PR makes use of:
1. avoid select followed by an update by using `UPDATE ... RETURNING` 
2. when 1. doesn't work use `SELECT ... FOR UPDATE` before the actual `UPDATE` call

### DIFF
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -851,7 +851,9 @@ func (s *Store) MigrateSector(root types.Hash256, hostKey types.PublicKey) (migr
 		err := tx.QueryRow(ctx, `
 			SELECT id, host_id, contract_sectors_map_id
 			FROM sectors
-			WHERE sector_root = $1`, sqlHash256(root)).Scan(&sectorID, &oldHostID, &contractMapID)
+			WHERE sector_root = $1
+			FOR UPDATE
+		`, sqlHash256(root)).Scan(&sectorID, &oldHostID, &contractMapID)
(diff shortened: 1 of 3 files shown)


persist/postgres/sectors.go
```diff
@@ -688,6 +691,7 @@ func (s *Store) MarkSectorsUnpinnable(threshold time.Time) error {
 				WHERE host_id IS NOT NULL
 					AND contract_sectors_map_id IS NULL
 					AND uploaded_at <= $1
+				FOR UPDATE
```
peterjan, line 709: Should we consider batching these? I guess it's probably fine but this potentially locks a decent amount of rows and we might run into the same issue we're trying to solve.

peterjan (dismissed):

chris124567 (approved):

peterjan (approved):

PR #821, opened 2026-02-19
Use unified host client in Pinner

### LINKED ISSUE #717: Use unified host client in Pinner
This one will be messy... The existing client has logic around contract selection that does not belong in an RPC client. It can be moved back into the pinner where it belongs, but that will require touching a lot of different places and verifying logic consistency.

chris124567:
Close #717


client/v2/contracts.go
```diff
@@ -134,58 +134,45 @@ func (c *Client) LatestRevision(ctx context.Context, hostKey types.PublicKey, co
...
-
 		result, err = rhp.RPCFreeSectors(ctx, transport, signer, chain.TipState(), prices, contract, indices)
 		return err
 	})
 	return
 }
 
 // AppendSectors appends sectors to the specified contract on the host.
-func (c *Client) AppendSectors(ctx context.Context, signer rhp.ContractSigner, chain ChainManager, revision rhp.ContractRevision, sectors []types.Hash256) (res rhp.RPCAppendSectorsResult, err error) {
+func (c *Client) AppendSectors(ctx context.Context, signer rhp.ContractSigner, chain ChainManager, prices proto.HostPrices, revision rhp.ContractRevision, sectors []types.Hash256) (res rhp.RPCAppendSectorsResult, err error) {
```
n8mgr: Why add the prices here? It makes it the caller's responsibility to refresh them when they expire, but we already cache them and have a helper for that exact purpose here.

ChrisSchinnerl (approved):

PR #822, opened 2026-02-19
Fix CORS handlers

### LINKED ISSUE #817: Browser requests blocked by CORS: Missing OPTIONS routes for DELETE and POST endpoints
Browser-based applications cannot call DELETE endpoints and some POST endpoints due to missing CORS preflight support. Requests fail with:
```
Access to fetch at 'https://app.sia.storage/objects/157aebb4a8febc041c8961e5583ea919295801a83fe25920bd6dc2204e8d6c11?sv=1771426356&sc=nmS95u9mPXfFkjj8fkVfCw28mgyVgq9IQOUQYBykQNs%3D&ss=Sb5uSPY9T-5teeh1i3TA09JxYcr2BL9tTCh9DV693F5B9HHFVIdrUCSKIS9t2r0KeOA6Yf9HXYCatyGWZB3HAA%3D%3D' from origin 'http://127.0.0.1:8080' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.

Access to fetch at 'https://app.sia.storage/slabs/bafec067801eed8ab2b5547450ecc67e24fb3826dde4acbdb9762febee0621b5' from origin 'http://127.0.0.1:8080' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.

Access to fetch at 'https://app.sia.storage/slabs' from origin 'http://127.0.0.1:8080' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

I believe this affects all DELETE endpoints and any POST with custom headers.

Browsers send an OPTIONS preflight request before "non-simple" requests (DELETE, POST with custom headers). The `jape.Mux` router does not automatically handle OPTIONS requests - they must be explicitly defined.

The fix might be as simple as explicitly defining `OPTIONS` endpoints for affecting endpoints.
eg,
```
"DELETE /slabs/:slabid": wrapCORS(wrapSignedAuth(a.handleDELETESlab)),
 "OPTIONS /slabs/:slabid": wrapCORS(func(jc jape.Context) {}),
```

See https://stackoverflow.com/questions/29954037/why-is-an-options-request-sent-and-can-i-disable-it

n8mgr:
Adds explicit per-route CORs handling.

Fixes #817

### DIFF
--- a/api/app/cors.go
+++ b/api/app/cors.go
@@ -1,0 +1,133 @@
+package app
+
+import (
+	"net/http"
+	"strings"
+
+	"go.sia.tech/jape"
+)
+
+type route struct {
+	last bool
+	sub  map[string]route
+}
+
+const wildcardParamSegment = ":param"
+
+// mergeOptionsSub merges src into dst, unioning keys recursively.
+func mergeOptionsSub(dst, src map[string]route) {
+	for key, val := range src {
+		if existing, ok := dst[key]; ok {
+			if val.last {
+				existing.last = true
+				dst[key] = existing
+			}
+			mergeOptionsSub(existing.sub, val.sub)
+		} else {
+			dst[key] = val
+		}
+	}
+}
+
+// addOptionsPath inserts a route path into the tree, normalizing
+// parameterized segments to a common wildcard key.
+func addOptionsPath(m map[string]route, segments []string) {
+	if len(segments) == 0 {
+		return
+	}
+	seg := segments[0]
+	if strings.HasPrefix(seg, ":") {
+		seg = wildcardParamSegment
+	}
+	node, ok := m[seg]
+	if !ok {
+		node = route{sub: make(map[string]route)}
+	}
+	if len(segments) == 1 {
+		node.last = true
+	}
+	m[seg] = node
+	addOptionsPath(m[seg].sub, segments[1:])
+}
+
+// collectOptionsPaths resolves wildcard/static conflicts at each tree
+// level by folding statics into the wildcard, then collects all leaf
+// paths. httprouter does not allow a static and parameterized segment
+// at the same position, so the wildcard is kept since it covers both.
+func collectOptionsPaths(m map[string]route, prefix string) []string {
+	// fold statics into wildcard at this level
+	if wildcard, ok := m[wildcardParamSegment]; ok {
+		for key, r := range m {
+			if key == wildcardParamSegment {
+				continue
+			}
+			if r.last {
+				wildcard.last = true
+			}
+			mergeOptionsSub(wildcard.sub, r.sub)
+			delete(m, key)
+		}
+		m[wildcardParamSegment] = wildcard
+	}
+
+	var paths []string
+	for seg, r := range m {
+		path := prefix + "/" + seg
+		if r.last {
+			paths = append(paths, path)
+		}
+		paths = append(paths, collectOptionsPaths(r.sub, path)...)
+	}
+	return paths
+}
+
+// corsMux is a helper that applies CORS middleware to the handlers in `enabledRoutes` and
+// prevents OPTIONS handling in `disabledRoutes`.
+//
+// If a route is covered by both maps, it panics.
+func corsMux(enabledRoutes map[string]jape.Handler, disabledRoutes map[string]jape.Handler) http.Handler {
+	// adds the CORS headers globally since most endpoints are expected to be called from the browser.
+	corsMiddleware := func(h jape.Handler) jape.Handler {
+		return func(jc jape.Context) {
+			jc.ResponseWriter.Header().Set("Access-Control-Allow-Origin", "*")
+			jc.ResponseWriter.Header().Set("Access-Control-Allow-Methods", "GET, POST, DELETE")
+			jc.ResponseWriter.Header().Set("Access-Control-Allow-Headers", "*")
+			h(jc)
+		}
+	}
+
+	optionsHandler := corsMiddleware(func(jc jape.Context) {
+		jc.ResponseWriter.WriteHeader(http.StatusNoContent)
+	})
+
+	disableOptionsHandler := func(jc jape.Context) {
+		if jc.Request.Method == http.MethodOptions {
+			jc.ResponseWriter.WriteHeader(http.StatusMethodNotAllowed)
+			return
+		}
+	}
+
+	optionsTree := make(map[string]route)
+	disabledOptionsTree := make(map[string]route)
+	routes := make(map[string]jape.Handler, len(enabledRoutes)+len(disabledRoutes))
+	for key, h := range enabledRoutes {
+		routes[key] = corsMiddleware(h)
+		addOptionsPath(optionsTree, strings.Split(strings.Fields(key)[1], "/")[1:])
+	}
+	for _, path := range collectOptionsPaths(optionsTree, "") {
+		routes["OPTIONS "+path] = optionsHandler
+	}
+	for key, h := range disabledRoutes {
+		routes[key] = h
+		addOptionsPath(disabledOptionsTree, strings.Split(strings.Fields(key)[1], "/")[1:])
+	}
+	// httprouter responds with a 200 by default. Add an explicit handler to reject it
+	for _, path := range collectOptionsPaths(disabledOptionsTree, "") {
+		optionsRoute := "OPTIONS " + path
+		if _, ok := routes[optionsRoute]; ok {
+			panic("disabled OPTIONS route " + path + " conflicts with a CORS-enabled OPTIONS route")
+		}
+		routes[optionsRoute] = disableOptionsHandler
+	}
+	return jape.Mux(routes)
+}
--- a/api/app/cors.go
+++ b/api/app/cors.go
@@ -14,7 +14,7 @@ type route struct {
 
 const wildcardParamSegment = ":param"
 
-// mergeOptionsSub merges src into dst, unioning keys recursively.
+// mergeOptionsSub merges src into dst
 func mergeOptionsSub(dst, src map[string]route) {
 	for key, val := range src {
 		if existing, ok := dst[key]; ok {
@@ -29,8 +29,8 @@ func mergeOptionsSub(dst, src map[string]route) {
 	}
 }
 
-// addOptionsPath inserts a route path into the tree, normalizing
-// parameterized segments to a common wildcard key.
+// addOptionsPath inserts an OPTIONS route path into the tree
+// normalizing parameter segments to a single wildcard segment.
 func addOptionsPath(m map[string]route, segments []string) {
 	if len(segments) == 0 {
 		return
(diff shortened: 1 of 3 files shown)


chris124567 (dismissed):
This seems a little overkill to me, especially given that we are still using wrappers for auth for instance.  I'm not entirely opposed so feel free to dismiss if you disagree but it feels inconsistent

Alrighttt (dismissed):
Works as expected. 

Added a "CORS diagnostic" section to my demo temporarily to sanity check this. 

https://alrighttt.github.io/indexd_wasm_demo/

n8mgr:
@chris124567 I agree it's a bit much, but is there a solution that doesn't require manually adding an additional `OPTIONS` handler for every route except the two where it's disabled? 

- We can't use a global CORS middleware without doing some sort of fragile string matching for disabled routes since it needs to add headers and short-circuit on OPTIONS (e.g. add headers for `/auth/connect/:requestID` but not `/auth/connect` or `/auth/connect/:requestID/status`.
- A global middleware would respond with 200 for any route instead of just routes actually registered in the mux.
- We can't use a wrapper on the existing routes since a `GET/POST/DELETE` route won't match an `OPTIONS` request.

Alrighttt:
```
diff --git a/api/app/app.go b/api/app/app.go
index 240b6f12..9ca465c9 100644
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -672,15 +672,11 @@ func NewAPI(advertiseURL string, store Store, am Accounts, contracts Contracts,
                        jc.ResponseWriter.Header().Set("Access-Control-Allow-Origin", "*")
                        jc.ResponseWriter.Header().Set("Access-Control-Allow-Methods", "GET, POST, DELETE")
                        jc.ResponseWriter.Header().Set("Access-Control-Allow-Headers", "*")
-                       if jc.Request.Method == http.MethodOptions {
-                               jc.ResponseWriter.WriteHeader(http.StatusNoContent)
-                               return
-                       }
                        h(jc)
                }
        }
 
-       return jape.Mux(map[string]jape.Handler{
+       mux := jape.Mux(map[string]jape.Handler{
                "GET /account": wrapCORS(wrapSignedAuth(a.handleGETAccount)),
 
                // auth is a multi-step process designed to protect privacy while allowing arbitrary apps to connect:
@@ -713,5 +709,19 @@ func NewAPI(advertiseURL string, store Store, am Accounts, contracts Contracts,
                "POST /slabs/prune":     wrapCORS(wrapSignedAuth(a.handlePOSTSlabsPrune)),
                "GET /slabs/:slabid":    wrapCORS(wrapSignedAuth(a.handleGETSlab)),
                "DELETE /slabs/:slabid": wrapCORS(wrapSignedAuth(a.handleDELETESlab)),
+       })
+
+       // Global OPTIONS handler: intercepts all preflight requests before
+       // httprouter's route matching, avoiding conflicts between parameterized
+       // and static routes (e.g. /slabs/:slabid vs /slabs/prune).
+       return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+               if r.Method == http.MethodOptions {
+                       w.Header().Set("Access-Control-Allow-Origin", "*")
+                       w.Header().Set("Access-Control-Allow-Methods", "GET, POST, DELETE")
+                       w.Header().Set("Access-Control-Allow-Headers", "*")
+                       w.WriteHeader(http.StatusNoContent)
+                       return
+               }
+               mux.ServeHTTP(w, r)
        }), nil
 }
 ```
 
This would also work. The only tradeoff is that `OPTIONS /nonexistent/path` returns 204 instead of 404.

**Comment by n8mgr** [comment]
> ```
> diff --git a/api/app/app.go b/api/app/app.go
> index 240b6f12..9ca465c9 100644
> --- a/api/app/app.go
> +++ b/api/app/app.go
> @@ -672,15 +672,11 @@ func NewAPI(advertiseURL string, store Store, am Accounts, contracts Contracts,
>                         jc.ResponseWriter.Header().Set("Access-Control-Allow-Origin", "*")
>                         jc.ResponseWriter.Header().Set("Access-Control-Allow-Methods", "GET, POST, DELETE")
>                         jc.ResponseWriter.Header().Set("Access-Control-Allow-Headers", "*")
> -                       if jc.Request.Method == http.MethodOptions {
> -                               jc.ResponseWriter.WriteHeader(http.StatusNoContent)
> -                               return
> -                       }
>                         h(jc)
>                 }
>         }
>  
> -       return jape.Mux(map[string]jape.Handler{
> +       mux := jape.Mux(map[string]jape.Handler{
>                 "GET /account": wrapCORS(wrapSignedAuth(a.handleGETAccount)),
>  
>                 // auth is a multi-step process designed to protect privacy while allowing arbitrary apps to connect:
> @@ -713,5 +709,19 @@ func NewAPI(advertiseURL string, store Store, am Accounts, contracts Contracts,
>                 "POST /slabs/prune":     wrapCORS(wrapSignedAuth(a.handlePOSTSlabsPrune)),
>                 "GET /slabs/:slabid":    wrapCORS(wrapSignedAuth(a.handleGETSlab)),
>                 "DELETE /slabs/:slabid": wrapCORS(wrapSignedAuth(a.handleDELETESlab)),
> +       })
> +
> +       // Global OPTIONS handler: intercepts all preflight requests before
> +       // httprouter's route matching, avoiding conflicts between parameterized
> +       // and static routes (e.g. /slabs/:slabid vs /slabs/prune).
> +       return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
> +               if r.Method == http.MethodOptions {
> +                       w.Header().Set("Access-Control-Allow-Origin", "*")
> +                       w.Header().Set("Access-Control-Allow-Methods", "GET, POST, DELETE")
> +                       w.Header().Set("Access-Control-Allow-Headers", "*")
> +                       w.WriteHeader(http.StatusNoContent)
> +                       return
> +               }
> +               mux.ServeHTTP(w, r)
>         }), nil
>  }
> ```
> 
> This would also work. The only tradeoff is that `OPTIONS /nonexistent/path` returns 204 instead of 404.

The browser will get a preflight response *with* the allow headers for disabled routes, send the request, then reject the response instead of failing up front at the preflight. It will work, but it's not ideal and relies on browser behavior being consistent.

**Review by Alrighttt — DISMISSED** [review]

**Review by chris124567 — DISMISSED** [review]

File: api/app/cors.go
Comment (ChrisSchinnerl) [inline api/app/cors.go:1]: I've been wondering if we can reuse the deduplication check of the http libs mux and came up with the following approach that is based on just replacing the method string and adding explicit OPTIONS routes for disabled routes as well. This passes your tests as well with the exception that it now consistently returns 204 and not 200 for disabled routes and 204 for enabled ones.

If you like this feel free to steal it.
If this doesn't work due to some edge case that I'm not aware of it would be great to extend the tests to hit that edge case. 

```
package app

import (
	"net/http"
	"strings"

	"go.sia.tech/jape"
)

// corsMux is a helper that applies CORS middleware to the handlers in `enabledRoutes` and
// prevents OPTIONS handling in `disabledRoutes`.
//
// Cleaner than manually adding OPTIONS handlers to supported routes or doing string matching
// in a global middleware.
//
// If a route overlaps in both maps, it panics.
func corsMux(enabledRoutes map[string]jape.Handler, disabledRoutes map[string]jape.Handler) http.Handler {
	corsMiddleware := func(h jape.Handler) jape.Handler {
		return func(jc jape.Context) {
			jc.ResponseWriter.Header().Set("Access-Control-Allow-Origin", "*")
			jc.ResponseWriter.Header().Set("Access-Control-Allow-Methods", "GET, POST, DELETE")
			jc.ResponseWriter.Header().Set("Access-Control-Allow-Headers", "*")
			h(jc)
		}
	}

	optionsNoCORSHandler := func(jc jape.Context) {
		jc.ResponseWriter.WriteHeader(http.StatusNoContent)
	}
	optionsHandler := corsMiddleware(optionsNoCORSHandler)

	// helper to replace the HTTP method in the route key with OPTIONS for the
	// OPTIONS handler
	replaceHTTPMethod := func(key string) string {
		switch {
		case strings.HasPrefix(key, http.MethodGet):
			return strings.Replace(key, http.MethodGet, http.MethodOptions, 1)
		case strings.HasPrefix(key, http.MethodHead):
			return strings.Replace(key, http.MethodHead, http.MethodOptions, 1)
		case strings.HasPrefix(key, http.MethodPost):
			return strings.Replace(key, http.MethodPost, http.MethodOptions, 1)
		case strings.HasPrefix(key, http.MethodPut):
			return strings.Replace(key, http.MethodPut, http.MethodOptions, 1)
		case strings.HasPrefix(key, http.MethodPatch):
			return strings.Replace(key, http.MethodPatch, http.MethodOptions, 1)
		case strings.HasPrefix(key, http.MethodDelete):
			return strings.Replace(key, http.MethodDelete, http.MethodOptions, 1)
		case strings.HasPrefix(key, http.MethodConnect):
			return strings.Replace(key, http.MethodConnect, http.MethodOptions, 1)
		case strings.HasPrefix(key, http.MethodTrace):
			return strings.Replace(key, http.MethodTrace, http.MethodOptions, 1)
		case strings.HasPrefix(key, http.MethodOptions):
			return key
		}
		return ""
	}

	routes := make(map[string]jape.Handler, len(enabledRoutes)+len(disabledRoutes))

	// add CORS headers for enabled routes
	for key, h := range enabledRoutes {
		routes[key] = corsMiddleware(h)
	}

	// add missing OPTIONS handlers for enabled routes
	for key := range enabledRoutes {
		optionsKey := replaceHTTPMethod(key)
		if _, ok := routes[optionsKey]; !ok {
			routes[optionsKey] = optionsHandler
		}
	}

	// add handlers without CORS for disabled routes
	for key, h := range disabledRoutes {
		// add handler without CORS for this route
		if _, ok := enabledRoutes[key]; ok {
			panic("route " + key + " cannot be both enabled and disabled")
		}
		routes[key] = h
	}

	// add missing OPTIONS handlers for disabled routes
	for key := range disabledRoutes {
		optionsKey := replaceHTTPMethod(key)
		if _, ok := routes[optionsKey]; !ok {
			routes[optionsKey] = optionsNoCORSHandler
		}
	}
	return jape.Mux(routes)
}
```

**Review by ChrisSchinnerl — CHANGES_REQUESTED** [review]
Feel free to resolve if the suggestion doesn't work. Just requesting changes to make sure you see this.

File: api/app/cors.go [hunk 1]
```diff
@@ -0,0 +1,173 @@
...
+		jc.ResponseWriter.WriteHeader(http.StatusNoContent)
+	})
+
+	optionsTree := make(map[string]route)
+	routes := make(map[string]jape.Handler, len(enabledRoutes)+len(disabledRoutes))
+	for key, h := range enabledRoutes {
+		// add CORS headers for this route
+		routes[key] = corsMiddleware(h)
+		// OPTIONs are collected separately since a parameterized route can overlap with a
+		// static route (e.g. POST /slabs/prune and GET /slabs/:slabid)
```
peterjan, line 156: `nit` could add a test case for these production routes in particular

api/app/cors.go
n8mgr: It does not work because OPTIONS /slabs/prune and OPTIONS /slabs/:slab overlap and cause httprouter to panic.

ChrisSchinnerl (approved):

PR #826, opened 2026-02-23
Add app stats endpoint

### LINKED ISSUE #820: Add app stats endpoint
Every app that a user links to the indexer has a unique id. We should add stats to get 

- number of active and total accounts that are linked to a given app
- amount of data pinned by a given app overall

chris124567:
Close #820

### DIFF
--- a/accounts/accounts_test.go
+++ b/accounts/accounts_test.go
@@ -51,13 +51,13 @@ func TestAccountFunding(t *testing.T) {
 	}
 
 	// create connect keys for each quota
-	if _, err := store.AddAppConnectKey(accounts.UpdateAppConnectKey{
+	if _, err := store.AddAppConnectKey(accounts.AppConnectKeyRequest{
 		Key:   connectKey1,
 		Quota: quota1Name,
 	}); err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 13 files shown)


ChrisSchinnerl (dismissed):

ChrisSchinnerl:
Reminder that this will also require an update to the deploy repo

chris124567:
@ChrisSchinnerl

> Reminder that this will also require an update to the deploy repo

Should I change it to make it return all of the apps like how `/api/stats/hosts/detailed` returns all the hosts?  Otherwise we'd need to add a per app endpoint to prometheus.yml.template.  Only asking because you approved this PR so I'm not 100% sure I should change it.

ChrisSchinnerl:
> @ChrisSchinnerl
> 
> > Reminder that this will also require an update to the deploy repo
> 
> Should I change it to make it return all of the apps like how `/api/stats/hosts/detailed` returns all the hosts? Otherwise we'd need to add a per app endpoint to prometheus.yml.template. Only asking because you approved this PR so I'm not 100% sure I should change it.

Oh damn you are right. I didn't consider that this won't work for prometheus. Yeah we should change it so we can add it to our dashboard. I'd add pagination similar to how we do for hosts.

ChrisSchinnerl (approved):

PR #831, opened 2026-02-25
Use GB instead of GiB for default quota

ChrisSchinnerl:
For consistency with the remaining values.


chris124567 (approved):

chris124567 (changes requested):
nvm

chris124567 (approved):

PR #841, opened 2026-02-27
Fix error checks, add back pressure to migrations and faster uploads

ChrisSchinnerl:
This PR does the following:

1. It changes migrations to not upload a batch of slabs, wait and then upload another batch but instead keep working on a full batch in parallel by making use of a worker pool
2. Fixes checks for `ErrOutOfFunds` and `ErrSectorNotFound` in the integrity check code which were previously not working leading to lost sectors failing the check instead of being marked as lost immediately
3. Update the v2 client to not count RPC errors if they are `ErrAbortedRPC` which can be passed when cancelling the context
4. Make sure stuck hosts are not used for migrations
5. Enforce per-host timeouts when pinning and pruning
6. Increase refresh and renewal timeouts to avoid aborting these RPCs on slow hosts

Imo 3. is probably the more controversial change in this PR so here is the reasoning behind it. When I started printing the errors for which we punish hosts in `ReadSector`, I noticed that 99.99% of them are `stream was gracefully closed`. This is quite unique to downloads because we cancel the context when a slab download is done. So this happens quite frequently. My thought was that it's quite silly to punish hosts so harshly for this since some hosts might have been scheduled as overdrive workers 1s before the download finished and got punished for no good reason at all.

The solution I implemented was 2 different contexts for the initial downloads and the overdrive ones. So that overdrive workers don't get punished when the RPC is interrupted and only the initial downloads do. At the time of writing this, this PR has gotten the migration speed up to over 1Gbps with the speed still climbing in the dashboard.

Alternatively, something that might be even more accurate from a measurement standpoint, we could just not cancel downloads and let them gracefully finish in the background. Then we don't need this not-so-nice mechanism because either a download finishes or it errors out. However, it means we end up with dangling goroutines in the background.


client/v2/candidates.go
```diff
@@ -15,6 +16,11 @@ import (
 
 const emaAlpha = 0.2
 
+// ErrAbortedRPC is a special error that can be used as the cancel for an RPC
+// context to indicate that the RPC was interrupted and that the corresponding
+// host should not have a failure recorded for that RPC.
+var ErrAbortedRPC = errors.New("aborted RPC")
```
n8mgr, line 22: Why not use context cancelled err?

ChrisSchinnerl: That's the default. Right now every RPC is cancelled with that. `ErrAbortedRPC` would allow for distinguishing from the default.

persist/postgres/hosts.go
```diff
@@ -129,7 +129,8 @@ WITH `+sqlGlobalsCTE+`, hosts AS (
 		settings_collateral, settings_storage_price, settings_ingress_price,
 		settings_egress_price, settings_free_sector_price, settings_tip_height, settings_valid_until, settings_signature,
 		last_successful_scan IS NOT NULL as has_settings,
-		(get_byte(settings_protocol_version, 0) << 16) + (get_byte(settings_protocol_version, 1) << 8) + (get_byte(settings_protocol_version, 2)) as settings_version
+		(get_byte(settings_protocol_version, 0) << 16) + (get_byte(settings_protocol_version, 1) << 8) + (get_byte(settings_protocol_version, 2)) as settings_version,
+		stuck_since IS NOT NULL as stuck
```
peterjan: We could extend the benchmark to cover `stuck` field.

slabs/manager.go
```diff
@@ -335,23 +335,46 @@ func (m *SlabManager) performSlabMigrations(ctx context.Context) error {
...
-	for !exhausted {
+	allHosts, goodContracts, err := m.migrationCandidates()
+	if err != nil {
+		return err
+	}
+
+	// start a worker pool that pulls slabs from a channel
+	slabCh := make(chan SlabID, m.migrationBatchSize)
+	var wg sync.WaitGroup
+	for range m.migrationBatchSize {
```
peterjan: I would rename `migrationBatchSize` to something like `migrationPoolThreads`

slabs/manager.go
```diff
@@ -335,23 +335,46 @@ func (m *SlabManager) performSlabMigrations(ctx context.Context) error {
 	log := m.log.Named("migrations")
 	log.Debug("starting slab migrations")
 
-	var exhausted bool
-	for !exhausted {
+	allHosts, goodContracts, err := m.migrationCandidates()
```
peterjan: we need to refresh these more often imo

client/v2/client.go
```diff
@@ -296,8 +303,12 @@ func (c *Client) ReadSector(ctx context.Context, accountKey types.PrivateKey, ho
 		result, err = rhp.RPCReadSector(ctx, transport, prices, token, w, root, offset, length)
 		return err
 	})
-	if err != nil {
-		c.hosts.AddFailedRPC(hostKey)
+	if err != nil && strings.Contains(err.Error(), proto.ErrSectorNotFound.Error()) {
```
peterjan, line 306: I would extract these into a helper function called `isErrSectorNotFound`. We are currently repeating these checks, see  slabs/downloads.go. If we do add a helper we should make it a pattern and apply it consistently. I don't really care about the duplicated checks, but we might eventually run into places where we mistakenly use `errors.Is` and don't think to do a string comparison and miss an edge.

peterjan: .. and then I encountered `isErrLostSector`

slabs/migrations.go
```diff
@@ -14,40 +14,37 @@ import (
...
 
-	// fetch all available hosts with contracts
 	const batchSize = 500
 	var allHosts []hosts.Host
 	for offset := 0; ; offset += batchSize {
-		batch, err := m.store.Hosts(offset, batchSize, hosts.WithBlocked(false), hosts.WithActiveContracts(true))
+		batch, err := m.store.Hosts(offset, batchSize,
+			hosts.WithBlocked(false),
+			hosts.WithActiveContracts(true),
+			hosts.WithUsable(true),
```
peterjan: Adding `usable` + `unstuck` is quite a big change here. Look at `sectorsToMigrate`, which receives the hosts returned by migration candidates. We call these `allHosts` but that name no longer make sense. We also perform operations on the list of hosts that no longer make sense, e.g. we turn it into a `hostsMap` that only has "good hosts", but with these changes all of them would be considered good. 


I think it's fine though because we don't seem to be using the "bad hosts" from the all hosts slice. However I do think we should update the naming, also consider dropping the IsGood check, or even extend the IsGood check with a `RemainingStorage` check.

slabs/migrations.go
```diff
@@ -14,40 +14,37 @@ import (
...
 
-	var wg sync.WaitGroup
-	for _, slabID := range slabIDs {
-		wg.Go(func() {
-			m.migrateSlab(ctx, slabID, allHosts, goodContracts, log.With(zap.Stringer("slab", slabID)))
-		})
-	}
-	wg.Wait()
-	return nil
+	allHosts = slices.DeleteFunc(allHosts, func(h hosts.Host) bool {
```
peterjan: I think this should be a filter. We can check this on the database level. If we make it a filter we can also expose GFU hosts as:

- usable
- not stuck
- not blocked
- have remaining storage

Which is (close) to the definition from the docs ` Number of hosts that are good for uploading to (hosts that are usable, not stuck and have remaining storage)`. We should probably fix that by the way. I think a blocked host can come up as GFU now.

peterjan (changes requested):

slabs/migrations.go
ChrisSchinnerl: You are right. This is actually not great because we don't want to migrate away from stuck hosts. I reverted it and added it to `sectorsToMigrate` to only affect the candidates for migration and not whether a sector is good or not.

slabs/manager.go
```diff
@@ -335,23 +335,53 @@ func (m *SlabManager) performSlabMigrations(ctx context.Context) error {
...
-		} else if len(batch) < m.migrationBatchSize {
-			exhausted = true
 		}
 
-		err = m.migrateSlabs(ctx, batch, log)
-		if errors.Is(err, context.Canceled) {
+		// update the candidates for every batch
+		allHosts, goodContracts, err := m.migrationCandidates()
+		if err != nil {
+			return err
```
chris124567, line 368: Do we need this here too?
```
close(slabCh)
wg.Wait()
```

peterjan (approved):

n8mgr (approved):

PR #843, opened 2026-03-03
Remove default account limit

ChrisSchinnerl:
By default we now no longer set the account limit to the connect key's limit but instead to `MaxInt64` to enforce no limit.


n8mgr (dismissed):

api/admin/admin.go
```diff
@@ -206,9 +207,10 @@ func NewAPI(chain ChainManager, accounts Accounts, contracts ContractManager, ho
 		"GET /consensus/network": a.handleGETConsensusNetwork,
 
 		// accounts endpoints
-		"GET    /accounts":            a.handleGETAccounts,
-		"GET    /account/:accountkey": a.handleGETAccount,
-		"DELETE /account/:accountkey": a.handleDELETEAccount,
+		"GET    /accounts":                  a.handleGETAccounts,
+		"GET    /account/:accountkey":       a.handleGETAccount,
+		"DELETE /account/:accountkey":       a.handleDELETEAccount,
+		"PUT    /account/:accountkey/limit": a.handlePUTAccountMaxPinnedData,
```
peterjan: I think it would be cleaner to expose `PATCH /account/:accountkey` and for the time being we only allow updating `maxPinnedData`. That way we get rid of the somewhat confusing "limit", which I don't think is a bad name for the route, but it is a little confusing. More importantly though, it might come in handy if we later want to allow updating app metadata (which is highly likely) or maybe even updating the connect key should we ever need some crazy migration or feature that would require such a thing.

api/admin/admin.go
```diff
@@ -605,6 +607,24 @@ func (a *admin) handleDELETEAccount(jc jape.Context) {
...
+func (a *admin) handlePUTAccountMaxPinnedData(jc jape.Context) {
+	var ak types.PublicKey
+	if jc.DecodeParam("accountkey", &ak) != nil {
+		return
+	}
+	var req accounts.UpdateMaxPinnedDataRequest
+	if jc.Decode(&req) != nil {
+		return
+	}
+	err := a.accounts.UpdateMaxPinnedData(jc.Request.Context(), ak, req.MaxPinnedData)
```
peterjan: There's no validation on `req.MaxPinnedData` here, do we ever want to set this to 0?

peterjan (changes requested):

api/admin/admin.go
ChrisSchinnerl: Yeah, you can do that if you want to. Similar to the 0 quotas.

accounts/account.go
```diff
@@ -83,6 +83,12 @@ type (
 		NextFund               time.Time
 	}
 
+	// UpdateMaxPinnedDataRequest is the request body for the
+	// [PUT] /account/:accountkey/limit endpoint.
+	UpdateMaxPinnedDataRequest struct {
```
peterjan: `nit` UpdateAccountRequest

accounts/manager.go
```diff
@@ -116,6 +117,11 @@ func (m *AccountManager) Accounts(ctx context.Context, offset, limit int, opts .
 	return m.store.Accounts(offset, limit, opts...)
 }
 
+// UpdateMaxPinnedData updates the max pinned data limit for the given account.
+func (m *AccountManager) UpdateMaxPinnedData(ctx context.Context, ak types.PublicKey, maxPinnedData uint64) error {
```
peterjan: `nit` `UpdateAccount(ctx, ak, maxPinnedData)` and same with store?

peterjan (dismissed):
I left two nits about renaming, current state is like 90% of the refactor you did to change it to a PATCH route imo. Approved it in case you don't want to rename it.

peterjan (approved):

PR #846, opened 2026-03-04
Split maintenance up into multiple loops

### LINKED ISSUE #844: More responsive account funding
Account funding can be blocked by other maintenance operations right now. We want to break these up to make sure we can immediately fund new user accounts to provide a good UX.

### LINKED ISSUE #845: Fix NDF in `TestApplicationAPI`
Sometimes it fails with `not enough funds`

ChrisSchinnerl:
This splits up the maintenance in
1. account funding loop
2. contract maintenance loop 
3. remaining maintenance loop (pruning, pinning etc)

They all use the same frequency right now which currently is 2 minutes.

Closes https://github.com/SiaFoundation/indexd/issues/844
Closes https://github.com/SiaFoundation/indexd/issues/845

### DIFF
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -54,8 +54,6 @@ type (
 	// to ensure the account is funded as soon as possible.
 	ContractManager interface {
 		TriggerAccountFunding(force bool) error
-		TriggerContractPruning() error
-		TriggerMaintenance()
 
 		MaintenanceSettings(ctx context.Context) (contracts.MaintenanceSettings, error)
 		UpdateMaintenanceSettings(ctx context.Context, ms contracts.MaintenanceSettings) error
@@ -285,7 +283,6 @@ func NewAPI(chain ChainManager, accounts Accounts, contracts ContractManager, ho
 	// debug endpoints
 	if a.debug {
 		routes["GET /debug/pprof/:handler"] = a.handleGETPProf
-		routes["POST /debug/trigger/:action"] = a.handlePOSTTrigger
 	}
 
 	return jape.Mux(routes)
@@ -325,32 +322,6 @@ func (a *admin) handleGETPProf(jc jape.Context) {
 	default:
 		pprof.Index(jc.ResponseWriter, jc.Request)
 	}
-}
-
-func (a *admin) handlePOSTTrigger(jc jape.Context) {
-	var action string
-	if jc.DecodeParam("action", &action) != nil {
-		return
-	}
-
-	switch action {
-	case "funding":
-		err := a.contracts.TriggerAccountFunding(true)
-		if jc.Check("failed to trigger account funding", err) != nil {
-			return
-		}
-	case "maintenance":
-		a.contracts.TriggerMaintenance()
-	case "pruning":
-		a.contracts.TriggerContractPruning()
-	case "scanning":
-		a.hosts.TriggerHostScanning()
-	default:
-		jc.Error(fmt.Errorf("unknown action: %q, available actions are 'funding', 'maintenance', 'pruning' or 'scanning'", action), http.StatusBadRequest)
-		return
-	}
-
-	jc.Encode(nil)
 }
 
 func (a *admin) handleGETAppConnectKeys(jc jape.Context) {
--- a/contracts/maintenance.go
+++ b/contracts/maintenance.go
@@ -137,6 +137,10 @@ func (cm *ContractManager) maintenanceLoop(ctx context.Context) {
 			case <-t.C:
 				log.Debug("starting scheduled contract maintenance")
 			}
+			walletLog := log.Named("wallet")
+			if err := cm.performWalletMaintenance(ctx, walletLog); err != nil {
+				log.Debug("maintenance failed", zap.Error(err)) // wallet maintenance is best-effort
+			}
 			maintenanceLog := log.Named("contracts")
 			logError(cm.performContractMaintenance(ctx, maintenanceLog), maintenanceLog)
 		}
@@ -165,13 +169,6 @@ func (cm *ContractManager) maintenanceLoop(ctx context.Context) {
 				log.Debug("starting scheduled maintenance")
 			}
 
-			// this is done first so that fragmenting the wallet can be prioritized before
-			// being used for other maintenance tasks
-			walletLog := log.Named("wallet")
-			if err := cm.performWalletMaintenance(ctx, walletLog); err != nil {
-				log.Debug("maintenance failed", zap.Error(err)) // wallet maintenance is best-effort
-			}
-
 			pruningLog := log.Named("pruning")
 			logError(cm.performContractPruning(ctx, false, pruningLog), pruningLog)
 			pinningLog := log.Named("pinning")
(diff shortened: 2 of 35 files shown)


contracts/maintenance.go
```diff
@@ -98,58 +98,92 @@ func (cm *ContractManager) performContractMaintenance(ctx context.Context, log *
...
+			select {
+			case <-ctx.Done():
+				return
+			case <-cm.triggerPruningChan:
+				log.Debug("triggering contract pruning")
+				if err := cm.performContractPruning(ctx, true, log); err != nil {
+					log.Error("contract pruning failed", zap.Error(err))
+				}
+				continue
+			case <-cm.triggerMaintenanceChan:
```
peterjan: This was meant to trigger contract maintenance. Although we should probably consider removing this trigger? Maybe we should even consider removing `/debug/trigger/:action` all together now that loops run independently from each other.

contracts/maintenance.go
```diff
@@ -98,58 +98,92 @@ func (cm *ContractManager) performContractMaintenance(ctx context.Context, log *
...
-		// this is done first so that fragmenting the wallet can be prioritized before
-		// being used for other maintenance tasks
-		walletLog := log.Named("wallet")
-		if err := cm.performWalletMaintenance(ctx, walletLog); err != nil {
-			log.Debug("maintenance failed", zap.Error(err)) // wallet maintenance is best-effort
-		}
+			// this is done first so that fragmenting the wallet can be prioritized before
+			// being used for other maintenance tasks
+			walletLog := log.Named("wallet")
+			if err := cm.performWalletMaintenance(ctx, walletLog); err != nil {
```
peterjan: I would've moved wallet maintenance to the contract maintenance loop.

contracts/maintenance.go
```diff
@@ -98,58 +98,92 @@ func (cm *ContractManager) performContractMaintenance(ctx context.Context, log *
...
-	}
+			pruningLog := log.Named("pruning")
+			logError(cm.performContractPruning(ctx, false, pruningLog), pruningLog)
+			pinningLog := log.Named("pinning")
+			logError(cm.performSectorPinning(ctx, pinningLog), pinningLog)
+
+			unpinnableLog := log.Named("unpinnable")
+			threshold := time.Now().Add(-unpinnableSectorThreshold)
+			logError(cm.store.MarkSectorsUnpinnable(threshold), unpinnableLog)
+			t.Reset(cm.maintenanceFrequency)
```
peterjan: I think this reset can be removed now that we went from timer to ticker. In any case its behaviour is inconsistent with the other loops.

contracts/manager.go
```diff
@@ -481,6 +482,7 @@ func newContractManager(renterKey types.PublicKey, accounts AccountManager, acco
 		store:  store,
 
 		client: client,
+		cl:     NewContractLocker(),
```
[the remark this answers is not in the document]
peterjan: The fact that you resolved it makes me second guess my comment here but... this is a critical issue, no? The funder and the manager must be on the same locker instance

contracts/revision.go
```diff
@@ -129,9 +129,10 @@ func (rm *RevisionManager) syncRevision(ctx context.Context, contractID types.Fi
 // reports an invalid signature, suggesting the local revision is out of sync,
 // it will synchronize with the host and retry the function using the updated
 // revision. Therefore, the revise function must be idempotent.
-func (rm *RevisionManager) WithRevision(ctx context.Context, contractID types.FileContractID, reviseFn func(contract rhp.ContractRevision) (rhp.ContractRevision, proto.Usage, error)) error {
+func (rm *RevisionManager) WithRevision(ctx context.Context, lc *LockedContract, reviseFn func(contract rhp.ContractRevision) (rhp.ContractRevision, proto.Usage, error)) error {
```
peterjan: `nit` update the docstring to reflect the new parameter type, considering it's such a special type

peterjan (changes requested):
The locking introduced here is definitely more extensive than what I would've gone for. I was going to split off the funding loop and use a simple `sync.Map` to skip over contracts. The `TryLock` vs `Lock` is nice though, but we have no real guarantees over what goroutine will end up winning the lock.. it might be that maintenance is blocked for a really long time now because other threads keep winning the same contract lock over and over again.

peterjan:
> The locking introduced here is definitely more extensive than what I would've gone for. I was going to split off the funding loop and use a simple `sync.Map` to skip over contracts. The `TryLock` vs `Lock` makes up for that, but we have no real guarantees over what goroutine will end up winning the lock.. it might be that maintenance is blocked for a really long time now because other threads keep winning the same contract lock over and over again.

Since we have quite a lot of complexity already, I would personally introduce a priority heap and make sure contract management has some sort of edge over all other threads that require a contract lock.

api/admin/admin.go
```diff
@@ -287,7 +283,6 @@ func NewAPI(chain ChainManager, accounts Accounts, contracts ContractManager, ho
 	// debug endpoints
 	if a.debug {
```
peterjan, line 288: remove the route from admin.yml

testutils/indexer.go
```diff
@@ -86,10 +86,12 @@ func defaultIndexerCfg(log *zap.Logger) *indexerCfg {
 		maintenanceSettings: MaintenanceSettings,
 		contractOpts: []contracts.ContractManagerOpt{
 			contracts.WithLogger(log.Named("contracts")),
-			contracts.WithMaintenanceFrequency(500 * time.Millisecond),
+			contracts.WithMaintenanceFrequency(100 * time.Millisecond),
```
peterjan, line 89: Out of curiosity, why did you lower this to 100ms?

api/admin/admin.go
peterjan: also check the client

peterjan:
> > The locking introduced here is definitely more extensive than what I would've gone for. I was going to split off the funding loop and use a simple `sync.Map` to skip over contracts. The `TryLock` vs `Lock` makes up for that, but we have no real guarantees over what goroutine will end up winning the lock.. it might be that maintenance is blocked for a really long time now because other threads keep winning the same contract lock over and over again.
> 
> Since we have quite a lot of complexity already, I would personally introduce a priority heap and make sure contract management has some sort of edge over all other threads that require a contract lock.

@ChrisSchinnerl do you disagree that lock starvation might be an issue here?

testutils/indexer.go
ChrisSchinnerl: To speed things up. All the maintenance operations have their own timeouts as well and i'ts just a bit slow if you have to wait for 1s in an unlucky case for pruning to happen because of the 500ms maintenance interval + the 500ms pruning interval.

ChrisSchinnerl:
> > > The locking introduced here is definitely more extensive than what I would've gone for. I was going to split off the funding loop and use a simple `sync.Map` to skip over contracts. The `TryLock` vs `Lock` makes up for that, but we have no real guarantees over what goroutine will end up winning the lock.. it might be that maintenance is blocked for a really long time now because other threads keep winning the same contract lock over and over again.
> > 
> > 
> > Since we have quite a lot of complexity already, I would personally introduce a priority heap and make sure contract management has some sort of edge over all other threads that require a contract lock.
> 
> @ChrisSchinnerl do you disagree that lock starvation might be an issue here?

I'd be more concerned about starvation if we had a priority heap for locking. Because then higher priority tasks could starve lower priority tasks. Go's `sync.Mutex` on the other hand is very fair. It uses a FIFO queue internally and will only starve a waiting goroutine for at most 1ms.

peterjan (approved):

PR #847, opened 2026-03-04
Expose PinObject

peterjan:
The Go SDK pins slabs in `Upload` (and not in `UploadPacked`) so it's internally inconsistent. The Rust SDK does not pin slabs for both regular and packed uploads but instead exposes `PinObject`. This PR makes the Go SDK consistent with the Rust SDK.

### DIFF
--- a/sdk/packing.go
+++ b/sdk/packing.go
@@ -35,9 +35,6 @@ type (
 		dataShards   uint8
 		parityShards uint8
 
-		// pin function, set by SDK
-		pinResultFn pinResultFn
-
 		// upload state
 		reader  *io.PipeReader
 		writer  *io.PipeWriter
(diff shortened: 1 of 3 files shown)


sdk/packing.go
```diff
@@ -27,14 +27,44 @@ var (
...
 type (
-	// A PackedUpload allows multiple objects to be uploaded together in a single
-	// upload. This can be more efficient than uploading each object separately
-	// if the size of the objects is less than the minimum slab size.
-	PackedUpload struct {
+	// PackedUpload defines the interface for a packed upload. A packedUpload allows
+	// multiple objects to be uploaded together in a single upload. This can be more
+	// efficient than uploading each object separately if the size of the objects is
+	// less than the minimum slab size.
+	PackedUpload interface {
```
ChrisSchinnerl: This interface isn't necessary. Interfaces should be declared where they are consumed not where they are produced.

ChrisSchinnerl (changes requested):

sdk/packing.go
peterjan: not sure what I was thinking

chris124567 (dismissed):

sdk/packing.go
```diff
@@ -257,6 +290,14 @@ func (s *SDK) UploadPacked(opts ...UploadOption) (*PackedUpload, error) {
 		dataShards:   uo.dataShards,
 		parityShards: uo.parityShards,
 
+		pinResultFn: func(ctx context.Context, result packedResult) error {
+			slabIDs, err := s.client.PinSlabs(ctx, s.appKey, result.PinParams()...)
```
ChrisSchinnerl: While going back to the Rust SDK to compare this code to the SDK I noticed that this actually breaks consistency. The Rust SDK explicitly doesn't pin packed uploads and expects you to explicitly call the save method after calling `finalize`. In the Go SDK we have `SaveObject` on the `SDK` for that.

ChrisSchinnerl (changes requested):

sdk/packing.go
peterjan: Hm damn, I should've checked. I assumed it would pin the slabs because the upload method does that and the packed version was inconsistent. So really Go's upload method was already inconsistent with the Rust SDK.

So we have two options:
a) remove pinning from Go's `Upload` method (makes it consistent with Rust, and internally consistent), and expose PinObject that pins + saves
2) update Rust to match Go

@ChrisSchinnerl can we confirm option a) - There's a comment in rust that says "The object can be sealed and pinned to the indexer when ready." which implies there's a good reason for not auto-pinning, but not sure what that would be. It looks to me like a bit of a foot gun to be honest, as an SDK user. If I forget to pin the object my data is gone all of a sudden...

We should probably also update that comment to explicitly mention `pin_object` should be called. Instead of referring to sealing + saving because saving is not even exposed on the SDK.

ChrisSchinnerl (approved):

chris124567 (approved):

PR #851, opened 2026-03-09
Add endpoint to prune orphaned slabs from an account

ChrisSchinnerl:
This will allow us to provide a dashboard function that users can use to prune garbage data from their accounts without having to go through their apps. 

It's also nice for manually pruning deleted objects. Our test server has a few lost slabs without objects that should be pruned.

### DIFF
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -623,6 +623,7 @@ func (a *admin) handleDELETEAccountPrune(jc jape.Context) {
 		jc.Error(err, http.StatusInternalServerError)
 		return
 	}
+	jc.Encode(nil)
 }
 
 func (a *admin) handlePATCHAccount(jc jape.Context) {
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -127,6 +127,7 @@ type (
 
 		DeleteContract(contractID types.FileContractID) error
 		LastScannedIndex() (types.ChainIndex, error)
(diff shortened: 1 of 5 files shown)


api/admin/admin.go
```diff
@@ -208,10 +209,11 @@ func NewAPI(chain ChainManager, accounts Accounts, contracts ContractManager, ho
...
 
 		// accounts endpoints
-		"GET    /accounts":            a.handleGETAccounts,
-		"GET    /account/:accountkey": a.handleGETAccount,
-		"DELETE /account/:accountkey": a.handleDELETEAccount,
-		"PATCH  /account/:accountkey": a.handlePATCHAccount,
+		"GET    /accounts":                  a.handleGETAccounts,
+		"GET    /account/:accountkey":       a.handleGETAccount,
+		"DELETE /account/:accountkey":       a.handleDELETEAccount,
+		"DELETE /account/:accountkey/prune": a.handleDELETEAccountPrune,
```
peterjan: This doesn't feel like a `DELETE` to me. That should be reserved for when you delete the resource itself. What about `POST /account/:accountkey/slabs/prune`

peterjan (changes requested):

api/admin/admin.go
peterjan: (note that we already have POST /slabs/prune so it would be consistent)

peterjan (approved):

PR #854, opened 2026-03-10
Fix subscriber

peterjan:
We removed `ctx` from some interfaces and left a `cancel` in there that now cancels the threadgroup context. Essentially causing us to sync up to a 100 blocks every time a block gets mined.

https://github.com/SiaFoundation/indexd/commit/b2ab75650d6fbae47ad7d00a50952d63cccefa79#diff-fb63e758ee6fcf8760a61967cce684177355cb624442a5028a6fefe004bb019cL184


subscriber/subscriber.go
```diff
@@ -201,7 +206,6 @@ func (s *Subscriber) Sync(ctx context.Context) error {
 			}
 			return nil
 		})
-		cancel()
```
ChrisSchinnerl, line 204: You can also get rid of the call to `s.tg.AddContext` at the beginning of `Sync` now. It doesn't spawn any goroutines itself so there is no point in having a `cancel` that gets deferred.

subscriber/subscriber.go
```diff
@@ -123,8 +123,13 @@ func New(cm ChainManager, hm HostManager, contracts ContractManager, wm WalletMa
 			select {
 			case <-reorgCh:
 				err := s.Sync(ctx)
-				if err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, threadgroup.ErrClosed) {
-					s.log.Panic("failed to sync database", zap.Error(err))
+				if err != nil {
+					if errors.Is(err, context.Canceled) {
+						s.log.Warn("sync cancelled", zap.Error(err))
```
ChrisSchinnerl: Do we need this change? The context should never get cancelled before the goroutine exits so that would be a developer error anyway.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (changes requested):
CI is failing. Please wait for pipelines to pass before assigning reviewers to save that extra roundtrip.

subscriber/subscriber.go
```diff
@@ -123,7 +123,7 @@ func New(cm ChainManager, hm HostManager, contracts ContractManager, wm WalletMa
 			select {
 			case <-reorgCh:
 				err := s.Sync(ctx)
-				if err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, threadgroup.ErrClosed) {
+				if err != nil && !errors.Is(err, threadgroup.ErrClosed) {
```
ChrisSchinnerl: Have you tried this locally? I tried to add a `Close` call to the subscriber in your test at the very end for good measure and it panics with a `context canceled`. So we might need to keep both anyway.

ChrisSchinnerl (changes requested):

subscriber/subscriber.go
peterjan: yeah I tried it but this was something I changed afterwards

peterjan: I already defer a call to `Close` in my tests right? Where did you add it? I'll revert the changes to the error check. I wanted to handle that differently because we had this in our code for 4 months without noticing. I guess the test accomplishes somewhat the same thing, better not to have log entries on shutdown for expected behaviour.

ChrisSchinnerl (approved):

PR #859, opened 2026-03-11
Extend accounts.Account type with Ready field

### LINKED ISSUE #856: Extend `accounts.Account` type with `Ready` field
The `Ready` field should serve as an indicator of whether the account is ready to be used. For now this should be false until an account has gone through 1 round of successful funding.

To determine this when fetching an account we should join with `account_hosts` and consider it `true` if `>= 30` rows exist in `account_hosts` with `consecutive_failed_funds == 0`.

That works because rows are inserted the first time we attempted to fund an account with a host which means no failed funds equals to at least 1 good one.

chris124567:
Close #856


persist/postgres/accounts.go
```diff
@@ -41,14 +41,20 @@ func (s *Store) Accounts(offset, limit int, opts ...accounts.QueryAccountsOpt) (
...
 		rows, err := tx.Query(ctx, `
-			SELECT a.public_key, ak.app_key, a.max_pinned_data, a.pinned_data, a.app_id, a.description, a.logo_url, a.service_url, a.last_used
+			SELECT a.public_key, ak.app_key, a.max_pinned_data, a.pinned_data, COALESCE(ahr.ready_hosts, 0) >= $4, a.app_id, a.description, a.logo_url, a.service_url, a.last_used
 			FROM accounts a
 			INNER JOIN app_connect_keys ak ON ak.id = a.connect_key_id
+			LEFT JOIN LATERAL (
+				SELECT COUNT(*) AS ready_hosts
+				FROM account_hosts ah
+				WHERE ah.account_id = a.id
+				  AND ah.consecutive_failed_funds = 0
```
peterjan: We don't have a dedicated index for this, `consecutive_failed_funds` could be included in a composite index for instance. We should benchmark it because this might need to scan all rows, especially since this lateral subquery runs for every row. Perhaps we could also bound the `COUNT`, e.g. we can stop counting at 30 ready hosts, we don't need to know we have 200 for instance if all we do is check it's higher than 30.

openapi/components.yml
```diff
@@ -415,6 +415,9 @@ components:
           type: number
           format: uint64
           description: Current amount of pinned data in bytes.
+        ready:
+          type: boolean
+          description: Whether the account is funded on enough hosts to be used
```
peterjan:
```suggestion
          description: Whether the account has been successfully funded on at least 30 hosts.
```
It's annoying that we have to update the docs when we update the threshold, but it's more annoying to have accounts that aren't ready and not know what the threshold is.

peterjan (changes requested):

openapi/components.yml
ChrisSchinnerl: This is fine. `ready` is already pretty vague because we only want devs to worry about whether an account is ready or not. So I'd change it so something like this:

`Whether the account is ready to be used. This only returns 'false' after an account is created. Once an account is ready it will remain ready. Getting ready can take up to a few minutes.`

ChrisSchinnerl (changes requested):

persist/postgres/accounts.go
chris124567:
```
goos: linux
goarch: amd64
BenchmarkAccountReady
BenchmarkAccountReady/Account
BenchmarkAccountReady/Account-4               56394            219209 ns/op
BenchmarkAccountReady/Accounts
BenchmarkAccountReady/Accounts-4                306          37624068 ns/op
PASS
ok         go.sia.tech/indexd/persist/postgres     79.186s
```

peterjan (dismissed):
raked up some conflicts but otherwise lgtm

ChrisSchinnerl (dismissed):
Looking good but got conflicts now.

peterjan (approved):

ChrisSchinnerl (approved):

PR #862, opened 2026-03-12
Extend testing around contract state updates

### LINKED ISSUE #853: Make sure revisions and renewals in the same block work as expected
Since we are using a `switch` in `applyV2ContractDiffs` similar to `hostd` did before https://github.com/SiaFoundation/hostd/pull/930 we should extend `TestApplyRevertDiff` to make sure we don't have the same issue in `indexd` and if we do, fix it.

peterjan:
This PR adds unit tests to make sure indexd doesn't suffer from the same bug as hostd that was fixed here https://github.com/SiaFoundation/hostd/pull/930.

Since nothing was fixed I didn't add a changelog.

Closes https://github.com/SiaFoundation/indexd/issues/853

### DIFF
--- a/contracts/chain_test.go
+++ b/contracts/chain_test.go
@@ -648,113 +648,351 @@ func (w *walletMock) ReleaseInputs(txns []types.Transaction, v2txns []types.V2Tr
 func (w *walletMock) SignV2Inputs(txn *types.V2Transaction, toSign []int)                  {}
 
 func TestApplyRevertDiff(t *testing.T) {
+	// create contract manager
 	cm := contracts.NewTestContractManager(types.PublicKey{}, nil, nil, nil, nil, nil, nil, nil, contracts.NewContractLocker(), nil, nil, nil)
 
-	// create a contract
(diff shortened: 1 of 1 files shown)


peterjan:
@ChrisSchinnerl couple of questions: 

- I originally wanted to refactor our tests to be more in line with the way we write these assertions in `renterd`. I came back from that though and figured it'd be better to stay consistent with the other tests. Do these work for you or do you want me to refactor them

- I noticed that we never update the state to `ContractStateExpired` in `indexd`, should we consider setting it instead of doing what we do now which is set it to resolved in all instances

- I considered updating `chain.go` to get rid of the switch, however the way we implemented the switch (using a `default` case which is always evaluated last), I opted not to change the code because we don't have the bug. Does that work for you, or should we be more explicit here and switch to an if/else or maybe add a NOTE.

ChrisSchinnerl (commented):
The test looks fine and we don't need to update anything that is not broken.

ChrisSchinnerl (approved):

PR #864, opened 2026-03-12
Re-enable QUIC

ChrisSchinnerl:
This PR enables QUIC in the client again. Unfortunately it's not as easy as uncommenting the code so this PR comes with a few more changes.

1. QUIC is enabled as the fallback and we try SiaMux first since it just seems to perform better
2. `shouldResetTransport` was updated to make sure we handle QUIC stream errors correctly
3. `rpcFn` now logs a debug line whenever we reset the transport to make sure we will be aware of degradations that stem from resetting the transport for the wrong reasons
4. `shardDownloadTicker` was added to the migration's download code to start launching new downloads sooner if possible. Waiting for 30s before launching another worker really kills downloads so this was changed to 5s. It's also nice because it's not quite obvious that changing the shard timeout has an impact and when the next worker gets launched.

At this point I'd also like to highlight a downside of QUIC I observed. The fact that a QUIC connection always times out and never fails immediately. In `coreutils` we have a 15s handshake timeout in `Dial` which is greater than the `5s` default in `quic.Config`. We might want to consider changing this to `3-5s` to not get stuck for `15s` on a host that is just offline.


peterjan:
@ChrisSchinnerl after our meeting from yesterday, do we still want to re-enable QUIC?

ChrisSchinnerl:
> @ChrisSchinnerl after our meeting from yesterday, do we still want to re-enable QUIC?

I'll leave that to @n8mgr . If we don't I can remove that from this PR and we could just merge the logging and ticker interval since that also helps when using the Siamux.

PR #867, opened 2026-03-15
Add ephemeral key to auth

n8mgr:
`GET /auth/connect/:requestID/status` and `POST /auth/connect/:requestID/register` previously accepted requests from any caller who knew the `requestID`. This meant observers could steal the `userSecret` or register an arbitrary key. It's unlikely to guess the random request ID, but it's also an easily avoidable vulnerability. This adds an ephemeral key pair to authenticate the request. The initial connect request is signed with it then all subsequent steps in the flow verify the signature in the same way app requests are verified.

Sia Storage and the Rust SDK will need to be updated similarly.

### DIFF
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -608,20 +608,24 @@ func (a *app) handleGETAuthConnectStatus(jc jape.Context) {
 
 	a.mu.Lock()
 	authReq, ok := a.authRequests[requestID]
-	a.mu.Unlock()
 	if !ok || time.Now().After(authReq.Expiration) {
+		a.mu.Unlock()
 		jc.Error(fmt.Errorf("request invalid or expired"), http.StatusNotFound)
 		return
 	} else if signerKey, ok := validateURLSignature(jc, a.hostname); !ok {
+		a.mu.Unlock()
 		return
 	} else if authReq.EphemeralKey != signerKey {
+		a.mu.Unlock()
 		jc.Error(fmt.Errorf("invalid request signature"), http.StatusUnauthorized)
 		return
 	}
-	jc.Encode(AuthConnectStatusResponse{
+	resp := AuthConnectStatusResponse{
 		Approved:   authReq.Approved,
 		UserSecret: authReq.UserSecret,
-	})
+	}
+	a.mu.Unlock()
+	jc.Encode(resp)
 }
 
 // 5. app registers the public key with indexd using a request signed
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -608,24 +608,20 @@ func (a *app) handleGETAuthConnectStatus(jc jape.Context) {
 
 	a.mu.Lock()
 	authReq, ok := a.authRequests[requestID]
+	a.mu.Unlock()
 	if !ok || time.Now().After(authReq.Expiration) {
-		a.mu.Unlock()
 		jc.Error(fmt.Errorf("request invalid or expired"), http.StatusNotFound)
 		return
 	} else if signerKey, ok := validateURLSignature(jc, a.hostname); !ok {
-		a.mu.Unlock()
 		return
 	} else if authReq.EphemeralKey != signerKey {
-		a.mu.Unlock()
 		jc.Error(fmt.Errorf("invalid request signature"), http.StatusUnauthorized)
 		return
 	}
-	resp := AuthConnectStatusResponse{
+	jc.Encode(AuthConnectStatusResponse{
 		Approved:   authReq.Approved,
 		UserSecret: authReq.UserSecret,
-	}
-	a.mu.Unlock()
-	jc.Encode(resp)
+	})
 }
 
 // 5. app registers the public key with indexd using a request signed
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -74,8 +74,10 @@ type (
 	}
 
 	authReq struct {
-		Request    RegisterAppRequest
-		Expiration time.Time
+		// EphemeralKey is the public key used to sign the initial connection request. It is used to authenticate future requests for the same connection request, so it must be unique for each connection request. It is not used for authentication after the app key is registered.
+		EphemeralKey types.PublicKey
+		Request      RegisterAppRequest
+		Expiration   time.Time
 
 		// set when the user approves the request
 		Approved   bool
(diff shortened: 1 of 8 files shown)


api/app/app.go
```diff
@@ -115,6 +117,14 @@ type (
 		Approve bool `json:"approve"`
 	}
 
+	// RegisterAppKeyRequest is the request body for registering an application key after the user approves the connection request.
+	// The request must be signed with the app key to prove ownership of it.
+	RegisterAppKeyRequest struct {
+		AppKey types.PublicKey `json:"appKey"`
+		// Signature is a signature of the request ID signed with the app key to prove ownership
+		Signature types.Signature `json:"signature"`
+	}
```
ChrisSchinnerl, line 137: Could you elaborate a bit on the attack vector you had in mind? It doesn't seem to be fully broken https but somewhat of a not full middleman observer that can get access to the request id but not see any requests or responses. Is that correct?

Because I'm wondering how far we need to take this. e.g. do we worry about a replay attack on the `/auth/connect/:requestID/status` endpoint? In theory knowing the signed request can expose the user secret if you send it before it expires or the app finishes the handshake. But if we assume the request can't be known it's fine.

ChrisSchinnerl (commented):
Left a question but looks good to me.

api/app/app.go
n8mgr: We were effectively treating the request ID as a pseudo-password. If it ever leaked somehow, the "handshake" could be broken by a timing attack and anyone could register an arbitrary app key. With this PR no part of the auth flow relies on the transport being secure. The only thing it requires now is that the client-side stuff stays client-side.

ChrisSchinnerl (dismissed):

api/app/app.go
```diff
@@ -588,16 +606,25 @@ func (a *app) handleGETAuthConnectStatus(jc jape.Context) {
 	jc.DecodeParam("requestID", &requestID)
 
 	a.mu.Lock()
-	defer a.mu.Unlock()
 	authReq, ok := a.authRequests[requestID]
 	if !ok || time.Now().After(authReq.Expiration) {
```
chris124567, line 612: Don't we only need the lock when we we get the `authReq`?

ChrisSchinnerl (dismissed):

api/app/client.go
```diff
@@ -323,13 +325,29 @@ func (c *Client) RequestStatus(ctx context.Context, statusURL string) (status Au
...
 // app key and registration URL.
-func (c *Client) RegisterApp(ctx context.Context, registerURL string, appKey types.PrivateKey) error {
+//
+// The request must be signed with the ephemeral key
+func (c *Client) RegisterApp(ctx context.Context, registerURL string, ephemeralKey, appKey types.PrivateKey) error {
 	u, err := url.Parse(registerURL)
 	if err != nil {
 		return fmt.Errorf("failed to parse register URL: %w", err)
 	}
+	// extract request ID from path
```
peterjan, line 335: This could be improved imo. Instead of having clients parse the request ID from the URL, perhaps we should return it as a separate field in `RegisterAppResponse`. If we don't want to do that perhaps we could trim known suffix and extract the ID that way, splitting on `/` and handling the pieces feels very brittle and is something that could fail silently if we ever change path prefix from /api/ to /api/v1 or something along those lines.

api/app/auth.go
```diff
@@ -52,7 +52,7 @@ const (
 // NOTE: This function does not check if the account exists, it only validates
 // the signature and expiration. If you need to check for account existence, use
 // validateSignedURLAuth instead.
-func ValidateURLSignature(req *http.Request, hostname, path string) (types.PublicKey, error) {
+func ValidateURLSignature(req *http.Request, hostname string) (types.PublicKey, error) {
```
peterjan, line 55: well spotted, this diff made me notice how we handle `maxRequestSize` 5 lines down... shouldn't we read limit+1 and then check the buffer to see if we exceed the max request size and return an error here? I was a little confused by those lines, they seem a little misplaced, might even be a leftover or something

api/app/app.go
```diff
@@ -74,8 +74,10 @@ type (
 	}
 
 	authReq struct {
-		Request    RegisterAppRequest
-		Expiration time.Time
+		// EphemeralKey is the public key used to sign the initial connection request. It is used to authenticate future requests for the same connection request, so it must be unique for each connection request. It is not used for authentication after the app key is registered.
```
peterjan, line 77: I'd consider dropping "so it must be unique for each connection request", we don't keep track of seen keys and it's really only a randomly generated key that must be the same in the scope of the connection request.

peterjan: This is a big `nit` of course but I checked the builder and its docstring says it's meant for connecting applications to the indexer. So currently we'll be connecting multiple applications using the same `ephemeralKey`, if people don't recreate the builder. Which is not that big of a deal but I just wanted to point it out.

```
// NewBuilder creates a new Builder for connecting applications to the indexer.
```

api/app/client.go
n8mgr: > fail silently

Auth would break. That’s far from silent.

peterjan: I'm sure you know what I meant by that. I still feel it's a little brittle and we could improve it by trimming the known suffix and considering the last part after trimming to be the request ID but I'll approve the PR.

peterjan (dismissed):

api/app/app.go
n8mgr: Changed the builder to be consumed after the SDK is created like it is in Rust. A random ed25519 key is effectively unique even if it's not tracked.

peterjan (dismissed):

api/app/auth.go
n8mgr: https://github.com/SiaFoundation/indexd/pull/872

sdk/connect.go
```diff
@@ -83,15 +105,19 @@ func (b *Builder) WaitForApproval(ctx context.Context) (bool, error) {
 // This key should be stored securely by the application and never
 // shared with anyone else. It can be regenerated using the same app
 // ID, user account, and seed phrase.
+//
+// // Panics if the builder has already created an SDK instance.
```
chris124567: nit: double comment

chris124567 (dismissed):

peterjan (approved):

sdk/connect.go
```diff
@@ -40,17 +41,38 @@ type (
...
 
 	request      app.RegisterAppRequest
 	registerResp *app.RegisterAppResponse
 	sharedSecret types.Hash256
+
+	consumed *atomic.Bool
+}
+
+// consume marks the builder as consumed, preventing further use. It panics if the builder has already been consumed.
+func (b *Builder) consume() {
```
ChrisSchinnerl, line 55: Insert *look what they have to do to mimic a fraction of our power" meme

ChrisSchinnerl (approved):

PR #871, opened 2026-03-17
Ignore offset and limit for prometheus responses

### LINKED ISSUE #604: Remove pagination from `/stats/hosts`

peterjan:
Not that important but just figured I'd get the issue closed since we're nearing release and want a stable API.

Closes https://github.com/SiaFoundation/indexd/issues/604

### DIFF
--- a/testutils/db.go
+++ b/testutils/db.go
@@ -19,6 +19,7 @@ import (
 	"go.sia.tech/indexd/persist/postgres"
 	"go.sia.tech/indexd/subscriber"
 	"go.uber.org/zap"
+	"lukechampine.com/frand"
 )
 
 const (
@@ -146,8 +147,7 @@ func (ts TestStore) AddTestAccount(t testing.TB, ak types.PublicKey) {
 	} else if err != nil {
 		t.Fatal(err)
 	}
-
-	if err := ts.RegisterAppKey(connectKey, ak, accounts.AppMeta{}); err != nil {
+	if err := ts.RegisterAppKey(connectKey, ak, accounts.AppMeta{ID: frand.Entropy256()}); err != nil {
 		t.Fatal(err)
 	}
 }
--- a/.changeset/ignore_offset_and_limit_for_prometheus_responses.md
+++ b/.changeset/ignore_offset_and_limit_for_prometheus_responses.md
@@ -1,0 +1,5 @@
+---
+default: minor
+---
+
+# Ignore offset and limit for prometheus responses.
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -2,6 +2,7 @@ package admin
 
 import (
 	"context"
+	"math"
 	"net/http"
 	"regexp"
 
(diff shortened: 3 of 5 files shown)


testutils/db.go
```diff
@@ -146,8 +147,7 @@ func (ts TestStore) AddTestAccount(t testing.TB, ak types.PublicKey) {
 	} else if err != nil {
 		t.Fatal(err)
 	}
-
-	if err := ts.RegisterAppKey(connectKey, ak, accounts.AppMeta{}); err != nil {
+	if err := ts.RegisterAppKey(connectKey, ak, accounts.AppMeta{ID: frand.Entropy256()}); err != nil {
```
peterjan, line 150: We group account stats on the `app_id` but it's not unique. We currently also allow empty values, we probably want to be more strict around the app ID and enforce it's set and even unique(?) cc @n8mgr @ChrisSchinnerl

ChrisSchinnerl: The API makes sure the app id is set to something that is not an empty hash. We can't really enforce uniqueness though. An app can use any app id it wants.

chris124567 (dismissed):

ChrisSchinnerl (commented):
Leaving it up to @n8mgr to decide whether to merge this since he was originally against it. 

Afaik this endpoint only returns hosts that we use but depending on the size of the indexer this number might still end up being pretty big.

It might also be worthwhile to check which graphs in our dashboard make use of this and whether we still find them useful or not and whether we can replace them with an aggregated value.

n8mgr:
90% of the time you want outliers or aggregates in your graphs. Displaying 500 lines is utterly useless. Removing pagination and making an api response unbounded is not a good experience.

chris124567 (approved):

peterjan:
> It might also be worthwhile to check which graphs in our dashboard make use of this and whether we still find them useful or not and whether we can replace them with an aggregated value.

I look at it more from a debugging perspective. It's not because we don't have a graph for it, that it might not prove useful if shit hits the fan and you have to debug production because all hosts lost all contracts for some reason. 

> Afaik this endpoint only returns hosts that we use but depending on the size of the indexer this number might still end up being pretty big.

It's hosts stats, but also app stats, which we added later.

> 90% of the time you want outliers or aggregates in your graphs. Displaying 500 lines is utterly useless. Removing pagination and making an api response unbounded is not a good experience.

I agree with that, I personally don't think it's that big of a deal to ignore it only for prometheus responses though.  Perhaps we should just close out the issue? I really don't feel strongly about it, I was just scanning through remaining issues (there's not that many) and found one we might want to pick up before finalising the API, that's all.

PR #872, opened 2026-03-17
Properly handle max request size

n8mgr:
Follow up to #867

### DIFF
--- a/api/app/auth.go
+++ b/api/app/auth.go
@@ -27,14 +27,6 @@ var (
 	// because of an unexpected issue.
 	ErrInternalError = errors.New("internal error")
 
-	// ErrSignatureExpired is returned when a signed URL can not be
-	// authenticated because the valid until timestamp is in the past.
-	ErrSignatureExpired = errors.New("signature expired")
-
-	// ErrSignatureInvalid is returned when a signed URL can not be
-	// authenticated because the signature was invalid.
-	ErrSignatureInvalid = errors.New("invalid signature")
-
 	// ErrUnknownAccount is returned when a signed URL can not be authenticated
 	// because the account does not exist in the account store.
 	ErrUnknownAccount = errors.New("unknown account")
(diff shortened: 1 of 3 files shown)


api/app/auth.go
```diff
@@ -47,59 +39,56 @@ const (
...
 // NOTE: This function does not check if the account exists, it only validates
 // the signature and expiration. If you need to check for account existence, use
 // validateSignedURLAuth instead.
-func ValidateURLSignature(req *http.Request, hostname string) (types.PublicKey, error) {
-	defer req.Body.Close()
-
-	buf, err := io.ReadAll(io.LimitReader(req.Body, maxRequestSize))
-	if err != nil {
-		return types.PublicKey{}, fmt.Errorf("failed to read request body: %w", err)
+func ValidateURLSignature(r *http.Request, w http.ResponseWriter, hostname string) (types.PublicKey, bool) {
```
peterjan, line 57: Would it not be cleaner to extract the `maxRequestSize` logic into a separate middleware. Similar to how we do with the `corsMux`, we could call it `limitBodyMux` or something and avoid passing the `http.ResponseWriter` here so on the call site we don't have to `http.Error` in some branches and simply return in others. 

This would clean up `ValidateURLSignature` signature, because we could have it return an error and return `http.StatusUnauthorized` for all errors, and not pass a response writer. It would also be more robust, because I think currently we are missing the check in `POST /auth/connect/:requestID`

Something like 

```go
// limitBodyMux wraps an http.Handler to limit POST request bodies to
// maxRequestSize. If the body exceeds the limit, it responds with 413.
// The body is fully read and buffered so downstream handlers can read
// it multiple times.
func limitBodyMux(h http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if r.Method == http.MethodPost {
			body := http.MaxBytesReader(w, r.Body, maxRequestSize)
			buf, err := io.ReadAll(body)
			body.Close()
			if _, ok := errors.AsType[*http.MaxBytesError](err); ok {
				http.Error(w, "request body too large", http.StatusRequestEntityTooLarge)
				return
			} else if err != nil {
				http.Error(w, "failed to read request body", http.StatusInternalServerError)
				return
			}
			r.Body = io.NopCloser(bytes.NewReader(buf))
		}
		h.ServeHTTP(w, r)
	})
}
```

peterjan (changes requested):

api/app/auth.go
ChrisSchinnerl: This feels a bit like a "why not both" kinda situation. We definitely want to limit it in `ValidateURLSignature` since that is an exposed method and a caller can't necessarily be expected to know the internals well enough to know to limit the body and what to limit it to.

On the other hand a middleware like the one in siastorage we wrap all routes in definitely gives peace of mind that all routes are protected. For reference in siastorage we have:

```go
func (s *server) maxBytesMiddleware(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		r.Body = http.MaxBytesReader(w, r.Body, maxHTTPRequestBodyBytes)
		next.ServeHTTP(w, r)
	})
}
```

peterjan: Good poin, I didn't consider the fact that `ValidateURLSignature` is exposed. That said, do we need `ValidateURLSignature` to check the request body? Is it so bad if it would only validate the signature and then if for whatever reason the request body is too big the user is made aware of that only when the request is executed. Perhaps `ValidateRequest` is a better name for the method in that case.

ChrisSchinnerl (approved):

PR #873, opened 2026-03-17
Use minimal object request for saving

n8mgr:
When sending a full object, the maximum object size given `maxRequestSize = 1MiB` is less than 10GiB at 10-of-30. This PR raises the limit to 10MiB and introduces a minimal `PinObjectRequest` upping the per request object size limit to over a TiB.

Just raising the limit to `maxRequestSize = 10MiB` is less than 100GiB per object. Just introducing the type is less than 500GiB per object. Both changes put us just under 4TiB. AWS S3 has an object size of 50TiB. I don't think we necessarily need to aim for that, but it's good context for expectations.

### DIFF
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -43,7 +43,7 @@ type (
 
 		Object(ctx context.Context, account proto.Account, key types.Hash256) (slabs.SealedObject, error)
 		DeleteObject(ctx context.Context, account proto.Account, objectKey types.Hash256) error
-		SaveObject(ctx context.Context, account proto.Account, obj slabs.SealedObject) error
+		SaveObject(ctx context.Context, account proto.Account, obj slabs.PinObjectRequest) error
 		ListObjects(ctx context.Context, account proto.Account, cursor slabs.Cursor, limit int) ([]slabs.ObjectEvent, error)
 		SharedObject(ctx context.Context, key types.Hash256) (slabs.SharedObject, error)
 	}
@@ -317,7 +317,7 @@ func (a *app) handleGETObjects(jc jape.Context, pk types.PublicKey) {
 }
 
 func (a *app) handlePOSTObjects(jc jape.Context, pk types.PublicKey) {
-	obj, ok := decodeRequest[slabs.SealedObject](jc)
+	obj, ok := decodeRequest[slabs.PinObjectRequest](jc)
 	if !ok {
 		return
 	}
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -323,7 +323,7 @@ func (a *app) handlePOSTObjects(jc jape.Context, pk types.PublicKey) {
 	}
 
 	err := a.slabs.SaveObject(jc.Request.Context(), proto.Account(pk), obj)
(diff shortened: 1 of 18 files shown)


slabs/objects.go
```diff
@@ -78,8 +78,89 @@ type (
...
+	h.E.Write(pr.EncryptedMetadataKey)
+	h.E.Write(pr.EncryptedMetadata)
+	return h.Sum()
+}
+
+// VerifyDataSignature verifies the object's data signature using the given
+// public key.
+func (pr *PinObjectRequest) VerifyDataSignature(pk types.PublicKey) error {
+	if !pk.VerifyHash(pr.DataSigHash(), pr.DataSignature) {
+		return fmt.Errorf("%w: invalid data signature", ErrInvalidObjectSignature)
```
peterjan, line 133: We have to extend the error check in `POST /objects` because currently `ErrInvalidObjectSignature` would return a 500

peterjan (changes requested):
Should we consider renaming `SaveObject` to `PinObject` and get rid of the extra type of action you can do with an object, SaveObject(... PinObjectRequest) is somewhat triggering.

slabs/objects.go
```diff
@@ -78,8 +78,89 @@ type (
...
+		// key, and encrypted_metadata) to attest that the object has not been
+		// tampered with.
+		MetadataSignature types.Signature `json:"metadataSignature"`
+	}
 )
 
+// ID returns the object's ID, which is a hash of its slabs. The
+// object ID is used to identify the object and retrieve it from
+// the indexer.
+func (pr *PinObjectRequest) ID() types.Hash256 {
```
ChrisSchinnerl: Considering that we now break the endpoint anyway how about we add this as a field to `PinObjectRequest` and then have the backend validate the ID you provided with what it thinks the ID should be.

api/app/app.go
```diff
@@ -307,7 +307,7 @@ func (a *app) handleGETObjects(jc jape.Context, pk types.PublicKey) {
 }
 
 func (a *app) handlePOSTObjects(jc jape.Context, pk types.PublicKey) {
-	var obj slabs.SealedObject
+	var obj slabs.PinObjectRequest
```
ChrisSchinnerl: Need to update the open api spec now

ChrisSchinnerl (changes requested):

slabs/objects.go
n8mgr: Why? To avoid defining ID again? That’s 64 extra hex encoded bytes for no reason.

ChrisSchinnerl: To make sure a client can't successfully create an object without knowing the ID to retrieve it with.

n8mgr: They can call ID like they do now? This doesn't introduce new behavior. This is the same behavior we already said was fine with SealedObject.

n8mgr: In the future, let's try to keep things scoped to the PR. Logic changes should be in a follow-up.

ChrisSchinnerl (approved):

PR #876, opened 2026-03-19
Make app ID required

peterjan:
When we fetch account stats, we group on the app ID. We can't make it unique, but perhaps we should make it required when registering an app. Maybe we should generate a random value ourselves if the user doesn't want to pass an ID...


peterjan:
Found 
```
case req.AppID == types.Hash256{}:
		jc.Error(errors.New("app ID is required"), http.StatusBadRequest)
		return
```

PR #877, opened 2026-03-19
Rejecting app connection shouldn't require app password

### LINKED ISSUE #850: Rejecting app connection shouldn't require app password
When I want to reject the app connection (for testing purposes), I'm still required to enter app password. I feel like this shouldn't be necessary, though.

Image

chris124567:
Close #850

### DIFF
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -560,18 +560,26 @@ func (a *app) handlePOSTAuthConnect(jc jape.Context) {
 		return
 	}
 
+	if !approveReq.Approve {
+		// user rejected the request, no password needed
+		a.mu.Lock()
+		delete(a.authRequests, requestID)
+		a.mu.Unlock()
+		jc.Encode(nil)
+		return
+	}
+
 	_, connectKey, ok := jc.Request.BasicAuth()
 	if !ok || connectKey == "" {
 		jc.Error(fmt.Errorf("missing basic auth password"), http.StatusBadRequest)
 		return
 	}
 
-	if !approveReq.Approve {
-		// user rejected the request
-		a.mu.Lock()
-		delete(a.authRequests, requestID)
-		a.mu.Unlock()
-		jc.Encode(nil)
+	if err := a.accounts.ValidAppConnectKey(jc.Request.Context(), connectKey); errors.Is(err, accounts.ErrKeyNotFound) {
+		jc.Error(errors.New("invalid app connect key"), http.StatusUnauthorized)
+		return
+	} else if err != nil {
+		jc.Error(ErrInternalError, http.StatusInternalServerError)
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


api/app/app.go
```diff
@@ -816,7 +804,7 @@ func NewAPI(advertiseURL string, store Store, am Accounts, contracts Contracts,
 	}, map[string]jape.Handler{
 		// CORS is disabled on these routes because we don't want to encourage programmatic access. It can't be
 		// blocked entirely, but it's less convenient without CORS support.
-		"GET /auth/connect/:requestID":  a.handleGETAuthConnectUI,               // UI for accept/reject connection requests
-		"POST /auth/connect/:requestID": wrapBasicAuth(a.handlePOSTAuthConnect), // API for accept/reject connection requests
+		"GET /auth/connect/:requestID":  a.handleGETAuthConnectUI, // UI for accept/reject connection requests
+		"POST /auth/connect/:requestID": a.handlePOSTAuthConnect,  // API for accept/reject connection requests
```
peterjan, line 808: It feels off to have authentication depend on the request body. That makes the API harder to reason about and to document, since the auth requirements aren’t clear from the route itself. It also breaks the usual request flow where authentication happens before we read or process the body, introducing a special case that could cause issues if we refactor later.

Would it make more sense to split this into separate routes, like /approve and /reject, so the auth requirements are explicit and consistent?

api/app/app.go
```diff
@@ -560,18 +560,26 @@ func (a *app) handlePOSTAuthConnect(jc jape.Context) {
 		return
 	}
 
+	if !approveReq.Approve {
+		// user rejected the request, no password needed
+		a.mu.Lock()
+		delete(a.authRequests, requestID)
+		a.mu.Unlock()
+		jc.Encode(nil)
+		return
```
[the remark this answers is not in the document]
peterjan: Might as well add this right? Considering we now conditionally set the auth in tests.

peterjan (changes requested):
PR looks good but the conditional auth triggered me. Explaining our auth is hard enough as it is, having a route that sometimes requires authentication and sometimes not does not make it better.

api/app/app.go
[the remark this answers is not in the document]
ChrisSchinnerl: `TestAppConnect` already does that. Search for `reject the request` in `app_test.go`.

api/app/app.go
ChrisSchinnerl: I want to say yeah that would be slightly more ergonomic but given that it breaks the API and requires updating pretty much every other repo including a release of Sia Mobile it might not be worth it.

ChrisSchinnerl (approved):

api/app/app.go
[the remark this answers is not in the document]
peterjan: Thanks

api/app/app.go
peterjan: Yeah it's now or never. I'm personally leaning towards now though. Approved but maybe we should run this by Nate as well. It's not great developer experience and it's a key route.

peterjan (approved):

api/app/app.go
ChrisSchinnerl: cc @n8mgr

PR #886, opened 2026-03-23
Add endpoint to delete slab

### LINKED ISSUE #882: Add debug route for deleting objects by slab id
We'd like to delete old slabs from the indexer which can't be repaired anymore. This is quite the dangerous task to do from the database since we have quite a few tables to modify whenever an object or slab is deleted. 

To make it easier it would be good to have a DELETE route under /debug which allows for deleting all objects that contain a given slab. This will require:

- a new lookup method on the store which returns a list of account + object id pairs for a slab
- a loop that calls the existing `store.DeleteObject`
- a call to `store.PruneSlabs` afterwards to prune the slabs from all accounts that had objects deleted

peterjan:
This PR adds a two debug endpoints to allow deleting a specific slab and prune slabs across all accounts.

Closes https://github.com/SiaFoundation/indexd/issues/882


api/admin/admin.go
```diff
@@ -328,6 +334,62 @@ func (a *admin) handleGETPProf(jc jape.Context) {
...
+		seen[obj.Account] = struct{}{}
+	}
+
+	// prune orphaned slabs for each affected account
+	for acc := range seen {
+		if err := a.store.PruneSlabs(acc); err != nil {
+			jc.Check("failed to prune slabs", err)
+			return
+		}
+	}
```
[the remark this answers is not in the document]
peterjan: I think this is fine (?), the issue described reusing existing store methods. There's no risk of leaving the DB in a partially modified state here since we want to get rid of the slab anyway and we're not performing any destructive actions that can lead to loss of data we want to keep. There's no need to make this atomic imo, but I don't mind updating the PR of course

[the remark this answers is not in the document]
ChrisSchinnerl: Yeah it should be fine. It's a debug endpoint. Pruning slabs itself is also not atomic and instead performs batching. You can always run it multiple times.

api/admin/admin.go
```diff
@@ -328,6 +334,62 @@ func (a *admin) handleGETPProf(jc jape.Context) {
...
+
+	// fetch all accounts referencing the slab
+	accounts, err := a.store.AccountsForSlab(slabID)
```
ChrisSchinnerl: If you do this you don't really need to return the account ids alongside the objects and have the previous `PruneSlabs` call right? This is effectively a superset since it also returns the id of an account that stores the slab but doesn't have an object that uses it.

api/admin/admin_test.go
ChrisSchinnerl, line 1070: Seems like TestSettingsAPI failed due to a constraint violation.

ChrisSchinnerl (changes requested):

api/admin/admin_test.go
peterjan: `cs.WantedContracts = frand.Uint64n(1e3)` can be 0 - that's my luck for ya

api/admin/admin.go
peterjan: I know it's a super set but I didn't see any harm in doing two loops that call `PruneSlabs` just to be on the safe side. The second pass would exit out early for most accounts. 

I updated though and got rid of the first loop. I'm not sure what you meant with not needing to return account IDs alongside object IDs because `DeleteObject` requires both? Since it's a debug endpoint I wanted to reuse as many methods as possible as well as keep the methods very simple and straightforward (I briefly considered a custom endpoint called `SlabReferences` or something with a more complex return type).

Updated the code and got rid of the first loop to prune slabs.

api/admin/admin.go
```diff
@@ -287,6 +292,7 @@ func NewAPI(chain ChainManager, accounts Accounts, contracts ContractManager, ho
 	// debug endpoints
 	if a.debug {
 		routes["GET /debug/pprof/:handler"] = a.handleGETPProf
+		routes["DELETE /debug/slab/:slabid"] = a.handleDELETESlab
```
ChrisSchinnerl, line 293: Let's break this up into 2 endpoints.

1. `/debug/slab/:slabid` to delete a slab and its objects
2. `/debug/accounts/prune` to fetch all accounts 

Then we don't need any new store methods apart from `ObjectsForSlab`. 

I tried this out on Zeus and the problem is that pruning is too slow to perform after every deletion. Deleting an object takes about 2-5ms and then we block for 30 seconds on pruning. 
Better to delete all objects we want to delete and then at the end prune all accounts once.

ChrisSchinnerl (changes requested):

api/admin/admin.go
peterjan: You mention "fetch all accounts" but later specify it should prune slabs for all accounts. Considering it takes such a long time and we already have `POST   /account/:accountkey/slabs/prune`. What if we have `/debug/slab/:slabid` returns the "affected" accounts instead so we can call the prune endpoint in a loop? I can see the benefit of having a prune-all endpoint though.. it will probably take a really long time and it won't prune accounts in any particular order so it'll take a long time. 

In that case maybe `POST /debug/slabs/prune` is a better endpoint. That is more consistent with the app API and covers the load better. We could rename PruneSlabs to PruneSlabsForAccount and add a PruneSlabs alonside it that does it across all accounts.

I think the latter is the best approach? So keep 1. but change 2 with `POST /debug/slabs/prune` with a custom store method that does everything in one go. That'll probably take a while, but it's a debug endpoint so that's fine (?)

peterjan: Discussed in DM - will update

ChrisSchinnerl (approved):

PR #888, opened 2026-03-23
Fix existingSlab var in PinSlabs never being used

### LINKED ISSUE #887: Fix existingSlab var in PinSlabs never being used
Right now `existingSlab` is always false since it is never updated after declaring it. This will always lead to the number of slabs being incremented even if it isn't actually a new slab.

We should first write a regression test to confirm this suspicion and then fix the actual bug.

chris124567:
Close #887

Removed `uploaded_at` assertions from `TestPinSlabs` that were only passing due to the bug causing sectors to be unnecessarily re-inserted for existing slabs.

### DIFF
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -270,8 +270,8 @@ func (s *Store) PinSlabs(account proto.Account, nextIntegrityCheck time.Time, to
 			INSERT INTO slabs (digest, encryption_key, min_shards)
 			VALUES ($1, $2, $3)
 			ON CONFLICT (digest) DO UPDATE SET pinned_at = NOW()
-			RETURNING id
-			`, sqlHash256(digest), sqlHash256(slab.EncryptionKey), slab.MinShards).Scan(&slabID)
+			RETURNING id, (xmax <> 0)
+			`, sqlHash256(digest), sqlHash256(slab.EncryptionKey), slab.MinShards).Scan(&slabID, &existingSlab)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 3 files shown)


persist/postgres/sectors.go
```diff
@@ -270,8 +270,8 @@ func (s *Store) PinSlabs(account proto.Account, nextIntegrityCheck time.Time, to
 			INSERT INTO slabs (digest, encryption_key, min_shards)
 			VALUES ($1, $2, $3)
 			ON CONFLICT (digest) DO UPDATE SET pinned_at = NOW()
-			RETURNING id
-			`, sqlHash256(digest), sqlHash256(slab.EncryptionKey), slab.MinShards).Scan(&slabID)
+			RETURNING id, (xmax <> 0)
```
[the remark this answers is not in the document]
chris124567: We are only scanning into two variables and the variable is called existingSlab I disagree

persist/postgres/sectors.go
ChrisSchinnerl, line 296: Usually I'd ask for a migration to recompute the stats since there is a chance they are inconsistent now (not just the number of slabs but also the other stats we set further down) but since we got another stat related PR open we can do that there as well.

ChrisSchinnerl (dismissed):

persist/postgres/sectors.go
peterjan: I'd do it here, the other stats PR might take a long time and it's better to keep the fix + reset in a single merge commit.

ChrisSchinnerl (approved):

peterjan (approved):
I would've added a migration that covers all sectors stats, but since Chris has approved it he probably doesn't mind so that works for me.

PR #892, opened 2026-03-24
Update integrity checks

### LINKED ISSUE #883: Set deadline for `performIntegrityChecksForHost`
Right now in `performIntegrityChecks` we process integrity checks in batches of 100 hosts at a time. Then we launch 50 goroutines that run `performIntegrityChecksForHost`.  However, we effectively pass the background context to each host.

The problem with that is we have hosts that time out. Even if each sector to check times out after the default stream timeout of 2 minutes or so, then 1 out of 100 hosts can prevent the next batch of 100 hosts from being fetched and checked for pretty much forever.

The solution is to create a context with a timeout of 1 minute or so and pass that to `performIntegrityChecksForHost`. That way a slow host will be interrupted and probably resumed in the next batch but at least we fetch 99 new hosts to check too.

NOTE: The tricky part here is to update `VerifySectors` to correctly handle the case when `v.hosts.ReadSector` returns a `context.DeadlineExceeded` and extend the tests to verify that we can gracefully interrupt integrity checks.

peterjan:
Deployed since yesterday, seems to be showing consistent integrity check speeds:

Screenshot 2026-03-31 at 08 47 46

Closes https://github.com/SiaFoundation/indexd/issues/883


slabs/manager.go
```diff
@@ -222,10 +241,14 @@ func newSlabManager(chain ChainManager, am AccountManager, cm ContractManager, h
 	for _, opt := range opts {
 		opt(m)
 	}
-	m.verifier = NewSectorVerifier(am, hosts, integrityAccount, m.log)
+	var err error
+	m.verifier, err = NewSectorVerifier(am, hosts, integrityAccount, 5*time.Minute, m.log)
+	if err != nil {
```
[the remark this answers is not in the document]
peterjan: I think 1 minute is a very low number to verify 1000 sectors. I considered 3 but decided to make a round number... we don't really care that much if we wait on a single host 3 or 5 minutes because we now have verifier workers that are working on other hosts and we 're no longer stalled.

slabs/manager.go
```diff
@@ -147,6 +149,20 @@ func WithIntegrityCheckIntervals(success, failure time.Duration) Option {
 	}
 }
 
+// WithNumIntegrityCheckGoroutines sets the number of hosts to check in
+// parallel during integrity checks. This directly impacts the number of
+// concurrent ReadSector RPCs performed when verifying sector integrity.
+//
+// The default is runtime.NumCPU().
```
chris124567: nit: default is 50

slabs/sectorverifier_test.go
```diff
@@ -124,4 +128,26 @@ func TestSectorVerifier(t *testing.T) {
...
+
+	// case 7: interruption via deadline exceeded on first sector
+	updateBalance(types.Siacoins(10))
+	client.integrityErrors[roots[0]] = context.DeadlineExceeded // deadline fires immediately
+	assertResults(roots[:1], nil, context.DeadlineExceeded)
+
+	// case 8: verify timeout fires on a slow host
+	updateBalance(types.Siacoins(10))
+	client.integrityErrors = make(map[types.Hash256]error)
+	client.slowHosts[hostKey.PublicKey()] = 2 * time.Second
```
chris124567: nit: use setSlowHost

slabs/manager.go
peterjan: stupid copilot auto impl made me change it, thanks for catching that

slabs/sectorverifier.go
```diff
@@ -89,19 +95,21 @@ func (v *SectorVerifier) VerifySectors(ctx context.Context, hostKey types.Public
 		return nil, fmt.Errorf("failed to get service account balance: %w", err)
 	}
 
+	// apply the verify timeout to the entire host interaction to prevent
+	// slow hosts from blocking indefinitely
+	verifyCtx, verifyCancel := context.WithTimeout(ctx, v.verifyTimeout)
```
ChrisSchinnerl: I think we need to use `WithTimeoutCause` here and pass `ErrAbortedRPC`.
Then `ctx.Err()` should still return `DeadlineExceeded` but `ctx.Cause` should correctly return the cause which prevents punishing the host for the failed RPC when it was us interrupting it. 

I'd also add the timeout on the context you pass in to `VerifySectors` rather than in here. Then you don't need to specify `verifyTimeout` as a field on the `SectorVerifier`.

slabs/sectorverifier.go
```diff
@@ -89,19 +95,21 @@ func (v *SectorVerifier) VerifySectors(ctx context.Context, hostKey types.Public
...
+	defer verifyCancel()
```
ChrisSchinnerl: Pretty sure that you don't need a separate name for this context because `UpdateBalance` and `ResetBalance` probably don't make use of their contexts and can be removed.

.gitignore
```diff
@@ -1,4 +1,5 @@
 bin/
+ts/
```
ChrisSchinnerl: what is this for?

ChrisSchinnerl (changes requested):

.gitignore
peterjan: It's funny that you ask that. I don't know is the answer, sometimes it pops up and appears in my unstaged files/dirs. I'll remove it.

slabs/sectorverifier.go
peterjan: We should have a static linter that checks for unused ctx, this had a big effect on how I implemented this PR. This makes it so the timeout shouldn't be on the verifier at all.

slabs/sectorverifier.go
peterjan: > I think we need to use WithTimeoutCause here and pass ErrAbortedRPC.
Then ctx.Err() should still return DeadlineExceeded but ctx.Cause should correctly return the cause which prevents punishing the host for the failed RPC when it was us interrupting it.

OK 

> I'd also add the timeout on the context you pass in to VerifySectors rather than in here. Then you don't need to specify verifyTimeout as a field on the SectorVerifier.

I don't think that's a good idea, we don't want to apply the timeout to `v.UpdateBalance` and `v.ResetBalance`

edit: ... ok the contexts are not used, I'll update it

slabs/sectorverifier.go
```diff
@@ -115,14 +112,14 @@ func (v *SectorVerifier) VerifySectors(ctx context.Context, hostKey types.Public
...
-			return nil, fmt.Errorf("failed to update service account balance: %w", err)
+		if err := v.UpdateBalance(hostKey, usage); err != nil {
+			return results, fmt.Errorf("failed to update service account balance: %w", err)
 		}
 
 		// check a random segment of the sector
 		segment := frand.Uint64n(proto.LeavesPerSector)
 		_, err = v.hosts.ReadSector(ctx, v.serviceAccount, hostKey, root, io.Discard, segment*proto.LeafSize, proto.LeafSize)
-		if errors.Is(err, context.Canceled) || errors.Is(err, mux.ErrClosedStream) {
+		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) || errors.Is(err, mux.ErrClosedStream) {
```
ChrisSchinnerl: Change this to `if err != nil && ctx.Err() != nil`. That's a better check since it will catch any interruption.

While deploying this branch on Zeus I noticed that this is actually failing with
```
indexd-1  | 2026-03-27T09:45:40Z	DEBUG	slabs.verifier	failed to verify sector	{"hostKey": "ed25519:019bb6d6a6c62f10c9f8b95f680ba2dc77766022725e02306274b19f7e430297", "sector": "043b8d49a38b35305b4bc1ba5c21b22d4a8abe4aa8d4d6028db9e4b910ea3bf2", "error": "failed to write request: i/o timeout"}
```

So interrupted reads will fail sector verification right now.

slabs/dataintegrity.go
```diff
@@ -40,8 +39,11 @@ func (m *SlabManager) performIntegrityChecksForHost(ctx context.Context, hostKey
 				break
 			}
 
-			batch, err := m.verifier.VerifySectors(ctx, hostKey, toCheck[len(results):])
-			if errors.Is(err, context.Canceled) || errors.Is(err, mux.ErrClosedStream) || errors.Is(err, errInsufficientServiceAccountBalance) || errors.Is(err, errHostUnreachable) {
+			verifyCtx, cancel := context.WithTimeoutCause(ctx, m.verifyTimeout, client.ErrAbortedRPC)
```
ChrisSchinnerl: I noticed that `rpcFn` doesn't always correctly decorate the error with `ErrAbortedRPC` in `rpcFn`.

So you'll also need to update `rpcFn` to have the following:
```
	// decorate the error with the context error to indicate whether the failed
	// rpc is a consequence of context cancellation or timeout
	if err != nil && context.Cause(ctx) != nil {
		err = fmt.Errorf("%w: %w", err, context.Cause(ctx))
	}
```

rather than checking for the graceful stream closure specifically

ChrisSchinnerl (changes requested):
Right now we are incorrectly considering sectors as failed. I left 2 comments to fix this.

slabs/dataintegrity.go
peterjan: nice find 👀

slabs/sectorverifier.go
peterjan: @ChrisSchinnerl I updated the PR but updated it to `if errors.Is(err, mux.ErrClosedStream) || (err != nil && ctx.Err() != nil)`, don't we have to check `mux.ErrClosedStream` as well or is it ok to accept that as a failure?

slabs/dataintegrity.go
```diff
@@ -40,8 +39,11 @@ func (m *SlabManager) performIntegrityChecksForHost(ctx context.Context, hostKey
...
+			batch, err := m.verifier.VerifySectors(verifyCtx, hostKey, toCheck[len(results):])
+			cancel()
+			results = append(results, batch...)
+			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) || errors.Is(err, mux.ErrClosedStream) || errors.Is(err, errInsufficientServiceAccountBalance) || errors.Is(err, errHostUnreachable) {
```
ChrisSchinnerl: Same issue here. We should change this to 
```
			if (err != nil && verifyCtx.Err() != nil) || errors.Is(err, errInsufficientServiceAccountBalance) || errors.Is(err, errHostUnreachable) {
```

slabs/sectorverifier.go
ChrisSchinnerl: We shouldn't need to check for `mux.ErrClosedStream` because whenever that error is returned `ctx.Err() != nil` is also true. 
If not, then reading the sector timed out on the stream timeout or something which also counts as a failure to retrieve the sector rather than us interrupting.

ChrisSchinnerl (changes requested):

slabs/dataintegrity.go
peterjan: Yeah I was actually just about to push a commit for this, but I added a manual `errors.Is(err, client.ErrAbortedRPC) ` instead of checking the verify ctx error

ChrisSchinnerl: It's safer to check for any err on the context since that will also cover stuff like failures due to shutting down or any other changes we might introduce in the future which cancel the context.

peterjan: Yeah I was unclear, what I meant to say was "ha that's funny I was just about to change that to something else but your suggestion is safer"

slabs/sectorverifier.go
peterjan: Hm ok, I still kind of feel we should return `errHostUnreachable` in the verifier and not consider this a failed sector because it's somewhat inconsistent with how we handle `hosts.Prices`. A stream timeout or even the host reloading/hanging up is not the same as it not being able to verify it has the sector.

ChrisSchinnerl: It is inconsistent by design. Fetching prices is something the host can't really infer an integrity check from but if a host can always just force a timeout on a stream whenever they have lost a sector to avoid punishment, they can bypass our integrity checks. Especially if it's obviously an integrity check due to us requesting exactly a leaf of data.

slabs/manager.go
```diff
@@ -295,35 +325,43 @@ func (m *SlabManager) performIntegrityChecks(ctx context.Context) error {
...
-				defer func() {
-					<-sem
-					wg.Done()
-				}()
-				m.performIntegrityChecksForHost(ctx, hostKey, logger)
-			}(host)
 		}
-		wg.Wait()
+		if len(batch) < batchSize {
+			break
```
ChrisSchinnerl, line 359: This is still not quite ideal. Because what we want is to continuously check 50 hosts in parallel until there are no more hosts with sectors to check.

But if we e.g. fetch 51 hosts then the first 50 run for 3 minutes, then the last one runs for 3 minutes and then we exit. Then we are idle for 4 minutes until a total of 10 minutes have passed for the next health check interval to start. However all of the 51 hosts we just went through probably have more sectors to check anyway. 

Instead we should rotate through all hosts that have sectors to check. Give each one 3 minutes or so and once we are through all of them start from the beginning. Keeping our 50 workers saturated all the time.

That also involves resetting `start` when we can't keep the workers saturated to see if we can fill the batch again.

ChrisSchinnerl (changes requested):

slabs/manager.go
peterjan: Did you notice we have 50 workers on a buffered channel? We fetch hosts in batches of 100, and feed them to the workers. The verify timeout is not applied on the host, but on a sector batch (this was not ideal). 

I updated the code to do the following: we apply a 5 minute timeout on a per-host basis, so every host gets the same budget. If sectors run out, the worker is freed and starts working on the next host. Sectors that weren't verified in this round, get picked up in the next round. 

I lowered the health check interval from 10 minutes to 1 minute. That means there's no long idle time on the tail-end, when we're checking the last (not-full) batch of hosts. We use this interval for integrity checks, as well as migrations. One minute is on the low side, because if the migration queue is empty we'd be querying the database for unhealthy slabs every minute... but maybe that's fine? 

If we apply the timeout on a per host basis, and we have 50 workers processing the hosts and the interval is lowered to one minute, we can avoid the re-feeding logic. I don't see the upside in resetting `start`, that timestamp is there to make sure consecutive calls don't return the same hosts but that's about it.

ChrisSchinnerl: If you look at the dashboard you'll see (set it to last 3 days) that I deployed this branch on Friday. It started strong similar to master branch at 7Gbps and then continuously dropped.

The following log line was printed once after I redeployed your latest changes today. So the loop didn't exit once since then.
```
{"level":"debug","ts":"2026-03-30T08:20:15Z","logger":"slabs.integrity","caller":"slabs/manager.go:367","msg":"finished integrity checks","elapsed":230738.827848183}
```

The slowdown probably stems from running through fast hosts and then being stuck on more than 100 slower hosts which get checked repeatedly since they never finish.

peterjan: Can you deploy the last commit? That commit changes everything. I made a mistake in putting a time bound on a sector batch, instead of applying it on a per host basis. I don't see an issue with the current implementation, there's 50 workers that are continuously working, until we get through the full batch, and then we start over.

client/v2/client.go
```diff
@@ -191,6 +191,12 @@ func (c *Client) rpcFn(ctx context.Context, hostKey types.PublicKey, fn func(ctx
 		if err != nil {
 			c.hosts.AddFailedRPC(hostKey, err)
 		}
+
+		// decorate the error with the context error to indicate whether the failed
+		// rpc is a consequence of context cancellation or timeout
+		if err != nil && context.Cause(ctx) != nil {
```
chris124567: Could we wrap the error before we call `AddFailedRPC`?  Then we wouldn't need the special case for `ErrClosedStream`

peterjan: Hm, I guess that's safe. This error handling is quite brittle :/

client/v2/client.go
```diff
@@ -191,12 +191,6 @@ func (c *Client) rpcFn(ctx context.Context, hostKey types.PublicKey, fn func(ctx
 		if err != nil {
 			c.hosts.AddFailedRPC(hostKey, err)
 		}
-
-		// decorate the error with the context error to indicate whether the failed
-		// rpc is a consequence of context cancellation or timeout
-		if err != nil && context.Cause(ctx) != nil {
```
ChrisSchinnerl: Removing this from the defer means that `c.hostTransport` is no longer covered. So we no longer punish a host for being unable to dial them.

ChrisSchinnerl (changes requested):
Up until commit [149c217](https://github.com/SiaFoundation/indexd/pull/892/commits/149c21760b8c4cea887d7cadfe124b655ed6bce4) this has been running successfully over night.

client/v2/client.go
chris124567: I meant still do it in the defer function but before we call AddFailedRPC

client/v2/client.go
peterjan: I'm an idiot.

client/v2/client.go
```diff
@@ -189,6 +189,11 @@ func (c *Client) rpcFn(ctx context.Context, hostKey types.PublicKey, fn func(ctx
 	defer func() {
 		// increment the failed RPC count for the host if the RPC failed
 		if err != nil {
+			// decorate the error with the context error to indicate whether the failed
+			// rpc is a consequence of context cancellation or timeout
+			if err != nil && context.Cause(ctx) != nil {
```
chris124567: nit: we don't need to check that `err` is nil again

chris124567 (dismissed):

ChrisSchinnerl (approved):

PR #894, opened 2026-03-24
Fix TestE2E NDF

### LINKED ISSUE #890: Fix TestE2E NDF
https://github.com/SiaFoundation/indexd/actions/runs/23454121263/job/68238686845?pr=889

chris124567:
Close #890

Seems the default timeout is too low

### DIFF
--- a/sdk/client_test.go
+++ b/sdk/client_test.go
@@ -286,7 +286,7 @@ func TestE2E(t *testing.T) {
 	// regular object upload
 	data := frand.Bytes(4096)
 	obj := NewEmptyObject()
-	err = client.Upload(t.Context(), &obj, bytes.NewReader(data), WithRedundancy(4, 11))
+	err = client.Upload(t.Context(), &obj, bytes.NewReader(data), WithRedundancy(4, 11), WithUploadHostTimeout(time.Minute))
 	if err != nil {
 		t.Fatalf("failed to upload: %v", err)
 	} else if _, err := client.Object(t.Context(), obj.ID()); err == nil || !strings.Contains(err.Error(), slabs.ErrObjectNotFound.Error()) {
@@ -295,7 +295,7 @@ func TestE2E(t *testing.T) {
 	assertShareable(obj, data)
 
 	// packed upload
-	packed, err := client.UploadPacked(WithRedundancy(4, 11))
+	packed, err := client.UploadPacked(WithRedundancy(4, 11), WithUploadHostTimeout(time.Minute))
 	if err != nil {
 		t.Fatalf("failed to create packed upload: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


sdk/client_test.go
```diff
@@ -286,7 +286,7 @@ func TestE2E(t *testing.T) {
 	// regular object upload
 	data := frand.Bytes(4096)
 	obj := NewEmptyObject()
-	err = client.Upload(t.Context(), &obj, bytes.NewReader(data), WithRedundancy(4, 11))
+	err = client.Upload(t.Context(), &obj, bytes.NewReader(data), WithRedundancy(4, 11), WithUploadHostTimeout(time.Minute))
```
peterjan: I think we should consider increasing the default timeout to 30s. I have been playing around with the SDK and passing a larger timeout is one of the first things I do when I use the SDK.

peterjan:
> Seems the default timeout is too low

Did you find out that increasing the timeout gets rid of the NDF? Or did you actually find errors indicating that the upload timeout is the culprit? It's a little awkward to increase the upload timeout to avoid an NDF in this one test and consider that to be the right fix. The upload timeout we have now really should not be timing out on the CI right.

peterjan:
> > Seems the default timeout is too low
> 
> Did you find out that increasing the timeout gets rid of the NDF? Or did you actually find errors indicating that the upload timeout is the culprit? It's a little awkward to increase the upload timeout to avoid an NDF in this one test and consider that to be the right fix. The upload timeout we have now really should not be timing out on the CI right.

Hm ok so I was a little apprehensive because even though NDFs do pop up all of a sudden, this one is getting close to being a deterministic failure almost if you run it on the CI... so that's probably because we recently changed something that led up to this. I think that happened in this PR https://github.com/SiaFoundation/indexd/pull/866

Instead of increasing the default timeout I would:
- add more hosts to give some margin, now we have exactly enough hosts I believe
- reduce redundancy 
- increase timeout

I'd probably go for a combination of 1 and 3. I also think that we should increase the default timeout to 30s, but have a default options for testing which I'd set at 5 or 8 seconds, but not that much more.

sdk/client_test.go
ChrisSchinnerl: Were you able to figure out why it is too low? This is a local test. The upload should be done in milliseconds and not require a minute per host. To me it seems like something else might be going on here that is causing an unexpected delay.

chris124567: I was unable to reproduce it locally despite running hundreds of times (with and without -race...) so I went based off the log from GitHub Actions and there was a line that said "failed to upgrade connection: could not read peer version: EOF" so I assumed it was due to the timeout.  I think the issue is caused by the GitHub Actions runner being slow.  I could probably lower the timeout by a lot.

peterjan (approved):
I would rather bump the default timeout to 30s. This will get tackled soon anyway because we have to make the upload options consistent with the Rust SDK, so probably fine for now.

ChrisSchinnerl (approved):

PR #896, opened 2026-03-25
Return the lower of the account and quota limit in the app api's `AccountResponse`

### LINKED ISSUE #895: Return the lower of the account and quota limit in the app api's `AccountResponse`
Apps can currently get confused about hitting quota limits when the account limit is still not reached. To fix this, `AccountResponse` which is returned in `handleGETAccount` should use the lower of the two limits (account and quota). Because realistically the app doesn't care about quota vs account limit. It just needs to know how much it can store.

To efficiently retrieve that information (right now it takes 3 queries with the methods we have) from the database I'd update the `Account` type to also return the quota limit as a separate field. That way we get the full picture from the admin API and a simplified few in the app API through the `AccountResponse`.

chris124567:
Close #895

### DIFF
--- a/accounts/account.go
+++ b/accounts/account.go
@@ -72,7 +72,6 @@ type (
 		MaxPinnedData        uint64        `json:"maxPinnedData"`
 		QuotaMaxPinnedData   uint64        `json:"quotaMaxPinnedData"`
 		ConnectKeyPinnedData uint64        `json:"connectKeyPinnedData"`
-		RemainingStorage     uint64        `json:"remainingStorage"`
 		Ready                bool          `json:"ready"`
 		PinnedData           uint64        `json:"pinnedData"`
 		PinnedSize           uint64        `json:"pinnedSize"`
--- a/accounts/account.go
+++ b/accounts/account.go
@@ -67,14 +67,17 @@ func WithConnectKey(connectKey string) QueryAccountsOpt {
 type (
 	// Account represents an account in the indexer.
 	Account struct {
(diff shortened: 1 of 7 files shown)


peterjan (dismissed):

accounts/account.go
```diff
@@ -67,14 +67,17 @@ func WithConnectKey(connectKey string) QueryAccountsOpt {
...
-		PinnedData    uint64        `json:"pinnedData"`
-		PinnedSize    uint64        `json:"pinnedSize"`
-		App           AppMeta       `json:"app"`
-		LastUsed      time.Time     `json:"lastUsed"`
+		AccountKey           proto.Account `json:"accountKey"`
+		ConnectKey           string        `json:"connectKey"`
+		MaxPinnedData        uint64        `json:"maxPinnedData"`
+		QuotaMaxPinnedData   uint64        `json:"quotaMaxPinnedData"`
+		ConnectKeyPinnedData uint64        `json:"connectKeyPinnedData"`
+		RemainingStorage     uint64        `json:"remainingStorage"`
```
ChrisSchinnerl: Considering that this can be computed from the other fields I'd drop it on the `Account` type. Then compute it on the fly within `handleGETAccount` for the app api.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (dismissed):

peterjan (approved):

ChrisSchinnerl (approved):

PR #897, opened 2026-03-26
Fix stats when unpinning slabs

### LINKED ISSUE #893: Make sure `unpinSlabs` correctly updates stats
`unpinSlabs` can delete from the `sectors` table but currently doesn't seem to update the corresponding `num_pinned_sectors`, `num_unpinnable_sectors` or `num_unpinned_sectors`. Same is probably true for the per-host sector stats. 

We should write a regression test to verify this is broken, fix it and write a migration to recalculate these stats.

Also double check that all stats are updated in all the right spots.

peterjan:
Closes https://github.com/SiaFoundation/indexd/issues/893

### DIFF
--- a/persist/postgres/migrations.go
+++ b/persist/postgres/migrations.go
@@ -846,4 +846,44 @@ WHERE ack.id = sub.connect_key_id;
 		_, err := tx.Exec(ctx, `UPDATE stats SET num_slabs = (SELECT COUNT(*) FROM slabs)`)
 		return err
 	},
+	// reset stats
+	func(ctx context.Context, tx *txn, _ *zap.Logger) error {
+		if _, err := tx.Exec(ctx, `
+		WITH counts AS (
+			SELECT
+				COUNT(*) FILTER (WHERE host_id IS NOT NULL AND contract_sectors_map_id IS NOT NULL)::bigint AS pinned,
+				COUNT(*) FILTER (WHERE host_id IS NOT NULL AND contract_sectors_map_id IS NULL)::bigint     AS unpinned,
+				COUNT(*) FILTER (WHERE host_id IS NULL     AND contract_sectors_map_id IS NULL)::bigint     AS unpinnable
+			FROM sectors
+		)
+		UPDATE stats s
+		SET
+			num_pinned_sectors     = counts.pinned,
+			num_unpinned_sectors   = counts.unpinned,
+			num_unpinnable_sectors = counts.unpinnable
+		FROM counts`); err != nil {
+			return fmt.Errorf("failed to reset sector stats: %w", err)
+		}
+
+		_, err := tx.Exec(ctx, `
+			UPDATE hosts h
+			SET unpinned_sectors = COALESCE(sub.cnt, 0)
+			FROM (
+				SELECT host_id, COUNT(*) as cnt
+				FROM sectors
+				WHERE contract_sectors_map_id IS NULL AND host_id IS NOT NULL
+				GROUP BY host_id
+			) sub
+			WHERE h.id = sub.host_id;
+
+			UPDATE hosts
+			SET unpinned_sectors = 0
+			WHERE id NOT IN (
+				SELECT DISTINCT host_id
+				FROM sectors
+				WHERE contract_sectors_map_id IS NULL AND host_id IS NOT NULL
+			);
+		`)
+		return err
+	},
 }
--- a/persist/postgres/migrations.go
+++ b/persist/postgres/migrations.go
@@ -909,12 +909,14 @@ WHERE ack.id = sub.connect_key_id;
 				COUNT(*) FILTER (WHERE host_id IS NULL     AND contract_sectors_map_id IS NULL)::bigint     AS unpinnable
 			FROM sectors
 		)
-		UPDATE stats s
-		SET
-			num_pinned_sectors     = counts.pinned,
-			num_unpinned_sectors   = counts.unpinned,
-			num_unpinnable_sectors = counts.unpinnable
-		FROM counts`); err != nil {
+		UPDATE stats
+		SET stat_value = CASE stat_name
+			WHEN 'num_pinned_sectors'     THEN counts.pinned
+			WHEN 'num_unpinned_sectors'   THEN counts.unpinned
+			WHEN 'num_unpinnable_sectors' THEN counts.unpinnable
+		END
+		FROM counts
+		WHERE stat_name IN ('num_pinned_sectors', 'num_unpinned_sectors', 'num_unpinnable_sectors')`); err != nil {
 			return fmt.Errorf("failed to reset sector stats: %w", err)
 		}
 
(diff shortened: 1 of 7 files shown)


persist/postgres/migrations.go
```diff
@@ -846,4 +846,44 @@ WHERE ack.id = sub.connect_key_id;
...
 		return err
 	},
+	// reset stats
+	func(ctx context.Context, tx *txn, _ *zap.Logger) error {
+		if _, err := tx.Exec(ctx, `
+		WITH counts AS (
+			SELECT
+				COUNT(*) FILTER (WHERE host_id IS NOT NULL AND contract_sectors_map_id IS NOT NULL)::bigint AS pinned,
+				COUNT(*) FILTER (WHERE host_id IS NOT NULL AND contract_sectors_map_id IS NULL)::bigint     AS unpinned,
+				COUNT(*) FILTER (WHERE host_id IS NULL     AND contract_sectors_map_id IS NULL)::bigint     AS unpinnable
```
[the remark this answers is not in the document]
peterjan: I considered this but figured migrations can be duplicated, extracting this into a function felt a bit awkward.

persist/postgres/sectors.go
```diff
@@ -456,24 +456,80 @@ RETURNING connect_key_id`, pinnedDataDelta, pinnedSizeDelta, accountID).Scan(&co
...
+	`
+
+	var pinned, unpinned, unpinnable int64
+	var unpinnedDeltas []unpinnedDelta
+	if len(toDelete) > 0 {
+		err := tx.QueryRow(ctx, candidateSectorsCTE+`
+			SELECT
+				COUNT(*) FILTER (WHERE s.host_id IS NOT NULL AND s.contract_sectors_map_id IS NOT NULL),
+				COUNT(*) FILTER (WHERE s.host_id IS NOT NULL AND s.contract_sectors_map_id IS NULL),
+				COUNT(*) FILTER (WHERE s.host_id IS NULL AND s.contract_sectors_map_id IS NULL)
```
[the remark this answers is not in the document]
peterjan: I think fixing this would unnecessarily introduce complexity and potentially slow the query way down and cause lock contention. These races are extremely unlikely in practice because `PinSectors` and `MarkSectorsUnpinnable` by nature don't conflict with `unpinSlabs`. It's highly unlikely we'd get a full transaction committed in between these statements which are executed microseconds apart from each other.

chris124567 (changes requested):
I looked at the diff and everything seems good (though it would nice to be have before and after benchmarks).

Codex found this which was a pre existing issue:
```
High: persist/postgres/sectors.go:708, because it only reads contract_sectors_map_id and relies on persist/postgres/sectors.go:1020 to classify rows. That means every contract_sectors_map_id IS NULL row is treated as “unpinned”, even if it is actually unpinnable (host_id IS NULL) or unpinned on a different host. The follow-up adjustment then decrements hosts.unpinned_sectors on the destination host at persist/postgres/sectors.go:732, so the source host stays overstated and num_unpinnable_sectors is never decremented. In the host_id IS NULL case this can even violate the non-negative host counter check. The PR’s “double check all stats” goal is not met while this path remains.
```

I think it is a real issue because if you add this test case, the final PinSectors call will fail with `failed to update number of unpinned sectors: ERROR: new row for relation "stats" violates check constraint"stats_num_unpinned_sectors_check"`

```
diff --git a/persist/postgres/stats_test.go b/persist/postgres/stats_test.go
index a3300039..2ef19749 100644
--- a/persist/postgres/stats_test.go
+++ b/persist/postgres/stats_test.go
@@ -280,6 +280,34 @@ func TestSectorStats(t *testing.T) {
  } else if len(unpinned) != 0 {
    t.Fatalf("expected 0 unpinned sectors, got %d", len(unpinned))
  }
+
+ // pinning an unpinnable sector should decrement num_unpinnable_sectors,
+ // not num_unpinned_sectors
+ unpinnableRoot := frand.Entropy256()
+ if _, err := store.PinSlabs(account, time.Time{}, slabs.SlabPinParams{
+   EncryptionKey: frand.Entropy256(),
+   MinShards:     1,
+   Sectors:       []slabs.PinnedSector{{HostKey: hk2, Root: unpinnableRoot}},
+ }); err != nil {
+   t.Fatal(err)
+ }
+ assertStats(0, 1, 3, 6)
+
+ var unpinnableUploadedAt time.Time
+ if err := store.pool.QueryRow(t.Context(), `
+   SELECT uploaded_at FROM sectors WHERE sector_root = $1
+ `, sqlHash256(unpinnableRoot)).Scan(&unpinnableUploadedAt); err != nil {
+   t.Fatal(err)
+ }
+ if err := store.MarkSectorsUnpinnable(unpinnableUploadedAt.Add(time.Second)); err != nil {
+   t.Fatal(err)
+ }
+ assertStats(0, 0, 4, 6)
+
+ if err := store.PinSectors(fcidHK2, []types.Hash256{unpinnableRoot}); err != nil {
+   t.Fatal(err)
+ }
+ assertStats(1, 0, 3, 6) // unpinnable decremented, not unpinned
 }
 
 func TestIntegrityCheckStats(t *testing.T) {
```

persist/postgres/sectors.go
```diff
@@ -456,24 +456,80 @@ RETURNING connect_key_id`, pinnedDataDelta, pinnedSizeDelta, accountID).Scan(&co
...
-					FROM slab_sectors ss
-					WHERE ss.slab_id = ANY($1) AND NOT EXISTS (
-						SELECT 1
-						FROM slab_sectors ss2
-						WHERE ss2.sector_id = ss.sector_id AND ss2.slab_id <> ANY($1)
-					)
-				)
-				DELETE FROM sectors WHERE id IN (SELECT sector_id FROM candidate_sectors);`, toDelete)
+	batch.Queue(candidateSectorsCTE+`
+		DELETE FROM sectors
```
ChrisSchinnerl: This is now not atomic anymore. You fetch the stats, then delete but in the meantime the rows might have changed.

persist/postgres/sectors.go
ChrisSchinnerl, line 1: would be nice to have benchmarks comparing the final changes to the query to the previous version to make sure we aren't degrading too much.

ChrisSchinnerl (changes requested):

persist/postgres/sectors.go
peterjan: `master`

```
goos: darwin
goarch: arm64
cpu: Apple M5
BenchmarkUnpinSlab
BenchmarkUnpinSlab-10                678           1756165 ns/op            5657 B/op        153 allocs/op
PASS
```

`DELETE FROM ... RETURNING`

```
goos: darwin
goarch: arm64
cpu: Apple M5
BenchmarkUnpinSlab
BenchmarkUnpinSlab-10                571           1865341 ns/op            8407 B/op        282 allocs/op
```

`SELECT FOR UPDATE`

```
goos: darwin
goarch: arm64
cpu: Apple M5
BenchmarkUnpinSlab
BenchmarkUnpinSlab-10                535           2181204 ns/op            9960 B/op        341 allocs/op
PASS
```

ChrisSchinnerl (approved):

chris124567 (approved):

PR #899, opened 2026-03-27
Benchmark migrateSlab

### LINKED ISSUE #885: Benchmark `migrateSlab`
It would be great to have a benchmark for `migrateSlab` that we could run whenever we make changes to the migration code. That covers downloading a slab for migration and uploading the bad sectors. That would help highlight certain regressions we have seen in the past. Such as a successful slab download waiting for 0-30s to be unblocked by the overdrive ticker. 

The benchmark should use a regular slab manager and the production database but can mock pretty much everything else. It would also be good to use realistic redundancy settings such as 10 data shards and 20 parity shards.

Ideally the benchmark allows for configuring 3 things
1. a number of bad shards that need to be migrated
2. a number of hosts that fail immediately
3. a number of hosts that just block forever

Then we can run the benchmark for 
1. the ideal cases where all hosts are good but 1, 5, 10, 20 shards need to be migrated -> should be very fast
2. the case where a lot of the "good" shards are on bad hosts -> should also be quite fast since hosts fail immediately
3. the case where a lot of the good shards are on blocking hosts -> should take 1 ticker interval per blocking host

chris124567:
Close #885

```
goos: linux
goarch: amd64
pkg: go.sia.tech/indexd/slabs
BenchmarkMigrateSlab/bad_1_fail_0_block_0-4                    5     228333244 ns/op      18.37 MB/s
BenchmarkMigrateSlab/bad_5_fail_0_block_0-4                    4     326961290 ns/op      64.14 MB/s
BenchmarkMigrateSlab/bad_10_fail_0_block_0-4                   3     441311463 ns/op      95.04 MB/s
BenchmarkMigrateSlab/bad_20_fail_0_block_0-4                   2     654274686 ns/op     128.21 MB/s
BenchmarkMigrateSlab/bad_5_fail_5_block_0-4                    3     343928105 ns/op      60.98 MB/s
BenchmarkMigrateSlab/bad_10_fail_10_block_0-4                  3     452493084 ns/op      92.69 MB/s
BenchmarkMigrateSlab/bad_5_fail_0_block_5-4                    2     806234808 ns/op      26.01 MB/s
BenchmarkMigrateSlab/bad_10_fail_0_block_5-4                   2     941555840 ns/op      44.55 MB/s
PASS
ok      go.sia.tech/indexd/slabs    35.895s
```


slabs/manager_test.go
```diff
@@ -85,8 +86,22 @@ type testStore struct {
...
+	log *zap.Logger
+}
+
+func withStoreLogger(log *zap.Logger) func(*mockStoreOpts) {
+	return func(o *mockStoreOpts) {
+		o.log = log
+	}
+}
+
+func newMockStore(t testing.TB, opts ...func(*mockStoreOpts)) *testStore {
```
peterjan, line 99: We should rename this to `newTestStore` imo, mock store makes it sound like there's no actual database behind it.

slabs/migrations_test.go
```diff
@@ -346,9 +349,241 @@ func newTestContract(hk types.PublicKey) contracts.Contract {
...
+				}
+				for _, h := range badParityHosts {
+					if _, err := db.Exec(ctx,
+						"UPDATE contracts SET good = FALSE WHERE host_id = (SELECT id FROM hosts WHERE public_key = $1)",
+						h.PublicKey[:]); err != nil {
+						b.Fatal(err)
+					}
+				}
+				// restore stats counters modified by MigrateSector
+				if _, err := db.Exec(ctx,
```
peterjan, line 505: perhaps the test store should have resetStats or something like it, we have several locations where we have to update the counters to avoid running into database constraints when the code decrements a stat field

peterjan (approved):
Results are what you'd expect I think. Benchmark looks good, exactly as described in the issue. In production the overdrive timeout is ~30s, that seems quite low actually (shard timeout / 4)... this is probably not the PR to change that but perhaps we should consider adding upload/download stats that include overdrive statistics at one point.

ChrisSchinnerl (approved):
Lgtm. PJ has some good points but we can follow up with that in a separate PR since it affects a lot more code than just this benchmark.

PR #901, opened 2026-03-30
Retry hosts on upload

### LINKED ISSUE #900: Fix failed uploads
Need to changes to the SDK to increase usability when hosts are slow. We already have these in the Rust SDK. Currently uploads just fail with “no more hosts” instead of completing slowly. 

1.	Retry timed-out hosts — call Candidates.Retry() to requeue hosts that fail due to timeout (not permanent errors), giving other shards a chance to use them
	2.	Progressive timeout — increase the timeout on retry attempts so that slow-but-reachable hosts eventually succeed

peterjan:
Closes https://github.com/SiaFoundation/indexd/issues/900

### DIFF
--- a/sdk/mock_test.go
+++ b/sdk/mock_test.go
@@ -28,6 +28,10 @@ type mockHostDialer struct {
 	delayMu   sync.Mutex
 	slowHosts map[types.PublicKey]time.Duration
 
+	flakyMu     sync.Mutex
+	flakyHosts  map[types.PublicKey]int // fail first N writes
+	writeCounts map[types.PublicKey]int
+
 	sectorsMu   sync.Mutex
 	hostSectors map[types.PublicKey]map[types.Hash256][]byte
 }
@@ -66,6 +70,16 @@ func (m *mockHostDialer) delay(ctx context.Context, hostKey types.PublicKey) err
 func (m *mockHostDialer) WriteSector(ctx context.Context, accountKey types.PrivateKey, hostKey types.PublicKey, data []byte) (rhp.RPCWriteSectorResult, error) {
 	if _, ok := m.hosts[hostKey]; !ok {
 		panic("host not found: " + hostKey.String()) // developer error
+	}
+
+	// simulate flaky hosts that fail their first N writes
+	m.flakyMu.Lock()
+	failCount := m.flakyHosts[hostKey]
+	m.writeCounts[hostKey]++
+	writes := m.writeCounts[hostKey]
+	m.flakyMu.Unlock()
+	if failCount > 0 && writes <= failCount {
+		return rhp.RPCWriteSectorResult{}, errors.New("simulated write failure")
 	}
 
 	// simulate i/o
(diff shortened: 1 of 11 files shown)


sdk/upload.go
```diff
@@ -186,6 +193,9 @@ func runUploadWorkers(ctx context.Context, client hostClient, accountKey types.P
 					}
 					return
 				}
+
+				// requeue failed hosts so other shards can use them
+				job.hosts.Retry(host)
```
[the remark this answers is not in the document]
peterjan: The issue mentions we want to retry hosts that timed out. That's not how the Rust SDK does it though, we retry the host on all errors there it seems. Figured I'd leave this open, do we want to retry on a condition such as `errors.Is(err, context.DeadlineExceeded) && ctx.Err() == nil `? Or maybe even limit the amount of retry attempts on a per host basis so we fail eventually?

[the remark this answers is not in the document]
peterjan: I ended up retrying hosts on `context.DeadlineExceeded`...

sdk/mock_test.go
```diff
@@ -79,12 +66,22 @@ func (m *mockHostDialer) delay(ctx context.Context, hostKey types.PublicKey) err
...
 	}
 
+	// simulate flaky hosts that fail their first N writes
+	m.flakyMu.Lock()
+	failCount := m.flakyHosts[hostKey]
+	m.writeCounts[hostKey]++
+	writes := m.writeCounts[hostKey]
+	m.flakyMu.Unlock()
+	if failCount > 0 && writes <= failCount {
+		return rhp.RPCWriteSectorResult{}, errors.New("simulated write failure")
```
chris124567: I think this should be `context.DeadlineExceeded`

sdk/upload.go
```diff
@@ -186,6 +193,11 @@ func runUploadWorkers(ctx context.Context, client hostClient, accountKey types.P
 					}
 					return
 				}
+
+				// requeue timed out hosts so other shards can use them
+				if errors.Is(err, context.DeadlineExceeded) && ctx.Err() == nil {
+					job.hosts.Retry(host)
```
chris124567, line 199: Should we stop trying after a certain number of attempts on a host?

peterjan: I'm trying to be consistent with the Rust SDK but yeah I think we should do that. I think I even mentioned it in the PR description at one point. I would even put it on a somewhat low number like 2 or 3.

sdk/mock_test.go
```diff
@@ -131,10 +145,28 @@ func (m *mockHostDialer) SetSlowHosts(n int, d time.Duration) {
...
+	defer m.flakyMu.Unlock()
+
+	var set int
+	for hostKey := range maps.Keys(m.hosts) {
+		if set >= n {
+			break
+		}
+		set++
+		m.flakyHosts[hostKey] = failCount
+	}
```
ChrisSchinnerl, line 163: I'd pass a `testing.TB` here and fail if `set < n`. Otherwise this will just silently succeed even when you don't have enough hosts to set flaky.

sdk/mock_test.go
```diff
@@ -68,6 +72,16 @@ func (m *mockHostDialer) WriteSector(ctx context.Context, accountKey types.Priva
 		panic("host not found: " + hostKey.String()) // developer error
 	}
 
+	// simulate flaky hosts that fail their first N writes
+	m.flakyMu.Lock()
+	failCount := m.flakyHosts[hostKey]
+	m.writeCounts[hostKey]++
+	writes := m.writeCounts[hostKey]
+	m.flakyMu.Unlock()
+	if failCount > 0 && writes <= failCount {
```
ChrisSchinnerl: I don't think you need `writeCounts` if you just subtract from `flakyHosts` every time until you hit 0.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

chris124567 (approved):

PR #904, opened 2026-03-31
Fetch all usable hosts in SDK

ChrisSchinnerl:
We used the default limit which is 100 up until now unlike the Rust SDK which paginates using a page size of 100 until it has fetch all hosts.

### DIFF
--- a/sdk/hosts.go
+++ b/sdk/hosts.go
@@ -9,6 +9,7 @@ import (
 	"go.sia.tech/core/types"
 	"go.sia.tech/coreutils/chain"
 	"go.sia.tech/coreutils/threadgroup"
+	"go.sia.tech/indexd/api"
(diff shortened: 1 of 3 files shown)


sdk/hosts.go
```diff
@@ -74,13 +75,19 @@ func (chs *cachedHostStore) Usable(hostKey types.PublicKey) (bool, error) {
 }
 
 func (chs *cachedHostStore) updateHosts(ctx context.Context) error {
-	hostInfos, err := chs.client.Hosts(ctx, chs.appKey)
-	if err != nil {
-		return err
-	}
+	const pageSize = 100
```
peterjan, line 78: `nit` rename to `batchSize`

peterjan (approved):
`nit` no 🧪

n8mgr (approved):

PR #917, opened 2026-04-03
Lower scan frequency

peterjan:
We want to check for hosts to be scanned more often than once every hour. When you set up an indexer for the first time it looks as if hosts aren't getting scanned.


chris124567 (approved):

ChrisSchinnerl:
This was already changed here: https://github.com/SiaFoundation/indexd/pull/914

PR #922, opened 2026-04-07
Add git lfs instructions to README

ChrisSchinnerl:
Indexd makes use of git lfs. This PR updates the README with instructions related to setting it up.


peterjan (dismissed):

chris124567 (approved):
I ran these commands on a newly created Linux user and was able to build indexd

PR #930, opened 2026-04-10
Use deltas for stats to reduce contention

### LINKED ISSUE #878: Update `stats` table very slow
`Update stats` queries frequently show up in the logs as slow. Probably related to lock contention around a single row. Sometimes taking seconds. e.g.

```
indexd-1  | 2026-03-20T08:56:48Z	DEBUG	postgres.transaction	slow exec	{"id": "07667abc", "attempt": 4, "query": "UPDATE stats SET num_pinned_sectors = num_pinned_sectors + $1", "elapsed": "13.956042566s", "stack": "go.sia.tech/indexd/persist/postgres.(*txn).Exec\n\tgo.sia.tech/indexd/persist/postgres/sql.go:76\ngo.sia.tech/indexd/persist/postgres.incrementNumPinnedSectors\n\tgo.sia.tech/indexd/persist/postgres/stats.go:32\ngo.sia.tech/indexd/persist/postgres.(*Store).PinSectors.func1\n\tgo.sia.tech/indexd/persist/postgres/sectors.go:672\ngo.sia.tech/indexd/persist/postgres.(*Store).doTransaction\n\tgo.sia.tech/indexd/persist/postgres/store.go:112\ngo.sia.tech/indexd/persist/postgres.(*Store).transaction\n\tgo.sia.tech/indexd/persist/postgres/store.go:64\ngo.sia.tech/indexd/persist/postgres.(*Store).PinSectors\n\tgo.sia.tech/indexd/persist/postgres/sectors.go:638\ngo.sia.tech/indexd/contracts.(*ContractManager).pinSectors.func1\n\tgo.sia.tech/indexd/contracts/pinning.go:221\ngo.sia.tech/indexd/contracts.(*ContractManager).pinSectors\n\tgo.sia.tech/indexd/contracts/pinning.go:232\ngo.sia.tech/indexd/contracts.(*ContractManager).performSectorPinningOnHost\n\tgo.sia.tech/indexd/contracts/pinning.go:97\ngo.sia.tech/indexd/contracts.(*ContractManager).performSectorPinning.func1.2\n\tgo.sia.tech/indexd/contracts/pinning.go:54\ngo.sia.tech/indexd/hosts.(*HostManager).WithScannedHost\n\tgo.sia.tech/indexd/hosts/manager.go:382\ngo.sia.tech/indexd/contracts.(*ContractManager).performSectorPinning.func1\n\tgo.sia.tech/indexd/contracts/pinning.go:53"}
```

Considering that we want stats to always be consistent I think there are 2 things we can try.
1. using 1 row per stat similar to hosts
2. if 1. doesn't alleviate the contention we could follow up with a deltas table that we insert to and a background thread that compresses those deltas into the actual stats table. If every insert is a new row there won't be any contention.

chris124567:
Close #878


persist/postgres/migrations.go
```diff
@@ -6,4 +6,15 @@ import (
...
-var migrations = []func(context.Context, *txn, *zap.Logger) error{}
+var migrations = []func(context.Context, *txn, *zap.Logger) error{
+	func(ctx context.Context, tx *txn, log *zap.Logger) error {
+		_, err := tx.Exec(ctx, `
+			CREATE TABLE stats_delta (
+				id BIGSERIAL PRIMARY KEY,
+				stat_name TEXT NOT NULL,
+				stat_delta BIGINT NOT NULL
+			);
+		`)
```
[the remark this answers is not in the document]
chris124567: TestMigrationConsistency passes...

cmd/indexd/run.go
```diff
@@ -59,6 +60,12 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
 	}
 	defer store.Close()
 
+	sm, err := stats.NewManager(store, stats.WithLogger(log.Named("stats")))
+	if err != nil {
+		return fmt.Errorf("failed to create stats manager: %w", err)
+	}
+	defer sm.Close()
```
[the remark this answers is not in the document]
chris124567: fixed

persist/postgres/stats.go
```diff
@@ -31,12 +31,7 @@ func incrementStat(ctx context.Context, tx *txn, name string, delta int64) error
...
 		return nil
 	}
-	result, err := tx.Exec(ctx, "UPDATE stats SET stat_value = stat_value + $1 WHERE stat_name = $2", delta, name)
-	if err != nil {
-		return err
-	} else if result.RowsAffected() == 0 {
-		return fmt.Errorf("stat %q does not exist", name)
-	}
+	_, err := tx.Exec(ctx, "INSERT INTO stats_delta (stat_name, stat_delta) VALUES ($1, $2)", name, delta)
 	return err
```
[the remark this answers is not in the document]
chris124567: this is the same as the issue below

persist/postgres/init.sql
```diff
@@ -387,6 +387,12 @@ CREATE TABLE stats (
     stat_value BIGINT NOT NULL DEFAULT 0 CHECK (stat_value >= 0)
 );
 
+CREATE TABLE stats_delta (
+    id BIGSERIAL PRIMARY KEY,
+    stat_name TEXT NOT NULL,
```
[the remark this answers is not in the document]
chris124567: This seems like it would undermine the purpose of this PR because every time a delta was inserted we'd have a lookup to the stats table

persist/postgres/migrations.go
```diff
@@ -6,4 +6,15 @@ import (
 	"go.uber.org/zap"
 )
 
-var migrations = []func(context.Context, *txn, *zap.Logger) error{}
+var migrations = []func(context.Context, *txn, *zap.Logger) error{
+	func(ctx context.Context, tx *txn, log *zap.Logger) error {
+		_, err := tx.Exec(ctx, `
+			CREATE TABLE stats_delta (
```
peterjan: `nit` use multiple `stats_deltas`

testutils/indexer.go
```diff
@@ -297,6 +303,9 @@ func NewIndexer(t testing.TB, c *ConsensusNode, log *zap.Logger, opts ...Indexer
 		if err := pm.Close(); err != nil {
 			t.Errorf("failed to close pin manager: %v", err)
 		}
+		if err := sm.Close(); err != nil {
```
peterjan: `nit` we should close the pin and stats manager with `closeWithTimeout` to be consistent

persist/postgres/init.sql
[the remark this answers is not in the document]
peterjan: I wonder if it matters, because you'd lookup the parent row and indeed lock it, but considering we now only have one thread updating these stats that might be perfectly fine? If `stat_name` is properly indexed I don't think it'll be any slower or introduce any weird lock contention. Not sure if it's worth finding out though

stats/manager.go
```diff
@@ -0,0 +1,99 @@
...
+	go func() {
+		defer cancel()
+		m.flushLoop(ctx)
+	}()
+
+	return m, nil
+}
+
+// Close stops the manager.
+func (m *Manager) Close() error {
```
peterjan, line 64: I would consider flushing on close here to leave the database in a nice state? Would check with Chris here. I think it's nice if you shut down for whatever reason that you have a clean stats table and don't have to take into account the deltas.

stats/manager.go
```diff
@@ -0,0 +1,99 @@
...
+// A Manager periodically flushes stat deltas into the stats table.
+type Manager struct {
+	store Store
+	tg    *threadgroup.ThreadGroup
+	log   *zap.Logger
+}
+
+// NewManager creates a new stats Manager. Any pending deltas left over from
+// a previous run are flushed immediately before the background loop starts.
+func NewManager(store Store, opts ...Option) (*Manager, error) {
```
peterjan, line 40: Consider adding a test that covers the functionality of the manager. A little test that confirms the manager flushes on open/close, that it properly works through the entire set of deltas.

peterjan (changes requested):
Requested changes but code looks good to me 👍

peterjan (dismissed):

persist/postgres/stats.go
```diff
@@ -123,6 +118,44 @@ func initStats(ctx context.Context, tx *txn) error {
...
+// them to the stats table atomically. It returns true if there may be
+// more rows to flush.
+func (s *Store) FlushStatsDelta(limit int) (more bool, err error) {
+	err = s.transaction(func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `
+			WITH batch AS (
+				SELECT id FROM stats_deltas
+				ORDER BY id
+				LIMIT $1
+				FOR UPDATE SKIP LOCKED
```
ChrisSchinnerl: The dangerous thing here is compressing stats out of order. If one stat adds 1, the next one adds 1 and the third one adds -2 everything is fine because it ends up at 0. However if the second stat is skipped, 1-2 = -1 which might lead to an underflow which causes the `CHECK` constraint on the `stats` table to fail.

So I'd remove the `SKIP LOCKED` here to avoid this edge case and because they should never be locked anyway. After all only a single thread flushes the deltas in production.

Also make sure to leave a comment that explains that the safety of this relies on the IDs reflecting the insertion order because
1. `BIGSERIAL` never reusing IDs even when rolling back a transaction
2. Only one thread ever calls FlushStatsDelta

stats/manager.go
```diff
@@ -0,0 +1,99 @@
...
+		m.flushLoop(ctx)
+	}()
+
+	return m, nil
+}
+
+// Close stops the manager and flushes any remaining deltas.
+func (m *Manager) Close() error {
+	m.tg.Stop()
+	m.flush()
```
ChrisSchinnerl: No need to flush here because it will only lead to a potential multithreaded flush which is something we don't want.

stats/manager.go
```diff
@@ -0,0 +1,99 @@
...
+	m := &Manager{
+		store: store,
+		tg:    threadgroup.New(),
+		log:   zap.NewNop(),
+	}
+	for _, opt := range opts {
+		opt(m)
+	}
+
+	m.flush()
```
ChrisSchinnerl: This here is fine since it runs before the loop is launched. Although it isn't really necessary since it will run in 5 seconds anyway.

ChrisSchinnerl (changes requested):

api/admin/admin_test.go
```diff
@@ -1276,6 +1276,9 @@ func TestHostsStatsAPI(t *testing.T) {
 		t.Fatalf("expected 0 hosts, got %d", len(res))
 	}
 
+	if _, err := cluster.Indexer.Store().FlushStatsDelta(math.MaxInt); err != nil {
```
peterjan: This is a lot of noise to flush stats. Can we add `cluster.FlushStats(t)` perhaps?

api/admin/admin_test.go
```diff
@@ -1276,6 +1276,9 @@ func TestHostsStatsAPI(t *testing.T) {
...
+		t.Fatal(err)
+	}
 	allStats, err := admin.StatsHosts(t.Context())
```
peterjan, line 1279: We should consider opening an issue to cleanup stats a bit. The admin has `HostStats` on both the store and the hosts manager. It calls sector stats directly from the store. All account related stats need to move to the account manager and so on.

api/admin/admin_test.go
```diff
@@ -1340,6 +1343,9 @@ func TestSectorStatsAPI(t *testing.T) {
 	cluster.WaitForContracts(t)
 
 	// assert 0 slabs
+	if _, err := indexer.Store().FlushStatsDelta(math.MaxInt); err != nil {
```
peterjan: I would seriously consider making reads include the deltas. That way we'd be treating the deltas as a sort of write-ahead log, and the stats table as the checkpoint. Lots of systems work like this. 

It of course comes at a trade-off but it would:
- not get in the way with the problem we are trying to solve
- there's no stats lag, great for prometheus and for testing
- no footguns, e.g. no thinking about it when we write tests
- we also don't have to take into account the order I think, we can just SUM all deltas

Definitely run this by Chris

peterjan (changes requested):

api/admin/admin_test.go
ChrisSchinnerl: Yeah that's fine. We do trade a bit of speed when reading the stats for accuracy but considering the stats are only fetched once every 5 minutes that should be alright. 
Getting rid of all the flushing in tests would be a nice benefit.

ChrisSchinnerl (commented):
I'd implement PJ's suggestion to then get rid of all the flushing in tests. Which cleans this up pretty nicely. We just still need to make sure to test that the stats deltas do still get cleared.

api/admin/admin_test.go
chris124567: https://github.com/SiaFoundation/indexd/issues/938

persist/postgres/init.sql
```diff
@@ -387,6 +387,12 @@ CREATE TABLE stats (
     stat_value BIGINT NOT NULL DEFAULT 0 CHECK (stat_value >= 0)
 );
 
+CREATE TABLE stats_deltas (
+    id BIGSERIAL PRIMARY KEY,
+    stat_name TEXT NOT NULL REFERENCES stats(stat_name),
```
peterjan, line 392: I think this needs an index

persist/postgres/sectors_test.go
```diff
@@ -161,7 +161,7 @@ func TestMigrateSector(t *testing.T) {
 		t.Helper()
 
 		var got int64
-		err = store.pool.QueryRow(t.Context(), `SELECT stat_value FROM stats WHERE stat_name = $1`, statMigratedSectors).Scan(&got)
+		err = store.pool.QueryRow(t.Context(), `SELECT `+sqlStatValue("$1"), statMigratedSectors).Scan(&got)
```
peterjan: Consider doing 

```go
func sqlStatValue(name string) string {
	return `(SELECT stat_value + COALESCE((SELECT SUM(stat_delta) FROM stats_deltas WHERE stat_name = '` + name + `'), 0) FROM stats WHERE stat_name = '` + name + `')`
}

func sqlStatSelect(names ...string) string {
	cols := make([]string, len(names))
	for i, name := range names {
		cols[i] = sqlStatValue(name)
	}
	return "SELECT " + strings.Join(cols, ", ")
}
```

That cleans it up a little bit imo:

```go
tx.QueryRow(ctx, sqlStatSelect(statScans, statScansFailed)).Scan(&stats.TotalScans, &stats.FailedScans)
```

peterjan (changes requested):

persist/postgres/sectors_test.go
ChrisSchinnerl: At that point do you even need `sqlStatValue` anymore? Might as well inline it in `sqlStatSelect` if possible.

persist/postgres/stats.go
```diff
@@ -123,20 +124,62 @@ func initStats(ctx context.Context, tx *txn) error {
 	return err
 }
 
+// FlushStatsDelta aggregates up to limit pending stat deltas and applies
+// them to the stats table atomically. It returns true if there may be
+// more rows to flush.
+//
+// This is safe because we apply the updates in the order they were inserted
+// (we order by ID).  BIGSERIAL never reuses IDs even when a transaction is
```
ChrisSchinnerl:
```suggestion
// (we order by ID). BIGSERIAL never reuses IDs even when a transaction is
```

persist/postgres/stats.go
```diff
@@ -123,20 +124,62 @@ func initStats(ctx context.Context, tx *txn) error {
...
+// rolled back, and only a single thread (the stats manager) ever calls
+// FlushStatsDelta.  This prevents underflows from out-of-order application.
```
ChrisSchinnerl:
```suggestion
// FlushStatsDelta. This prevents underflows from out-of-order application.
```

peterjan (approved):

ChrisSchinnerl (approved):

PR #937, opened 2026-04-15
Adjust max fund limit to exclude uploads when account remaining storage is 0

### LINKED ISSUE #933: Adjust max fund limit to exclude uploads when account remaining storage is 0
Adjust max fund limit to exclude uploads when account remaining storage is 0

chris124567:
Close #933

### DIFF
--- a/contracts/fund_test.go
+++ b/contracts/fund_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 	"time"
 
+	proto "go.sia.tech/core/rhp/v4"
 	"go.sia.tech/core/types"
 	"go.sia.tech/indexd/accounts"
 	"go.sia.tech/indexd/contracts"
@@ -19,6 +20,8 @@ const testFundTargetBytes = uint64(1 << 30) // 1 GiB
 type fundAccountsCall struct {
 	host        hosts.Host
 	contractIDs []types.FileContractID
+	accounts    []accounts.HostAccount
+	target      types.Currency
 }
 
 type accountsManagerMock struct {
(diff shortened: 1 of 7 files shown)


contracts/funding.go
```diff
@@ -86,23 +87,45 @@ OUTER:
...
-				log.Debug("not all accounts could be funded, no more contracts available", zap.String("quota", quota.Key))
-				break OUTER
+			// fund each group with the appropriate target
+			for _, batch := range []struct {
+				accs   []accounts.HostAccount
+				target types.Currency
+			}{
+				{uploadAccs, fundTarget},
+				{fullStorageAccs, readFundTarget},
+			} {
```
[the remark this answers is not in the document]
chris124567: fixed

contracts/funding.go
```diff
@@ -86,23 +87,45 @@ OUTER:
...
+				if len(batch.accs) == 0 {
+					continue
+				}
```
ChrisSchinnerl, line 110: Check if `target.IsZero()` here and then continue if it is. Then you can remove the `if fundTarget.IsZero()` check above.
Right now if the read target is 0 you will still attempt funding.

persist/postgres/accounts.go
ChrisSchinnerl, line 1: Could you extend the `postgres` package tests we have with some assertions as well?

ChrisSchinnerl (changes requested):

peterjan (approved):

contracts/fund_test.go
```diff
@@ -86,9 +89,13 @@ type accountFunderMock struct {
 func (f *accountFunderMock) FundAccounts(ctx context.Context, host hosts.Host, contractIDs []types.FileContractID, accs []accounts.HostAccount, target types.Currency, log *zap.Logger) (funded int, drained int, err error) {
 	f.mu.Lock()
 	defer f.mu.Unlock()
+	accsCopy := make([]accounts.HostAccount, len(accs))
```
ChrisSchinnerl, line 92: (nit) `slices.Clone`

ChrisSchinnerl (approved):

PR #939, opened 2026-04-15
Add consolidated metrics endpoint

n8mgr:
Adds a consolidated metrics endpoint to make scraper configuration easier.

### DIFF
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -1321,7 +1321,7 @@ func (a *admin) handleGETPrometheusMetrics(jc jape.Context) {
 	// unlike AppStats this returns the hosts package type that doesn't
 	// implement prometheus.Marshaller.
 	//
-	// TODO: fix this sillyness
+	// TODO: fix this silliness
 	hosts, err := a.hosts.Stats(jc.Request.Context(), 0, 1000)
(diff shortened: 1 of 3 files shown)


api/admin/admin.go
```diff
@@ -1250,6 +1253,104 @@ func (a *admin) handleGETStatsSectors(jc jape.Context) {
...
+
+	apps, err := a.store.AppStats(0, 1000)
+	if jc.Check("failed to retrieve app stats", err) != nil {
+		return
+	}
+
+	// unlike AppStats this returns the hosts package type that doesn't
+	// implement prometheus.Marshaller.
+	//
+	// TODO: fix this silliness
```
ChrisSchinnerl, line 1324: In addition to this should we also follow up with a PR that removes the "prometheus" query arg altogether considering that we have this endpoint now?

ChrisSchinnerl (approved):

peterjan (approved):

api/admin/admin.go
ChrisSchinnerl: Created an issue to discuss this further here: https://github.com/SiaFoundation/indexd/issues/940

PR #942, opened 2026-04-16
Call managers instead of the store from APIs

### LINKED ISSUE #938: Move stats functions to respective managers
- Admin API calls the host manager's `Stats` function but then uses `AggregatedHostStats` from the store
- Admin API calls `AccountStats`, `SectorStats`, `ContractsStats`, `ConnectKeyStats`, `AppStats` from the store

chris124567:
Close #938

### DIFF
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -1215,11 +1215,7 @@ func (a *admin) handleGETStatsApps(jc jape.Context) {
 	if jc.Check("failed to retrieve app stats", err) != nil {
 		return
 	}
-	resp := make(AppStatsResponse, 0, len(stats))
-	for _, s := range stats {
-		resp = append(resp, AppStats(s))
-	}
-	writeResponse(jc, resp)
+	writeResponse(jc, AppStatsResponse(stats))
 }
 
 func (a *admin) handleGETStatsContracts(jc jape.Context) {
@@ -1329,11 +1325,6 @@ func (a *admin) handleGETPrometheusMetrics(jc jape.Context) {
 		return
 	}
 
-	appStats := make(AppStatsResponse, 0, len(apps))
-	for _, s := range apps {
-		appStats = append(appStats, AppStats(s))
-	}
-
 	jc.ResponseWriter.Header().Set("Content-Type", "text/plain; version=0.0.4")
 	enc := prometheus.NewEncoder(jc.ResponseWriter)
 	for _, m := range []prometheus.Marshaller{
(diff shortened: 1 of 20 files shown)


peterjan (approved):
All our managers accept a context parameter but most don't do anything with it because our store methods don't take a context. We should consider removing all of those from the managers as well

api/admin/admin.go
```diff
@@ -1331,17 +1342,22 @@ func (a *admin) handleGETPrometheusMetrics(jc jape.Context) {
 		hostStats = append(hostStats, HostStats(s))
 	}
 
+	appStats := make(AppStatsResponse, 0, len(apps))
```
peterjan: this is introducing the same silliness we have with host stats (edit: I'm looking at an issue that involves this part, will loop back here with a better comment - just wanted to revoke my approval)

peterjan (changes requested):

api/admin/types.go
```diff
@@ -49,50 +39,19 @@ type (
...
 	// AppStats contains per-app statistics.
-	AppStats struct {
-		AppID      types.Hash256 `json:"appID"`
-		Name       string        `json:"name"`
-		Accounts   uint64        `json:"accounts"`
-		Active     uint64        `json:"active"`
-		PinnedData uint64        `json:"pinnedData"`
-		PinnedSize uint64        `json:"pinnedSize"`
-	}
+	AppStats accounts.AppStats
```
peterjan: I would get rid of this wrapper type. We really don't want the for loop in the prometheus metrics endpoint. Instead I'd implement `PrometheusMetric` on the `AppStatsResponse` and loop over the items and call a method that returns the prometheus metrics for that item. (see https://github.com/SiaFoundation/indexd/pull/945)

e.g. 

```go
func (a AppStatsResponse) PrometheusMetric() (metrics []prometheus.Metric) {
	for _, s := range a {
		metrics = append(metrics, appStatsMetrics(s)...)
	}
	return
}

func appStatsMetrics(a accounts.AppStats) []prometheus.Metric {
```

For me that is the cleanest approach with minimal code changes. The store is not exporting any api types, instead the dedicated packages have a `Stats` struct and the  managers expose a stats method. The api package is the only one that needs to care about prometheus and defines stats responses object that implement the marshaler interface.

ChrisSchinnerl (dismissed):
Lgtm. It doesn't fix the "silliness" as we have somehow started to call it but that's fine imo. It gets rid of the store fields and that's really the point of this PR.

ChrisSchinnerl (approved):

peterjan (approved):

PR #943, opened 2026-04-16
Don't consider hosts on "bad" QUIC ports usable

### LINKED ISSUE #941: Don't consider hosts on "bad" QUIC ports usable
Chrome and Firefox block a bunch of ports for various reasons. We've added a check to the troubleshooter, but we might as well also consider hosts on these ports unhealthy.

https://github.com/SiaFoundation/troubleshootd/commit/f741cf5971c7f8a53773f680b0d99117adbbf22e

chris124567:
Close #941

### DIFF
--- a/.changeset/dont_consider_hosts_on_bad_quic_ports_usable.md
+++ b/.changeset/dont_consider_hosts_on_bad_quic_ports_usable.md
@@ -3,7 +3,3 @@ default: patch
 ---
 
 # Don't consider hosts on "bad" QUIC ports usable
-
-#943 by @chris124567
-
-Close #941
--- a/.changeset/dont_consider_hosts_on_bad_quic_ports_usable.md
+++ b/.changeset/dont_consider_hosts_on_bad_quic_ports_usable.md
@@ -1,0 +1,9 @@
+---
+default: patch
+---
+
+# Don't consider hosts on "bad" QUIC ports usable
+
+#943 by @chris124567
+
+Close #941
--- a/hosts/manager.go
+++ b/hosts/manager.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math"
 	"net"
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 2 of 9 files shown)


persist/postgres/migrations.go
```diff
@@ -6,4 +6,23 @@ import (
 	"go.uber.org/zap"
 )
 
-var migrations = []func(context.Context, *txn, *zap.Logger) error{}
+var migrations = []func(context.Context, *txn, *zap.Logger) error{
+	func(ctx context.Context, tx *txn, log *zap.Logger) error {
+		const query = `
+DELETE FROM host_addresses
```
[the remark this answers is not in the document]
chris124567: fixed

hosts/manager.go
```diff
@@ -753,5 +776,14 @@ func validateAddress(na chain.NetAddress) error {
 	if na.Address == "" {
 		return fmt.Errorf("empty address")
 	}
+	if na.Protocol == quic.Protocol {
+		_, port, err := net.SplitHostPort(na.Address)
```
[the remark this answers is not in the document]
chris124567: fixed

hosts/manager.go
```diff
@@ -48,6 +48,29 @@ var (
...
 
 	alertStuckHostsID = alerts.RandomAlertID()
+
+	// badQUICPorts is the set of ports blocked by browsers for QUIC/WebTransport
+	// connections. Hosts announcing QUIC on these ports will be unreachable from
+	// browsers.
+	//
+	// https://fetch.spec.whatwg.org/#port-blocking
+	// https://searchfox.org/firefox-release/source/netwerk/base/nsIOService.cpp#122-206
+	badQUICPorts = map[string]bool{
```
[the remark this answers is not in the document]
chris124567: We don't want to do this because if we update this list that would mean the migration would have a different effect

persist/postgres/migrations.go
```diff
@@ -6,4 +6,23 @@ import (
...
+  AND substring(net_address from ':(\d+)$') IN (
+    '1', '7', '9', '11', '13', '15', '17', '19', '20', '21', '22', '23',
+    '25', '37', '42', '43', '53', '69', '77', '79', '87', '95', '101',
+    '102', '103', '104', '109', '110', '111', '113', '115', '117', '119',
+    '123', '135', '137', '139', '143', '161', '179', '389', '427', '465',
+    '512', '513', '514', '515', '526', '530', '531', '532', '540', '548',
+    '554', '556', '563', '587', '601', '636', '989', '990', '993', '995',
+    '1719', '1720', '1723', '2049', '3659', '4045', '4190', '5060', '5061',
+    '6000', '6566', '6665', '6666', '6667', '6668', '6669', '6679', '6697',
+    '10080'
```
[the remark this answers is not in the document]
chris124567: fixed

peterjan (dismissed):

hosts/manager.go
```diff
@@ -749,9 +773,21 @@ func isPrivateIP(ip net.IP) bool {
 func validateAddress(na chain.NetAddress) error {
 	if !(na.Protocol == siamux.Protocol || na.Protocol == quic.Protocol) {
 		return fmt.Errorf("unknown protocol %q", na.Protocol)
-	}
-	if na.Address == "" {
+	} else if na.Address == "" {
 		return fmt.Errorf("empty address")
+	} else if na.Protocol == quic.Protocol {
+		_, portStr, err := net.SplitHostPort(na.Address)
```
n8mgr: The port range validation should be done on both protocols

n8mgr (changes requested):

hosts/manager.go
chris124567: Done but what's the benefit if we can't use Siamux from a browser anyways and it could potentially result in migrations?

n8mgr: I meant the 1-65535 check

n8mgr (dismissed):

peterjan (dismissed):

.changeset/dont_consider_hosts_on_bad_quic_ports_usable.md
```diff
@@ -0,0 +1,5 @@
+---
+default: patch
+---
+
+# Don't consider hosts on "bad" QUIC ports usable
```
ChrisSchinnerl, line 5: One thing I'd like to highlight is that this PR doesn't actually cause hosts on bad ports to be unusable. It only ignores the QUIC announcement.

If that is fine @n8mgr we can merge this but if we want to make sure a host without any QUIC announcement is bad we still need to add that check in the scanning code to close this issue.

ChrisSchinnerl (changes requested):

.changeset/dont_consider_hosts_on_bad_quic_ports_usable.md
n8mgr: I think that’s fine, right? If they don’t have a valid Siamux and quic address they aren’t consider usable during scan right?

n8mgr (dismissed):

.changeset/dont_consider_hosts_on_bad_quic_ports_usable.md
chris124567: I could've sworn we only considered hosts with both quic and siamux to be usable but I looked and it seems we don't.  Maybe an issue for a future PR or I guess we could do it here.  Do we want to just add a field to usability called `ValidQUIC` or something like that?

n8mgr: A field in usability is probably the correct choice.

Dealers choice on here or a f/u. Idc.

chris124567: Will do it in a follow up

peterjan (approved):

Issue #929, opened 2026-04-10
Add option to filter contracts by id


ChrisSchinnerl:
We can already filter contracts by host public key but it would be good to also add a contract filter to quickly find specific contracts.

One small nit related to the existing behavior I have is the UI doesn't care if a public key doesn't have a proper "ed25519:"  prefix when filtering hosts or contracts. It might be a nice UX improvement to check if the user provided that and if they posted the key without a prefix implicitly add it. That's because usually when you copy a word you double click it to highlight it and then copy + paste. But the `:` causes the prefix to not be highlighted and you end up with this slightly infuriating extra step of either typing the prefix yourself after pasting or dragging the cursor to highlight the whole key.

peterjan:
Is this a UI ticket? We already allow filtering by contract IDs in the backend. I'll open a PR to implicitly handle the prefix for host keys.

ChrisSchinnerl:
> Is this a UI ticket? We already allow filtering by contract IDs in the backend. I'll open a PR to implicitly handle the prefix for host keys.

Yes, that's why it's marked with `Domain: frontend`.

peterjan:
> > Is this a UI ticket? We already allow filtering by contract IDs in the backend. I'll open a PR to implicitly handle the prefix for host keys.
> 
> Yes, that's why it's marked with `Domain: frontend`.

<img width="392" height="466" alt="Image" src="https://github.com/user-attachments/assets/020a2b7d-bed1-46e9-89e5-87a2ac8c1e11" />

not immediately obvious from the issue

ChrisSchinnerl:
I recommend using the "TODO" project board and selecting "backend"

PR #944, opened 2026-04-20
Handle host key prefix gracefully

peterjan:
References https://github.com/SiaFoundation/indexd/issues/929


n8mgr:
Why would we want to weaken that contract? This is an API. Handle it in the UI.

PR #945, opened 2026-04-20
Remove `api.HostStats`

### LINKED ISSUE #940: Remove prometheus arg and fix silliness
Related to this PR: https://github.com/SiaFoundation/indexd/pull/939

Considering that all prometheus metrics are now served from one endpoint we should be able to drop the flag we have on the native stats endpoint.

We also want to fix the silliness as indicated by the TODO.

peterjan:
Closes https://github.com/SiaFoundation/indexd/issues/940

### DIFF
--- a/api/admin/admin.go
+++ b/api/admin/admin.go
@@ -1229,12 +1229,7 @@ func (a *admin) handleGETStatsHostsDetailed(jc jape.Context) {
 	if jc.Check("failed to retrieve host stats", err) != nil {
 		return
 	}
-
-	var res []HostStats
-	for _, s := range stats {
-		res = append(res, HostStats(s))
-	}
-	writeResponse(jc, HostStatsResponse(res))
+	writeResponse(jc, HostStatsResponse(stats))
 }
 
 func (a *admin) handleGETStatsHostsAggregated(jc jape.Context) {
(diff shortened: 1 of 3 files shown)


.changeset/remove_prometheus_response_format_from_stats_endpoints.md
ChrisSchinnerl, line 1: This is `major` to make sure it's listed in the changelog as breaking.

ChrisSchinnerl (changes requested):
Looking good but before we merge this we should update the deploy repo to pull in the stats from the new route.

api/admin/admin.go
```diff
@@ -732,7 +732,7 @@ func (a *admin) handleGETState(jc jape.Context) {
 	}
 
 	ts := a.chain.TipState()
-	writeResponse(jc, State{
```
n8mgr, line 735: Why?

api/admin/admin.go
```diff
@@ -1092,7 +1092,7 @@ func (a *admin) handleGETWallet(jc jape.Context) {
 		return
 	}
 
-	writeResponse(jc, WalletResponse{
```
n8mgr, line 1095: Why?

n8mgr (changes requested):
Why are you removing the response format from the other endpoints?

chris124567 (approved):

ChrisSchinnerl (approved):

PR #951, opened 2026-04-27
Update last_used field on Account upon authentication

ChrisSchinnerl:
Previously we only updated `last_used` when pinning a slab or fetching it. The problem with that approach is that we switched over to syncing object events rather than fetching slabs as the preferred approach. So someone who uploads data and only syncs it rather than explicitly fetching slabs will see their account become inactive and accounts to no longer be funded.

I ran into the same issue with my CLI tool that doesn't even pin slabs and instead only uploads and downloads.

This PR fixes this by updating `last_used` in `HasAccount` which is called during app authentication. So as long as the app connects periodically to the indexer in any way, the account will remain active.


testutils/host.go
```diff
@@ -209,7 +209,7 @@ func (c *ConsensusNode) NewHost(t testing.TB, pk types.PrivateKey, log *zap.Logg
 	t.Cleanup(func() { rhp4Listener.Close() })
 	go siamux.Serve(rhp4Listener, rs, log)
 
-	udpAddr, err := net.ResolveUDPAddr("udp", ":0")
+	udpAddr, err := net.ResolveUDPAddr("udp", "127.0.0.1:0")
```
ChrisSchinnerl, line 212: This fixes an issue with IPv6 since it will result in a wildcard address like `[::]:12345` which can't be dialed instead of a localhost address.

chris124567 (approved):

n8mgr (approved):

PR #954, opened 2026-04-28
Don't reset transport on invalid proof error

ChrisSchinnerl:
Noticed a lot of "resetting transport" logging due to one host on Zeus returning invalid proofs.

### DIFF
--- a/client/v2/client.go
+++ b/client/v2/client.go
@@ -431,6 +431,11 @@ func shouldResetTransport(err error) bool {
 		// os.ErrDeadlineExceeded indicates that the stream hit a timeout which was set
 		// using SetDeadline. In this case, the mux is still healthy.
 		return false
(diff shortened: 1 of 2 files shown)


client/v2/client.go
```diff
@@ -431,6 +431,11 @@ func shouldResetTransport(err error) bool {
 		// os.ErrDeadlineExceeded indicates that the stream hit a timeout which was set
 		// using SetDeadline. In this case, the mux is still healthy.
 		return false
+	case errors.Is(err, rhp.ErrInvalidProof):
```
n8mgr, line 434: Shouldn’t we just make sure this isn’t returning the transport error code?

ChrisSchinnerl: How about a new error code? `ErrorCodeClientError` as opposed to `ErrorCodeHostError`? To indicate that it was caused by something on the client side?

PR #959, opened 2026-05-06
Truncate object event timestamps in database to seconds

### LINKED ISSUE #956: Truncate timestamps in database to seconds
We are running into issues with object event pagination due to microsecond precision in postgres but millisecond precision in our SDKs. Since a second precision is more than enough anyway we should:

1. Change all timestamps in the database to use second precision (especially object events)
2. Write a migration to truncates all existing timestamps to seconds

We still want to keep the timestamp type that Postgres gives us. But we could add a custom timestamp type that implements Valuer and Scanner and implicitly truncates.

chris124567:
Close #956

### DIFF
--- a/persist/postgres/init.sql
+++ b/persist/postgres/init.sql
@@ -350,7 +350,7 @@ CREATE TABLE object_events (
     object_key BYTEA NOT NULL CHECK(LENGTH(object_key) = 32), -- not a FK since deletions need to hang around
     account_id BIGINT NOT NULL REFERENCES accounts(id) ON DELETE CASCADE,
     was_deleted BOOLEAN NOT NULL, -- true if deleted, false otherwise
-    updated_at TIMESTAMP(0) WITH TIME ZONE DEFAULT date_trunc('second', NOW()), -- last time the object was created/updated/deleted
+    updated_at TIMESTAMP WITH TIME ZONE DEFAULT NOW(), -- last time the object was created/updated/deleted truncated to second precision
     PRIMARY KEY (account_id, object_key)
 );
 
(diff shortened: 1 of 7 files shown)


persist/postgres/init.sql
```diff
@@ -105,7 +105,7 @@ CREATE INDEX hosts_location_gist_idx ON hosts USING GIST (location);
 CREATE TABLE account_hosts (
     account_id INTEGER NOT NULL REFERENCES accounts(id) ON DELETE CASCADE,
     host_id INTEGER NOT NULL REFERENCES hosts(id) ON DELETE CASCADE,
-    next_fund TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    next_fund TIMESTAMP(0) WITH TIME ZONE NOT NULL DEFAULT date_trunc('second', NOW()),
```
ChrisSchinnerl: Do we even need any of the `date_trunc` calls when the type is changed to `TIMESTAMP(0)`? Because if not that would really limit the changes for this PR to `init.sql` and `migrations.go` which would be pretty fantastic.

chris124567: TIMESTAMP(0) rounds instead of truncating

ChrisSchinnerl: Hm I think that's fine? Trying to think of anything apart from maybe specific tests where that doesn't work.

chris124567: Pushed it... is a little cleaner but still not pretty

ChrisSchinnerl:
Ok yeah this is still a bit awkward. It might be best to just do it for the object events table to fix the query and leave the rest

persist/postgres/objects.go
```diff
@@ -269,7 +269,7 @@ func (s *Store) DeleteObject(account proto.Account, objectKey types.Hash256) err
 			return fmt.Errorf("failed to delete object: %w", err)
 		}
 		_, err = tx.Exec(ctx, `
-                       UPDATE object_events SET was_deleted = TRUE, updated_at = NOW()
+                       UPDATE object_events SET was_deleted = TRUE, updated_at = date_trunc('second', NOW())
```
peterjan, line 272: These don't hurt but I wonder if we need them? I'm not sure if it's really worth documenting this in these queries... I don't think we'll ever need to think about this again

peterjan (changes requested):
PR looks fine but the issue mentioned updating all timestamps, this only updates `object_events` which is likely the most important one but it would be kind of nice if all timestamps behaved the same way.

ChrisSchinnerl (approved):

PR #963, opened 2026-05-11
Use `ErrorCodeClientError` 

peterjan:
Updates shouldResetTransport to use `ErrorCodeClientError` and avoid resetting the transport for client errors.


client/v2/client.go
```diff
@@ -431,10 +431,8 @@ func shouldResetTransport(err error) bool {
 		// os.ErrDeadlineExceeded indicates that the stream hit a timeout which was set
 		// using SetDeadline. In this case, the mux is still healthy.
 		return false
-	case errors.Is(err, rhp.ErrInvalidProof):
-		// An invalid proof error indicates that the host sent us a response
-		// with an invalid proof. This error is raised on our side and therefore
-		// not transport related.
+	case proto.ErrorCode(err) == proto.ErrorCodeClientError:
```
ChrisSchinnerl: You shouldn't need to add that case since `promo.ErrCode(err) == promo.ErrorCodeTransport` will fail anyway.

ChrisSchinnerl (changes requested):

go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/oschwald/geoip2-golang v1.13.0
 	github.com/shopspring/decimal v1.4.0
 	go.sia.tech/core v0.21.0
-	go.sia.tech/coreutils v0.22.1-0.20260513143633-9856f6e895bd
+	go.sia.tech/coreutils v0.22.0
```
ChrisSchinnerl: This is a version downgrade

ChrisSchinnerl (changes requested):

go.mod
peterjan: omg

peterjan: I updated it, don't we want to keep fixed versions though in indexd and not set it to `master`

ChrisSchinnerl (approved):

Issue #971, opened 2026-05-13
Investigate migration code


ChrisSchinnerl:
SiaStorage had to migrate some data over the last couple of days. A few TB of data (~10%) of the total capacity which is not that concerning.
What was however strange is that the logs reported `toMigrate: "14"` or similar large numbers. When realistically it should be closer to 3 shards requiring migration.

So we should double check the code and figure out if
- we are migrating shards that are not bad
- we start migrating too late when the slab's health has already degraded more than we thought
- we already upload slabs with a lot of bad sectors
- something else maybe

What seemed to have triggered the migrations in the first place was 14 contracts going bad and then a bit later 13 going bad. While only 2 hosts were blocked. Indicating that those contracts belonged to 2 hosts which should only cause 1-2 shards to be migrated.

This would strongly indicate that multiple contracts with the same host hold shards of the same slab. Which could happen in multiple places. Either ~~the client already uploaded the slab that way~~ or the migration code caused it.

NOTE: We should also make sure that when `sectorsToMigrate` encounters a slab like that (with multiple shards that have the same host) it will migrate all but one of those sectors away to other hosts.

PR #972, opened 2026-05-13
Prevent pinning of slabs with duplicate hosts

ChrisSchinnerl:
Related to https://github.com/SiaFoundation/indexd/issues/971

This doesn't fix the issue but it hardens the backend against clients that might attempt to pin slabs like that.


chris124567 (approved):

chris124567 (changes requested):
Is this not sufficient: https://github.com/SiaFoundation/indexd/blob/8dab484f9e7a802df4f8e59f19aff082de8f776b/slabs/slabs.go#L150

ChrisSchinnerl:
@chris124567 nice I completely forgot about that. So we can rule out client error already.

PR #977, opened 2026-05-14
Add basic pool support

### LINKED ISSUE #962: Basic Pool Support
Instead of funding one account per connected app, we want to only link app accounts to a pool and then fund one pool per app connect key.

So what we need is:

1. A pools table to hold information about the pools we maintain.
2. A reference to a pool in the app_connect_keys table since every app gets its own pool
3. A pool_hosts table which is the same as account_hosts but for funding pools
4. A attached_pools table which references pool_hosts by pool id and host id as well as an account by account id. A row in attached_pools means that a pool has been attached to a given account.

The usual flow is then:

1. create connect key and pool for connect key
2. populate pool_hosts when asking for pools to fund for a host like we do for account_hosts
3. fund the pools which also ensures they are created
4. query attached_pools to figure out which pools still require attaching for the given host link the pools in batches and insert the linked pools in attached_pools

peterjan:
This PR adds basic pool support. 

Closes https://github.com/SiaFoundation/indexd/issues/962


persist/postgres/init.sql
```diff
@@ -112,6 +112,29 @@ CREATE TABLE account_hosts (
 CREATE INDEX account_hosts_host_id_next_fund_idx ON account_hosts (host_id, next_fund);
 CREATE INDEX account_hosts_account_id_consecutive_failed_funds_idx ON account_hosts (account_id, consecutive_failed_funds);
 
+CREATE TABLE pools (
```
peterjan, line 115: We don't really need a `pools` table considering it's 1-1 with connect keys and we don't have any fields here. I decided to add one though, for clarity and also to have one to potentially extend with a description or other fields. Wanted to point that out though

accounts/manager.go
```diff
@@ -16,14 +16,13 @@ import (
 )
 
 const (
-	// AccountFundBatch is the number of host accounts we will fund in one
-	// batch.  It is equivalent to the max batch size used in replenish RPC.
-	AccountFundBatch = proto.MaxAccountBatchSize
-	// ReadyHostThreshold is the number of successfully funded host accounts
-	// required before an account is considered ready for use.
+	// ReadyHostThreshold is the number of hosts with a funded and attached
```
peterjan: I chose to update the `ready` field to look at pools instead of accounts. I'm wondering though whether we should get rid of the field all together since pools should solve this issue all together.

n8mgr: It still requires a RPC to every host on app connect to attach the new key to the pool.

.changeset/add_basic_pool_support.md
```diff
@@ -0,0 +1,7 @@
+---
+default: patch
```
ChrisSchinnerl: This probably deserves to be major

accounts/accounts_test.go
ChrisSchinnerl, line 1: Got a few tests that are still skipped

accounts/account.go
```diff
@@ -88,18 +88,40 @@ type (
...
 
+	// PoolFundInfo contains funding info for a single pool.
+	PoolFundInfo struct {
+		FundTargetBytes uint64
+		ActiveAccounts  uint64
+	}
+
 	// QuotaFundInfo contains funding info for a quota including the number
 	// of active accounts.
 	QuotaFundInfo struct {
```
ChrisSchinnerl, line 116: This type is returned by `AccountFundingInfo` which doesn't seem to be called in prod anymore apart from testing. Which is odd since we still want to fund accounts for legacy hosts that haven't upgraded yet.

contracts/funding.go
```diff
@@ -22,10 +27,66 @@ const (
 	serviceAccountFundTargetBytes = uint64(16 << 30) // 16 GiB
 )
 
-// FundAccounts attempts to fund all accounts for the given host key. It does so
+// FundAccounts funds service accounts for the given host. User accounts draw
```
ChrisSchinnerl: We also need to fund accounts for all hosts before `5.1.0` or whatever the cutoff is.

persist/postgres/init.sql
ChrisSchinnerl: We should store the pool secret here

persist/postgres/types.go
```diff
@@ -307,6 +308,22 @@ func (pk *sqlPublicKey) Scan(src any) error {
...
+// sqlPoolKey scans a user secret and derives the pool private key.
+type sqlPoolKey types.PrivateKey
+
+func (pk *sqlPoolKey) Scan(src any) error {
+	switch src := src.(type) {
+	case []byte:
+		if len(src) != 32 {
+			return fmt.Errorf("failed to scan pool key: expected 32 bytes, got %d", len(src))
+		}
+		*pk = sqlPoolKey(keys.DerivePrivateKey(types.PrivateKey(src), "pool"))
```
ChrisSchinnerl: Let's just give every pool its own random key. This is very sneaky to have in a serialization helper and will bite us if we ever want to recycle pools because they aren't independent of users.

testutils/cluster.go
```diff
@@ -190,24 +190,33 @@ func (c *Cluster) NewHosts(t testing.TB, n int) []*Host {
...
 func (c *Cluster) WaitForAccountFunding(t *testing.T, pk proto.Account) {
 	t.Helper()
 	for _, h := range c.Hosts {
-		var funded bool
-		for i := 0; i < 120; i++ {
-			b, err := c.Indexer.Client().AccountBalance(t.Context(), h.PublicKey(), pk)
+		hk := h.PublicKey()
+		var attached bool
+		for range 120 {
+			// TODO: replace this raw query with a proper store method
```
ChrisSchinnerl: Leftover TODO

accounts/manager.go
```diff
@@ -197,16 +232,15 @@ func HostFundTarget(host hosts.Host, fundTargetBytes uint64) types.Currency {
...
-// HostFundTarget: as a combined read+write target. Since these accounts can
-// only download, we fund only the read portion of that target.
-func HostReadFundTarget(host hosts.Host, fundTargetBytes uint64) types.Currency {
-	if fundTargetBytes == 0 {
-		return types.ZeroCurrency
-	}
-	sectors := (fundTargetBytes + proto.SectorSize - 1) / proto.SectorSize
-	return host.Settings.Prices.RPCReadSectorCost(proto.SectorSize).RenterCost().Mul64(sectors).Div64(2)
+// PoolFundTarget calculates the fund target for a pool on the given host. The
+// target scales with the number of active accounts drawing from the pool,
```
ChrisSchinnerl: The target doesn't scale with the number of active accounts. We always fund the pool at the same rate. After all the point of the pools is that we can remove the limit on the apps a user can connect with all apps sharing the same rate.

contracts/funding.go
```diff
@@ -22,10 +27,66 @@ const (
...
+// active accounts, plus the service account fund target.
+func (cm *ContractManager) ContractFundTarget(ctx context.Context, host hosts.Host, minAllowance types.Currency) (types.Currency, error) {
+	poolInfos, err := cm.accounts.PoolFundingInfo()
+	if err != nil {
+		return types.ZeroCurrency, err
+	}
+
+	var target types.Currency
+	for _, pi := range poolInfos {
+		target = target.Add(accounts.PoolFundTarget(host, pi.FundTargetBytes, pi.ActiveAccounts))
```
ChrisSchinnerl: We will need to make sure the same rules apply as before but updated for pools.

1. if a user has reached their global storage limit we don't fund their pool for upload anymore
2. if a pool has 0 active accounts we don't fund it

ChrisSchinnerl (changes requested):
This PR seems to be missing proper backwards compatibility

accounts/manager.go
peterjan: I updated the code but we have to take into account the size of the pool one way or another right? Either by dynamically adjusting the pool fund interval, or by adjusting the funding amount. If a pool has 100 accounts drawing from it, we can't fund the pool at the same rate as a pool with 1000 accounts?

n8mgr: We can because the 500Mbps limit turns into a more useful and easy to understand per user limit rather than a per app limit.

contracts/maintenance.go
```diff
@@ -94,6 +52,72 @@ func (cm *ContractManager) performContractMaintenance(ctx context.Context, log *
...
+			// fetch contracts for funding
+			contractIDs, err := cm.store.ContractsForFunding(hostKey, 10)
+			if err != nil {
+				log.Error("failed to fetch contracts for funding", zap.Error(err))
+				return
+			}
+
+			// fund accounts
+			if len(contractIDs) > 0 {
+				cm.hosts.WithScannedHost(ctx, hostKey, func(host hosts.Host) error {
```
chris124567: WithScannedHost can fail too if prices are out of date and we can't scan the host so I think you should do a debug log here too

persist/postgres/migrations.go
```diff
@@ -86,4 +88,52 @@ UPDATE hosts SET has_quic = EXISTS (
...
+		}
+		defer rows.Close()
+
+		for rows.Next() {
+			var id int64
+			if err := rows.Scan(&id); err != nil {
+				return err
+			}
+			poolKey := types.GeneratePrivateKey()
+			if _, err := tx.Exec(ctx, `INSERT INTO pools (connect_key_id, pool_key) VALUES ($1, $2)`, id, []byte(poolKey)); err != nil {
```
chris124567, line 149: I remember recently with my explored postgres port that you have to read all the rows first with pgx before executing another transaction

accounts/manager.go
peterjan: Oh right that makes sense! For whatever reason I was thinking we want to make sure the pool never drains fully but we're using this as a rate limiter, kindof.

accounts/account.go
```diff
@@ -88,12 +88,35 @@ type (
 		FullStorage            bool
 	}
 
+	// HostPool represents a balance pool on a host.
+	HostPool struct {
+		ConnectKey             string
```
ChrisSchinnerl: Don't need the `ConnectKey` field on this type I think.

accounts/manager_test.go
```diff
@@ -25,37 +25,37 @@ func newTestStore(t testing.TB) testStore {
...
 
-// TestUpdateFundedAccounts is a unit test that covers the functionality of
-// updating the funded accounts. It asserts that the consecutive failed funds
+// TestUpdateFundedPools is a unit test that covers the functionality of
+// updating the funded pool. It asserts that the consecutive failed funds
 // and next fund time are updated correctly based on the number of funded
-// accounts.
-func TestUpdateFundedAccounts(t *testing.T) {
+// pools.
+func TestUpdateFundedPools(t *testing.T) {
```
ChrisSchinnerl, line 118: Why not keep both rather than not testing legacy behavior anymore and only testing the new one?

contracts/fund_test.go
```diff
@@ -205,69 +245,85 @@ func TestPerformAccountFunding(t *testing.T) {
...
+	// verify no call was made for the pool host
+	for _, call := range funderMock.calls {
+		if call.host.PublicKey == hk5 {
+			t.Fatal("pool host should not have been funded via FundAccounts")
+		}
+	}
 }
 
-func TestFundAccountsFullStorage(t *testing.T) {
+func TestPerformAccountFundingFullStorage(t *testing.T) {
```
ChrisSchinnerl, line 280: Missing a counterpart for pools?

contracts/funding.go
```diff
@@ -22,10 +27,88 @@ const (
...
+// ContractFundTarget calculates the fund target for a contract on the given
+// host. For hosts that support pools it sums the pool fund targets. For legacy
+// hosts it uses per-account targets. Service accounts are always included.
+func (cm *ContractManager) ContractFundTarget(ctx context.Context, host hosts.Host, minAllowance types.Currency) (types.Currency, error) {
+	var target types.Currency
+
+	if host.HasPoolSupport() {
+		// pool-funded host
+		threshold := time.Now().Add(-accounts.AccountActivityThreshold)
+		poolInfos, err := cm.accounts.PoolFundingInfo(threshold)
```
ChrisSchinnerl: This is a lot less efficient than `AccountFundingInfo` since it returns one value for each pool rather than a summary for each quota. Ideally both the `AccountFundingInfo` and `PoolFundingInfo` methods return a `QuotaFundInfo`. 
Maybe change `ActiveAccounts` and `FullStorageAccounts` to `Active` and `FullStorage` to be generic but then this logic cleans up a bit and is the same for both.

1 active pool == 1 active legacy account in terms of funding

ChrisSchinnerl (changes requested):

accounts/manager_test.go
peterjan: It's an oversight, I initially wanted to keep the diff as small as possible and initially was only funding service accounts.

accounts/account.go
peterjan: You are right, that's odd, I used it at one point. Maybe I didn't have the PoolKey on there at that point

contracts/fund_test.go
peterjan: I originally removed full storage on pools but then added it back later. I thought it wasn't necessary because the funding logic would be different.

persist/postgres/funding.go
```diff
@@ -0,0 +1,483 @@
...
+
+		rows, err := tx.Query(ctx, `
+SELECT a.public_key, p.pool_key
+FROM pool_hosts ph
+INNER JOIN pools p ON p.id = ph.pool_id
+INNER JOIN app_connect_keys ack ON ack.id = p.connect_key_id
+INNER JOIN hosts h ON h.id = ph.host_id
+INNER JOIN accounts a ON a.connect_key_id = ack.id AND a.deleted_at IS NULL
+LEFT JOIN pool_attachments pa ON pa.pool_id = p.id AND pa.host_id = ph.host_id AND pa.account_id = a.id
+WHERE h.public_key = $1 AND pa.pool_id IS NULL AND ph.consecutive_failed_funds = 0
```
peterjan, line 248: @ChrisSchinnerl is `ph.consecutive_failed_funds = 0` too tight here? we might be excluding funded pools currently in a backoff for whatever reason but otherwise usable

ChrisSchinnerl: Might want to track whether a pool should exist using a bool that is true when the pool has been funded at least once. 
checking against 0 might indeed exclude pools which isn't great but >0 will cause the whole rpc to fail which is also not great. Unless I'm mistaken and we don't fail the whole RPC.

peterjan: Instead of adding a field for this, can't we get away with allowing 0 or 1 consecutive failed funds? That's a little out there perhaps but have a field for this feels equally silly.

ChrisSchinnerl: I'd go with 0 for now. If we get stuck on a single host which is unlikely it shouldn't be a big deal. It will eventually create the pool and attach all accounts.

peterjan: hm, ok, I'll add a comment though because I feel this is not obvious and it has my name on it so I'm pretty sure I'll one day get the blame for this

persist/postgres/apps.go
```diff
@@ -55,7 +55,14 @@ func (s *Store) AddAppConnectKey(meta accounts.AppConnectKeyRequest) (key accoun
 			case pgerrcode.UniqueViolation:
 				return accounts.ErrKeyAlreadyExists
 			}
```
ChrisSchinnerl, line 59: Missing a `default` case here where we return an error. Right now any `PgError` that doesn't match is treated as a success.

persist/postgres/funding_test.go
```diff
@@ -0,0 +1,838 @@
...
+	// now h2 should also have one pending
+	pending, err = store.PendingPoolAttachments(hk2, 10)
+	if err != nil {
+		t.Fatal(err)
+	} else if len(pending) != 1 {
+		t.Fatal("expected one pending for h2", len(pending))
+	}
+
+	// attach on h1, h2 should be unaffected
+	if err := store.InsertPoolAttachments(hk1, pending[:0]); err != nil {
```
ChrisSchinnerl: What's the idea behind attaching an empty slice here?

ChrisSchinnerl (changes requested):

chris124567 (commented):
Nothing to add to Chris's comments other than fixing merge conflicts

chris124567 (dismissed):

peterjan:
The failing OpenAPI check is getting annoying.

ChrisSchinnerl (approved):

PR #978, opened 2026-05-14
Remove ErrAbortedRPC

ChrisSchinnerl:
With `AddFailedRPC` being exposed now for the Go SDK to use it the pattern we used with `ErrAbortedRPC` becomes more confusing. 

To clean this up, this PR removes `ErrAbortedRPC` and updates the client to only count failed RPCs if they were not caused by contexts. If they were, the client leaves it up to the caller to know what their intentions were.

### DIFF
--- a/slabs/downloads.go
+++ b/slabs/downloads.go
@@ -12,7 +12,6 @@ import (
 
 	proto "go.sia.tech/core/rhp/v4"
 	"go.sia.tech/core/types"
-	"go.sia.tech/indexd/client/v2"
 	"go.sia.tech/mux/v3"
 	"go.uber.org/zap"
 )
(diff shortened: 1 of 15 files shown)


slabs/downloads.go
```diff
@@ -103,32 +92,43 @@ func (m *SlabManager) downloadShards(ctx context.Context, slab Slab, log *zap.Lo
...
+				log.Debug("shard download failed", zap.Error(err))
+				// non-blocking send to indicate a failure
+				select {
+				case failedCh <- struct{}{}:
+				default:
+				}
+				// a host gets demoted if either
+				// 1. it hit the shard timeout
+				// 2. it was part of the initial batch of hosts and was interrupted
+				if (timeoutCtx.Err() != nil && ctx.Err() == nil) || (initial && ctx.Err() != nil) {
```
chris124567, line 112: Why should an initial host being interrupted result in it being demoted?  I guess that means it's slow but that's different than failed

ChrisSchinnerl: That only happens if it doesn't finish before the download is done. So if it is just slow it won't get demoted. If it is so slow that another worker that was launched later finished first, it will. Because we thought we selected the best hosts for the job but apparently there was a better option.

Unfortunately demoting it is our only choice for killing workers when the download finishes. We can't let goroutines leak after the download is done to get an accurate estimate of how fast it would have been and we also can't ignore it because then we would just keep using it. This is essentially what we did before by passing different contexts to the initial hosts and the racing ones but more explicit.

peterjan (approved):

PR #980, opened 2026-05-18
Update sectors to migrate

### LINKED ISSUE #971: Investigate migration code

peterjan:
In [this PR](https://github.com/SiaFoundation/indexd/pull/700) we added `ContractsForAppend` to determine whether a sector's contract is healthy. That method filters out contracts in the renew window, at max size, or low on funds. Sectors on those contracts are fine where they are but were getting flagged for migration anyway. This explains the inflated `toMigrate` counts we saw in production. This PR also tightens `PinSectors` to skip sectors that moved to a different host since they were queried.

Closes https://github.com/SiaFoundation/indexd/issues/971

### DIFF
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -700,8 +700,8 @@ func (s *Store) PinSectors(contractID types.FileContractID, roots []types.Hash25
 		rows, err := tx.Query(ctx, `
 			SELECT id, host_id, contract_sectors_map_id
 			FROM sectors
-			WHERE sector_root = ANY($1) AND (host_id IS NULL OR host_id = $3) AND (contract_sectors_map_id IS NULL OR contract_sectors_map_id != $2)
-			FOR UPDATE`, sqlRoots, contractMapID, hostID)
+			WHERE sector_root = ANY($1) AND (host_id IS NULL OR host_id = $2) AND (contract_sectors_map_id IS NULL OR contract_sectors_map_id != $3)
+			FOR UPDATE`, sqlRoots, hostID, contractMapID)
 		if err != nil {
 			return fmt.Errorf("failed to query sectors: %w", err)
 		}
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -670,10 +670,9 @@ ORDER BY ss.slab_index ASC`, slabID).Query(func(rows pgx.Rows) error {
 	return results, err
 }
 
-// PinSectors pins a batch of sector roots to a given contract. This also
-// updates the host the sector is associated with to the host that we have the
-// contract with. That way, we can avoid a race where the host changes in the
-// meantime and the contract then no longer matches the host.
+// PinSectors pins a batch of sector roots to a given contract. Only sectors
+// that are currently on the contract's host or have no host set are updated.
+// Sectors that were migrated to a different host in the meantime are skipped.
 func (s *Store) PinSectors(contractID types.FileContractID, roots []types.Hash256) error {
 	if len(roots) == 0 {
 		return nil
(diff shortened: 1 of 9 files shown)


slabs/migrations.go
```diff
@@ -139,7 +143,10 @@ func (m *SlabManager) migrateSlab(ctx context.Context, slabID SlabID, allHosts [
 // sectors that require migration together with hosts that can be used to
 // migrate bad sectors to. These hosts are guaranteed to be at least
 // minHostDistance apart from each other and are returned in random order.
-func sectorsToMigrate(slab Slab, allHosts []hosts.Host, goodContracts []contracts.Contract, minHostDistanceKm float64) ([]int, []types.PublicKey) {
+// healthyContracts is the full set of revisable good contracts used to decide
+// whether a sector's contract is healthy. migrationContracts is the narrower
+// subset eligible for uploading migrated sectors, used to select candidate hosts.
+func sectorsToMigrate(slab Slab, allHosts []hosts.Host, healthyContracts, migrationContracts []contracts.Contract, minHostDistanceKm float64) ([]int, []types.PublicKey) {
```
ChrisSchinnerl: The `migrationContracts` are the subset of `healthyContracts` for which `GoodForAppend == nil`. So you can filter them down directly in here since it's also the only place we need them. Which is kinda nice because it makes the relation between these more obvious and doesn't pollute the remaining codebase with the concept of both.

Then you can also remove `ContractsForAppend` on the `ContractManager` altogether.

persist/postgres/sectors.go
```diff
@@ -701,8 +700,8 @@ func (s *Store) PinSectors(contractID types.FileContractID, roots []types.Hash25
 		rows, err := tx.Query(ctx, `
 			SELECT id, host_id, contract_sectors_map_id
 			FROM sectors
-			WHERE sector_root = ANY($1) AND (contract_sectors_map_id IS NULL OR contract_sectors_map_id != $2)
-			FOR UPDATE`, sqlRoots, contractMapID)
+			WHERE sector_root = ANY($1) AND (host_id IS NULL OR host_id = $3) AND (contract_sectors_map_id IS NULL OR contract_sectors_map_id != $2)
```
ChrisSchinnerl: (nit) this is a simple enough query to keep the args in order. So `$1`, `$2` and `$3` and then `sqlRoots, hostID, contractMapID`.

ChrisSchinnerl (changes requested):

slabs/migrations.go
```diff
@@ -13,40 +13,58 @@ import (
...
+// migrationState holds the hosts, contracts, and chain state needed for
+// slab migrations.
+type migrationState struct {
+	Height              uint64
+	HealthyContracts    []contracts.Contract
+	Hosts               []hosts.Host
+	MaintenanceSettings contracts.MaintenanceSettings
+}
 
+// migrationState fetches all available hosts, contracts, maintenance settings,
```
chris124567: nit: `fetchMigrationState fetches...`

chris124567 (dismissed):

chris124567 (approved):

ChrisSchinnerl (approved):

PR #983, opened 2026-05-27
Rebind slabs

n8mgr:
Repinning a slab now rebinds host sectors that have been lost. Also adds a check and test that makes sure we're calling `Validate` when pinning slabs. This change requires Postgres 18 for `old` in the `RETURNING` clause.

### DIFF
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -261,7 +261,7 @@ func (s *Store) PinSlabs(account proto.Account, nextIntegrityCheck time.Time, to
 			}
 
 			digest := slab.Digest()
-			digests = append(digests, slab.Digest())
+			digests = append(digests, digest)
 
 			// insert slab
 			var slabID int64
@@ -292,17 +292,15 @@ func (s *Store) PinSlabs(account proto.Account, nextIntegrityCheck time.Time, to
 				newPinnedSize += slab.Size()
 			}
 
-			// if the slab already existed, we don't need to insert the sectors
-			if existingSlab {
-				continue
+			if !existingSlab {
+				if err := incrementNumSlabs(ctx, tx, 1); err != nil {
+					return fmt.Errorf("failed to increment number of slabs: %w", err)
+				}
 			}
 
-			// update slab stats
-			if err := incrementNumSlabs(ctx, tx, 1); err != nil {
-				return fmt.Errorf("failed to increment number of slabs: %w", err)
-			}
-
-			// insert slab's sectors in a single batch
+			// insert the slab's sectors. For a slab that already
+			// exists this may rebind any sectors that were marked
+			// lost since it was pinned.
 			batch := &pgx.Batch{}
 			for _, sector := range slab.Sectors {
 				batch.Queue(`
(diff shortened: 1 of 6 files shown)


ChrisSchinnerl (dismissed):

persist/postgres/sectors.go
```diff
@@ -292,33 +292,33 @@ func (s *Store) PinSlabs(account proto.Account, nextIntegrityCheck time.Time, to
...
 				INSERT INTO sectors (sector_root, host_id, next_integrity_check)
 				SELECT $1, h.id, $3
 				FROM hosts h
 				WHERE h.public_key = $2
-				ON CONFLICT (sector_root) DO UPDATE SET uploaded_at=NOW()
-				RETURNING id, host_id, (xmax = 0) AS inserted`,
+				ON CONFLICT (sector_root) DO UPDATE SET
+					uploaded_at = NOW(),
+					host_id = COALESCE(sectors.host_id, EXCLUDED.host_id)
+				RETURNING id, host_id, (OLD.id IS NULL) AS inserted, (OLD.id IS NOT NULL AND OLD.host_id IS NULL) AS rebound`,
```
peterjan, line 314: If this requires Postgres 18 we should add that to the readme and update the docker files in this repo to specify the version.

persist/postgres/sectors.go
```diff
@@ -292,33 +292,33 @@ func (s *Store) PinSlabs(account proto.Account, nextIntegrityCheck time.Time, to
...
 			for _, sector := range slab.Sectors {
 				batch.Queue(`
 				INSERT INTO sectors (sector_root, host_id, next_integrity_check)
 				SELECT $1, h.id, $3
 				FROM hosts h
 				WHERE h.public_key = $2
-				ON CONFLICT (sector_root) DO UPDATE SET uploaded_at=NOW()
-				RETURNING id, host_id, (xmax = 0) AS inserted`,
+				ON CONFLICT (sector_root) DO UPDATE SET
+					uploaded_at = NOW(),
```
peterjan, line 312: Are we sure we want to unconditionally update uploaded_at here? That will push the sector to the end of the unpinned sectors queue and also postpone marking it as unpinnable.

persist/postgres/sectors_test.go
```diff
@@ -1145,6 +1145,92 @@ func TestPinSlabsDuplicate(t *testing.T) {
 	assertNumSlabs(1)
 }
 
+func TestPinSlabsRebindLostSector(t *testing.T) {
```
peterjan, line 1148: If you decide to update the above remarks I'd extend the test here to check that healthy sectors weren't updated.

peterjan (changes requested):

persist/postgres/sectors.go
n8mgr: If we don’t push it to the back, a malicious user could use that to upload missing sectors and clog our pinning queue.

peterjan (approved):

peterjan (approved):

ChrisSchinnerl (approved):

PR #989, opened 2026-06-02
pull unhealthy slab in separate goroutine

ChrisSchinnerl:
This PR prefetches more slabs for repair from a separate goroutine to parallelize fetching slabs and migrating them. Right now we run into gaps because the workers migrate faster than we can migrate slabs.

### DIFF
--- a/slabs/manager.go
+++ b/slabs/manager.go
@@ -383,12 +383,17 @@ func (m *SlabManager) performSlabMigrations(ctx context.Context) error {
 	log := m.log.Named("migrations")
 	log.Debug("starting slab migrations")
 
-	// start a worker pool that pulls slabs from a channel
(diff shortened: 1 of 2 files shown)


slabs/manager.go
```diff
@@ -398,41 +403,54 @@ func (m *SlabManager) performSlabMigrations(ctx context.Context) error {
...
-			return err
-		}
+	// fetch unhealthy slabs and feed them to the workers in a separate
+	// goroutine so the slow UnhealthySlabs / fetchMigrationState calls
+	// happen in parallel with worker processing rather than blocking it.
+	producerErrCh := make(chan error, 1)
+	go func() {
+		defer close(slabCh)
+		for {
+			log.Debug("processing batch")
```
peterjan, line 413: this line feels redundant if we have a duration line below it

peterjan (approved):

Issue #879, opened 2026-03-20
Debug slow queries


ChrisSchinnerl:
This is a tracking issues of queries appearing slow on Zeus. We should keep an eye on them and check if there is anything we can do to improve how long they take to execute. 

For some queries that might adding a missing index, batching or an architectural overhaul. 

```
indexd-1  | 2026-03-20T13:32:30Z        DEBUG   postgres.transaction    slow exec       {"id": "3ee49935", "attempt": 1, "query": "UPDATE contracts SET good = FALSE WHERE proof_height <= $1", "elapsed": "1.011385703s", "stack": "go.sia.tech/indexd/persist/postgres.(*txn).Exec\n\tgo.sia.tech/indexd/persist/postgres/sql.go:76\ngo.sia.tech/indexd/persist/postgres.(*Store).MarkUnrenewableContractsBad.func1\n\tgo.sia.tech/indexd/persist/postgres/contracts.go:708\ngo.sia.tech/indexd/persist/postgres.(*Store).doTransaction\n\tgo.sia.tech/indexd/persist/postgres/store.go:112\ngo.sia.tech/indexd/persist/postgres.(*Store).transaction\n\tgo.sia.tech/indexd/persist/postgres/store.go:64\ngo.sia.tech/indexd/persist/postgres.(*Store).MarkUnrenewableContractsBad\n\tgo.sia.tech/indexd/persist/postgres/contracts.go:707\ngo.sia.tech/indexd/contracts.(*ContractManager).performContractMaintenance\n\tgo.sia.tech/indexd/contracts/maintenance.go:80\ngo.sia.tech/indexd/contracts.(*ContractManager).maintenanceLoop.func2\n\tgo.sia.tech/indexd/contracts/maintenance.go:145\nsync.(*WaitGroup).Go.func1\n\tsync/waitgroup.go:258"}
```

```
[...]

PR #991, opened 2026-06-03
slabs: drop redundant slabs_digest_idx

peterjan:
This PR drops a redundant index on our slab digest, this switches the index we use to the unique index which should be faster, it should also give us a little performance improvement on insert/update. Unfortunately though the benchmark has high variance and I could not consistently get a clear performance improvement out of it. The planner clearly shows an improvement though (see hits 7)

Planner:

```
Before:

    Update on public.slabs  (cost=0.28..8.29 rows=0 width=0) (actual time=0.047..0.047 rows=0.00 loops=1)
      Buffers: shared hit=27 dirtied=1
      ->  Index Scan using slabs_digest_idx on public.slabs
    Execution Time: 0.054 ms

After:

    Update on public.slabs  (cost=0.28..8.29 rows=0 width=0) (actual time=0.009..0.010 rows=0.00 loops=1)
      Buffers: shared hit=7
      ->  Index Scan using slabs_digest_key on public.slabs
    Execution Time: 0.013 ms
```    

Benchmarks:

```
Before:
    ...
    BenchmarkUnhealthySlabs/250-10             	     120	  10141471 ns/op	   73555 B/op	    1782 allocs/op
    BenchmarkUnhealthySlabs/250-10             	     100	  10128730 ns/op	   73817 B/op	    1787 allocs/op
    BenchmarkUnhealthySlabs/250-10             	     100	  10891780 ns/op	   74157 B/op	    1793 allocs/op
    BenchmarkUnhealthySlabs/250-10             	     144	   7926760 ns/op	   73233 B/op	    1774 allocs/op
    BenchmarkUnhealthySlabs/250-10             	     165	   7660943 ns/op	   74136 B/op	    1794 allocs/op

After:
    ...
    BenchmarkUnhealthySlabs/250-10             	     147	   7219438 ns/op	   73990 B/op	    1792 allocs/op
    BenchmarkUnhealthySlabs/250-10             	     261	   4680302 ns/op	   73496 B/op	    1779 allocs/op
    BenchmarkUnhealthySlabs/250-10             	     246	   4928943 ns/op	   73865 B/op	    1787 allocs/op
    BenchmarkUnhealthySlabs/250-10             	     229	   5348066 ns/op	   73879 B/op	    1789 allocs/op
    BenchmarkUnhealthySlabs/250-10             	     222	   5434267 ns/op	   73612 B/op	    1781 allocs/op
```

edit: verified the drop on zeus, the index switches over fine

References #879


ChrisSchinnerl (changes requested):
Got a conflict now

persist/postgres/migrations.go
```diff
@@ -86,6 +86,10 @@ UPDATE hosts SET has_quic = EXISTS (
 		_, err := tx.Exec(ctx, `ALTER TABLE hosts DROP COLUMN has_bad_quic_port`)
 		return err
 	},
+	func(ctx context.Context, tx *txn, log *zap.Logger) error {
```
ChrisSchinnerl, line 89: Newer migration needs to go last.

ChrisSchinnerl (changes requested):

persist/postgres/migrations.go
peterjan: urgh, crazy, I even checked I think :/

peterjan: wait.. I did check, and it is correct?

ChrisSchinnerl (approved):

PR #992, opened 2026-06-03
slabs: update object_events once per migrated slab

peterjan:
This PR moves updating `object_events` out of `MigrateSector` because we were doing that for every single sector. I also found out we can cleanup the query a little bit by getting of a redundant JOIN. 

Benchmarks:

```
Before:

    BenchmarkMigrateSector-10    	     823	   1228398 ns/op	3414.45 MB/s	    2883 B/op	      65 allocs/op
    BenchmarkMigrateSector-10    	     933	   1110778 ns/op	3776.00 MB/s	    2882 B/op	      65 allocs/op
    BenchmarkMigrateSector-10    	     985	   1127339 ns/op	3720.53 MB/s	    2882 B/op	      65 allocs/op
    BenchmarkMigrateSector-10    	     782	   1285995 ns/op	3261.52 MB/s	    2884 B/op	      65 allocs/op
    BenchmarkMigrateSector-10    	    1005	   1027657 ns/op	4081.42 MB/s	    2881 B/op	      65 allocs/op

After:

    BenchmarkMigrateSector-10    	    1077	   1094869 ns/op	3830.87 MB/s	    2714 B/op	      60 allocs/op
    BenchmarkMigrateSector-10    	    1022	   1038303 ns/op	4039.58 MB/s	    2715 B/op	      60 allocs/op
    BenchmarkMigrateSector-10    	    1044	   1136143 ns/op	3691.70 MB/s	    2714 B/op	      60 allocs/op
    BenchmarkMigrateSector-10    	    1086	   1017254 ns/op	4123.16 MB/s	    2713 B/op	      60 allocs/op
    BenchmarkMigrateSector-10    	    1086	   1000083 ns/op	4193.96 MB/s	    2714 B/op	      60 allocs/op
```

The big win here of course is that we call this once per slab instead of once per migrated sector.

References #879

### DIFF
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -934,8 +934,8 @@ func (s *Store) RecordSlabMigrated(slabID slabs.SlabID) error {
 		_, err := tx.Exec(ctx, `
 			UPDATE object_events
 			SET updated_at = date_trunc('second', NOW())
-			WHERE (account_id, object_key) IN (
-				SELECT DISTINCT o.account_id, o.object_key
+			WHERE object_key IN (
+				SELECT DISTINCT o.object_key
 				FROM object_slabs os
 				INNER JOIN objects o ON os.object_id = o.id
 				WHERE os.slab_digest = $1
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -926,22 +926,45 @@ func (s *Store) UnhealthySlabs(limit int) (unhealthy []slabs.SlabID, err error) 
 	return
 }
 
+// RecordSlabMigrated updates the database to reflect that a slab was migrated.
+// This should be called after a slab was migrated and one or more of its
+// sectors were updated.
+func (s *Store) RecordSlabMigrated(slabID slabs.SlabID) error {
+	return s.transaction(func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `
+			UPDATE object_events
+			SET updated_at = date_trunc('second', NOW())
+			WHERE (account_id, object_key) IN (
+				SELECT DISTINCT o.account_id, o.object_key
+				FROM object_slabs os
+				INNER JOIN objects o ON os.object_id = o.id
+				WHERE os.slab_digest = $1
+			)
+		`, sqlHash256(slabID))
+		if err != nil {
+			return fmt.Errorf("failed to update affected object events: %w", err)
+		}
+		return nil
+	})
+}
+
 // MigrateSector updates a sector that was just migrated in the database to be
 // linked to the new host identified by 'hostKey'. This will reset the contract
 // ID since a freshly migrated sector isn't pinned yet. To pin a sector
 // 'PinSectors' is used. If the host is not found, e.g. due to being deleted in
-// the meantime, this operation is a no-op.
+// the meantime, this operation is a no-op. The caller is responsible for
+// invoking RecordSlabMigrated once per slab after the batch completes to bump
+// the corresponding object_events rows.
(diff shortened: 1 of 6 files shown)


persist/postgres/sectors.go
```diff
@@ -926,22 +926,45 @@ func (s *Store) UnhealthySlabs(limit int) (unhealthy []slabs.SlabID, err error)
...
+// RecordSlabMigrated updates the database to reflect that a slab was migrated.
+// This should be called after a slab was migrated and one or more of its
+// sectors were updated.
+func (s *Store) RecordSlabMigrated(slabID slabs.SlabID) error {
+	return s.transaction(func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `
+			UPDATE object_events
+			SET updated_at = date_trunc('second', NOW())
+			WHERE (account_id, object_key) IN (
+				SELECT DISTINCT o.account_id, o.object_key
```
ChrisSchinnerl: I think you can just ignore the `account_id` in this query? We want to update all events regardless of account anyway.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (dismissed):

persist/postgres/sectors.go
```diff
@@ -926,22 +926,45 @@ func (s *Store) UnhealthySlabs(limit int) (unhealthy []slabs.SlabID, err error)
...
 
+// RecordSlabMigrated updates the database to reflect that a slab was migrated.
+// This should be called after a slab was migrated and one or more of its
+// sectors were updated.
+func (s *Store) RecordSlabMigrated(slabID slabs.SlabID) error {
+	return s.transaction(func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `
+			UPDATE object_events
+			SET updated_at = date_trunc('second', NOW())
+			WHERE object_key IN (
```
chris124567, line 949: There's no index on object_key
```
CREATE TABLE object_events (
    object_key BYTEA NOT NULL CHECK(LENGTH(object_key) = 32), -- not a FK since deletions need to hang around
    account_id BIGINT NOT NULL REFERENCES accounts(id) ON DELETE CASCADE,
    was_deleted BOOLEAN NOT NULL, -- true if deleted, false otherwise
    updated_at TIMESTAMP WITH TIME ZONE DEFAULT NOW(), -- last time the object was created/updated/deleted truncated to second precision
    PRIMARY KEY (account_id, object_key)
);

-- fast sorting by update time and key
CREATE INDEX object_events_updated_at_object_key_idx ON object_events(updated_at ASC, object_key ASC);
```

Can we make this faster by adding an index on object_key or leave the `WHERE (account_id, object_key) IN` in place?

peterjan: It's covered by the PK, verified the plan with `EXPLAIN` to confirm

ChrisSchinnerl: Is that still as fast as before considering that the primary key is `(account_id, object_key)`?

peterjan: It's 26% slower, sorry for dismissing your comment @chris124567 , I just didn't realize I changed the query in such a meaningful way

chris124567: Do we need a new index or can we just do `(account_id, object_key)`?

peterjan: I considered both but I think it's cleaner to add a new index so we can have the query in its simplest form... Not sure if I made the right decision there though but I still think that's the best way

ChrisSchinnerl (approved):

chris124567 (approved):

PR #996, opened 2026-06-03
Don't renew a contract with capacity 0 if there is another active contract with capacity 0 already

### LINKED ISSUE #995: Don't renew a contract with capacity 0 if there is another active contract with capacity 0 already
That way we only ever keep one empty contract for potential uploads rather than locking up collateral on multiple.

chris124567:
Close #995

### DIFF
--- a/contracts/renew.go
+++ b/contracts/renew.go
@@ -19,6 +19,14 @@ func (cm *ContractManager) performContractRenewals(ctx context.Context, period, 
 	newProofHeight := bh + period
 
 	var eligible, attempted, successful int
+	// renew at most one empty contract per host to keep a single empty contract
+	// for potential uploads instead of locking up collateral on several. emptiness is
+	// determined by Size rather than Capacity: a renewal sets the new contract's
+	// capacity to the old contract's size, so a contract with no stored data
+	// always renews into a capacity-0 contract. empty contracts are deferred and
+	// renewed after the scan, once every active empty contract is known.
+	var emptyEligible []Contract
+	hostsWithActiveEmptyContract := make(map[types.PublicKey]struct{})
 	batchSize := 50
 	for offset := 0; ; offset += batchSize {
 		contracts, err := cm.store.Contracts(offset, batchSize, WithGood(true), WithRevisable(true))
@@ -28,9 +36,15 @@ func (cm *ContractManager) performContractRenewals(ctx context.Context, period, 
 		eligible += len(contracts)
 		for _, contract := range contracts {
 			if contract.ProofHeight > minProofHeight {
-				continue // too early to renew
-			} else if !contract.Good {
-				continue // contract is bad
+				// too early to renew; an empty one is the host's retained empty contract
+				if contract.Size == 0 {
+					hostsWithActiveEmptyContract[contract.HostKey] = struct{}{}
+				}
+				continue
+			} else if contract.Size == 0 {
+				// defer empty contracts so at most one per host is renewed
+				emptyEligible = append(emptyEligible, contract)
+				continue
 			}
 			attempted++
 			log := log.With(zap.Stringer("contractID", contract.ID), zap.Stringer("host", contract.HostKey))
@@ -45,6 +59,25 @@ func (cm *ContractManager) performContractRenewals(ctx context.Context, period, 
 			break
 		}
 	}
+
+	// renew at most one empty contract per host, skipping hosts that already have
+	// an active empty contract. the rest are left to expire.
+	for _, contract := range emptyEligible {
+		log := log.With(zap.Stringer("contractID", contract.ID), zap.Stringer("host", contract.HostKey))
+		if _, ok := hostsWithActiveEmptyContract[contract.HostKey]; ok {
+			log.Debug("skipping empty contract renewal, host already has an active empty contract")
+			continue
+		}
+		attempted++
+		if err := cm.renewContract(ctx, contract, newProofHeight, log); err != nil {
+			log.Error("failed to renew contract", zap.Error(err))
+		} else {
+			// the renewal is also empty and is now the host's retained empty contract
+			hostsWithActiveEmptyContract[contract.HostKey] = struct{}{}
+			successful++
+		}
+	}
+
 	log.Debug("renewals finished", zap.Int("eligible", eligible), zap.Int("attempted", attempted), zap.Int("successful", successful))
 	return nil
 }
--- a/contracts/renew.go
+++ b/contracts/renew.go
@@ -68,12 +68,14 @@ func (cm *ContractManager) performContractRenewals(ctx context.Context, period, 
 			log.Debug("skipping empty contract renewal, host already has an active empty contract")
 			continue
 		}
+		// mark the host before attempting: renewContract can fail after the renewal
+		// has already been negotiated/broadcast, and we must not negotiate a second
+		// empty contract for the same host in this run.
+		hostsWithActiveEmptyContract[contract.HostKey] = struct{}{}
 		attempted++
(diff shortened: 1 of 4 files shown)


contracts/renew.go
```diff
@@ -28,9 +36,15 @@ func (cm *ContractManager) performContractRenewals(ctx context.Context, period,
...
-				continue // too early to renew
-			} else if !contract.Good {
-				continue // contract is bad
+				// too early to renew; an empty one is the host's retained empty contract
+				if contract.Size == 0 {
+					hostsWithActiveEmptyContract[contract.HostKey] = struct{}{}
+				}
+				continue
+			} else if contract.Size == 0 {
+				// defer empty contracts so at most one per host is renewed
```
ChrisSchinnerl: Could we defer all renewals and have this first loop just collect eligible contracts? Seems a bit nicer to me to only call `renewContract` in one place and would probably also fix the `eligible` counter currently counting all contracts as eligible regardless of whether they are early or not.

This also slightly changes the behavior to keep trying to renew contracts if renewing an empty contract failed because even if it succeeded on the host side at this point it still becomes unusable to us. 

e.g. like so
```
diff --git a/contracts/renew.go b/contracts/renew.go
index 8a4cce38..a815ed79 100644
--- a/contracts/renew.go
+++ b/contracts/renew.go
@@ -18,14 +18,12 @@ func (cm *ContractManager) performContractRenewals(ctx context.Context, period,
 	minProofHeight := bh + renewWindow
 	newProofHeight := bh + period
 
-	var eligible, attempted, successful int
-	// renew at most one empty contract per host to keep a single empty contract
-	// for potential uploads instead of locking up collateral on several. emptiness is
-	// determined by Size rather than Capacity: a renewal sets the new contract's
-	// capacity to the old contract's size, so a contract with no stored data
-	// always renews into a capacity-0 contract. empty contracts are deferred and
-	// renewed after the scan, once every active empty contract is known.
-	var emptyEligible []Contract
+	var attempted, successful int
+
+	// collect all contracts elligible for renewal and remember hosts that have
+	// an active empty contract pre-renewal as we only want to keep around one
+	// empty contract per host.
+	var eligible []Contract
 	hostsWithActiveEmptyContract := make(map[types.PublicKey]struct{})
 	batchSize := 50
@@ -33,7 +31,6 @@ func (cm *ContractManager) performContractRenewals(ctx context.Context, period,
 		if err != nil {
 			return fmt.Errorf("failed to fetch contracts for renewal: %w", err)
 		}
-		eligible += len(contracts)
 		for _, contract := range contracts {
 			if contract.ProofHeight > minProofHeight {
 				// too early to renew; an empty one is the host's retained empty contract
@@ -41,18 +38,8 @@ func (cm *ContractManager) performContractRenewals(ctx context.Context, period,
 					hostsWithActiveEmptyContract[contract.HostKey] = struct{}{}
 				}
 				continue
-			} else if contract.Size == 0 {
-				// defer empty contracts so at most one per host is renewed
-				emptyEligible = append(emptyEligible, contract)
-				continue
-			}
-			attempted++
-			log := log.With(zap.Stringer("contractID", contract.ID), zap.Stringer("host", contract.HostKey))
-			if err := cm.renewContract(ctx, contract, newProofHeight, log); err != nil {
-				log.Error("failed to renew contract", zap.Error(err))
-			} else {
-				successful++
 			}
+			eligible = append(eligible, contract)
 		}
 
 		if len(contracts) < batchSize {
@@ -60,27 +47,29 @@ func (cm *ContractManager) performContractRenewals(ctx context.Context, period,
 		}
 	}
 
-	// renew at most one empty contract per host, skipping hosts that already have
-	// an active empty contract. the rest are left to expire.
-	for _, contract := range emptyEligible {
+	// perform renewals on eligible contracts
+	for _, contract := range eligible {
 		log := log.With(zap.Stringer("contractID", contract.ID), zap.Stringer("host", contract.HostKey))
 		if _, ok := hostsWithActiveEmptyContract[contract.HostKey]; ok {
 			log.Debug("skipping empty contract renewal, host already has an active empty contract")
 			continue
 		}
-		// mark the host before attempting: renewContract can fail after the renewal
-		// has already been negotiated/broadcast, and we must not negotiate a second
-		// empty contract for the same host in this run.
-		hostsWithActiveEmptyContract[contract.HostKey] = struct{}{}
 		attempted++
 		if err := cm.renewContract(ctx, contract, newProofHeight, log); err != nil {
 			log.Error("failed to renew contract", zap.Error(err))
-		} else {
-			successful++
+			continue
+		}
+
+		// success
+		successful++
+
+		// remember that this host now has an active empty contract
+		if contract.Size == 0 {
+			hostsWithActiveEmptyContract[contract.HostKey] = struct{}{}
 		}
 	}
 
-	log.Debug("renewals finished", zap.Int("eligible", eligible), zap.Int("attempted", attempted), zap.Int("successful", successful))
+	log.Debug("renewals finished", zap.Int("eligible", len(eligible)), zap.Int("attempted", attempted), zap.Int("successful", successful))
 	return nil
 }
```

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

contracts/renew.go
```diff
@@ -18,34 +18,58 @@ func (cm *ContractManager) performContractRenewals(ctx context.Context, period,
...
 		if len(contracts) < batchSize {
 			break
 		}
 	}
-	log.Debug("renewals finished", zap.Int("eligible", eligible), zap.Int("attempted", attempted), zap.Int("successful", successful))
+
+	// perform renewals on eligible contracts
+	for _, contract := range eligible {
+		log := log.With(zap.Stringer("contractID", contract.ID), zap.Stringer("host", contract.HostKey))
+		if _, ok := hostsWithActiveEmptyContract[contract.HostKey]; ok {
```
peterjan: we should check `contract.Size` here right? Aren't we skipping over an important renewal here if contract.Size is not zero?

contracts/renew_test.go
```diff
@@ -131,6 +131,126 @@ func TestPerformContractRenewals(t *testing.T) {
 	}
 }
 
+func TestPerformContractRenewalsEmptyContracts(t *testing.T) {
```
peterjan, line 134: if my previous comment checked out, we should update this test so it fails

peterjan (changes requested):

contracts/renew.go
peterjan: Hm actually it's probably correct but the logic is confusing to me. I think the error message is quite misleading because it says "skipping empty contract renewal, host already has an active empty contract", but this can be printed for a contract that is not in its renew window and is the only (or first) empty contract we encounter for that host.. right? 

I would change it so you filter down contracts to eligible contracts, based on the `minProofHeight`. Once you have those you can have the second loop that keeps track of a map of empty contracts on a per host basis. If you find an empty contract and it's the first time, you renew it, set the map, any consecutive empty contracts you find you know you don't have to renew it.

ChrisSchinnerl: No, you are correct. What you describe is what should be happening here but it's missing the size == 0 check. The first loop filters them down using `minProofHeight`. This loop then renews them unless they have a size of 0 and have another size 0 contract that is not up for renewal.

Definitely want the test to capture this though.

ChrisSchinnerl (changes requested):

contracts/renew.go
chris124567: Thank you for catching this

peterjan (approved):
logic is still quite tricky but LGTM

ChrisSchinnerl (approved):

PR #1006, opened 2026-06-11
Register additional sharing account for each user

### LINKED ISSUE #1004: Register additional sharing account for each user
Right now each user or rather connect key has 1 associated account pool.
Each registered app with that connect key is 1 new account that gets linked to that pool.

What we want next is an extra account that gets created alongside the pool for every user. It is attached to a user's pool just like the app account keys but the secret (signing key) is stored alongside the connect key in the database.

NOTE: This is a preparation for API keys. Which are keys users can share publicly with people who don't have access to an indexer themselves. The API key authenticates them with the API and allows them to obtain account tokens created from the secret key above. File shared like that will drain the sharing user's pool instead of requiring the person the object was shared with to have an account themselves.

chris124567:
Close #1004

### DIFF
--- a/accounts/connect.go
+++ b/accounts/connect.go
@@ -174,3 +174,11 @@ func (m *AccountManager) AppSecret(connectKey string, appID types.Hash256) (type
 	}
 	return types.Hash256(keys.Derive(secret[:], appID[:], []byte("server app secret"), 32)), nil
 }
+
+// DeriveSharingAccountKey deterministically derives a connect key's sharing
+// account key pair from its user secret.
+func DeriveSharingAccountKey(userSecret types.Hash256) types.PrivateKey {
+	seed := keys.Derive(userSecret[:], []byte("sharing"), nil, 32)
+	defer clear(seed)
+	return types.NewPrivateKeyFromSeed(seed)
+}
--- a/persist/postgres/init.sql
+++ b/persist/postgres/init.sql
@@ -123,9 +123,13 @@ CREATE TABLE pool_hosts (
     host_id INTEGER NOT NULL REFERENCES hosts(id) ON DELETE CASCADE,
     next_fund TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
     consecutive_failed_funds INTEGER NOT NULL DEFAULT 0,
+    sharing_attached BOOLEAN NOT NULL DEFAULT FALSE, -- whether the pool's derived sharing account has been attached on this host
     CONSTRAINT pool_hosts_pk PRIMARY KEY (pool_id, host_id)
 );
 CREATE INDEX pool_hosts_host_id_next_fund_idx ON pool_hosts (host_id, next_fund);
+-- partial index for SharingPoolAttachments: only indexes pools whose sharing
+-- account still needs attaching
+CREATE INDEX pool_hosts_sharing_unattached_idx ON pool_hosts (host_id) WHERE sharing_attached = FALSE;
 
 CREATE TABLE pool_attachments (
     pool_id INTEGER NOT NULL,
--- a/.changeset/register_additional_sharing_account_for_each_user.md
+++ b/.changeset/register_additional_sharing_account_for_each_user.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+# Register additional sharing account for each user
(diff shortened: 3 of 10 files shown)


persist/postgres/apps.go
```diff
@@ -36,14 +36,15 @@ func scanConnectKey(s scanner) (key accounts.ConnectKey, err error) {
 func (s *Store) AddAppConnectKey(meta accounts.AppConnectKeyRequest) (key accounts.ConnectKey, err error) {
 	err = s.transaction(func(ctx context.Context, tx *txn) error {
 		userSecret := frand.Bytes(32)
+		sharingAccountPrivateKey := types.GeneratePrivateKey()
 		key, err = scanConnectKey(tx.QueryRow(ctx, `
```
peterjan, line 39: might make sense to return the database ID of the key here as a third parameter, I think other call sites end up doing `SELECT id FROM app_connect_keys WHERE app_key = $1` so we need it there as well

peterjan: (feel free to ignore it if it feels awkward though)

peterjan (dismissed):

persist/postgres/apps.go
chris124567: meh it only saves us a query in one place

accounts/connect.go
```diff
@@ -165,6 +165,14 @@ func (m *AccountManager) Quotas(ctx context.Context, offset, limit int) ([]Quota
 	return m.store.Quotas(offset, limit)
 }
 
+// SharingAccountPrivateKey returns the private signing key of the sharing
+// account created alongside the given connect key. The key must never be
+// exposed to the user. If the connect key is not found, it returns
+// [ErrKeyNotFound].
+func (m *AccountManager) SharingAccountPrivateKey(ctx context.Context, connectKey string) (types.PrivateKey, error) {
```
ChrisSchinnerl: Unused method. I'd remove this for now since we'll probably expose a method that returns account tokens instead of the key.

persist/postgres/apps_test.go
```diff
@@ -242,6 +245,95 @@ func TestAppConnectKey(t *testing.T) {
...
+	// the sharing account's private signing key should be stored alongside the
+	// connect key
+	sk, err := store.SharingAccountPrivateKey(connectKey)
+	if err != nil {
+		t.Fatal("failed to get sharing account private key:", err)
+	}
+
+	// an account should exist for the sharing account's public key, but it
+	// should not show up in account listings
+	if acc, err := store.Account(sk.PublicKey()); err != nil {
```
ChrisSchinnerl: Is that even necessary? We might be able to get away with completely hiding the account I think. Except for funding of course.

ChrisSchinnerl (changes requested):
It's a shame we have to touch every account query but I guess there is no good way around it.

persist/postgres/accounts.go
```diff
@@ -108,7 +110,7 @@ LEFT JOIN LATERAL (
 		LIMIT $2
 	) ready
 ) ahr ON TRUE
-WHERE public_key = $1`, sqlPublicKey(ak), accounts.ReadyHostThreshold))
+WHERE a.public_key = $1 AND NOT a.sharing`, sqlPublicKey(ak), accounts.ReadyHostThreshold))
```
n8mgr: You can derive from `user_secret` and skip registering it in the database.

persist/postgres/init.sql
```diff
@@ -38,13 +38,17 @@ CREATE TABLE accounts (
     logo_url TEXT NOT NULL DEFAULT '',
     service_url TEXT NOT NULL DEFAULT '',
     last_used TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
-    deleted_at TIMESTAMP WITH TIME ZONE
+    deleted_at TIMESTAMP WITH TIME ZONE,
+    sharing BOOLEAN NOT NULL DEFAULT FALSE -- the connect key's sharing account rather than a registered app; it must never own pinned objects or slabs so DeleteAppConnectKey can hard delete it
```
n8mgr: My point with the derivation was that you can probably drop the `sharing` field entirely, similarly to how we don't add service accounts to the database.

peterjan (approved):
One aspect of this updated version that bugs me a bit is that a `sharing_attached` bool doesn't scale past a single sharing account and also loosely references the key.. if we ever end up adding another type of key, or change the derivation schema, we'll have to migrate the database anyway. In other words we could easily end up with false statements in the database, if we update the sharing domain to `sharing-foo` we'd have rows falsely stating that sharing key is attached. 

PR looks good so I approved it.

ChrisSchinnerl (approved):

n8mgr (approved):

PR #1010, opened 2026-06-12
Ensure that pruning slabs/sectors in parallel doesn't leave orphaned slabs/sectors

### LINKED ISSUE #1009: Test that pruning slabs/sectors in parallel doesn't leave orphaned slabs/sectors
We recently discovered a good number of orphaned sectors on our staging indexer. There is a good chance that our account pruning and other pruning code can lead to slabs or sectors not being pruned correctly due to db locking. Among the orphaned sectors were pinned, unpinned and also unpinnable sectors. 

Investigate this and add some concurrency testing to try and reproduce this.

chris124567:
Close #1009

Test fails on master and passes on this branch

### DIFF
--- a/persist/postgres/accounts.go
+++ b/persist/postgres/accounts.go
@@ -131,14 +131,31 @@ func (s *Store) HasAccount(ak types.PublicKey) (bool, error) {
 	return exists, nil
 }
 
-// DeleteAccount deletes the account in the database with given account key.
+// DeleteAccount soft deletes the account in the database with the given account
+// key. The account's pinned totals are reclaimed from its connect key
+// immediately so freed space is reflected right away; the account's objects,
+// slabs and the row itself are cleaned up later by PruneAccounts.
 func (s *Store) DeleteAccount(acc proto.Account) error {
 	return s.transaction(func(ctx context.Context, tx *txn) error {
-		res, err := tx.Exec(ctx, `UPDATE accounts SET deleted_at = NOW() WHERE public_key = $1 AND deleted_at IS NULL`, sqlPublicKey(acc))
-		if err != nil {
-			return fmt.Errorf("failed to delete account: %w", err)
-		} else if res.RowsAffected() == 0 {
+		// mark the account deleted and decrement the connect key's pinned totals
+		// by the account's current usage
+		var deleted bool
+		err := tx.QueryRow(ctx, `
+WITH deleted AS (
+	UPDATE accounts SET deleted_at = NOW()
+	WHERE public_key = $1 AND deleted_at IS NULL
+	RETURNING connect_key_id, pinned_data, pinned_size
+)
+UPDATE app_connect_keys ack
+SET pinned_data = ack.pinned_data - deleted.pinned_data,
+	pinned_size = ack.pinned_size - deleted.pinned_size
+FROM deleted
+WHERE ack.id = deleted.connect_key_id
+RETURNING TRUE`, sqlPublicKey(acc)).Scan(&deleted)
+		if errors.Is(err, sql.ErrNoRows) {
 			return accounts.ErrNotFound
+		} else if err != nil {
+			return fmt.Errorf("failed to delete account: %w", err)
 		}
 		return nil
 	})
@@ -225,7 +242,7 @@ RETURNING o.object_key;`, accountID, remaining)
 				)
 				RETURNING slab_id
 			), queued AS (
-				INSERT INTO account_slabs_for_deletion (slab_id)
+				INSERT INTO slab_deletion_queue (slab_id)
 				SELECT slab_id FROM deleted
 			)
 			SELECT COUNT(*) FROM deleted;`, accountID, remaining).Scan(&queued); err != nil {
@@ -235,14 +252,10 @@ RETURNING o.object_key;`, accountID, remaining)
 			return nil
 		}
 
-		// no slabs left; decrement the connect key's pinned totals and hard
-		// delete the account
-		var connectKeyID int64
-		var pinnedData, pinnedSize uint64
-		if err := tx.QueryRow(ctx, `DELETE FROM accounts WHERE id = $1 RETURNING connect_key_id, pinned_data, pinned_size`, accountID).Scan(&connectKeyID, &pinnedData, &pinnedSize); err != nil {
+		// no slabs left; hard delete the account. its connect key's pinned
+		// totals were already reclaimed when the account was soft deleted
+		if _, err := tx.Exec(ctx, `DELETE FROM accounts WHERE id = $1`, accountID); err != nil {
 			return fmt.Errorf("failed to delete account: %w", err)
-		} else if _, err := tx.Exec(ctx, `UPDATE app_connect_keys SET pinned_data = pinned_data - $1, pinned_size = pinned_size - $2 WHERE id = $3`, pinnedData, pinnedSize, connectKeyID); err != nil {
-			return fmt.Errorf("failed to update connect key pinned data: %w", err)
 		} else if err := incrementNumAccounts(ctx, tx, -1); err != nil {
 			return fmt.Errorf("failed to decrement account count: %w", err)
 		}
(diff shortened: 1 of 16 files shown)


persist/postgres/sectors.go
```diff
@@ -423,6 +423,24 @@ func (s *Store) PinSlabs(account proto.Account, nextIntegrityCheck time.Time, to
...
 
 func (s *Store) unpinSlabs(ctx context.Context, tx *txn, accountID int64, sIDs []int64) error {
+	if len(sIDs) == 0 {
+		return nil
+	}
+
+	// lock the slabs to serialize with other transactions unpinning the same
+	// slabs. The deletion checks below rely on committed account_slabs rows, so
+	// two transactions each removing one of the final references must not run
+	// their checks at the same time.
```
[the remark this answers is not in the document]
chris124567: fixed

persist/postgres/slabs_test.go
```diff
@@ -390,6 +391,200 @@ func TestSlabPruning(t *testing.T) {
...
+				}
+			}()
+		}
+		close(start)
+
+		go func() {
+			wg.Wait()
+			close(errCh)
+		}()
+
```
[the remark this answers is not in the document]
chris124567: fixed

peterjan (dismissed):
Verified test fails on master and there's no noticeable regression in BenchmarkUnpinSlab

persist/postgres/sectors.go
```diff
@@ -423,6 +423,24 @@ func (s *Store) PinSlabs(account proto.Account, nextIntegrityCheck time.Time, to
...
+	if _, err := tx.Exec(ctx, `
```
ChrisSchinnerl: As discussed in the meeting today we want to move unpinning slabs and sectors out of the hot path. So when a user prunes their account we limit the deletion to the `account_slabs` table and prune slabs and sectors in a background loop which should then avoid the concurrency issue.

ChrisSchinnerl (changes requested):

ChrisSchinnerl:
@chris124567 reminder to update this

chris124567:
Ok I will do this on Wednesday got tied up with s3d stuff

persist/postgres/slabs_test.go
```diff
@@ -463,3 +764,68 @@ func BenchmarkPruneSlabs(b *testing.B) {
...
+			select {
+			case <-stopPrune:
+				return
+			default:
+			}
+			if _, err := s.PruneDeletedSlabs(pruneBatch); err != nil {
+				errCh <- err
+				return
+			}
+		}
```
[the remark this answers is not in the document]
chris124567: Ran these multiple times and did not have reliability issues even with -race

persist/postgres/sectors.go
chris124567: I missed that meeting but is what I just pushed roughly what we wanted to do?

persist/postgres/sectors.go
```diff
@@ -910,6 +947,7 @@ func (s *Store) UnhealthySlabs(cursor int64, limit int) (unhealthy []slabs.SlabI
 				SELECT DISTINCT ss.slab_id AS id
 				FROM slab_sectors ss
 				WHERE ss.sector_id IN (SELECT id FROM batch)
+					AND EXISTS (SELECT 1 FROM account_slabs a WHERE a.slab_id = ss.slab_id)
```
ChrisSchinnerl: This seems like it would potentially slow down this query for hardly any benefit. Sure we might consider orphaned slabs for repairs now but they don't stick around for long.

persist/postgres/sectors.go
```diff
@@ -531,8 +562,14 @@ RETURNING connect_key_id`, pinnedDataDelta, pinnedSizeDelta, accountID).Scan(&co
...
-	// prune the slabs
-	if _, err := tx.Exec(ctx, `DELETE FROM slabs WHERE id = ANY($1)`, toDelete); err != nil {
+	// prune the slabs, counting rows actually deleted since the queue can hold
+	// a slab twice
+	var deletedSlabs int64
+	if err := tx.QueryRow(ctx, `
+		WITH deleted AS (
+			DELETE FROM slabs WHERE id = ANY($1) RETURNING id
+		)
+		SELECT COUNT(*) FROM deleted`, toDelete).Scan(&deletedSlabs); err != nil {
```
ChrisSchinnerl: This seems needlessly fancy compared to a `tx.Exec` with the `DELETE` and then looking at the affected rows. It also materializes the id which isn't necessary since we only care about the count.

persist/postgres/init.sql
```diff
@@ -393,6 +393,11 @@ CREATE TABLE account_slabs (
 -- speed up query used when unpinning slabs
 CREATE INDEX account_slabs_slab_id_idx ON account_slabs(slab_id);
 
+CREATE TABLE account_slabs_for_deletion (
```
ChrisSchinnerl: (nit) the name makes it sound like the account slabs in here should be deleted but they are actually already deleted and this is more of a queue for slabs to check.

persist/postgres/accounts.go
```diff
@@ -212,43 +212,102 @@ RETURNING o.object_key;`, accountID, remaining)
...
+			return nil
+		}
+
+		// no slabs left; decrement the connect key's pinned totals and hard
+		// delete the account
+		var connectKeyID int64
+		var pinnedData, pinnedSize uint64
+		if err := tx.QueryRow(ctx, `DELETE FROM accounts WHERE id = $1 RETURNING connect_key_id, pinned_data, pinned_size`, accountID).Scan(&connectKeyID, &pinnedData, &pinnedSize); err != nil {
+			return fmt.Errorf("failed to delete account: %w", err)
+		} else if _, err := tx.Exec(ctx, `UPDATE app_connect_keys SET pinned_data = pinned_data - $1, pinned_size = pinned_size - $2 WHERE id = $3`, pinnedData, pinnedSize, connectKeyID); err != nil {
```
ChrisSchinnerl: From a UX perspective it might be better to do that right when we set `deleted_at` on the account deletion. That way reclaiming space by deleting an app feels instant to the user. Rather than either
1. what we have on master where the pinned data slowly drains as slabs get unpinned
2. this PR where nothing happens until the account is empty and then the numbers get updated

ChrisSchinnerl (changes requested):

persist/postgres/accounts.go
```diff
@@ -212,43 +212,102 @@ RETURNING o.object_key;`, accountID, remaining)
...
+			), queued AS (
+				INSERT INTO account_slabs_for_deletion (slab_id)
+				SELECT slab_id FROM deleted
+			)
+			SELECT COUNT(*) FROM deleted;`, accountID, remaining).Scan(&queued); err != nil {
+			return fmt.Errorf("failed to queue account slabs for deletion: %w", err)
+		} else if queued == int64(remaining) {
+			// filled the batch, so more may remain
+			return nil
+		}
```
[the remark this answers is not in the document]
chris124567: simplified query

persist/postgres/accounts.go
```diff
@@ -212,43 +229,98 @@ RETURNING o.object_key;`, accountID, remaining)
...
+					WHERE account_id = $1
+					ORDER BY slab_id
+					LIMIT $2
+				)
+				RETURNING slab_id
+			), queued AS (
+				INSERT INTO slab_deletion_queue (slab_id)
+				SELECT slab_id FROM deleted
+			)
+			SELECT COUNT(*) FROM deleted;`, accountID, remaining).Scan(&queued); err != nil {
```
ChrisSchinnerl: This is also slightly more complicated than it needs to be. The queued expression already returns the rows that were queued through `RowsAffected`. 
So you can do

```
diff --git a/persist/postgres/accounts.go b/persist/postgres/accounts.go
index 3c9fd03f..a0d0ca50 100644
--- a/persist/postgres/accounts.go
+++ b/persist/postgres/accounts.go
@@ -230,8 +230,7 @@ RETURNING o.object_key;`, accountID, remaining)
 		}
 
 		// queue up to `remaining` slab associations for background deletion
-		var queued int64
-		if err := tx.QueryRow(ctx, `
+		res, err := tx.Exec(ctx, `
 			WITH deleted AS (
 				DELETE FROM account_slabs
 				WHERE account_id = $1 AND slab_id IN (
@@ -241,13 +240,12 @@ RETURNING o.object_key;`, accountID, remaining)
 					LIMIT $2
 				)
 				RETURNING slab_id
-			), queued AS (
-				INSERT INTO slab_deletion_queue (slab_id)
-				SELECT slab_id FROM deleted
 			)
-			SELECT COUNT(*) FROM deleted;`, accountID, remaining).Scan(&queued); err != nil {
+			INSERT INTO slab_deletion_queue (slab_id)
+			SELECT slab_id FROM deleted;`, accountID, remaining)
+		if err != nil {
 			return fmt.Errorf("failed to queue account slabs for deletion: %w", err)
-		} else if queued == int64(remaining) {
+		} else if res.RowsAffected() == int64(remaining) {
 			// filled the batch, so more may remain
 			return nil
 		}
```

ChrisSchinnerl (commented):
One last nit

peterjan (approved):

ChrisSchinnerl (approved):

PR #1011, opened 2026-06-13
add reconnecting field

n8mgr:
Adds a `reconnecting` field to auth connect status so apps can tell if a user is returning or connecting for the first time.


peterjan (approved):
`nit` m

edit: no sure what I was going to nit about.. so approved

chris124567 (approved):

PR #1014, opened 2026-06-22
Add network read and write throughput estimates

peterjan:
This PR exposes a global read/write estimate on the client. Required for https://github.com/SiaFoundation/sia-storage-go/issues/27 also going to be introduced in https://github.com/SiaFoundation/indexd/pull/1013 so figured I'd extract it into a separate PR.

### DIFF
--- a/.changeset/add_network_throughput_estimates.md
+++ b/.changeset/add_network_throughput_estimates.md
@@ -1,0 +1,7 @@
+---
+default: minor
+---
+
+# Add network read and write throughput estimates to the client.
+
+The client now exposes `ReadEstimate` and `WriteEstimate`, returning the expected duration to transfer a given number of bytes from the network-wide observed throughput. Both fall back to a default rate before any bulk transfers have been sampled.
--- a/client/v2/client.go
+++ b/client/v2/client.go
@@ -334,6 +334,20 @@ func (c *Client) Prioritize(hosts []types.PublicKey) []types.PublicKey {
 	return c.hosts.Prioritize(hosts)
 }
 
+// ReadEstimate returns the expected time to read the given number of bytes
+// based on the network-wide observed read throughput. See
+// [Provider.ReadEstimate].
+func (c *Client) ReadEstimate(bytes uint64) time.Duration {
+	return c.hosts.ReadEstimate(bytes)
+}
+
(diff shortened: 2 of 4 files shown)


chris124567 (commented):
```
package client_test

import (
    "testing"
    "time"

    "go.sia.tech/indexd/client/v2"
    "go.sia.tech/indexd/hosts"
)

func TestProviderReadEstimateBias(t *testing.T) {
    s := newTestStore(t)
    store := hosts.NewHostStore(s.Store)
    hostA := s.addUsableHost(t)
    provider := client.NewProvider(store)

    const sector = 4 << 20 // 4 MiB sector

    // one fast read: 4 MiB in 40ms == 100 MiB/s
    provider.AddReadSample(hostA, sector, 40*time.Millisecond)

    // the network then degrades: every subsequent 4 MiB read takes a full
    // 4 seconds (1 MiB/s).
    const observed = 4 * time.Second
    for i := 0; i < 5; i++ {
        provider.AddReadSample(hostA, sector, observed)
    }

    got := provider.ReadEstimate(sector)
    t.Logf("most recent observed transfer time for a 4 MiB sector: %v", observed)
    t.Logf("ReadEstimate for the same 4 MiB sector:                %v", got)
}
```

```
estimate_poc_test.go:30: most recent observed transfer time for a 4 MiB sector: 4s
estimate_poc_test.go:31: ReadEstimate for the same 4 MiB sector:                119.61608ms
```

I feel like this does not update very well

peterjan:
> I feel like this does not update very well

Hm that's fair but it's an example to highlight the worst case possible of course. This is what the Rust SDK does currently. In practice this blends many hosts and many reads so it would self-correct pretty quickly. If you go with pace instead of speed you get the opposite on tiny reads that take a little longer.

That said I am now wondering why we want with speed over pace, but considering all the benchmarking we did on the Rust SDK I'm inclined to say it was a very conscious decision at the time. Perhaps we should make it a byte-weighted rate though, that makes the most sense and would handle all these edge cases the best.

edit: probably just fine though considering it averages all hosts and lots of reads/writes

chris124567 (approved):
Yeah if it's global it's probably fine

ChrisSchinnerl (approved):

PR #1015, opened 2026-06-22
Add slab versioning

n8mgr:
Adds a version byte to the slab. This breaks Sia encoding for slabs and objects, but I don't believe anywhere is actually using it. The use case is for changing slab behavior. More specifically, we need to change how the object encryption works so that we can overwrite data without needing to re-encrypt the entire object.

### DIFF
--- a/slabs/slabs.go
+++ b/slabs/slabs.go
@@ -19,6 +19,9 @@ const (
 
 	// maxTotalShards is the maximum number of total shards (data + parity) allowed in a slab.
 	maxTotalShards = 256
+
+	// maxSlabVersion is the maximum slab version supported by the indexer
+	maxSlabVersion = 1
 )
 
 var (
@@ -36,6 +39,10 @@ var (
 	// number of minimum shards, for example if `MinShards` exceeds the number
 	// of sectors.
 	ErrMinShards = errors.New("slab has invalid number of minimum shards")
+
+	// ErrUnsupportedSlabVersion is returned when attempting to pin a slab with
+	// a version that is not yet supported.
+	ErrUnsupportedSlabVersion = errors.New("unsupported slab version")
 )
 
 type (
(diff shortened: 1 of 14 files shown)


peterjan (dismissed):
LGTM 🚀

slabs/slabs.go
```diff
@@ -82,6 +91,7 @@ type (
 	// A PinnedSlab is a slab that has been pinned to hosts.
 	PinnedSlab struct {
 		ID            SlabID         `json:"id"`
+		Version       uint8          `json:"version"`
```
chris124567, line 94: nit: It seems like we don't strictly have to but why not add `Version` to the encoding functions for PinnedSlab

chris124567 (dismissed):

peterjan:
raked up some conflicts, otherwise LGTM

slabs/slabs_test.go
```diff
@@ -87,6 +87,25 @@ func TestSlabPinParamsDigest(t *testing.T) {
 	}
 }
 
+func TestSlabVersionDigest(t *testing.T) {
```
ChrisSchinnerl, line 90: (nit) would be nice to have a known hash for legacy slab pin params and then compare the digest of the new params with version 0 to that known hash.

ChrisSchinnerl (changes requested):
S3d currently uses the Sia encoding (see `sqlSiaObject`) to persist a sealed object. So that would break without either backwards compatibility here or a migration plan for s3d.

ChrisSchinnerl (approved):

slabs/slabs.go
ChrisSchinnerl: @n8mgr any thoughts? We could also drop the endpoint since there is a chance we don't rely on it anywhere anymore but keeping it as is will allow a user to fetch a version 1 slab with octet stream encoding that gets decoded as version 0.

ChrisSchinnerl (changes requested):

slabs/slabs.go
n8mgr: Is this not already fixed?

chris124567 (approved):

ChrisSchinnerl (approved):

PR #1020, opened 2026-06-24
Make integrity checks safe to run in parallel 

ChrisSchinnerl:
Similar to the PR related to migrations. With this both migrations and integrity checks should be able to run on multiple nodes in parallel.

### DIFF
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -22,6 +22,13 @@ const (
 	// maxBadParityShards is the maximum proportion of parity shards that can be
 	// on bad hosts when pinning a slab.
 	maxBadParityShards = 0.2
+	// integrityCheckClaimInterval is how far into the future
+	// SectorsForIntegrityCheck pushes the next_integrity_check of the sectors
+	// it hands out. It acts as a lease so that concurrent integrity-check
+	// workers never fetch the same sectors. If a check is interrupted before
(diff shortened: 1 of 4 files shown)


persist/postgres/sectors.go
```diff
@@ -143,14 +156,20 @@ func (s *Store) SectorsForIntegrityCheck(hostKey types.PublicKey, limit int) ([]
...
-			SELECT sector_root
-			FROM sectors
-			WHERE
-				host_id = (SELECT id FROM hid)
-				AND next_integrity_check <= NOW()
-			ORDER BY next_integrity_check ASC
-			LIMIT $2
-		`, sqlPublicKey(hostKey), limit)
+			UPDATE sectors
+			SET next_integrity_check = NOW() + make_interval(secs => $3)
```
peterjan, line 160: why build the interval in SQL?

peterjan (approved):

persist/postgres/sectors.go
ChrisSchinnerl: To avoid clock skew across different servers. Now the database clock is the source of truth for integrity checks.

peterjan: we probably have more instances that fall under the same umbrella, perhaps we should open an issue to update all of those as well

PR #1021, opened 2026-06-26
Add `remote` subcommand for spinning up a minimal remote indexer

### LINKED ISSUE #1012: Repair mode
The current bottleneck when scaling a node is a single's server ability to migrate slabs. To fix this we want to introduce a repair mode for indexd that launches `indexd` with just the capabilities it needs to fetch slabs to migrate, migrate them and then update the metadata of those slabs. No http API, contract maintenance etc. 

To perform a migration a node needs to at least know:
- the private key of a funded account (e.g. the service account)
- the list of hosts that make a sector a "healthy sector" (Good for upload hosts)
- the list of hosts that can be used to migrate a sector to (Good for append hosts)
- slab metadata 

The easiest solution to achieve that is by granting an indexer in repair mode access to the database and then run the slab migration loop on that indexer.

However, we need to make sure the migration loop is safe to use from multiple threads. Meaning that 2 indexers performing migrations don't end up with the same slabs. In theory `UnhealthySlabs` already updates `next_repair_attempt` optimistically so that shouldn't be an issue but we might need a `FOR UPDATE` or some other update to the slab fetching query to make this atomic.

ChrisSchinnerl:
This PR adds the `remote` subcommand for spinning up another indexer which fetches repair jobs from a primary indexer. 

It adds 2 new endpoints. One for fetching jobs and one for reporting results. To avoid reimplementing most of the migration code a lot of the logic is isolated into a new `Migrator` type which is used by both the main as well as the remote node.

Closes https://github.com/SiaFoundation/indexd/issues/1012


peterjan (dismissed):
I understand the appeal of running a remote indexer, but looking at Zeus it doesn’t seem like we’re close to saturating the available bandwidth. From what I can tell, we’re mostly I/O-bound and waiting on hosts, while CPU usage remains fairly relaxed.

Is the main reason for running the indexer separately to isolate migration or integrity-check throughput from other actions? It feels a bit like a shortcut around concurrency control, when a better scheduling/concurrency model would likely improve the overall throughput of migrations.

I'm sure you tested all of that I'm just asking 🤷‍♂️  currently the indexer has scaling knobs that focus on CPU rather than the dynamic scaling we've been adding to the SDK recently.

peterjan (dismissed):

cmd/indexd/remote.go
```diff
@@ -0,0 +1,111 @@
...
+	walletHash := types.HashBytes(walletKey[:])
+	if err := store.VerifyWalletKey(walletHash); errors.Is(err, wallet.ErrDifferentSeed) {
+		return errors.New("wallet seed does not match the primary node's database")
+	} else if err != nil {
+		return fmt.Errorf("failed to verify wallet key: %w", err)
+	}
+
+	alerter := alerts.NewManager()
+
+	am, err := accounts.NewManager(store, accounts.WithLogger(log.Named("accounts")))
```
chris124567: The accounts manager has a maintenance loop too (for `PruneAccounts`)
Did you check that it is fine to run from multiple instances?

ChrisSchinnerl:
Turning this back into a draft while I'm refactoring it to use an API approach rather than direct access to the database.

slabs/remote.go
```diff
@@ -0,0 +1,302 @@
...
+
+// Usable implements client.Store. It mirrors the primary node's DB-backed
+// store: the host must be known (the state only carries unblocked hosts) and
+// pass its usability checks, so recovery skips hosts the primary's own
+// recovery would skip.
+func (s *cachedHostStore) Usable(hostKey types.PublicKey) (bool, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	h, ok := s.hosts[hostKey]
+	return ok && h.IsGood(), nil
```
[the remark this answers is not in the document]
ChrisSchinnerl: That should be fine. I ran a test a few weeks back where we prefer bad hosts for migrations but the problem is that since they don't get funded, they don't end up getting used after the first couple of minutes.
So making that extra distinction doesn't seem to be worth it.

api/admin/admin_test.go
```diff
@@ -1781,6 +1781,27 @@ func TestPruneAccounts(t *testing.T) {
 	}
 }
 
+func TestMigrationBatchEmpty(t *testing.T) {
+	c := testutils.NewConsensusNode(t, zap.NewNop())
+	indexer := testutils.NewIndexer(t, c, zap.NewNop())
+	adminClient := indexer.Admin
+
+	// no unhealthy slabs on a fresh indexer
+	resp, err := adminClient.MigrationBatch(context.Background(), 0, 100)
```
peterjan:
```suggestion
	resp, err := adminClient.MigrationBatch(t.Context(), 0, 100)
```

slabs/downloads.go
```diff
@@ -219,29 +225,14 @@ func (r *slabRecovery) recoverChunk(ctx context.Context, offset, length uint64)
...
-		prices, err := m.hosts.Prices(ctx, hostKey)
-		if err != nil {
-			return fmt.Errorf("failed to fetch host prices: %w", err)
-		}
-
-		// debit the service account for the read since the host may charge for it
-		// even if it is cancelled quickly. This is best effort, it's fine to
-		// log the error and continue on failure.
-		cost := prices.RPCReadSectorCost(length).RenterCost()
-		if err = m.am.DebitServiceAccount(hostKey, m.migrationAccount, cost); err != nil {
```
peterjan, line 231: why is it ok to get rid of this? we never read from it but we no longer have correct internal accounting for the service accounts?

api/admin/admin.go
```diff
@@ -768,6 +780,32 @@ func (a *admin) handleGETState(jc jape.Context) {
...
+
+// handlePOSTMigrationResults persists the outcomes of migrations reported by a
+// remote node.
+func (a *admin) handlePOSTMigrationResults(jc jape.Context) {
+	var results []slabs.MigrationResult
+	if jc.Decode(&results) != nil {
+		return
+	}
+	a.slabs.ApplyMigrationResults(results)
+	jc.Encode(nil)
```
peterjan, line 883: this always returns 200 OK? even in case of db failure?

slabs/downloads.go
ChrisSchinnerl: Even if we leave this in we don't have correct tracking because the remote node uses the migration service account too. It seemed unnecessary to go through great lengths to track some spending accurately if we never consume it anyway. e.g. by funding a separate account for each remote node.

api/admin/admin.go
ChrisSchinnerl: Yeah I don't think the client has a good way to handle it anyway apart from logging a potential chain of errors that the main server already logs.

chris124567 (dismissed):

peterjan (dismissed):

chris124567:
Those changes were just from rebasing right?

api/admin/admin.go
peterjan: Could we not report the number of migration results that failed to be persisted and return an non 200 error if that number is not zero? I'm just thinking of the following unlikely scenario where reads would work but writes don't, I think the migrator would be repairing the same slabs over and over again at full throttle, because the backoff wouldn't be kicking in. I agree it's a bit of a `nit` but I think the way it is currently we have a backoff mechanism in the migrator for exactly problems such as this one, that now would never throttle down the speed with which the migrator migrates slabs.

peterjan (dismissed):

slabs/uploads.go
```diff
@@ -113,40 +118,17 @@ top:
...
-
-	// record the slab got migrated so object events gets updated
-	if err := m.store.RecordSlabMigrated(slab.ID); err != nil {
-		log.Debug("failed to record slab migration", zap.Error(err))
-	}
-	return int(migrated), nil
-}
-
-func (m *SlabManager) uploadShard(ctx context.Context, hostKey types.PublicKey, shard []byte) (rhp.RPCWriteSectorResult, error) {
-	usable, err := m.hm.Usable(ctx, hostKey)
```
chris124567, line 145: I feel like we should still check usability

ChrisSchinnerl: Usability should be checked at the time the upload state is prepared. So if a host becomes unusable it affects only one batch of slabs.

n8mgr (dismissed):

chris124567 (dismissed):

PR #1028, opened 2026-07-08
Unblock usable hosts that were automatically blocked due to being unusable

### LINKED ISSUE #1019: Unblock usable hosts that were automatically blocked due to being unusable
Right now we block hosts if they are unusable and have >0 contracts. These hosts should be fine to unblock if
1. they become usable again
2. they have 0 active contracts

Then they get another chance to be included in the set of contracts again assuming we still need more.

A host should only have the reasons discarded that were applied by indexd itself. e.g. if a host was blocked manually for "foo" and by indexd for "egress price" we only want to remove the "egress price" reason but keep "foo".

chris124567:
Close #1019

### DIFF
--- a/contracts/manager.go
+++ b/contracts/manager.go
@@ -416,10 +416,10 @@ func (cm *ContractManager) unblockUsableHosts(ctx context.Context) error {
 
 	for hk, reasons := range toUnblock {
 		if err := cm.hosts.RemoveBlocklistReasons(ctx, []types.PublicKey{hk}, reasons); err != nil {
-			log.Error("failed to unblock host", zap.Stringer("hostKey", hk), zap.Error(err))
+			log.Error("failed to remove usability blocklist reasons", zap.Stringer("hostKey", hk), zap.Error(err))
 			continue
 		}
-		log.Info("unblocking usable host", zap.Stringer("hostKey", hk), zap.Strings("usability", reasons))
+		log.Info("removed usability blocklist reasons", zap.Stringer("hostKey", hk), zap.Strings("usability", reasons))
 	}
 	return nil
 }
--- a/contracts/manager.go
+++ b/contracts/manager.go
@@ -100,7 +100,6 @@ type (
 		HostsForPruning(ctx context.Context) ([]types.PublicKey, error)
 		HostsForPinning(ctx context.Context) ([]types.PublicKey, error)
 		BlockHosts(ctx context.Context, hostKeys []types.PublicKey, reasons []string) error
-		RemoveBlocklistReasons(ctx context.Context, hostKeys []types.PublicKey, reasons []string) error
 		HostsWithUnpinnableSectors(ctx context.Context) ([]types.PublicKey, error)
 		UsabilitySettings(ctx context.Context) (hosts.UsabilitySettings, error)
 
@@ -370,47 +369,6 @@ func (cm *ContractManager) blockBadHosts(ctx context.Context) error {
 		}
 		log.Warn("blocking unusable host", zap.Stringer("hostKey", hk), zap.Strings("usability", reasons))
 	}
-	return nil
-}
-
-// unblockUsableHosts unblocks hosts that were blocked for failing usability
-// checks but that are usable again and no longer have any active contracts.
-// Only reasons that correspond to usability checks are removed; any other
-// blocklist reasons are left intact.
-func (cm *ContractManager) unblockUsableHosts(ctx context.Context) error {
-	ctx, cancel := context.WithTimeout(ctx, time.Minute)
-	defer cancel()
-	log := cm.log.Named("unblockhosts")
-
-	const batchSize = 100
-	var toUnblock []types.PublicKey
-	for offset := 0; ; offset += batchSize {
-		hosts, err := cm.hosts.Hosts(ctx, offset, batchSize,
-			hosts.WithBlocked(true),
-			hosts.WithUsable(true),
-			hosts.WithActiveContracts(false))
-		if err != nil {
-			return fmt.Errorf("failed to fetch hosts to unblock: %w", err)
-		}
-		for _, host := range hosts {
-			toUnblock = append(toUnblock, host.PublicKey)
-		}
-		if len(hosts) < batchSize {
-			break
-		}
-	}
-
-	if len(toUnblock) == 0 {
-		return nil
-	}
-
-	// removing reasons not present on a host is a no-op, so we can strip all
-	// usability reasons unconditionally. Non-usability reasons (e.g. manual
-	// blocks) are left intact.
-	if err := cm.hosts.RemoveBlocklistReasons(ctx, toUnblock, hosts.AllUsabilityChecks()); err != nil {
-		return fmt.Errorf("failed to remove usability blocklist reasons: %w", err)
-	}
-	log.Info("removed usability blocklist reasons", zap.Int("hosts", len(toUnblock)))
 	return nil
 }
 
--- a/contracts/manager.go
+++ b/contracts/manager.go
@@ -382,13 +382,8 @@ func (cm *ContractManager) unblockUsableHosts(ctx context.Context) error {
 	defer cancel()
 	log := cm.log.Named("unblockhosts")
 
-	usabilityChecks := make(map[string]struct{})
-	for _, name := range hosts.AllUsabilityChecks() {
-		usabilityChecks[name] = struct{}{}
-	}
-
 	const batchSize = 100
-	toUnblock := make(map[types.PublicKey][]string)
+	var toUnblock []types.PublicKey
 	for offset := 0; ; offset += batchSize {
 		hosts, err := cm.hosts.Hosts(ctx, offset, batchSize,
 			hosts.WithBlocked(true),
@@ -398,29 +393,24 @@ func (cm *ContractManager) unblockUsableHosts(ctx context.Context) error {
 			return fmt.Errorf("failed to fetch hosts to unblock: %w", err)
 		}
 		for _, host := range hosts {
-			// skip hosts blocked solely for non-usability reasons; we have nothing to remove
-			var remove []string
-			for _, r := range host.BlockedReasons {
-				if _, ok := usabilityChecks[r]; ok {
-					remove = append(remove, r)
-				}
-			}
-			if len(remove) > 0 {
-				toUnblock[host.PublicKey] = remove
-			}
+			toUnblock = append(toUnblock, host.PublicKey)
 		}
 		if len(hosts) < batchSize {
 			break
 		}
 	}
 
-	for hk, reasons := range toUnblock {
-		if err := cm.hosts.RemoveBlocklistReasons(ctx, []types.PublicKey{hk}, reasons); err != nil {
-			log.Error("failed to remove usability blocklist reasons", zap.Stringer("hostKey", hk), zap.Error(err))
-			continue
-		}
-		log.Info("removed usability blocklist reasons", zap.Stringer("hostKey", hk), zap.Strings("usability", reasons))
+	if len(toUnblock) == 0 {
+		return nil
 	}
+
+	// removing reasons not present on a host is a no-op, so we can strip all
+	// usability reasons unconditionally. Non-usability reasons (e.g. manual
+	// blocks) are left intact.
+	if err := cm.hosts.RemoveBlocklistReasons(ctx, toUnblock, hosts.AllUsabilityChecks()); err != nil {
+		return fmt.Errorf("failed to remove usability blocklist reasons: %w", err)
+	}
+	log.Info("removed usability blocklist reasons", zap.Int("hosts", len(toUnblock)))
 	return nil
 }
 
--- a/contracts/manager.go
+++ b/contracts/manager.go
@@ -100,6 +100,7 @@ type (
 		HostsForPruning(ctx context.Context) ([]types.PublicKey, error)
 		HostsForPinning(ctx context.Context) ([]types.PublicKey, error)
 		BlockHosts(ctx context.Context, hostKeys []types.PublicKey, reasons []string) error
+		RemoveBlocklistReasons(ctx context.Context, hostKeys []types.PublicKey, reasons []string) error
 		HostsWithUnpinnableSectors(ctx context.Context) ([]types.PublicKey, error)
 		UsabilitySettings(ctx context.Context) (hosts.UsabilitySettings, error)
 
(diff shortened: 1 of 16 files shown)


contracts/manager.go
```diff
@@ -372,6 +373,57 @@ func (cm *ContractManager) blockBadHosts(ctx context.Context) error {
 	return nil
 }
 
+// unblockUsableHosts unblocks hosts that were blocked for failing usability
+// checks but that are usable again and no longer have any active contracts.
+// Only reasons that correspond to usability checks are removed; any other
+// blocklist reasons are left intact.
+func (cm *ContractManager) unblockUsableHosts(ctx context.Context) error {
```
peterjan: Should we consider doing this fully in SQL? 

I was looking into simplifying or avoiding `RemoveBlocklistReasons` because that feels a little clunky, and then realised we can probably replace this whole thing with a query considering it's host filters + host blocklist update/delete. 

If you can get rid of that store method, you can also remove `fullyUnblockHost` again, which is fine I think because we only want to unblock hosts that no longer have any active contracts anyway. That should make this simpler overall.

Perhaps we should move this over to the host manager and trigger the query when we finished host scanning, because that is the trigger from a previously blocked host to became usable again.

peterjan (changes requested):

contracts/manager.go
peterjan: This is an open question by the way, not everything has to be a query, sometimes it's more readable if you do it from code but in this case I think the query might be worth it. It would look something like this:

```
WITH candidates AS (
	SELECT
		h.id AS host_id,
		hb.public_key,
		hb.reasons AS old_reasons,

		ARRAY(
			SELECT reason
			FROM unnest(hb.reasons) AS reason
			WHERE reason <> ALL($1::text[])
		) AS new_reasons,

		ARRAY(
			SELECT reason
			FROM unnest(hb.reasons) AS reason
			WHERE reason = ANY($1::text[])
		) AS removed_reasons
	FROM hosts_blocklist AS hb
	JOIN hosts AS h ON h.public_key = hb.public_key
	CROSS JOIN globals
	WHERE hb.reasons && $1::text[]

	-- usable host filter
	AND recent_uptime >= 0.9
	AND has_settings
	AND settings_max_contract_duration >= globals.contracts_period
	AND settings_max_collateral >= settings_collateral * globals.one_tb * globals.contracts_period
	AND settings_version >= globals.host_min_version
	AND settings_valid_until >= last_successful_scan + INTERVAL '15 minutes'
	AND settings_accepting_contracts
	AND has_quic
	AND has_siamux
	AND settings_contract_price <= globals.one_sc
	AND settings_collateral >= globals.hosts_min_collateral
	AND settings_collateral >= 2 * settings_storage_price
	AND settings_storage_price <= globals.hosts_max_storage_price
	AND settings_ingress_price <= globals.hosts_max_ingress_price
	AND settings_egress_price <= globals.hosts_max_egress_price
	AND settings_free_sector_price <= globals.one_sc / globals.sectors_per_tb

	-- no active contracts
	AND NOT EXISTS (
		SELECT 1
		FROM contracts AS c
		WHERE c.host_id = h.id
		  AND c.state IN (0, 1)
		  AND c.renewed_to IS NULL
		  AND c.proof_height > globals.scanned_height
	)
),

deleted AS (
	DELETE FROM hosts_blocklist AS hb
	USING candidates AS c
	WHERE hb.public_key = c.public_key
	  AND cardinality(c.new_reasons) = 0
	RETURNING
		hb.public_key,
		c.old_reasons,
		c.removed_reasons,
		ARRAY[]::text[] AS new_reasons,
		'deleted'::text AS action
),

updated AS (
	UPDATE hosts_blocklist AS hb
	SET reasons = c.new_reasons
	FROM candidates AS c
	WHERE hb.public_key = c.public_key
	  AND cardinality(c.new_reasons) > 0
	RETURNING
		hb.public_key,
		c.old_reasons,
		c.removed_reasons,
		hb.reasons AS new_reasons,
		'updated'::text AS action
)

SELECT * FROM updated
UNION ALL
SELECT * FROM deleted;
```

chris124567: That would simplify things quite a bit though I don't know if readability improves that much with a giant query.  I went this route partially because it's how the existing `blockBadHosts` works.  cc @ChrisSchinnerl

ChrisSchinnerl: I do agree that the host manager is probably the right spot for this rather than the contract manager.

I'm pretty indifferent about whether we go with a single specialized query or a fetch hosts + remove reasons approach. Although the former is probably not a single query either since it misses the part where we mark contracts good again.

This PR could probably be cleaned up a bit doing the following though:

1. `unblockUsableHosts` could just always remove all usability reasons instead of checking which ones a host has. 
2. `RemoveUnblockReasons` can perform its update in a batch rather than in a loop

Especially the calling code becomes shorter that way. The database code remains similar but without the loop. 

e.g. 
```
diff --git a/contracts/manager.go b/contracts/manager.go
index eb786918..f3fb71d4 100644
--- a/contracts/manager.go
+++ b/contracts/manager.go
@@ -382,13 +382,8 @@ func (cm *ContractManager) unblockUsableHosts(ctx context.Context) error {
 	defer cancel()
 	log := cm.log.Named("unblockhosts")
 
-	usabilityChecks := make(map[string]struct{})
-	for _, name := range hosts.AllUsabilityChecks() {
-		usabilityChecks[name] = struct{}{}
-	}
-
 	const batchSize = 100
-	toUnblock := make(map[types.PublicKey][]string)
+	var toUnblock []types.PublicKey
 		hosts, err := cm.hosts.Hosts(ctx, offset, batchSize,
 			hosts.WithBlocked(true),
@@ -398,29 +393,24 @@ func (cm *ContractManager) unblockUsableHosts(ctx context.Context) error {
 			return fmt.Errorf("failed to fetch hosts to unblock: %w", err)
 		}
 		for _, host := range hosts {
-			// skip hosts blocked solely for non-usability reasons; we have nothing to remove
-			var remove []string
-			for _, r := range host.BlockedReasons {
-				if _, ok := usabilityChecks[r]; ok {
-					remove = append(remove, r)
-				}
-			}
-			if len(remove) > 0 {
-				toUnblock[host.PublicKey] = remove
-			}
+			toUnblock = append(toUnblock, host.PublicKey)
 		}
 		if len(hosts) < batchSize {
 			break
 		}
 	}
 
-	for hk, reasons := range toUnblock {
-		if err := cm.hosts.RemoveBlocklistReasons(ctx, []types.PublicKey{hk}, reasons); err != nil {
-			log.Error("failed to remove usability blocklist reasons", zap.Stringer("hostKey", hk), zap.Error(err))
-			continue
-		}
-		log.Info("removed usability blocklist reasons", zap.Stringer("hostKey", hk), zap.Strings("usability", reasons))
+	if len(toUnblock) == 0 {
+		return nil
+	}
+
+	// Removing reasons not present on a host is a no-op, so we can strip all
+	// usability reasons unconditionally. Non-usability reasons (e.g. manual
+	// blocks) are left intact.
+	if err := cm.hosts.RemoveBlocklistReasons(ctx, toUnblock, hosts.AllUsabilityChecks()); err != nil {
+		return fmt.Errorf("failed to remove usability blocklist reasons: %w", err)
 	}
+	log.Info("removed usability blocklist reasons", zap.Int("hosts", len(toUnblock)))
 	return nil
 }
 
diff --git a/persist/postgres/hosts.go b/persist/postgres/hosts.go
index 2752915c..64240205 100644
--- a/persist/postgres/hosts.go
+++ b/persist/postgres/hosts.go
@@ -426,49 +426,62 @@ func (s *Store) UnblockHost(hk types.PublicKey) error {
 // entry, leaving any other reasons untouched. A host left with no remaining
 // reasons is removed from the blocklist and its contracts are marked good again.
 func (s *Store) RemoveBlocklistReasons(hks []types.PublicKey, reasons []string) error {
+	if len(hks) == 0 || len(reasons) == 0 {
+		return nil
+	}
+	sqlHks := make([]sqlPublicKey, len(hks))
+	for i, hk := range hks {
+		sqlHks[i] = sqlPublicKey(hk)
+	}
 	return s.transaction(func(ctx context.Context, tx *txn) error {
-		for _, hk := range hks {
-			var current []string
-			err := tx.QueryRow(ctx, `
-				SELECT reasons
-				FROM hosts_blocklist
-				WHERE public_key = $1`, sqlPublicKey(hk)).Scan(&current)
-			if errors.Is(err, sql.ErrNoRows) {
-				continue // host not blocked
-			} else if err != nil {
-				return fmt.Errorf("failed to fetch blocklist entry for host %q: %w", hk, err)
-			}
-
-			// keep any reasons not being removed
-			var remaining []string
-			changed := false
-			for _, r := range current {
-				if slices.Contains(reasons, r) {
-					changed = true
-					continue
-				}
-				remaining = append(remaining, r)
-			}
-
-			if !changed {
-				continue // nothing to remove
+		// Delete the entries that would be left with no reasons at all, i.e.
+		// where none of the current reasons survive the removal. Collect their
+		// host keys so we can mark their contracts good again.
+			DELETE FROM hosts_blocklist
+			WHERE public_key = ANY($1)
+			  AND NOT EXISTS (
+			      SELECT 1 FROM unnest(reasons) AS r WHERE r <> ALL($2::text[])
+			  )
+			RETURNING public_key`, sqlHks, reasons)
+		if err != nil {
+			return fmt.Errorf("failed to remove fully-unblocked hosts: %w", err)
+		}
+		var unblocked []sqlPublicKey
+		for rows.Next() {
+			var hk sqlPublicKey
+			if err := rows.Scan(&hk); err != nil {
+				rows.Close()
+				return err
 			}
+			unblocked = append(unblocked, hk)
+		}
+		rows.Close()
+			return err
+		}
 
-			if len(remaining) == 0 {
-				// no reasons remain, fully unblock the host
-				if err := fullyUnblockHost(ctx, tx, hk); err != nil {
-					return err
-				}
-				continue
+		// Mark the fully-unblocked hosts' contracts good again.
+		if len(unblocked) > 0 {
+			if _, err := tx.Exec(ctx, `
+				UPDATE contracts c
+				SET good = TRUE
+				FROM hosts h
+				WHERE c.host_id = h.id AND h.public_key = ANY($1)`, unblocked); err != nil {
+				return fmt.Errorf("failed to mark contracts good: %w", err)
 			}
+		}
 
-			_, err = tx.Exec(ctx, `
-				UPDATE hosts_blocklist
-				SET reasons = $2
-				WHERE public_key = $1`, sqlPublicKey(hk), remaining)
-			if err != nil {
-				return fmt.Errorf("failed to update blocklist entry for host %q: %w", hk, err)
-			}
+		// Strip the reasons from the entries that still have some left; the
+		// fully-emptied entries were already deleted above. Reasons not present
+		// on an entry are a no-op.
+			UPDATE hosts_blocklist
+			SET reasons = ARRAY(
+				SELECT r FROM unnest(reasons) AS r WHERE r <> ALL($2::text[])
+			)
+			WHERE public_key = ANY($1)`, sqlHks, reasons); err != nil {
+			return fmt.Errorf("failed to update blocklist reasons: %w", err)
 		}
 		return nil
 	})
```

ChrisSchinnerl: Another approach that would clean things up a lot is normalizing the block reasons. So that instead of an array we use one row per reason. This might slow down the hot read path though so I'm not sure it's worth it. But the clear advantage is that the queries get a lot simpler.

ChrisSchinnerl (changes requested):

contracts/manager.go
chris124567: OK I applied that and moved the logic to the host manager.  Maybe we can do that in a followup if performance hit isn't too big

peterjan (dismissed):

persist/postgres/hosts.go
```diff
@@ -395,23 +395,93 @@ func (s *Store) HostsWithUnpinnableSectors() ([]types.PublicKey, error) {
...
+	}
+	return nil
+}
+
 // UnblockHost removes the given host key from the blocklist and marks its
 // contracts as good again.
 func (s *Store) UnblockHost(hk types.PublicKey) error {
 	return s.transaction(func(ctx context.Context, tx *txn) error {
-		_, err := tx.Exec(ctx, "DELETE FROM hosts_blocklist WHERE public_key = $1", sqlPublicKey(hk))
+		return fullyUnblockHost(ctx, tx, hk)
```
ChrisSchinnerl: This can be inlined again since it is only called in one spot

hosts/manager.go
```diff
@@ -630,6 +631,53 @@ loop:
...
+			toUnblock = append(toUnblock, host.PublicKey)
+		}
+		if len(hosts) < batchSize {
+			break
+		}
+	}
+
+	if len(toUnblock) == 0 {
+		return nil
+	} else if err := ctx.Err(); err != nil {
```
ChrisSchinnerl: The context is only used here as a check between 2 functions that should be pretty fast. I'd remove it as an argument.

hosts/manager.go
```diff
@@ -630,6 +631,53 @@ loop:
 	wg.Wait()
 
 	m.log.Debug("host scans finished", zap.Int("hosts", len(hosts)), zap.Duration("duration", time.Since(start)))
+	if err := m.unblockUsableHosts(ctx); err != nil {
+		m.log.Error("failed to unblock usable hosts after scanning", zap.Error(err))
+	}
```
ChrisSchinnerl, line 638: (nit) might be nice to return the number of actually unblocked hosts to log them here on DEBUG level.

ChrisSchinnerl (changes requested):
Tiny cleanup would be nice but apart from that it looks good. Also seems to work as expected on Zeus right now.

hosts/manager.go
```diff
@@ -630,6 +631,52 @@ loop:
 	wg.Wait()
 
 	m.log.Debug("host scans finished", zap.Int("hosts", len(hosts)), zap.Duration("duration", time.Since(start)))
+	if unblocked, err := m.unblockUsableHosts(ctx); err != nil {
+		m.log.Error("failed to unblock usable hosts after scanning", zap.Error(err))
+	} else {
```
peterjan: `nit` `else if unblocked > 0` ?

peterjan (dismissed):

hosts/manager.go
```diff
@@ -630,6 +631,52 @@ loop:
...
+	if len(toUnblock) == 0 {
+		return 0, nil
+	}
+
+	// removing reasons not present on a host is a no-op, so strip all usability
+	// reasons unconditionally while preserving manual blocklist reasons.
+	if err := m.store.RemoveBlocklistReasons(toUnblock, Usability{}.FailedChecks()); err != nil {
+		return 0, fmt.Errorf("failed to remove usability blocklist reasons: %w", err)
+	}
+	return len(toUnblock), nil
```
ChrisSchinnerl: Just because the host was in `toUnblock` doesn't mean it was actually unblocked right? Only `RemoveBlocklistReasons` knows if all reasons were removed.

hosts/manager.go
```diff
@@ -630,6 +631,53 @@ loop:
...
+		m.log.Debug("unblocked usable hosts", zap.Int("hosts", unblocked))
+	}
+}
+
+// unblockUsableHosts unblocks hosts that were blocked for failing usability
+// checks but that are usable again and no longer have any active contracts.
+// Only reasons that correspond to usability checks are removed; any other
+// blocklist reasons are left intact.
+func (m *HostManager) unblockUsableHosts(ctx context.Context) (int, error) {
+	ctx, cancel := context.WithTimeout(ctx, time.Minute)
```
peterjan: what's the reason for this arbitrary timeout here?

peterjan (dismissed):

hosts/manager.go
ChrisSchinnerl: It do be a bit strange.

ChrisSchinnerl (approved):

PR #1036, opened 2026-07-14
Support pre-authorized keys

n8mgr:
Pre-authorized keys let developers connect apps without requiring users to complete an interactive approval flow first.

### DIFF
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -57,8 +57,8 @@ type (
 	Accounts interface {
 		ValidAppConnectKey(context.Context, string) error
 		RegisterAppKey(string, types.PublicKey, accounts.AppMeta) error
-		AppSecret(connectKey string, appID types.Hash256) (types.Hash256, error)
-		HasAppAccount(connectKey string, appID types.Hash256) (bool, error)
+		AuthorizeAppConnectKey(connectKey string, appID types.Hash256) (accounts.AppAuthorization, error)
+		AuthorizePreAuthorizedKey(publicKey types.PublicKey, appID types.Hash256) (accounts.AppAuthorization, error)
 
 		HasAccount(context.Context, types.PublicKey) (bool, error)
 		Account(context.Context, types.PublicKey) (accounts.Account, error)
@@ -95,6 +95,11 @@ type (
 		LogoURL     string        `json:"logoURL"`
 		ServiceURL  string        `json:"serviceURL"`
 		CallbackURL string        `json:"callbackURL"`
+
+		// PreAuthorizedKey and PreAuthorizationSignature allow the request to
+		// bypass interactive approval when signed by a pre-authorized key.
+		PreAuthorizedKey          types.PublicKey `json:"preAuthorizedKey,omitzero"`
+		PreAuthorizationSignature types.Signature `json:"preAuthorizationSignature,omitzero"`
 	}
 
 	// AuthConnectStatusResponse is the response body for checking the status of an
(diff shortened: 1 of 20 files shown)


persist/postgres/apps.go
```diff
@@ -167,6 +191,177 @@ func (s *Store) AppConnectKeys(offset, limit int) (keys []accounts.ConnectKey, e
...
+}
+
+// ConsumePreAuthorizedKey validates and consumes one use of a pre-authorized
+// key. The update and authorization lookup are atomic, preventing concurrent
+// callers from exceeding the key's use limit.
+func (s *Store) ConsumePreAuthorizedKey(publicKey types.PublicKey, appID types.Hash256) (connectKey string, userSecret types.Hash256, reconnecting bool, err error) {
+	err = s.transaction(func(ctx context.Context, tx *txn) error {
+		err := tx.QueryRow(ctx, `
+			UPDATE preauthorized_keys pak
+			SET remaining_uses = pak.remaining_uses - 1, last_used = NOW()
```
peterjan, line 298: should reconnecting consume a use? with a normal connect key re-auth is free, even when the key is exhausted, but here an app that already has an account still burns a use when it redoes the handshake...

n8mgr: I think so. I feel like the app limit is a different thing entirely. If you limit it like we do apps, developers cant be sure a one time key is actually a one time key. If you want to allow reconnection, issue a new key, set a high limit, or go through the interactive flow.

api/app/client.go
```diff
@@ -299,7 +315,21 @@ func (c *Client) SharedObject(ctx context.Context, sharedURL string) (slabs.Shar
...
+	var opts requestAppConnectionOptions
+	for _, option := range options {
+		option(&opts)
+	}
+	if opts.preAuthorizedKey != nil && len(opts.preAuthorizedKey) != ed25519.PrivateKeySize {
+		return RegisterAppResponse{}, errors.New("invalid pre-authorized private key")
+	}
+
+	if opts.preAuthorizedKey != nil {
+		request.PreAuthorizedKey = opts.preAuthorizedKey.PublicKey()
```
ChrisSchinnerl, line 329: This is a bit odd considering that `PreAuthorizedKey` is an exported field on the request but nothing we expect the user to set explicitly and instead have them go through `WithPreAuthorizedKey` which then overwrites the `PreAuthorizedKey` here.

What if we changed the `request` parameter to an `AppInfo` type that doesn't have `PreAuthorizedKey` and `PreAuthorizationSignature` so that users don't get the option to set it and then we create the `RegisterAppRequest` in this method.

api/app/app.go
```diff
@@ -508,13 +513,47 @@ func (a *app) handleAuthRequest(jc jape.Context) {
...
+			a.log.Debug("failed to authorize pre-authorized key", zap.Error(err))
+			jc.Error(ErrInternalError, http.StatusInternalServerError)
+			return
+		}
+
+		authRequest.Approved = true
+		authRequest.Reconnecting = authorization.Reconnecting
+		authRequest.UserSecret = authorization.UserSecret
+		authRequest.ConnectKey = authorization.ConnectKey
+	}
```
ChrisSchinnerl, line 558: Reminder that this handler is never actually hit when going through Sia Storage because it reimplements the handshake and then just calls an admin route to add the app to the indexer.

Which makes me wonder if the indexer even needs pre-authorized keys. If you control the indexer you can also create a connect key with 1 use or something for someone to connect an app. It's really more useful for Sia Storage.

n8mgr: Sia Storage and indexd expose the same API surface. We're not changing that.

ChrisSchinnerl (approved):

peterjan (approved):

PR #1038, opened 2026-07-15
Clamp value of failureRate at sane epsilon to make sure hosts go back to 0 failures eventually

chris124567:
Close #1035

### DIFF
--- a/client/v2/hostqueue.go
+++ b/client/v2/hostqueue.go
@@ -88,19 +88,20 @@ func (fr *failureRate) AddSample(success bool) {
 		}
 		fr.value = emaAlpha*sample + (1.0-emaAlpha)*fr.value
 	}
-	if fr.value < failureRateZeroThreshold {
-		fr.value = 0
-	}
 	fr.lastDecay = now
 }
 
 // Value returns the failure rate, decaying it for idle time since the last
-// sample. Mutates fr; not safe for concurrent use.
+// sample and reporting rates below failureRateZeroThreshold as zero. Mutates
+// fr; not safe for concurrent use.
 func (fr *failureRate) Value() float64 {
 	if fr.init && fr.value != 0 {
 		if elapsed := time.Since(fr.lastDecay); elapsed >= failureRateHalfLife {
 			fr.decay(elapsed)
 		}
+	}
+	if fr.value < failureRateZeroThreshold {
+		return 0
 	}
 	return fr.value
 }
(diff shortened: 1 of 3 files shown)


client/v2/hostqueue.go
```diff
@@ -72,20 +83,40 @@ func (fr *failureRate) AddSample(success bool) {
...
 }
 
+// decay halves the failure rate for every idle half-life. This decay schedule is
+// independent of the sample EMA (emaAlpha), though emaAlpha still affects the
+// rate immediately after samples are applied.
+func (fr *failureRate) decay(elapsed time.Duration) {
+	halfLives := int(elapsed / failureRateHalfLife)
+	fr.value = math.Ldexp(fr.value, -halfLives)
+	fr.lastDecay = fr.lastDecay.Add(time.Duration(halfLives) * failureRateHalfLife)
+	if fr.value < failureRateZeroThreshold {
```
peterjan: `nit` is this redundant? we already do this in AddSample as well

peterjan (dismissed):
LGTM

client/v2/hostqueue.go
```diff
@@ -72,20 +83,40 @@ func (fr *failureRate) AddSample(success bool) {
...
 		fr.init = true
 	} else {
+		if elapsed := now.Sub(fr.lastDecay); fr.value != 0 && elapsed >= failureRateHalfLife {
+			fr.decay(elapsed)
+		}
 		fr.value = emaAlpha*sample + (1.0-emaAlpha)*fr.value
 	}
-	fr.lastAttempt = time.Now()
+	if fr.value < failureRateZeroThreshold {
+		fr.value = 0
```
ChrisSchinnerl: I'd not actually set it to 0 and instead just return the clamped value in `Value` if it is below the threshold to avoid influencing the actual decay and ema.

client/v2/hostqueue.go
ChrisSchinnerl: same thing. Let's just clamp in `Value`.

ChrisSchinnerl (changes requested):

client/v2/hostqueue.go
```diff
@@ -72,20 +83,38 @@ func (fr *failureRate) AddSample(success bool) {
 		fr.value = sample
 		fr.init = true
 	} else {
+		if elapsed := now.Sub(fr.lastDecay); fr.value != 0 && elapsed >= failureRateHalfLife {
+			fr.decay(elapsed)
+		}
 		fr.value = emaAlpha*sample + (1.0-emaAlpha)*fr.value
 	}
-	fr.lastAttempt = time.Now()
+	fr.lastDecay = now
```
ChrisSchinnerl: Can't you drop this line now since `fr.decay` sets `lastDecay`? Having it here as well will effectively prevent a decay from happening as long as samples come in. 

I'd also consider moving the `if` condition that the decay is wrapped both times into `decay` and return early if the time for a decay hasn't come yet or the value hasn't been initialized yet.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

PR #1040, opened 2026-07-17
Take into account whether contract reached proof height when marking sector lost

ChrisSchinnerl:
Once a contract reaches its proof height a host can submit a storage proof and therefore fulfill their data storage obligation. When the host then drops the sectors of that contract while we also repair slabs that contain those sectors we will attempt to use that host for retrieving the sectors. The host will report that they don't have the sectors anymore and the indexer will mark the sectors as lost.

This is fixed by this PR by further restricting `MarkSectorsLost` to only mark a sector as lost if it doesn't belong to a contract that hasn't reached its proof height yet. Once it does, we have to expect that a host might drop the sectors at any moment. 

Closes https://github.com/SiaFoundation/indexd/issues/1030

### DIFF
--- a/.changeset/dont_consider_sectors_lost_when_they_are_pinned_to_a_contract_beyond_its_proof_height.md
+++ b/.changeset/dont_consider_sectors_lost_when_they_are_pinned_to_a_contract_beyond_its_proof_height.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+# Don't consider sectors lost when they are pinned to a contract beyond its proof height.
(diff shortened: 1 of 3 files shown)


peterjan (approved):
looks good, also ran the benchmark to verify we weren't introducing any db regression but seems fine, it even improved a bit allocations wise

peterjan (changes requested):
hm actually, looking at `updateSectorStats` I don't think we're updating the global stats counter correctly, because it passes unpinnableDelta into incrementNumSectorsLost, so `num_sectors_lost` still counts the post-proof-height sectors we just excluded from the host counter

chris124567 (commented):
Shouldn’t `PruneContractSectorsMap` also set `host_id=NULL`? Otherwise the expired sectors become unpinned, and if the host has deleted them, MarkSectorsLost will still count them against the host.

ChrisSchinnerl:
> I think a similar change is required for `DeleteContract` too. Other than that this LGTM

`DeleteContract` intentionally keeps the host id because we use it mostly to delete stuck contracts (that fail to refresh for silly reasons) and repin the data to another contract with the same host. To avoid reuploading it all to a new host.

PR #1041, opened 2026-07-21
Check tip height of host against local tip height when appending sectors

ChrisSchinnerl:
This PR adds a sanity check before appending sectors to make sure a host's height isn't more than 6 hours in the past relative to our synced height.

### DIFF
--- a/client/v2/contracts.go
+++ b/client/v2/contracts.go
@@ -186,16 +186,17 @@ func (c *Client) AppendSectors(ctx context.Context, signer rhp.ContractSigner, c
 		if err != nil {
 			return fmt.Errorf("failed to get host prices: %w", err)
 		}
-		if c.checkPricesHeight(settings.Prices, chain.TipState().Index.Height) != nil {
+		tipState := chain.TipState()
+		if c.checkPricesHeight(settings.Prices, tipState.Index.Height) != nil {
 			settings, err = c.refreshSettings(ctx, hostKey, transport)
 			if err != nil {
 				return fmt.Errorf("failed to refresh host prices: %w", err)
 			}
 		}
-		if err := c.checkPricesHeight(settings.Prices, chain.TipState().Index.Height); err != nil {
+		if err := c.checkPricesHeight(settings.Prices, tipState.Index.Height); err != nil {
 			return err
 		}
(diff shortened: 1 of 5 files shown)


n8mgr (dismissed):

client/v2/contracts.go
```diff
@@ -181,11 +181,22 @@ func (c *Client) AppendSectors(ctx context.Context, signer rhp.ContractSigner, c
...
-		settings, _, err := c.settings(ctx, revision.Revision.HostPublicKey, transport)
+		hostKey := revision.Revision.HostPublicKey
+		settings, _, err := c.settings(ctx, hostKey, transport)
 		if err != nil {
 			return fmt.Errorf("failed to get host prices: %w", err)
 		}
-		res, err = rhp.RPCAppendSectors(ctx, transport, signer, chain.TipState(), settings.Prices, revision, sectors)
+		tipState := chain.TipState()
+		if c.checkPricesHeight(settings.Prices, tipState.Index.Height) != nil {
+			settings, err = c.refreshSettings(ctx, hostKey, transport)
```
chris124567, line 191: `settings` will refresh for you if it's more than 30s old... given that we have 6 hours of tolerance I would say we probably don't need to refresh here

peterjan (commented):
aside from Christopher's remark this LGTM

client/v2/contracts.go
chris124567: Ok actually I was a bit off it will refresh if there is less than thirty seconds until ValidUntil and in hostd price tables are valid for 30 minutes by default.  I guess you can make a better case for refreshing then

peterjan (approved):

n8mgr:
@peterjan did you approve while also requesting changes?

peterjan:
> @peterjan did you approve while also requesting changes?

on the [other PR,](https://github.com/SiaFoundation/indexd/pull/1040) yes... I did a double pass on an open tab and noticed the stats issue. This one I approved, right? 

edit: oh if you mean "aside from Christopher's remark this LGTM" - that seems to have resolved itself, I though this comment made sense but the timeout was incorrect and Christopher ended up resolving it

PR #1043, opened 2026-07-21
Prune slabs after deleting an object 

ChrisSchinnerl:
This PR causes slabs of an object to be scheduled for deletion after the object was deleted.

It also closes a gap we had where it was possible to unpin a slab that was currently referenced by an object.

Reviewers please pay close attention to the locking. When reference counting is involved things get a bit tricky. There is a good amount of test coverage for concurrent operations but still.

### DIFF
--- a/persist/postgres/sectors.go
+++ b/persist/postgres/sectors.go
@@ -702,9 +702,9 @@ func (s *Store) UnpinSlab(account proto.Account, slabID slabs.SlabID) error {
 			return fmt.Errorf("failed to get account ID: %w", err)
 		}
 
-		// lock the slab before any checks so they serialize against
+		// lock the slab before the checks so they serialize against
 		// concurrent unpins and against PinObject attaching the slab to a
-		// new object; the checks run as separate statements after the lock,
+		// new object; the checks run in a separate statement after the lock,
 		// so they see the state committed by whoever held the lock before us
 		var sID int64
 		err = tx.QueryRow(ctx, `SELECT id FROM slabs WHERE digest = $1 FOR UPDATE`, sqlHash256(slabID)).Scan(&sID)
@@ -714,28 +714,23 @@ func (s *Store) UnpinSlab(account proto.Account, slabID slabs.SlabID) error {
 			return fmt.Errorf("failed to lock slab: %w", err)
 		}
 
-		var pinned bool
-		err = tx.QueryRow(ctx, `SELECT EXISTS (
-			SELECT 1 FROM account_slabs WHERE account_id = $1 AND slab_id = $2
-		)`, id, sID).Scan(&pinned)
+		// the slab must still be pinned by the account and must not be
+		// referenced by one of its objects; refusing the unpin while an object
+		// references the slab stops the account from freeing quota while the
+		// object keeps the slab alive
+		var pinned, inUse bool
+		err = tx.QueryRow(ctx, `SELECT
+			EXISTS (SELECT 1 FROM account_slabs WHERE account_id = $1 AND slab_id = $2),
+			EXISTS (
+				SELECT 1
+				FROM objects o
+				JOIN object_slabs os ON o.id = os.object_id
+				WHERE o.account_id = $1 AND os.slab_digest = $3
+			)`, id, sID, sqlHash256(slabID)).Scan(&pinned, &inUse)
 		if err != nil {
-			return fmt.Errorf("failed to check if slab is pinned: %w", err)
+			return fmt.Errorf("failed to check slab pin state: %w", err)
 		} else if !pinned {
 			return slabs.ErrSlabNotFound
-		}
-
-		// refuse to unpin a slab that is still referenced by one of the
-		// account's objects; otherwise the account would free up quota while
-		// the object keeps the slab alive
-		var inUse bool
-		err = tx.QueryRow(ctx, `SELECT EXISTS (
-			SELECT 1
-			FROM objects o
-			JOIN object_slabs os ON o.id = os.object_id
-			WHERE o.account_id = $1 AND os.slab_digest = $2
-		)`, id, sqlHash256(slabID)).Scan(&inUse)
-		if err != nil {
-			return fmt.Errorf("failed to check if slab is in use: %w", err)
 		} else if inUse {
 			return slabs.ErrSlabInUse
 		}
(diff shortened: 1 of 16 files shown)


chris124567:
@ChrisSchinnerl

```
func TestABC(t *testing.T) {
    store := initPostgres(t, zap.NewNop())

    acc := proto.Account{1}
    store.addTestAccount(t, types.PublicKey(acc))

    hk := store.addTestHost(t)
    store.addTestContract(t, hk)

    slab := newTestSlab(hk)
    store.pinTestSlabs(t, acc, slab)
    obj := store.pinRandomObject(t, acc, []slabs.SlabSlice{slab.Slice(0, 100)})

    if err := store.DeleteAccount(acc); err != nil {
        t.Fatal(err)
    }

    // without special handling for soft-deleted accounts, this tries to reclaim
    // the connect-key usage a second time and violates its nonnegative constraint
    if err := store.DeleteObject(acc, obj.ID()); err != nil {
        t.Fatal(err)
    }
}
```

```
$ POSTGRES_USER=postgres POSTGRES_PASSWORD=postgres POSTGRES_DB=postgres go test ./persist/postgres -v -count=1 -run="TestABC"
=== RUN   TestABC
    objects_test.go:1007: failed to update connect key pinned data: ERROR: new row for relation "app_connect_keys" violates check constraint "app_connect_keys_pinned_data_check" (SQLSTATE 23514)
--- FAIL: TestABC (0.77s)
FAIL
FAIL    go.sia.tech/indexd/persist/postgres     0.779s
FAIL
```

Fix:
```
index 0ffe3762..91308290 100644
@@ -551,15 +551,19 @@ RETURNING a.slab_id`, accountID, sIDs)
        }
 
        var connectKeyID int64
+       var active bool
        err = tx.QueryRow(ctx, `UPDATE accounts
 SET pinned_data = pinned_data - $1, pinned_size = pinned_size - $2
 WHERE id = $3
-RETURNING connect_key_id`, pinnedDataDelta, pinnedSizeDelta, accountID).Scan(&connectKeyID)
+RETURNING connect_key_id, deleted_at IS NULL`, pinnedDataDelta, pinnedSizeDelta, accountID).Scan(&connectKeyID, &active)
        if err != nil {
                return fmt.Errorf("failed to update account's pinned data: %w", err)
        }
-       if _, err := tx.Exec(ctx, `UPDATE app_connect_keys SET pinned_data = pinned_data - $1, pinned_size = pinned_size - $2 WHERE id = $3`, pinnedDataDelta, pinnedSizeDelta, connectKeyID); err != nil {
-               return fmt.Errorf("failed to update connect key pinned data: %w", err)
+
+       if active {
+               if _, err := tx.Exec(ctx, `UPDATE app_connect_keys SET pinned_data = pinned_data - $1, pinned_size = pinned_size - $2 WHERE id = $3`, pinnedDataDelta, pinnedSizeDelta, connectKeyID); err != nil {
+                       return fmt.Errorf("failed to update connect key pinned data: %w", err)
+               }
        }
 
        // queue the slabs for background deletion; the queue is append-only so a
```

n8mgr:
@chris124567 Chris is OOO, mind pushing your diff?

persist/postgres/sectors.go
```diff
@@ -608,22 +690,54 @@ func (s *Store) deleteOrphanedSlabs(ctx context.Context, tx *txn, sIDs []int64)
...
+		err = tx.QueryRow(ctx, `SELECT id FROM slabs WHERE digest = $1 FOR UPDATE`, sqlHash256(slabID)).Scan(&sID)
 		if errors.Is(err, sql.ErrNoRows) {
 			return slabs.ErrSlabNotFound
 		} else if err != nil {
-			return fmt.Errorf("failed to check if slab exists: %w", err)
+			return fmt.Errorf("failed to lock slab: %w", err)
+		}
+
+		var pinned bool
+		err = tx.QueryRow(ctx, `SELECT EXISTS (
```
peterjan: we could save a round trip here and do the two exists checks in a single query

```
index 91308290..85d84247 100644
@@ -702,9 +702,9 @@ func (s *Store) UnpinSlab(account proto.Account, slabID slabs.SlabID) error {
 			return fmt.Errorf("failed to get account ID: %w", err)
 		}
 
-		// lock the slab before any checks so they serialize against
+		// lock the slab before the checks so they serialize against
 		// concurrent unpins and against PinObject attaching the slab to a
-		// new object; the checks run as separate statements after the lock,
+		// new object; the checks run in a separate statement after the lock,
 		// so they see the state committed by whoever held the lock before us
 		var sID int64
 		err = tx.QueryRow(ctx, `SELECT id FROM slabs WHERE digest = $1 FOR UPDATE`, sqlHash256(slabID)).Scan(&sID)
@@ -714,28 +714,23 @@ func (s *Store) UnpinSlab(account proto.Account, slabID slabs.SlabID) error {
 			return fmt.Errorf("failed to lock slab: %w", err)
 		}
 
-		var pinned bool
-		err = tx.QueryRow(ctx, `SELECT EXISTS (
-			SELECT 1 FROM account_slabs WHERE account_id = $1 AND slab_id = $2
-		)`, id, sID).Scan(&pinned)
+		// the slab must still be pinned by the account and must not be
+		// referenced by one of its objects; refusing the unpin while an object
+		// references the slab stops the account from freeing quota while the
+		// object keeps the slab alive
+		var pinned, inUse bool
+		err = tx.QueryRow(ctx, `SELECT
+			EXISTS (SELECT 1 FROM account_slabs WHERE account_id = $1 AND slab_id = $2),
+			EXISTS (
+				SELECT 1
+				FROM objects o
+				JOIN object_slabs os ON o.id = os.object_id
+				WHERE o.account_id = $1 AND os.slab_digest = $3
+			)`, id, sID, sqlHash256(slabID)).Scan(&pinned, &inUse)
 		if err != nil {
-			return fmt.Errorf("failed to check if slab is pinned: %w", err)
+			return fmt.Errorf("failed to check slab pin state: %w", err)
 		} else if !pinned {
 			return slabs.ErrSlabNotFound
-		}
-
-		// refuse to unpin a slab that is still referenced by one of the
-		// account's objects; otherwise the account would free up quota while
-		// the object keeps the slab alive
-		var inUse bool
-		err = tx.QueryRow(ctx, `SELECT EXISTS (
-			SELECT 1
-			FROM objects o
-			JOIN object_slabs os ON o.id = os.object_id
-			WHERE o.account_id = $1 AND os.slab_digest = $2
-		)`, id, sqlHash256(slabID)).Scan(&inUse)
-		if err != nil {
-			return fmt.Errorf("failed to check if slab is in use: %w", err)
 		} else if inUse {
 			return slabs.ErrSlabInUse
 		}

```

peterjan (dismissed):
pretty solid overall, left one minor optimisation

peterjan (approved):

PR #1045, opened 2026-07-22
Skip redundant settings validation

peterjan:
We re-validate the settings under lock for every RPC. Under heavy concurrency that serializes every RPC across every host behind ~28µs of crypto and 1600 B of allocations per call. It's a very tiny change but it came up during benchmarking and I figured it's probably worth fixing.

### DIFF
--- a/client/v2/settings_test.go
+++ b/client/v2/settings_test.go
@@ -1,0 +1,48 @@
+package client
+
+import (
+	"testing"
+	"time"
+
+	proto "go.sia.tech/core/rhp/v4"
+	"go.sia.tech/core/types"
+)
+
+func signedSettings() (types.PublicKey, proto.HostSettings) {
+	priv := types.GeneratePrivateKey()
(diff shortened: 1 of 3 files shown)


client/v2/settings_test.go
```diff
@@ -154,3 +154,42 @@ func TestRefreshSettingsInvalidPrices(t *testing.T) {
...
+// regression that reintroduces the ed25519 verify under the client mutex.
+func BenchmarkSettingsCacheHit(b *testing.B) {
+	hostKey, settings := signedSettings()
+	c := &Client{cachedSettings: map[types.PublicKey]proto.HostSettings{hostKey: settings}}
+	ctx := b.Context()
+	b.ReportAllocs()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			if _, cached, err := c.settings(ctx, hostKey, nil); err != nil || !cached {
+				b.Fatal("expected cache hit")
```
chris124567, line 184: nit: I think you are supposed to use `b.Error` if you use `RunParallel`

PR #1046, opened 2026-07-22
Support adding and removing sharing keys and shared objects

n8mgr:
This adds the app methods required to manage sharing keys and shared objects. It does not include shared auth logic or the methods needed for sharing keys to download data.

### DIFF
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -21,6 +21,7 @@ import (
 	"go.sia.tech/indexd/accounts"
 	"go.sia.tech/indexd/api"
 	"go.sia.tech/indexd/hosts"
+	"go.sia.tech/indexd/sharing"
 	"go.sia.tech/indexd/slabs"
 	"go.sia.tech/jape"
 	"go.uber.org/zap"
@@ -46,6 +47,18 @@ type (
 		PinObject(ctx context.Context, account proto.Account, obj slabs.PinObjectRequest) error
 		ListObjects(ctx context.Context, account proto.Account, cursor slabs.Cursor, limit int) ([]slabs.ObjectEvent, error)
 		SharedObject(ctx context.Context, key types.Hash256) (slabs.SharedObject, error)
+	}
+
+	// Sharing defines the sharing key management interface for the application
+	// API.
+	Sharing interface {
+		AddSharingKey(account proto.Account, req sharing.KeyRequest) (sharing.Key, error)
+		SharingKey(publicKey types.PublicKey) (sharing.Key, error)
+		SharingKeys(account proto.Account, offset, limit int) ([]sharing.Key, error)
+		DeleteSharingKey(account proto.Account, publicKey types.PublicKey) error
+		AddSharedObject(account proto.Account, sharingKey types.PublicKey, req sharing.SharedObjectRequest) error
+		DeleteSharedObject(account proto.Account, sharingKey types.PublicKey, objectKey types.Hash256) error
+		OwnedSharedObjects(account proto.Account, sharingKey types.PublicKey, offset, limit int) ([]slabs.SealedObject, error)
 	}
 
 	// Hosts defines the hosts interface for the application API.
@@ -159,6 +172,7 @@ type (
 		accounts  Accounts
 		contracts Contracts
 		slabs     Slabs
+		sharing   Sharing
 		log       *zap.Logger
 		rl        RateLimiter
 
@@ -359,6 +373,142 @@ func (a *app) handleDELETEObjects(jc jape.Context, pk types.PublicKey) {
 
 	err := a.slabs.DeleteObject(jc.Request.Context(), proto.Account(pk), key)
 	if errors.Is(err, slabs.ErrObjectNotFound) {
+		jc.Error(err, http.StatusNotFound)
+		return
+	} else if err != nil {
+		jc.Error(err, http.StatusInternalServerError)
+		return
+	}
+	jc.Encode(nil)
+}
+
+func (a *app) handlePOSTSharing(jc jape.Context, pk types.PublicKey) {
+	req, ok := decodeRequest[sharing.KeyRequest](jc)
+	if !ok {
+		return
+	}
+
+	key, err := a.sharing.AddSharingKey(proto.Account(pk), req)
+	if errors.Is(err, sharing.ErrInvalidRequest) {
+		jc.Error(err, http.StatusBadRequest)
+		return
+	} else if errors.Is(err, sharing.ErrSharingKeyExists) {
+		jc.Error(err, http.StatusConflict)
+		return
+	} else if err != nil {
+		jc.Error(err, http.StatusInternalServerError)
+		return
+	}
+	jc.Encode(key)
+}
+
+func (a *app) handleGETSharing(jc jape.Context, pk types.PublicKey) {
+	offset, limit, ok := api.ParseOffsetLimit(jc)
+	if !ok {
+		return
+	}
+
+	keys, err := a.sharing.SharingKeys(proto.Account(pk), offset, limit)
+	if jc.Check("failed to list sharing keys", err) != nil {
+		return
+	}
+	jc.Encode(keys)
+}
+
+func (a *app) handleGETSharingKey(jc jape.Context, pk types.PublicKey) {
+	var key types.PublicKey
+	if jc.DecodeParam("key", &key) != nil {
+		return
+	}
+
+	sk, err := a.sharing.SharingKey(key)
+	// scope to the caller: a key owned by another account reads as not found
+	if errors.Is(err, sharing.ErrSharingKeyNotFound) || (err == nil && sk.Account != pk) {
+		jc.Error(sharing.ErrSharingKeyNotFound, http.StatusNotFound)
+		return
+	} else if err != nil {
+		jc.Error(err, http.StatusInternalServerError)
+		return
+	}
+	jc.Encode(sk)
+}
+
+func (a *app) handleDELETESharing(jc jape.Context, pk types.PublicKey) {
+	var key types.PublicKey
+	if jc.DecodeParam("key", &key) != nil {
+		return
+	}
+
+	err := a.sharing.DeleteSharingKey(proto.Account(pk), key)
+	if errors.Is(err, sharing.ErrSharingKeyNotFound) {
+		jc.Error(err, http.StatusNotFound)
+		return
+	} else if err != nil {
+		jc.Error(err, http.StatusInternalServerError)
+		return
+	}
+	jc.Encode(nil)
+}
+
+func (a *app) handlePOSTSharingObject(jc jape.Context, pk types.PublicKey) {
+	var key types.PublicKey
+	if jc.DecodeParam("key", &key) != nil {
+		return
+	}
+
+	req, ok := decodeRequest[sharing.SharedObjectRequest](jc)
+	if !ok {
+		return
+	}
+
+	err := a.sharing.AddSharedObject(proto.Account(pk), key, req)
+	if errors.Is(err, sharing.ErrInvalidRequest) {
+		jc.Error(err, http.StatusBadRequest)
+		return
+	} else if errors.Is(err, sharing.ErrSharingKeyNotFound) || errors.Is(err, slabs.ErrObjectNotFound) {
+		jc.Error(err, http.StatusNotFound)
+		return
+	} else if err != nil {
+		jc.Error(err, http.StatusInternalServerError)
+		return
+	}
+	jc.Encode(nil)
+}
+
+func (a *app) handleGETSharingObjects(jc jape.Context, pk types.PublicKey) {
+	var key types.PublicKey
+	if jc.DecodeParam("key", &key) != nil {
+		return
+	}
+
+	offset, limit, ok := api.ParseOffsetLimit(jc)
+	if !ok {
+		return
+	}
+
+	objects, err := a.sharing.OwnedSharedObjects(proto.Account(pk), key, offset, limit)
+	if errors.Is(err, sharing.ErrSharingKeyNotFound) {
+		jc.Error(err, http.StatusNotFound)
+		return
+	} else if err != nil {
+		jc.Error(err, http.StatusInternalServerError)
+		return
+	}
+	jc.Encode(objects)
+}
+
+func (a *app) handleDELETESharingObject(jc jape.Context, pk types.PublicKey) {
+	var key types.PublicKey
+	if jc.DecodeParam("key", &key) != nil {
+		return
+	}
+	var objectKey types.Hash256
+	if jc.DecodeParam("objectkey", &objectKey) != nil {
+		return
+	}
+
+	err := a.sharing.DeleteSharedObject(proto.Account(pk), key, objectKey)
+	if errors.Is(err, sharing.ErrSharedObjectNotFound) {
 		jc.Error(err, http.StatusNotFound)
 		return
 	} else if err != nil {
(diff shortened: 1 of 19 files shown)


openapi/components.yml
```diff
@@ -241,6 +241,113 @@ components:
           items:
             $ref: "#/components/schemas/SlabSlice"
 
+    Nonce:
+      type: string
+      pattern: "^[0-9a-fA-F]{64}$"
+      description: A 32-byte value (hex-encoded) used as the HKDF salt to derive a sharing key from an app key
```
peterjan, line 247: `nit` Nonce is quite a broad term if the description references sharing and app keys

sharing/manager.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	return m.store.AddSharedObject(account, sharingKey, req)
+}
+
+// DeleteSharedObject detaches an object from one of the account's sharing keys.
+func (m *Manager) DeleteSharedObject(account proto.Account, sharingKey types.PublicKey, objectKey types.Hash256) error {
+	return m.store.DeleteSharedObject(account, sharingKey, objectKey)
+}
+
+// OwnedSharedObjects returns a paginated list of the object IDs attached to a
```
peterjan: `nit` update docstring, returns way more than object IDs

api/app/app.go
```diff
@@ -368,6 +382,142 @@ func (a *app) handleDELETEObjects(jc jape.Context, pk types.PublicKey) {
...
+	jc.Encode(keys)
+}
+
+func (a *app) handleGETSharingKey(jc jape.Context, pk types.PublicKey) {
+	var key types.PublicKey
+	if jc.DecodeParam("key", &key) != nil {
+		return
+	}
+
+	sk, err := a.sharing.SharingKey(key)
```
peterjan: we do the auth check inline here in the handler, but seems like all other handlers either perform the check in the manager or even in the store, we should probably try and keep the ownership in one layer

persist/postgres/sharing.go
```diff
@@ -0,0 +1,315 @@
...
+		if len(req.EncryptedMetadata) > 0 {
+			encryptedMeta = req.EncryptedMetadata
+		}
+
+		_, err = tx.Exec(ctx, `
+			INSERT INTO shared_objects (object_id, sharing_key_id, encrypted_data_key, encrypted_meta_key, encrypted_metadata, data_signature, meta_signature, size, pinned_data, pinned_size)
+			VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10)
+			ON CONFLICT (object_id, sharing_key_id) DO UPDATE SET (encrypted_data_key, encrypted_meta_key, encrypted_metadata, data_signature, meta_signature, size, pinned_data, pinned_size, updated_at) = (EXCLUDED.encrypted_data_key, EXCLUDED.encrypted_meta_key, EXCLUDED.encrypted_metadata, EXCLUDED.data_signature, EXCLUDED.meta_signature, EXCLUDED.size, EXCLUDED.pinned_data, EXCLUDED.pinned_size, NOW())`,
+			objectID, sharingKeyID, req.EncryptedDataKey, encryptedMetaKey, encryptedMeta, sqlSignature(req.DataSignature), sqlSignature(req.MetadataSignature), size, pinnedData, pinnedSize)
+		if err != nil {
```
peterjan, line 213: should we check for unique violations like we do in `AddSharingKey`, it would surface as a raw 500 while AddSharingKey handles it more gracefully with `sharing.ErrSharingKeyExists`

persist/postgres/sharing.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+		return nil
+	})
+}
+
+// PruneExpiredSharingKeys deletes all sharing keys that expired on or before
+// the cutoff.
+func (s *Store) PruneExpiredSharingKeys(cutoff time.Time) error {
+	return s.transaction(func(ctx context.Context, tx *txn) error {
+		_, err := tx.Exec(ctx, `DELETE FROM sharing_keys WHERE expires_at IS NOT NULL AND expires_at <= $1`, cutoff)
```
peterjan, line 228: we have an index, `preauthorized_keys_expires_at_idx`, for this on pre-authed keys but not here, should we consider adding one, does a full table scan every interval

sharing/manager.go
```diff
@@ -0,0 +1,154 @@
...
+
+// WithLogger sets the logger for the Manager.
+func WithLogger(l *zap.Logger) Option {
+	return func(m *Manager) {
+		m.log = l
+	}
+}
+
+// WithPruneInterval sets the interval at which expired sharing keys are pruned.
+func WithPruneInterval(interval time.Duration) Option {
```
peterjan: currently not used

persist/postgres/objects.go
```diff
@@ -105,37 +105,7 @@ func (s *Store) Object(account proto.Account, key types.Hash256) (obj slabs.Seal
...
-		}
-		if err := rows.Err(); err != nil {
-			return err
-		}
-
-		if err := decorateSectors(ctx, tx, &obj, slabIDs); err != nil {
-			return fmt.Errorf("failed to decorate sectors of slab: %w", err)
-		}
-		return nil
+		return loadObjectSlabs(ctx, tx, objID, &obj)
```
peterjan, line 108: I think we can also reuse this in `ListObjects`

sharing/sharing.go
```diff
@@ -0,0 +1,163 @@
...
+	// NonceSize is the required length of a sharing key's nonce, used as the
+	// HKDF salt when deriving the sharing key from the app key.
+	NonceSize = 32
+	// EncryptionKeySize is the length of a sealed encryption key (xchacha20
+	// nonce + key + tag).
+	EncryptionKeySize = 72
+	// MaxDescriptionSize is the maximum length of a sharing key description.
+	MaxDescriptionSize = 1024
+	// MaxMetadataSize is the maximum length of a shared object's encrypted
+	// metadata, mirroring the limit on a normal object's metadata.
```
peterjan: `nit` could export the slab metadatasize and re-export it here so they are linked

persist/postgres/sharing.go
```diff
@@ -0,0 +1,315 @@
...
+	"github.com/jackc/pgx/v5/pgconn"
+	proto "go.sia.tech/core/rhp/v4"
+	"go.sia.tech/core/types"
+	"go.sia.tech/indexd/accounts"
+	"go.sia.tech/indexd/sharing"
+	"go.sia.tech/indexd/slabs"
+)
+
+func scanSharingKey(s scanner) (key sharing.Key, err error) {
+	var expiresAt sql.NullTime
```
peterjan: `nit` unnecessary? it'll encode a nil time to NULL automatically

sharing/sharing.go
```diff
@@ -0,0 +1,163 @@
...
+		PinnedData uint64 `json:"pinnedData"`
+		// PinnedSize is the size of all objects attached to this key on the network, including redundancy
+		PinnedSize uint64     `json:"pinnedSize"`
+		ExpiresAt  *time.Time `json:"expiresAt,omitempty"`
+		CreatedAt  time.Time  `json:"createdAt"`
+		UpdatedAt  time.Time  `json:"updatedAt"`
+	}
+
+	// A KeyRequest contains the fields required to create a sharing key.
+	KeyRequest struct {
```
peterjan, line 71: should we consider requiring a signature here?

peterjan (changes requested):

sharing/sharing.go
```diff
@@ -0,0 +1,199 @@
...
+// VerifySignature verifies that the sharing key owner signed the request.
+func (r KeyRequest) VerifySignature() error {
+	if !r.PublicKey.VerifyHash(r.SigHash(), r.Signature) {
+		return fmt.Errorf("%w: invalid signature", ErrInvalidRequest)
+	}
+	return nil
+}
+
+func (r KeyRequest) validate() error {
+	switch {
```
chris124567, line 123: nit: You could check `ExpiresAt` too

persist/postgres/init.sql
```diff
@@ -398,6 +398,76 @@ CREATE INDEX object_events_updated_at_object_key_idx ON object_events(updated_at
...
+    size BIGINT NOT NULL, -- logical size of the object (object.Size()), captured at attach time for trigger
+    pinned_data BIGINT NOT NULL, -- data size of the object before redundancy, captured at attach time for trigger
+    pinned_size BIGINT NOT NULL, -- size of the object including redundancy, captured at attach time for trigger
+    PRIMARY KEY (object_id, sharing_key_id)
+);
+CREATE INDEX shared_objects_sharing_key_id_idx ON shared_objects(sharing_key_id);
+
+-- maintain sharing_keys.{object_count, pinned_data, pinned_size} as objects are
+-- attached and detached.
+CREATE FUNCTION shared_objects_maintain_totals() RETURNS TRIGGER AS $$
```
chris124567, line 437: Are we absolutely sure we want to introduce triggers for this?  We maintain all other stats manually.

n8mgr: This one felt odd to do manually since it needs to catch cascades off the main account. I think introducing a select to do it manually is a worse option, but honestly unsure and open to doing it manually.

n8mgr: Maybe we can refine it to a statement level trigger instead so it's not fired per row?

https://www.postgresql.org/docs/current/sql-createtrigger.html

persist/postgres/sharing.go
```diff
@@ -0,0 +1,315 @@
...
+		if err != nil {
+			return err
+		}
+
+		rows, err := tx.Query(ctx, `
+			SELECT a.public_key, sk.public_key, sk.nonce, sk.use_description, sk.object_count, sk.size, sk.pinned_data, sk.pinned_size, sk.expires_at, sk.created_at, sk.updated_at
+			FROM sharing_keys sk
+			INNER JOIN accounts a ON a.id = sk.account_id
+			WHERE sk.account_id = $1 AND (sk.expires_at IS NULL OR sk.expires_at > NOW())
+			ORDER BY sk.created_at DESC
```
chris124567: nit: add `sk.id` tiebreaker (edit: or just use `sk.id`?)

persist/postgres/migrations.go
```diff
@@ -206,6 +206,119 @@ CREATE TABLE preauthorized_keys (
 );
 CREATE INDEX preauthorized_keys_connect_key_id_idx ON preauthorized_keys(connect_key_id);
 CREATE INDEX preauthorized_keys_expires_at_idx ON preauthorized_keys(expires_at);
+`)
```
chris124567, line 216: We should check triggers in  `TestMigrationConsistency`

sharing/sharing.go
```diff
@@ -0,0 +1,201 @@
...
+	case len(r.EncryptedMetadataKey) != 0 && len(r.EncryptedMetadataKey) != EncryptionKeySize:
+		return fmt.Errorf("%w: encrypted metadata key must be %d bytes", ErrInvalidRequest, EncryptionKeySize)
+	case len(r.EncryptedMetadata) > MaxMetadataSize:
+		return fmt.Errorf("%w: encrypted metadata exceeds %d bytes", ErrInvalidRequest, MaxMetadataSize)
+	}
+	return nil
+}
+
+func (r SharedObjectRequest) dataSigHash() types.Hash256 {
+	h := types.NewHasher()
```
chris124567, line 151: Do we need a prefix for these?

chris124567:
Edge case with objects having same slab multiple times

``` 
func TestSharingKeyTotalsRepeatedSlab(t *testing.T) {
   store := initPostgres(t, zap.NewNop())

   acc := proto.Account(types.GeneratePrivateKey().PublicKey())
   store.addTestAccount(t, types.PublicKey(acc))
   hk := store.addTestHost(t)
   store.addTestContract(t, hk)

   params := slabs.SlabPinParams{
       EncryptionKey: frand.Entropy256(),
       MinShards:     1,
       Sectors:       []slabs.PinnedSector{{Root: frand.Entropy256(), HostKey: hk}},
   }
   if _, err := store.PinSlabs(acc, time.Time{}, params); err != nil {
       t.Fatalf("failed to pin slabs: %v", err)
   }

   // an object may reference the same slab at more than one index
   obj := slabs.SealedObject{
       EncryptedDataKey:  frand.Bytes(72),
       DataSignature:     types.Signature(frand.Bytes(64)),
       MetadataSignature: types.Signature(frand.Bytes(64)),
       Slabs:             []slabs.SlabSlice{params.Slice(0, 100), params.Slice(100, 100)},
   }
   if err := store.PinObject(acc, obj.PinRequest()); err != nil {
       t.Fatalf("failed to pin object: %v", err)
   }

   var accountPinnedData, accountPinnedSize uint64
   if err := store.pool.QueryRow(t.Context(),
       `SELECT pinned_data, pinned_size FROM accounts WHERE public_key = $1`,
       sqlPublicKey(types.PublicKey(acc))).Scan(&accountPinnedData, &accountPinnedSize); err != nil {
       t.Fatal(err)
   }

   sk := store.addTestSharingKey(t, acc, "repeated slab")
   attachTestObject(t, store, acc, sk, obj.ID())

   // both slices share one slab, so the key's storage totals must match what
   // the account was charged; only the logical size counts each slice
   assertKeyTotals(t, store, sk, 1, 200, accountPinnedData, accountPinnedSize)
}
```

```
=== RUN   TestSharingKeyTotalsRepeatedSlab
    sharing_test.go:531: unexpected totals: count=1 objectSize=200 pinnedData=8388608 pinnedSize=8388608, want 1/200/4194304/4194304
```

Fix:
```
diff --git a/persist/postgres/sharing.go b/persist/postgres/sharing.go
index 2fc690ba..e8ac99a4 100644
--- a/persist/postgres/sharing.go
+++ b/persist/postgres/sharing.go
@@ -173,16 +173,20 @@ func (s *Store) AddSharedObject(account proto.Account, sharingKey types.PublicKe
        // capture the object's sizes at attach time so the trigger can
        // maintain the sharing key's totals: size is the object's logical size
        // (sum of slab slice lengths), pinned_data/pinned_size are its storage
-       // footprint before and after redundancy.
+       // footprint before and after redundancy. a slab referenced by more than
+       // one slice is only stored once, so it only counts once towards them.
        var objectSize, minShards, sectorCount int64
        err = tx.QueryRow(ctx, `
+           WITH object_slab_ids AS (
+               SELECT DISTINCT s.id, s.min_shards
+               FROM object_slabs os
+               INNER JOIN slabs s ON s.digest = os.slab_digest
+               WHERE os.object_id = $1
+           )
            SELECT
-               COALESCE(SUM(os.slab_length), 0)::bigint,
-               COALESCE(SUM(s.min_shards), 0)::bigint,
-               COALESCE(SUM((SELECT COUNT(*) FROM slab_sectors ss WHERE ss.slab_id = s.id)), 0)::bigint
-           FROM object_slabs os
-           INNER JOIN slabs s ON s.digest = os.slab_digest
-           WHERE os.object_id = $1
+               (SELECT COALESCE(SUM(slab_length), 0)::bigint FROM object_slabs WHERE object_id = $1),
+               (SELECT COALESCE(SUM(min_shards), 0)::bigint FROM object_slab_ids),
+               (SELECT COUNT(*)::bigint FROM slab_sectors WHERE slab_id IN (SELECT id FROM object_slab_ids))
        `, objectID).Scan(&objectSize, &minShards, &sectorCount)
        if err != nil {
            return fmt.Errorf("failed to compute object size: %w", err)
```

api/app/sharing_test.go
```diff
@@ -0,0 +1,243 @@
...
+	"go.sia.tech/indexd/slabs"
+	"go.sia.tech/indexd/testutils"
+	"go.uber.org/zap"
+	"lukechampine.com/frand"
+)
+
+func assertStatus(t *testing.T, err error, status int) {
+	t.Helper()
+	var he *app.HTTPError
+	if !errors.As(err, &he) {
```
ChrisSchinnerl: (nit) use the new `errors.AsType`.

persist/postgres/init.sql
```diff
@@ -398,6 +398,117 @@ CREATE INDEX object_events_updated_at_object_key_idx ON object_events(updated_at
...
+            FROM new_rows
+            GROUP BY sharing_key_id
+        ) agg
+        WHERE sharing_keys.id = agg.sharing_key_id;
+    ELSIF (TG_OP = 'UPDATE') THEN
+        UPDATE sharing_keys SET
+            size = sharing_keys.size + agg.size,
+            pinned_data = sharing_keys.pinned_data + agg.pinned_data,
+            pinned_size = sharing_keys.pinned_size + agg.pinned_size,
+            updated_at = NOW()
```
ChrisSchinnerl, line 462: If a shared object has its sharing key id updated during an update, the object_count actually goes down for one sharing key and up for another. This is not a problem right now since we don't do that and maybe never will but it's a tiny change that makes sure all updates are correct.

PR #1047, opened 2026-07-22
Add support for sharing keys

n8mgr:
SDKs will be able to access objects using the `/shared` endpoints and a valid sharing key

### DIFF
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -1054,11 +1054,11 @@ func NewAPI(advertiseURL string, hm Hosts, am Accounts, contracts Contracts, sla
 		"DELETE /sharing/:key/objects/:objectkey": wrapSignedAuth(a.handleDELETESharingObject),
 
 		// shared-key endpoints, authenticated with a sharing key
-		"GET /shared":                     wrapSharedAuth(a.handleGETShared),
-		"GET /shared/objects":             wrapSharedAuth(a.handleGETSharedObjects),
-		"GET /shared/objects/:id":         wrapSharedAuth(a.handleGETSharedObject),
-		"GET /shared/hosts":               wrapSharedAuth(a.handleGETSharedHosts),
-		"GET /shared/hosts/:pubkey/token": wrapSharedAuth(a.handleGETSharedHostToken),
+		"GET /shared":                      wrapSharedAuth(a.handleGETShared),
+		"GET /shared/objects":              wrapSharedAuth(a.handleGETSharedObjects),
+		"GET /shared/objects/:id":          wrapSharedAuth(a.handleGETSharedObject),
+		"GET /shared/hosts":                wrapSharedAuth(a.handleGETSharedHosts),
+		"GET /shared/hosts/:hostkey/token": wrapSharedAuth(a.handleGETSharedHostToken),
 
 		"GET /slabs":            wrapSignedAuth(a.handleGETSlabs),
 		"POST /slabs":           wrapSignedAuth(a.handlePOSTSlabs),
--- a/api/app/app.go
+++ b/api/app/app.go
@@ -55,10 +55,14 @@ type (
 		AddSharingKey(account proto.Account, req sharing.KeyRequest) (sharing.Key, error)
 		DeleteSharingKey(account proto.Account, publicKey types.PublicKey) error
 		SharingKeys(account proto.Account, offset, limit int) ([]sharing.Key, error)
+		SharingKey(publicKey types.PublicKey) (sharing.Key, error)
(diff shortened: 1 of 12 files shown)


api/app/auth.go
```diff
@@ -120,6 +123,27 @@ func validateSignedURLAuth(jc jape.Context, hostname string, store Accounts) (ty
...
+// request must be signed with a sharing key's private key; the signer's public
+// key must resolve to an existing, unexpired sharing key. On success it returns
+// that sharing key.
+func validateSharedURLAuth(jc jape.Context, hostname string, store Sharing) (sharing.Key, bool) {
+	pk, ok := ValidateURLSignature(jc.Request, jc.ResponseWriter, hostname)
+	if !ok {
+		return sharing.Key{}, false
+	}
+
+	key, err := store.SharingKey(pk)
```
peterjan, line 136: might be nicer UX to return `ErrSharingKeyExpired` in case we know the key but it has expired?

api/app/app.go
```diff
@@ -1033,6 +1050,13 @@ func NewAPI(advertiseURL string, hm Hosts, am Accounts, contracts Contracts, sla
 		"GET /sharing/:key/objects":               wrapSignedAuth(a.handleGETSharingObjects),
 		"DELETE /sharing/:key/objects/:objectkey": wrapSignedAuth(a.handleDELETESharingObject),
 
+		// shared-key endpoints, authenticated with a sharing key
+		"GET /shared":                     wrapSharedAuth(a.handleGETShared),
+		"GET /shared/objects":             wrapSharedAuth(a.handleGETSharedObjects),
+		"GET /shared/objects/:id":         wrapSharedAuth(a.handleGETSharedObject),
+		"GET /shared/hosts":               wrapSharedAuth(a.handleGETSharedHosts),
+		"GET /shared/hosts/:pubkey/token": wrapSharedAuth(a.handleGETSharedHostToken),
```
peterjan: `hostkey` for consistency

api/app/shared.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	obj, err := a.sharing.SharedObject(key.PublicKey, objectKey)
+	if errors.Is(err, sharing.ErrSharingKeyNotFound) {
+		jc.Error(err, http.StatusUnauthorized)
+		return
+	} else if errors.Is(err, sharing.ErrSharedObjectNotFound) {
+		jc.Error(err, http.StatusNotFound)
+		return
+	} else if err != nil {
```
peterjan: `nit` all other handlers you added do a `jc.Check(` here with an error message

peterjan (approved):

api/app/auth.go
n8mgr: An expired key is around for 10m before pruning. Seems like a lot of effort to provide an error that will be hittable for at most 10m.

chris124567 (commented):
Is there anything stopping an account token from being used for writing to temporary storage?

peterjan (approved):

api/app/shared.go
```diff
@@ -49,22 +58,28 @@ func (a *app) handleGETSharedObject(jc jape.Context, key sharing.Key) {
...
 		return
-	} else if jc.Check("failed to create account token", err) != nil {
+	} else if jc.Check("failed to create account tokens", err) != nil {
 		return
 	}
-	jc.Encode(token)
+
+	sharedHosts := make([]SharedHost, len(usable))
+	for i, h := range usable {
+		sharedHosts[i] = SharedHost{HostInfo: h, Token: tokens[i]}
```
peterjan, line 82: this is not documented properly in the openapi spec, the commit that introduced it didn't touch the docs so we're likely missing a SharedHost component plus an update to the route

api/app/shared.go
```diff
@@ -0,0 +1,85 @@
...
+		jc.Error(err, http.StatusNotFound)
+		return
+	} else if jc.Check("failed to get shared object", err) != nil {
+		return
+	}
+	jc.Encode(obj)
+}
+
+func (a *app) handleGETSharedHosts(jc jape.Context, key sharing.Key) {
+	usable, ok := a.usableHosts(jc)
```
chris124567, line 62: Don't the hosts need to have the `sharing_attached=TRUE` for the owner's pool?

ChrisSchinnerl: Theoretically yes. You need hosts with pool support + an attached key. But the worst that can happen is that the host responds with an error so it's probably fine and we can handle it in the app in case the download fails.

